PR #1, opened 2015-06-25
s/resolution scope/base URI/

Acubed:
So let's try making a pull request to the draft.

Not to push things along faster than they must, of course, let's be sure of the PR model we want to use before merging.

---

This fixes a number of vocabulary confusion, and inconsistencies with the URI definitions and ABNF terms. I think there's still a few more, though.


willcrick:
these changes look good to me.

PR #2, opened 2015-06-29
`profile` in Accept header must be a quoted-string

jdesrosiers:
I discovered recently that the usage of the `profile` Accept header extension
described in the documentation is not a valid accept header extension.

```
Content-Type: application/json; profile=/schema-for-this-data
```

The schema should be within quotes in order the valid.
```
Content-Type: application/json; profile="/schema-for-this-data"
```

This is how the `Content-Type` header is defined
```
Content-Type   = "Content-Type" ":" media-type

media-type     = type "/" subtype *( ";" parameter )
type           = token
subtype        = token

parameter      = attribute "=" value
attribute      = token
value          = token | quoted-string

token          = 1*<any CHAR except CTLs or separators>
separators     = "(" | ")" | "<" | ">" | "@"
               | "," | ";" | ":" | "\" | <">
               | "/" | "[" | "]" | "?" | "="
               | "{" | "}" | SP | HT
```
http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.17
http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7
http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.6
http://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2

Therefore, `/schema-for-this-data` is not a valid token because it contains a
`/`.  It needs to be expressed as a `quoted-string`.  This occurs anywhere
`profile` appears in the documentation.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -380,7 +380,7 @@
                 <t>
                     JSON Schema uses <xref target="json-reference">JSON Reference</xref> as a
                     mechanism for schema addressing. It extends this specification in two ways:
-                    
+
                     <list>
                         <t>JSON Schema offers facilities to alter the base URI against which a
                         reference must resolve by the means of the "id" keyword;</t>
@@ -388,7 +388,7 @@
                     </list>
 
                 </t>
-                
+
                 <t>
                     Altering the URI within a schema is called defining a new resolution scope.
                     The initial resolution scope of a schema is the URI of the schema itself, or a suitable substitute URI if none is known.
@@ -396,7 +396,7 @@
             </section>
 
             <section title='URI resolution scope alteration with the "id" keyword'>
-                
+
                 <section title="Valid values">
                     <t>
                         The value for this keyword MUST be a string, and MUST be a valid URI (relative or absolute).
@@ -455,7 +455,7 @@
                         Subschemas at the following URI-encoded <xref target="json-pointer">JSON
                         Pointer</xref>s (starting from the root schema) define the following
                         resolution scopes:
-    
+
                         <list style="hanging">
                             <t hangText="# (document root)">http://x.y.z/rootschema.json#</t>
                             <t hangText="#/schema1">http://x.y.z/rootschema.json#foo</t>
@@ -465,7 +465,7 @@
                             <t hangText="#/schema3">some://where.else/completely#</t>
                         </list>
                     </t>
-    
+
                 </section>
 
                 <section title="Methods for dereferencing">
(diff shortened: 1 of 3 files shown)


Acubed:
Oh wow, I think I make this mistake even in my own code. Nice catch.

+1

Acubed:
@jdesrosiers Note that RFC2616 is no longer current, that's defined in RFC7230 now: http://httpwg.github.io/specs/rfc7230.html#rule.token.separators

jdesrosiers:
Oops.  I actually wrote this up a while back shortly before I learned about RFC7230.  I was lazy and didn't think to update it when I recreated it for this pull request.

But, is it true that RFC2616 is not longer current?  Everything I have found on RFC7230 seems to call it a "proposed standard" or it is on a "standards track".  This leads me to believe that it is still a work in progress and is not official yet.  In that case RFC2616 would still be current.  I've heard others refer to RFC2616 as obsolete as well, so am I missing something?  Or, are we all jumping the gun because we are excited about the update?

Acubed:
@jdesrosiers https://www.mnot.net/blog/2014/06/07/rfc2616_is_dead

The designations are... complex and often misleading. I'm not sure what the distinction is. But what's important is that RFC7230, RFC7231, etc list RFC2616 as "Obsolete" which means it's no longer relevant in any capacity. The original .txt is never changed once published (which was in 1999), but kept for historical purposes of course.

jdesrosiers:
Thanks @Acubed.  That blog post puts it in no uncertain terms.  Given the author, I'm sure he knows what he's talking about, but I was hoping for something more official than a blog post.

Acubed:
@jdesrosiers The official word is going to be the publication of a new RFC that lists the old one as obsolete. e.g. [RFC 7230](http://www.rfc-editor.org/rfc/rfc7230.txt) specifies "Obsoletes: 2145, 2616". An index of the obsoleted RFCs is maintained at rfc-editor.org too. Additionally, you can see the brown header (=obsolete) at the IETF website, along with published errata: <https://tools.ietf.org/html/rfc7230>

It's just little bit confusing, I didn't actually realize rfc-editor.org was official.

jdesrosiers:
@Acubed Yes, it is confusing.  Thanks for shedding some light on it for me.

Acubed:
Hmm, do we have a policy on trailing whitespace on otherwise empty lines?

jdesrosiers:
I don't think there are any official policies at this point, but removing trailing whitespace is a pretty widely accepted practice.

PR #3, opened 2015-07-08
Standardise use of hyphens with 'non'

pdl:
The only "non interactive" without hyphenation I have left in is the title of an earlier publication.


Acubed:
Good idea (+1)

PR #11, opened 2016-03-17
README changes (80 chars/line; use recommended single top level heading;

brettz9:
more specific licenses; dev. steps);
allow building via npm (avoids Make problems/dependency on Windows and established build pattern).

I didn't know the authors/contributors to add or whether my change to a more specific license was accurate, and let me know whatever other changes you might want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,34 @@
-## Welcome to JSON Schema
+# JSON Schema
+
+## Welcome
 
 This repository contains the current, and historical, JSON Schema
 specifications.
 
 ## Call for reviews
 
-Specifications are starting to get written. Reviews, comments and suggestions
-are of paramount importance to JSON Schema. It is humbly asked to you, dear
-reader, that you bring your contribution.
+Specifications are starting to get written. Reviews, comments, and suggestions
+are of paramount importance to JSON Schema. It is humbly asked of you, dear
+reader, that you bring your contributions.
 
 ## The website
 
-The JSON Schema web site is at http://json-schema.org/
+The JSON Schema web site is at <http://json-schema.org/>.
 
 ## License
 
-The source material in this repository is licensed under the AFL or BSD license.
+The source material in this repository is licensed under the AFL v3.0
+or BSD 3 revised license.
+
+## Development
+
+1.  [Download and install Python](https://www.python.org/downloads/), being
+    sure to add Python to the system path.
+
+2.  Follow the [install instructions for xml2rfc](https://pypi.python.org/pypi/xml2rfc/).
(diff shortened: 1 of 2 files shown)


brettz9:
If you're ok with the changes, I'd recommend your publishing the repo on npm before it is taken (will be under "json-schema-spec" per the current `package.json`).

Acubed:
This is work under/a submission to the IETF, so is subject to the IETF IP copyright practices, I believe the current document describing that is <https://www.ietf.org/rfc/rfc3979.txt>

As for building, a Makefile is about as established as a build tool can get. Can you elaborate a bit why Makefile doesn't work out for you?

brettz9:
Windows does not support it out natively and I've just had trouble getting GNU Make for Windows to work for me.

brettz9:
It's also convenient to have a `package.json` anyways for apps to use `npm` for incorporating the current schema into their apps (e.g., the like [JSON Editor](https://github.com/jdorn/json-editor/)).

brettz9:
The README makes mention of BSD and AFL licenses, but that document mentions neither, and seems to be a more general document.

Acubed:
Package publication sort of a difficult issue for me to comment on as a maintainer of the '[jsonschema](https://github.com/tdegrunt/jsonschema)' package, which already ships the meta-schemas.

But I don't think it'd be appropriate to publish a specification as a software package, on any platform except the IETF or RFC editor. At the very least, we'd need to decide why to support one platform and not another, or under what cases to we add support for a platform.

I'd highly suggest getting a `make` program to work, though, lots of other ECMAScript/JavaScript packages depend on it for building, it's a pretty common thing, and definitely supported on Windows.

brettz9:
Ok, thanks for prompting me about `Make`. In this case, it did work with Cygwin, but using GNU Make alone did not work with the "clean" command (since rm is not recognized by it).

I think the ubiquity of JavaScript and the platform-neutral nature of the web brings a pretty compelling reason to use `npm` which is widely used by those working with JavaScript, including for browser scripts.

I'm not sure how your status as maintainer of the jsonschema package is of concern here, as the idea is not to prescribe any particular software implementation, and libraries such as yours could readily include the meta-schemas as dependencies, and in a more modular, cleaner way.

And as long as people don't mind a `node_modules` sub-directory, other languages can obtain the files as well. The only other competition of which I'm aware is `bower`, and a number of packages seem to have been discontinuing their support for it.

awwright:
As mentioned, the package.json file doesn't really make sense, so I'll close this.

If you wanna submit a PR for the README changes, or whatever else you have, we can definitely consider that.

PR #25, opened 2016-05-02
fixing typos

dret:


erosb:
I don't mind this change, but "refering" vs. "referring" sounds to be more like a difference between dialects than one being right while the other is wrong. See the HTTP "Referer" header.

Acubed:
Nice catch!

The HTTP `Referer` header is a known misspelling, but nonetheless probably where the current version (erroneously) gets its spelling from.

Issue #26, opened 2016-05-02
Profiles are not Schemas


dret:
the current draft talks about a profile media type parameter, and recommends using it to link to a JSON schema. the existing https://tools.ietf.org/html/rfc6906 about profiles (that also talks about "profile" media type parameters) is not referenced, but at least there is some overlap. please consider that the idea of profiles (as specified in RFC 6906) is not to interlink instances and schemas. instead, the idea is that a profile identifies a set of rules applied to an instance which can be processed with or without knowing the profile-specific processing rules.

Acubed:
I agree the spec should be a bit more clear about this.

The "profile" media type parameter doesn't form a Web link as such, if we want to also provide the path to download the schema, that needs to be done in a Link header.

This also has the effect of allowing schemas to be defined at alternate locations, or allowing you to use schemas that aren't network-addressable (like a URN).

Ideally, it seems, you should use both.

Acubed:
Also, JSON Schema suggests using the "describedby" link relation, but this appears to be overloading POWDER, a vaguely similar specification, as "describedby" is specific to application/powder+xml.

dret:
On 2016-05-03 12:56, Austin Wright wrote:
> Also, JSON Schema suggests using the "describedby" link relation, but
> this appears to be overloading POWDER, a vaguely similar specification,
> as "describedby" is specific to application/powder+xml.

POWDER does not constrain the link target to be POWDER resources (and 
note that RFC 5988 guidelines specifically say that registered link 
relations must not constrain the media type of the context or the target 
resources):

https://www.w3.org/TR/powder-dr/#assoc-linking

"This is a generic relationship type that does not of itself imply that 
the link points to a POWDER document."

but apart from that, i find the definition to be much more about 
providing metadata about a resource (such as "who's the author" and so 
on), but the language is so extremely vague that you can pretty much 
read into it whatever you like.

azaroth42:
Also, the `describedby` link relation was updated by LDP to be explicitly less specific:
    https://www.w3.org/TR/ldp/#link-relation-describedby

handrews:
@azaroth42 , @dret :  I think it's fine for `describedBy` to be vague about what it requires on the other end of the link.  It's not just the `describedBy` relation that tells clients that the target is a schema.  It is the media type `application/schema+json` that is set for the target representation that tells you that you now have a JSON Schema document that you can and should use to validate the data that supplied the `describedBy` link.

[...]

Issue #9, opened 2015-12-18
Link relation for pointing to network-addressable schema


awwright:
JSON Schema presently specifies to use the "profile" link relation, e.g.:

    Link: <http://example.com/Schema.json>;rel="profile"

However, the "profile" link relation is not supposed to be dereferenced. There's no way for an automated user agent to be able to follow this link and actually download the schema if it doesn't already have it.

Perhaps publish a link relation type, similar to `profile`, that is supposed to be downloaded?

handrews:
From RFC 6906:

>Thus, clients SHOULD treat profile URIs as
   identifiers and not as links, but profiles MAY be defined in a way
   that the URIs do identify retrievable profile description and thus
   can be accessed by clients by dereferencing the profile URI.  For
   profiles intended for use in environments where clients may encounter
   unknown profile URIs, profile maintainers SHOULD consider to make the
   profile URI dereferencable and provide useful documentation at that
   URI.  The design and representation of such profile descriptions,
   however, is outside the scope of this specification.

Which I take to mean that a client cannot depend on a profile URI being a dereferencable, but as profile authors are encourage to provide a derferencable URI for the profile then it certainly can be dereferenced and used.  It's just not guaranteed to be dereferencable.

So I think the only problem here is that it's not something anyone can rely on, but it is valid for it to work the way the json-schema specification suggests.

Plus, json-schema also provides for use of the describedBy link relation type, so isn't this effectively already fixed?  The describedBy relation is explicitly dereferencable.

awwright:
Take a look at what I wrote for the new JSON Schema website yesterday:
https://awwright.github.io/json-schema-org-pages/guide/http

Does this mesh well with everyone's understanding?

handrews:
Thanks for the quick reply!  The new site looks great, btw.

This paragraph does not match my reading of the RFC

>The “profile” media-type parameter is never supposed to be dereferenced or used to download a schema, it is strictly an identifier provided for the benefit of clients that have pre-programmed knowledge of what that profile means.

although I am certainly open to being convinced that my reading is wrong.  A revision of this paragraph that fits my (current) understanding would be something like:

>The “profile” media-type parameter is an identifier provided for the benefit of clients, and while it is a URI is not guaranteed to be dereferencable or to result in a downloadable schema.  However, implementations MAY provide a downloadable schema at the profile URI.  To unambiguously provide a downloadable URI, use the describedBy header.

[...]

PR #45, opened 2016-09-12
Rewrite/expand HTTP/profile/describedby sections

awwright:
Addresses #9, #26 

Rendered draft available at [jsonschema-core.txt](https://github.com/json-schema-org/json-schema-spec/files/466582/jsonschema-core.txt)

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -4,6 +4,7 @@
 <!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
 <!ENTITY RFC5988 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5988.xml">
 <!ENTITY RFC6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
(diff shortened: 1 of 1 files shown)


handrews:
Nice!  Since you recommend that profile and describedBy have the same URI for simplicity, is it worth calling out any sort of reason why they would not be similar?  If so, a one-liner might be something like:

It may be preferable for `describedBy` link to use a different URI if clients are likely to run on a network that cannot access the identifying profile URI.

I do not feel at all strongly about this- it's fine and possibly less confusing as-is.

awwright:
Good point. Maybe something like... hmm, lemme think.

Maybe:

> Since "profile" is opaque and doesn't necessarily need to point to a network location, a separate link relation is used for linking to a downloadable schema.  However, for simplicity, schema authors should make these URIs point to the same resource when possible.

handrews:
Yes, that works.  It doesn't explain why someone might use something other than a network location (or use a non-accessible network location), but that's a potentially large topic and therefore probably outside of the scope of this page.

awwright:
I used the following language:

>  The profile URI is opaque and SHOULD NOT automatically be dereferenced. If the implementation does not understand the semantics of the provided profile, the implementation can instead follow the "describedby" links, if any, which may provide information on how to handle the profile.
> Since "profile" doesn't necessarily point to a network location, the "describedby" relation is used for linking to a downloadable schema.
> However, for simplicity, schema authors should make these URIs point to the same resource when possible.

PR #80, opened 2016-10-11
Updated the readme

Relequestual:
to better explain what master in this repo is and is not

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 This repository contains the current, and historical, JSON Schema
 specifications.
 
+It should be noted that the master branch of this repo does not constitute a currently released version.
+For draft 3 or 4, please see <https://tools.ietf.org/html/draft-zyp-json-schema-03> and <https://tools.ietf.org/html/draft-zyp-json-schema-04> respectivly
 
 ## Call for reviews
 


awwright:
Even though we just had this conversation I'm a little bit unclear as to what this is clarifying...?

Or maybe that's because you just clarified it for me.

In any event, this repository doesn't contain built .txt files for anything, and it should contain the meta-schemas for every released version (which it does now, I forgot I only `touch`ed the files when I created this).

Relequestual:
The landing branch of a repo, generally, is expected to be where you can find the latest stable release. This is the standard that many repos use, with development work going on in other branches. The clarification avoids the ambiguity where some people might come here and look at the files in root, believing that they are related draft 4.

awwright:
Perhaps mention instead that the latest releases are all found in <archive/> or in the list of Internet-Drafts.

I don't think it really makes sense for this repository to say there's a "stable" branch.

Relequestual:
Right. I'll add that they can be found in the archive directory, and link to the I-Ds

I didn't say there's a stable branch, just that master shouldn't be considered as such (as is normally expected for any repo)

README.md
```diff
@@ -3,6 +3,8 @@
 This repository contains the current, and historical, JSON Schema
 specifications.
 
+It should be noted that the master branch of this repo does not constitute a currently released version.
+For draft 3 or 4, please see <https://tools.ietf.org/html/draft-zyp-json-schema-03> and <https://tools.ietf.org/html/draft-zyp-json-schema-04> respectivly
```
handrews: Should we instead refer folks to this repository's tags to find released versions in the change history, and the web site to find links to the current drafts?  This PR is already out of date and it would be nice to only need to update the web site and not this README as well when a new draft is released.

Relequestual: Maybe we could do both? I mean, personally, if I'm looking for the latest version, I'd rather be directed to the draft document on ietf, so I think this is fine. Plus I'm not sure we have tags for 3 and 4 in this repo.
(Probably should add draft 5, right?)

handrews:
> Plus I'm not sure we have tags for 3 and 4 in this repo.

We don't have tags, and the old spec files are in the web site repo which seems wrong.
But yeah. just adding draft 5 would be fine with me here if you'd rather do that.

Relequestual:
@handrews so you're suggesting we now also link to the draft 5 tagged release on github in the readme? (I think that's a good idea, but confirming that's what you mean)

handrews:
@Relequestual yes, although that was before I learned that @awwright views the schema versions completely differently than I do.  The archive folders only have meta-schemas, and there is no draft 05 meta-schema and there never will be, so really I have no idea.  Do whatever.

Relequestual:
@awwright Any opinions on this?

awwright:
@Relequestual I guess I'm still not sure what's intended, all files in this repository are drafts in the same way the I-D publication is.

handrews:
@awwright No, the I-D publications are published.  The stuff committed to "master" here is unpublished- it's just the current working state, between published drafts.

Relequestual:
I have directly modified the README to better reflect what is actually in this repo. Please raise issue if you dissagree with the change @awwright @handrews, and feel free to immediatly revert if so.
I forgot this PR was outstanding, and discussion has stalled anyway.

As such, I'm going to close this issue.

Issue #103, opened 2016-10-19
additionalProperties behavior for non-object instances


handrews:
This is a pretty subtle wording issue, and easily fixed, if we agree on the desirabilities of the underlying principles.

The wording for validating the additionalProperties keyword against an object (independent of its children) has changed from the following in [v4 (fge-00)](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.4.4):

```
Successful validation of an object instance against these three
keywords depends on the value of "additionalProperties":

if its value is boolean true or a schema, validation succeeds;

if its value is boolean false, the algorithm to determine
validation success is described below.
```

To this in [v5 (wright-00)](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.18):

```
The value of "additionalProperties" MUST be a boolean or a schema.

If "additionalProperties" is absent, it may be considered present
with an empty schema as a value.

If "additionalProperties" is true, validation always succeeds.

If "additionalProperties" is false, validation succeeds only if the
instance is an object and all properties on the instance were covered
by "properties" and/or "patternProperties".

If "additionalProperties" is an object, validate the value as a
schema to all of the properties that weren't validated by
"properties" nor "patternProperties".
```

Oddly, the wording of "additionalItems" has not been changed to have this problem- it still preserves the behavior it had in v4.

The old wording had the following properties, which are lost in the new wording:
- Whether or not the object itself validated could be calculated independent of the validation outcomes of its children.
- The validation effects of `properties`, `patternProperties`, and `additionalProperties` with respect to the object itself were independent:
  - `properties` did not affect validation - as an object validation keyword, it always succeeds
  - `patternProperties` did not affect validation - as an object validation keyword, it always succeeds
  - `additionalProperties` always fails for the value `false` and always succeeds otherwise, no matter the value of the other two keywords

This language change significantly affects the theoretical model and the principles discussed in #55 as it is no longer possible to return a validation result of the object without calculating the validation results of its children.

I previously complained about the confusion of separating child validation from object validation but _I was completely wrong to do so_.  Once I started working through a lot of schema algebra in an effort to solve the reuse-with-additionalProperties-false problem, the value of the separation became very apparent.

The correct solution is to more clearly correlate the object and child validation without tying them together.  The v4 wording read strangely because it appeared to completely ignore child values, rather than explaining how and why object (and array) validation is independent of child validation, and fit it into the overall framework of many independent validation decisions which are only combined after their evaluation.

[...]

Issue #101, opened 2016-10-17
Allow true and false for all schemas (except maybe the root schema)


handrews:
Currently, `additionalItems` and `additionalProperties` allow the booleans `true` and `false` as values, with the following equivalences:

`true` is equivalent to `{}`
`false` is equivalent to `{"not": {}}`

These are helpful because they more clearly indicate the intent of "this can't exist" and "this can exist in any form", particularly when it comes to `false`.  They are purely syntactical convenience, but they become **very** convenient when doing more complex schema algebra to work with complicated schema re-use.

We should allow `true` and `false` anywhere a schema is allowed (although requiring an object schema for the root would be reasonable, if we don't want booleans on their own to be legal schema documents).

This will allow expressing things like "all properties except..." and "only these properties with any value..." very clearly.  In particular, `false` is much more immediately understandable than `{"not": {}}`.  Allowing `true` in all schema places also means that we can say `{"not": true}` == `false`.

``` JSON
{
    "properties": {
        "foo": false,
        "bar": false
    }
}
```

``` JSON
{
    "properties": {
        "foo": true,
        "bar": true,
    }
    "additionalProperties": false
}
```

It also makes it a bit more clear to build a schema that forbids additional properties based on two sets of properties combined (I've inlined the constituent schemas instead of using $ref which would be more normal):

``` JSON
{
    "definitions": {
        "firstSet": {
            "properties": {
                 "a": {...},
                 "b": {...},
                 "c": {...},
             }
        },
        "lastSet": {
             "properties": {
                 "x": {...},
                 "y": {...},
                 "z": {...}
             }
        }
    },
    "allOf": [
        {"$ref": "#/definitions/firstSet"},
        {"$ref": "#/definitions/lastSet"},
    ]
    "properties": {
        "a": true,
        "b": true,
        "c": true,
        "x": true,
        "y": true,
        "z": true
    },
    "additionalProperties": false
}
```

This reads a bit more clearly as "allow only these things, which have their real schemas defined in the allOf".

Of course this is just a syntax of convenience, but that's true of the existing `true` and `false` use as well.  It's just as convenient in other places.

Relequestual:
It looks like you're wanting to add a different route to existing functionality. As you say, for convenience. Is that the case? Surely what's needed is better documentation in that case?

handrews:
[...]

PR #111, opened 2016-10-25
Make properties/items wording consistent.

handrews:
This is a fix for #103 , although I went a bit further and made an effort to make all of the object/array property/item keywords consistent.  I expect the wording might need a bit of tweaking even if the general approach is accepted.  Also, I'll happily drop this if a different approach is preferred.

Separate "items" and "additionalItems" for consistency with the separate "properties",
"patternProperties", and "additionalProperties".

Fix wording bug:  non-objects always successfully validate against "additionalProperties".

Arrange the wording for each of these to follow this pattern:
- keyword value type requirements
- value to assume if absent (if relevant)
- validation against the instance (independent of child validation)
- validation of children (possibly first explaining to which children this keyword applies).

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -168,6 +168,20 @@
                 </t>
             </section>
 
+            <section title="Validation of primitive types and child values">
+                <t>
+                    Two of the primitive types, array and object, allow for child values.  The validation of
+                    the primitive type is considered separately from the validation of child instances.
+                </t>
+                <t>
+                    For arrays, primitive type validation consists of validating restrictions on length.
+                </t>
+                <t>
+                    For objects, primitive type validation consists of validating restrictions on the presence
+                    or absence of property names.
+                </t>
+            </section>
+
             <section title="Missing keywords">
                 <t>
                     Validation keywords that are missing never restrict validation.
@@ -306,7 +320,8 @@
                     If absent, it can be considered present with an empty schema.
                 </t>
                 <t>
-                    Validation of this keyword against the instance always succeeds.
+                    This keyword controls child instance validation.  Validation of the
+                    primitive instance type against this keyword always succeeds.
                 </t>
                 <t>
                     If "items" is a schema, child validation succeeds if all elements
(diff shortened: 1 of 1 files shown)


handrews:
@awwright @Relequestual any feedback?  I have the use-boolean-values-for-child-schema wording ready to add but it builds on this and I don't want to combine them in case this wording is rejected.

jsonschema-validation.xml
```diff
@@ -297,31 +297,59 @@
...
+                    Each object MUST be a valid JSON Schema.
                 </t>
                 <t>
-                    The value of "items" MUST be either a schema or array of schemas.
+                    If absent, it can be considered present with an empty schema.
                 </t>
                 <t>
-                    Successful validation of an array instance with regards to these two
-                    keywords is determined as follows:
+                    Validation of this keyword against the instance always succeeds.
```
Relequestual: I don't really understand this sentence.

jsonschema-validation.xml
```diff
@@ -415,6 +443,14 @@
                 <t>
                     If absent, it can be considered the same as an empty object.
                 </t>
+                <t>
+                    Validation of this keyword against the instance always succeeds.
```
Relequestual: Again I don't understand this =/

jsonschema-validation.xml
handrews: That's because it's weird :-P

While working out schema algebra rules for the whole `"additionalProperties": false"` thing, it became clear that it's very useful to have separate notions of whether the immediate instance validates against the keyword and whether any children validate.

For an array, immediate instance validation is just about about the array length.

So in this case, the "items" keyword cannot place restrictions on the length, so the array (as an array, independent of its elements) cannot fail validation based on this keyword.  Even if "items" is itself an array, it does not on its own restrict the length of the array- that is only done by "minItems" and "maxItems".

Even `"additionalItems": false` does not impose length restrictions, it just ensures that child validation fails for all elements after a certain point.  Actually, I think the wording can be improved around that now that I've tried to explain it.

jsonschema-validation.xml
handrews: For an object, immediate instance validation is just about what property names are or are not present (as determined by `required`, `minProperties`, `maxProperties`).

`properties` tells us things about certain property names, but does not restrict the set of property names.

Although now that I think about it, technically even `"additionalProperties": false` does not restrict the set of property names, it just ensures that all child values for names not otherwise accounted for fail their validation.

Hmm... that may lead to a better phrasing.  Let me think about it and I'll submit an update.

handrews:
@Relequestual I've tried a slightly different approach.  I feel like this makes things much more clear by eliminating the most convoluted language around additionalItems and additionalProperties, but we'll see if it's clear to anyone else :-)

This also gets rid of the need to talk about "true" and "false" in special ways, as the behavior follows cleanly from how they are applied to child instance values.

I'm happy to keep reworking this as needed, as it's been a persistent source of confusion and complexity.

Relequestual:
I'm going to merge this as I don't see it makes any changes, just makes a few bits clearer!

Relequestual:
Curious. Any idea why this commit doesn't come up in the commits at https://github.com/json-schema-org/json-schema-spec/graphs/contributors ? I don't appear, and you only have one commit

handrews:
I think those graphs aren't quite real-time.

jsonschema-validation.xml
```diff
@@ -427,27 +463,39 @@
...
+                    the child instance for that name successfully validates against each
+                    schema that corresponds to a matching regular expression.
+                </t>
             </section>
 
             <section title="additionalProperties">
                 <t>
-                    The value of "additionalProperties" MUST be a boolean or a schema.
+                    The value of "additionalProperties" MUST be a boolean or an
+                    object.  If it is an object, the object MUST be a valid JSON Schema.
```
awwright: This has the effect of only permitting schemas that happen to be objects; it's entirely possible schemas might also take the form of a boolean, and in earlier drafts they could be a string, too.

awwright: Ok. Iirc I was trying to make it consistent the other way, saying "If it's an object, then it's a schema" seems redundant when a schema is already _defined_ to be an object. I'll consider making that change if you don't see any problems with that.

handrews: I was about to put up a pull request for #101 which updates all of this a little bit by no longer requiring the boolean case to be specially explained in two places.

I think my wording for that accomplishes your goal, and if not, definitely feel free to fix it in another way.   I'll have the PR up later today.

Issue #69, opened 2016-10-02
Allow empty "required" array


gibson042:
(copied from https://github.com/json-schema/json-schema/issues/232 )

Draft v4 specifies that a `required` array "MUST have at least one element", which makes it impossible to explicitly indicate that all properties are optional (and therefore also impossible to distinguish intentional optionality from accidental omission of `required`). Empty values (`"required": []`) should be permitted.

erosb:
> which makes it impossible to explicitly indicate that all properties are optional

I don't think it is a problem. Being able to express the same thing in multiple ways (an explicit and an implicit notation) is not a goal of the specification.

gibson042:
Then why is there an `integer` type, to point out just one from a _multitude_ of counterexamples? And why not make `required` required, so author intention is always clear?

Put another way, the default value of `minItems` is 0. Why deviate from that default for `required`?

erosb:
> And why not make required required, so author intention is always clear?
That should be a different proposal.

gibson042:
I was pointing out the inconsistency in claiming that JSON Schema is good about preventing multiple ways of expressing the same thing, not seriously suggesting requiring `required`. _This_ (allowing `required: []`) is the serious proposal, because an empty value conveys information that is not possible if the property is absent (and in fact I've begun defining schemas with `_required: []` as a workaround).

awwright:
I would personally try to leave out a keyword if it's not going to have any effect on validation (since keywords generally add constraints to an instance), but not being overly pedantic about these sorts of things seems pretty reasonable.

There might be a good reason to want to leave in a keyword even if it's not going to do anything.

handrews:
I agree with @awwright about trying to leave it out, but as long as `required` doesn't become required (no keyword should be required, as {} should always be valid and validate everything) I like the idea of allowing `"required": []`

I think mostly we allow putting in the default explicitly, so `required` is currently an exception to that pattern.

Relequestual:
If it says in the spec, that if `required` is not included, then all are optional, then I don't see that there's a problem.

epoberezkin:
No objection here to allowing `required: []`, it's backward compatible :)...

awwright:
Er, as I mentioned in #112, it turns out "dependencies" has this behavior too.

I'd like to just modify "stringArray" so both "dependencies" properties and "required" allows multiple items. They're both used for the same purpose, we should probably modify this behavior together.

handrews:
[...]

PR #112, opened 2016-10-28
Allow an empty array for "required".

handrews:
This addresses the enhancement requested in issue #69.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -21,7 +21,6 @@
         "stringArray": {
             "type": "array",
             "items": { "type": "string" },
-            "minItems": 1,
             "uniqueItems": true
         }
     },
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -435,9 +435,8 @@
 
             <section title="required">
                 <t>
-                    The value of this keyword MUST be an array. This array MUST have at
-                    least one element. Elements of this array MUST be strings, and MUST be
-                    unique.
+                    The value of this keyword MUST be an array. Elements of this array,
+                    if any, MUST be strings, and MUST be unique.
                 </t>
                 <t>
                     An object instance is valid against this keyword if its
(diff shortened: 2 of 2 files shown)


handrews:
Paging everyone on issue #69:  @awwright @Relequestual @gibson042 @epoberezkin @erosb

This seemed uncontroversial by the time the discussion stopped.  Any three out of the five of you willing to give approval so this can get merged?

epoberezkin:
I agree. But why don't we allow array in dependencies to be empty too in this case? For consistency sake?

handrews:
> But why don't we allow array in dependencies to be empty too in this case? For consistency sake?

Feel free to make another pull request, or I can make one after this is accepted, but please don't block this one because it doesn't do more stuff.  The issue was about allowing an empty "required" array.  We should still do this even if someone else objects about dependencies.

gibson042:
> Paging everyone on issue #69: @awwright @Relequestual @gibson042 @epoberezkin @erosb

I doubt my approval matters, but +1.

erosb:
Okey, +1

schema.json
```diff
@@ -117,7 +119,7 @@
             "additionalProperties": {
                 "anyOf": [
                     { "$ref": "#" },
-                    { "$ref": "#/definitions/stringArray" }
+                    { "$ref": "#/definitions/nonEmptyStringArray" }
```
awwright: Why is "dependencies" being modified here?

handrews: Search and replace error, good catch. (although some want this done, I'd rather keep this focused on requirements unless there's immediate agreement to do it for dependencies as well- otherwise I'll fix this to just be for requirements as intended).

Relequestual: I think this change makes sense.

awwright: We could also consider if both `dependencies` and `required` should allow any number of items.

They effectively do the same thing, after all.

awwright:
I'd like to just modify "stringArray" so both "dependencies" properties and "required" allows multiple items. They're both used for the same purpose, we should probably modify this behavior together.

handrews:
@awwright @epoberezkin dependencies now also allows for an empty array, which makes all string array definitions consistent.

Relequestual:
Should the wording make clear that the array may be an empty array?

handrews:
> Should the wording make clear that the array may be an empty array?

Done, for both keywords.

Issue #63, opened 2016-09-19
v6 validation: "contains"


handrews:
Originally written by @geraintluff at https://github.com/json-schema/json-schema/wiki/contains-(v5-proposal)
### Proposed keywords
- `contains`

We also might want an equivalent for objects (like `containsProperty`).
### Purpose

Specifying that an array must contain _at least one_ matching item is awkward.  It can currently be done, but only using some inside-out syntax:

``` json
{
    "type": "array",
    "not": {
        "items": {
            "not": {... whatever ...}
        }
    }
}
```

This would replace it with the much neater:

``` json
{
    "type": "array",
    "contains": {... whatever ...}
}
```

It would also enable us to specify multiple schemas that must be matched by _distinct_ items (which is currently not supported).
### Values

The value of `contains` would be either a schema, or an array of schemas.
### Validation

If the value of `contains` is a schema, then validation would only succeed if at least one of the items in the array matches the provided sub-schema.

If the value of `contains` is an array, then validation would only succeed if it is possible to map each sub-schema in `contains` to a _distinct_ array item matching that sub-schema.  Two sub-schemas in `contains` cannot be mapped to the same array index.
#### Plain schema

``` json
{
    "type": "array",
    "contains": {
        "type": "string"
    }
}
```

Valid: `["foo"]`, `[5, null, "foo"]`  
Invalid: `[]`, `[5, null]`
#### Array of schemas

``` json
{
    "type": "array",
    "items": {"type": "object"},
    "contains": [
        {"required": ["propA"]},
        {"required": ["propB"]}
    ]
}
```

Valid:
- `[{"propA": true}, {"propB": true}]`
- `[{"propA": true}, {"propA": true, "propB": true}]`

Invalid:
- `[]`
- `[{"propA": true}]` - no match for second entry
- `[{"propA": true, "propB": true}]` - entries in `contains` must describe different items
##### Implementation

The plain-schema case is simple.

The array case is equivalent to Hall's Marriage Theorem.  There are [relatively efficient](http://www.mcs.csueastbay.edu/~simon/handouts/4245/hall.html) solutions for the general problem - but, I suspect a brute-force search will be surprisingly effective and efficient (due to the relatively small number of entries in `contains`).

It may or may not be worth warning schema authors about stuffing hundreds of entries into `contains`, because a naive implementation could easily end up having O(n<sup>3</sup>m) complexity.
##### Complexity of understanding (for humans)

[...]

Issue #32, opened 2016-05-26
Add `contains` keyword


BigBlueHat:
Just surfacing this proposal from https://github.com/json-schema/json-schema/wiki/contains-%28v5-proposal%29

It would greatly simplify JSON Schemas such as this one:
https://github.com/w3c/web-annotation-tests/blob/master/common/contextValue.json#L15-L26

It has been implemented in [ajv](https://github.com/epoberezkin/ajv).

/cc @awwright @epoberezkin

handrews:
@BigBlueHat , @awwright : I just realized I re-filed this as issue #63 as I did not think anyone had migrated it over.  Since I copied the full text of the proposal in that issue, can I request that we close this one and keep #63 ?  Or I could copy stuff over here.

handrews:
Done!  Please close unless we missed something (I don't have the permissions to close this).

Relequestual:
Closed as merged.

PR #113, opened 2016-10-28
Add the "contains" keyword for array instances.

handrews:
This addresses the enhancement requested in issues #32 and #63.
Only the single-schema form from #63 is added here as the multi-schema form
did not gather significant support in the absence of a clear use case.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -367,6 +367,17 @@
                 </t>
             </section>
 
+            <section title="contains">
+                <t>
+                    The value of this keyword MUST be an object.  This object MUST be
+                    a valid JSON Schema.
+                </t>
+                <t>
+                    An array instance is valid against "contains" if at least one of
+                    its elements is valid against the given schema.
+                </t>
+            </section>
+
             <section title="maxProperties">
                 <t>
                     The value of this keyword MUST be an integer. This integer MUST be
--- a/schema.json
+++ b/schema.json
@@ -87,6 +87,7 @@
             "type": "boolean",
             "default": false
         },
+        "contains": { "$ref": "#" },
         "maxProperties": { "$ref": "#/definitions/positiveInteger" },
         "minProperties": { "$ref": "#/definitions/positiveIntegerDefault0" },
         "required": { "$ref": "#/definitions/stringArray" },


jsonschema-validation.xml
```diff
@@ -367,6 +367,17 @@
...
 
+            <section title="contains">
+                <t>
+                    The value of this keyword MUST be an object.  This object MUST be
+                    a valid JSON Schema.
+                </t>
+                <t>
+                    An array instance is valid against "contains" if at least one of
+                    its elements is valid against the given schema.
+                </t>
```
epoberezkin: or array is empty

handrews: The "or array is empty" is easily handled by:

```JSON
{
    "type": "array",
    "items": {...},
    "oneOf": [
        {"contains": {...}},
        {"maxItems": 0}
    ]
}
```

epoberezkin: It follows from contain being equivalent to:

```json
{
    "type": "array",
    "not": {
        "items": {
            "not": {... whatever ...}
        }
    }
}
```

epoberezkin: or it doesn't? :)

epoberezkin: yes, ok, should have at least one item

handrews: No, an empty array fails that schema, because an empty array passes the schema `"items": {"not": {...}}`, so the outer `not` causes it to fail overall.

handrews:
@awwright @Relequestual I think all outstanding concerns here are resolved, any chance of getting this merged?

Relequestual:
I'm in favour of saying, with 3 change approvals, merging is OK to go.

handrews:
> I'm in favour of saying, with 3 change approvals, merging is OK to go.

Do you mean I need to get two more formal approvals?  @awwright , @epoberezkin ?

Relequestual:
Actually, given the "official" people involved, that could be a bit tricky. Can non members submit reviews?

handrews:
@Relequestual : Yup, as you probably noticed because I just tried it out by commenting on your PR :-)

jdesrosiers:
Let's do this.  JSON Schema can do this functionality without `contains`, but it's ugly, verbose, and difficult to work with.  I think `contains` is very welcome spoonful of syntactic sugar.

Relequestual:
OK. if a contributor reviews and agrees then I'll merge!

handrews:
@epoberezkin , @awwright any chance I can convince either of you to be approver #3?

awwright:
I'd normally wait 14 days but this seems relatively uncontroversial

handrews:
Thanks @awwright .  I had figured since this was was a pretty direct implementation of something that had gotten approval on the issue, that the 14 day counter included the period of agreement on the issue.

If you want pull requests to always stay open for 14 days that's good to know.  I'd like to try to get the requisite approval comments as soon as possible but I'm fine with leaving things open for a minimum period.  We should document that somewhere, perhaps a "contributors" section of the site?  Or a wiki page in this repo?

Relequestual:
I would personally like to see this info (on what will happen and how) in the repo readme. Even if JUST "we will wait 14 days before merging, if we like it"

handrews:
@Relequestual I filed issue #137 to collect stuff like "we will wait 14 days" before merging.  I'm tired of getting criticized for not following rules that aren't written down :-(

awwright:
I was explaining what I do to help manage expectations, it's not really a
rule for people to follow.

handrews:
> I was explaining what I do to help manage expectations, it's not really a
rule for people to follow.

Whether it's a "rule" or some other word is not the point.  The point is that it's not documented, and there are a lot of these things that are not documented.  Tripping over them when you suddenly say that something I've been doing is counter-productive or pointless or trying to move too fast (and really I'm not trying to rush anything) is very frustrating.  I could have just not spent time on the thing you don't want people to work on or aren't going to look at anytime remotely soon.  And I don't want to be pestering people when there's a minimum deadline that hasn't been met.  Etc. etc. etc.

Issue #122, opened 2016-11-01
id and $schema for meta-schemas still says "draft-04"


handrews:
Should this be changed to "draft-05", "draft-wright-00", "v5", or something else?

awwright:
The latest draft doesn't change the meta-schema being used.

handrews:
It should have.

Regular meta-schema:
* Added `format` (bug fix)
* Should have changed several formats from "uri" to the newly added "uriref"

meta-hyper-schema:
* removed `fragmentResolution`
* added `base`
* changed `method` such that its description is no longer valid and needs updating

awwright:
The explicit purpose of the latest I-D, before anything else, was to keep the existing meta-schema.

Perhaps there's some changes that could have been avoided, or withheld for a new meta-schema, but I'm not really sure how. All the changes made were in an effort to update references, clarify existing behavior when it contradicted other specifications (and declare the specifications as being the "correct" one), or otherwise settle fundamental problems that prevented development from moving forward; for the most part none of this impacts how an implementation is coded (except for a small handful of new features, and the "rel" magic values that to the best of my knowledge literally nobody implemented).

In any event, an I-D can't be changed, instead it'll be replaced in its entirety when a new publication is released, just like how the latest release replaced the previous I-Ds.

If we _really_ want a new schema to describe the one new "base" property (plus "format" bugfix), I suppose we can do that. But I don't really see any need for it.

handrews:
I'm sorry, @awwright , but none of that makes any sense to me.  We're going for rough consensus and working code, and working code should include meta-schemas that correspond to the actual specification.

No matter how many times you claim that it was just a correction, you removed two keywords, added a keyword, and made an incompatible change to yet another keyword (I know you consider your  change to `method` to be compatible, but as far as I can tell no one else agrees with you on that and the description in the meta-schema is absolutely not compatible with your new definition).

Your tolerance for confusingly contradictory publications is much higher than many of the rest of us.  To me, the meta-schemas and published drafts *must* be kept in sync for potential implementors to easily be able to follow along with what we are doing.

While I can easily ignore meta-schemas or hack up my own, why should we make everyone do that?  It costs us practically nothing to keep the meta-schemas in sync with the drafts, and that's what most people expects to find.

awwright:
[...]

PR #127, opened 2016-11-04
Update meta-schemas for draft-wright-00

handrews:
Addresses issue #122  and (sort of) json-schema-org/json-schema-org.github.io#38

I went ahead and made this a pull request so that we can talk about something concrete.
This is a branch from the draft-wright-json-schema-00 to avoid pulling in other work.  We'd presumably also merge this into master.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -1,9 +1,9 @@
 {
-    "$schema": "http://json-schema.org/draft-04/hyper-schema#",
-    "id": "http://json-schema.org/draft-04/hyper-schema#",
+    "$schema": "http://json-schema.org/draft-wright-00/hyper-schema#",
+    "id": "http://json-schema.org/draft-wright-00/hyper-schema#",
     "title": "JSON Hyper-Schema",
     "allOf": [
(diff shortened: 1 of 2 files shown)


awwright:
I suppose one name is as arbitrary as any other, at least to a computer; but for us humans I've been thinking the next meta-schema will be named `http://json-schema.org/draft-06/schema#`

As seen here, I don't see any problem with skipping numbers if it represents an I-D release that didn't change the meta-schema. If there's a good argument we could just call it draft-05, but that was already used to refer to the latest release.

handrews:
Totally on board with draft-06.

Do we want to release this as a new draft of just a meta-schema update, or should I drop this and open a new pull request rebased out to the current master and also include the seem-likely-to-be-approved "contains", "examples", and required-can-be-empty changes?  I think all of that together is quite useful.

That would be a new draft with matching meta-schema and a bit of updated content.

handrews:
It's clear that this pull request is not doing anything allowable, so I'm closing it.  I will re-submit the relevant changes on top of the current master using a draft-06 number.

PR #128, opened 2016-11-04
Allow boolean subschemas everywhere.

handrews:
This implements issue #101

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -189,7 +189,7 @@
                 <t>
                     A JSON Schema document, or simply a schema, is a JSON document used to describe an instance.
                     A schema is itself interpreted as an instance.
-                    A JSON schema MUST be an object.
+                    A JSON Schema document MUST consist of an object, known as the root schema.
                 </t>
                 <t>
                     Properties that are used to describe the instance are called keywords, or schema keywords.
@@ -231,6 +231,31 @@
                     In this example document, the schema titled "array item" is a subschema,
                     and the schema titled "root" is the root schema.
                 </t>
+                <t>
+                    A subschema MUST be an object or a boolean, where boolean values are equivalent to object schemas as follows:
+                </t>
+                <t>
+                    true:
+                </t>
+                <figure>
+                    <artwork>
+<![CDATA[
+{}
+]]>
+                    </artwork>
+                </figure>
+                <t>
+                    false:
+                </t>
+                <figure>
+                    <artwork>
+<![CDATA[
+{
+    "not": {}
+}
+]]>
+                    </artwork>
+                </figure>
             </section>
 
         </section>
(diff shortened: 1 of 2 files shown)


handrews:
Added a commit that puts JSON Reference in the meta-schema.  This addition does not change any specified behavior at all, it just further brings the meta-schema in line with the latest draft.  I added it here because it was easier than having to merge two different subschema fixes.

handrews:
@awwright @Relequestual any concerns holding this up?

epoberezkin:
lgtm

as a side comment - we are on the right path of differentiating schemas and subschemas :)

hyper-schema.json
```diff
@@ -69,9 +59,17 @@
         }
     },
     "definitions": {
+        "subSchema": {
+            "anyOf": [
+                {"$ref": "#"},
+                {"type": "boolean"},
+                {"$ref": "http://json-schema.org/draft/hyper-schema#/definitions/jsonReference"}
```
epoberezkin: I think it should be http://json-schema.org/draft/schema#/definitions/jsonReference

handrews: Good catch, thanks!  Fixed.

schema.json
```diff
@@ -3,10 +3,24 @@
     "$schema": "http://json-schema.org/draft/schema#",
     "description": "Core schema meta-schema",
     "definitions": {
+        "jsonReference": {
```
CroniD: Shouldn't the property "$ref" be required?

handrews: Yeah, I'll fix that, thanks!

handrews: Fixed.

handrews:
@awwright @Relequestual @epoberezkin 

I have taken the meta-schema changes out of this and will re-add once PR #168 is accepted, at which point the changes will be simpler.

I will do the same with PR #167 (the version of this that includes root schemas), which I definitely prefer to this version now.  So if there's any consensus on that please comment so I can close this version.

handrews:
@awwright if you're pretty sure that #167 is the better approach (and at this point I prefer it) I can close this one out.

jsonschema-core.xml
```diff
@@ -189,7 +189,7 @@
                 <t>
                     A JSON Schema document, or simply a schema, is a JSON document used to describe an instance.
                     A schema is itself interpreted as an instance.
-                    A JSON schema MUST be an object.
+                    A JSON Schema document MUST consist of an object, known as the root schema.
                 </t>
```
epoberezkin: This language is not correct. "Consists of" means "has parts". Previous "must be" is correct.

jsonschema-validation.xml
```diff
@@ -535,12 +532,11 @@
                 </t>
                 <t>
                     This keyword's value MUST be an object. Each property specifies a dependency.
-                    Each dependency value MUST be an object or an array.
+                    Each dependency value MUST be an array or a valid subschema.
                 </t>
```
epoberezkin: I would add "array of strings"?

jsonschema-core.xml
handrews: I think it's fixed in #167, which currently has more support than this PR.  If that changes I will backport @awwright's language fixes to this PR.

awwright:
#167 was used instead.

Issue #52, opened 2016-09-16
v6 hyper-schema: extended templating (vars+template)


handrews:
Originally written by @geraintluff at https://github.com/json-schema/json-schema/wiki/Extended-templating-syntax-(v5-proposal)
### Proposed keywords

Uses existing keyword - proposes extension of [`href`](href) in [LDOs](Link-description-object).

_(Should this syntax also be allowed inside `"$ref"` to allow templating of references?)_
### Purpose

Currently, the only values available for templating in `href` are the object itself, and the immediate children of the object (which must be referred to by their exact name).

This proposed new syntax would allow more powerful templating, specifying values from the data using [Relative JSON Pointer](http://tools.ietf.org/html/draft-luff-relative-json-pointer-00).

It would also allow the re-naming of template variables.  This is useful because in some templates, the variable name is actually included in the results, e.g.
- `/prefix/{?foo,bar,baz}` -> `/prefix/?foo=1&bar=2&baz=3`
### Values

In addition to the existing string values/behaviour for `href`, the following is proposed:

The value of `href` may be an object, containing the following properties:
- `template` - containing a URI template
- `vars` - an object, where all the values are [Relative JSON Pointers](http://tools.ietf.org/html/draft-luff-relative-json-pointer-00)
### Behaviour

To obtain the URI for the link, the URI template in `template` is expanded.  When a variable is referenced by the template, its value is obtained like so:
- if the variable name is a defined property in `vars`, then:
  - the corresponding value in `vars` is interpreted as a Relative JSON Pointer, and resolved relative to the current data instance.
  - the result of resolving the relative pointer is used as the value for the variable in the template
- otherwise, the variable name is percent-decoded and taken as the name of an immediate property.

(Note the complete lack of pre-processing rules - they are not needed here, due to the expressive power of Relative JSON Pointers.)
### Example

Data:

``` json
{
    "author": {"id": 241, "name": "Jenny"},
    ...
}
```

Schema:

``` json
{
    "links": [
        {
            "rel": "author",
            "href": {
                "template": "/users/{authorId}",
                "vars": {
                    "authorId": "0/author/id"
                }
            }
        }
    ]
}
```
##### Mismatched pre-processing rules

This syntax is in many ways _much simpler_ than the existing syntax, because there is no need for escaping rules.  (The current syntax does pre-processing using `(`, `)` and `$`.)

[...]

PR #129, opened 2016-11-04
Replace template preprocessing with re-mapping.

handrews:
This change attempts to minimally address the awkwardness and
shortcomings of the current URI template preprocessing approach.

In addition to its dubious aesthetics, preprocessing does not
help with using values other than the current instance or an
immediate property (for an object instance) or element (for an
array instance).

Instead, use an object to map the variable names to locations in
the instance with relative JSON pointers.  This neatly solves both
the UTF-8/illegal variable name problem and the complex data
structure problem with the same mechanism.

This is close to what is proposed in #52, but that proposal includes
several other related features which are understandably more
controversial.  I think this is actually closer to the original
proposal that preceded #52 in the old repo.

The change from `vars` to `hrefVars` was inspired by the terminology
being used in the JSON Home project.  While compatibility with JSON
Home is not a goal, I like how this clarifies exactly what the keyword
is intended to do.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -50,6 +50,11 @@
             "description": "URI Template resolved as for the 'href' keyword in the Link Description Object.  The resulting URI Reference is resolved against the current URI base and sets the new URI base for URI references within the instance.",
             "type": "string"
         },
+        "baseVars": {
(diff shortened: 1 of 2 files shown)


handrews:
Even if we go this route, I suspect we'll need to re-do this based on #179 so I'm just going to close it out for now to reduce confusion.

PR #132, opened 2016-11-05
Update meta-schemas for changes in the last draft.

handrews:
These changes were made for draft 05 but as that was intentionally published
with no new meta-schema, we are going straight to draft 06 with the
meta-schema identifiers.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -46,12 +46,13 @@
         "oneOf": {"$ref": "#/definitions/schemaArray"},
         "not": { "$ref": "#" },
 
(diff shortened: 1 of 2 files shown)


handrews:
@Relequestual @awwright there's nothing new here, just catching the meta-schemas up to the spec.  Unless there's a problem with how I've implemented it, can we get this merged?  There's other stuff that would interact with this so I'd like to go ahead and get it out of the way.

Relequestual:
As you said, there's nothing new here, just bringing it in line.

Issue #97, opened 2016-10-15
Hyper-Schema examples: wrong, confusing, or changed?


jdesrosiers:
This example is either wrong, confusing, or significantly changes the semantics of Hyper-Schema.  All of the examples look like this, so I'm concerned.

> For example, if a schema is defined:
> 
> ``` json
>  {
>     "links": [{
>         "rel": "self",
>         "href": "{id}"
>    }, {
>         "rel": "up",
>         "href": "{upId}"
>     }]
> }
> ```
> 
> And if a collection of instance resources were retrieved with JSON
> representation:
> 
> GET /Resource/
> 
> ``` json
> [{
>     "id": "thing",
>     "upId": "parent"
> }, {
>     "id": "thing2",
>     "upId": "parent"
> }]
> ```

**Wrong**
This Hyper-Schema and JSON Document do not match up in a meaningful way.  The links apply to the array, not the items in the list.  An array has no property "id" or "upId", so neither of these links will apply.

**Confusing**
Maybe you intended that the Hyper-Schema was a segment of the full Hyper-Schema describing this resource and that the LDO applied to each item in the list.  In that case, the example is confusing and misleading.

**Changed**
The other option is that you changed the semantics of how LDO's work in a way that is not well documented.

A Hyper-Schema that correctly describes the JSON Document in the example would look like this.

``` json
{
    "type": "array",
    "items": {
        "links": [{
            "rel": "self",
            "href": "{id}"
        }, {
            "rel": "up",
            "href": "{upId}"
        }]
    }
}
```

handrews:
Or just make the example have two instances, each of which fits the current schema.  Instead of stuffing them in an array.

awwright:
@jdesrosiers Yeah, it looks like this needs a lot of cleanup. Unfortunately Hyper-schema didn't get nearly as much love as "core" did.

The only interpretation that makes sense is it's providing an array of instances, each can be described by the schema.

The ["base" keyword](https://tools.ietf.org/html/draft-wright-json-schema-hyperschema-00#section-4.1) has the most up-to-date example in the document, does that form sense you you?

jdesrosiers:
Yes, that one looks right.

Relequestual:
If you think the PR doesn't "fix" this issue, please reopen.

PR #138, opened 2016-11-08
Fix parent link example.

handrews:
This addresses issue #97, where it was noted that "The only
interpretation that makes sense is it's providing an array
of instances, each can be described by the schema."

This makes the schema an array schema, and replaces the "self"
link with the more correct (or at least less confusing in this
context) "item" link.

Also fix a "the the" typo in the security considerations section,
and make "hyper-schema" consistent.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -514,16 +514,19 @@
                 </t>
 
                 <figure>
-                    <preamble>For example, if a schema is defined:</preamble>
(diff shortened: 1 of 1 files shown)


Relequestual:
Giving those who were involved in https://github.com/json-schema-org/json-schema-spec/issues/97 a few days to object, but it seems pretty clear cut to me.

Relequestual:
Actually, I'm happy with this change, and don't think it requires further feedback.

Issue #58, opened 2016-09-18
v6 validation: "constant"


handrews:
Originally written by @geraintluff at https://github.com/json-schema/json-schema/wiki/constant-(v5-proposal)
### Proposed keywords
- `constant`
### Purpose

For ordinary use, this would be equivalent to a single-valued `enum`, simply tidier.

The only real difference comes in its behaviour with `$data`-substitution.  `$data`-subsitution would be allowed for this keyword, which means that this keyword is _not_ capable of specifying literal values of the form: `{"$data":...}`, because these would be interpreted as `$data`-substitutions.

However, literal values of this form can still be specified using `enum`, so there is no loss of functionality.
### Values

The value of this keyword would be any value - however, it would be subject to `$data`-substitution.
### Validation

Instances are only valid if they are exactly equal to the value of this keyword.
#### Simple constant

``` json
{
    "type": "object",
    "properties": {
        "five": {
            "constant": 5
        }
    }
}
```

Valid: `{}`, `{"five": 5}`
Invalid: `{"five": 0}`, `{"five": "5"}`
#### Using `$data` to specify equality

``` json
{
    "type": "object",
    "properties": {
        "a": {"type": "string"},
        "b": {
            "constant": {"$data": "1/a"}
        }
    },
    "required": ["a", "b"]
}
```

Valid: `{"a": "foo", "b": "foo"}`, `{"a": "bar", "b": "bar"}`
Invalid: `{"a": "foo", "b": "bar"}`
#### Similarity to `enum`

Unless `$data` is being used, the same effect can be obtained using fewer actual characters:
- `{"constant":"whatever"}` - 23 characters
- `{"enum":["whatever"]}` - 21 characters

However, when used in combination with `$data`, it opens up possibilities that are not otherwise available.

handrews:
I personally find `constant` to be useful independent of `$data` as it more clearly expresses the intent of the schema author to require a fixed value, and not an enum that (at the moment) happens to only have one entry.

HotelDon:
Personally, I get the most use out of constant when combining it with $data, but you bring up an excellent point I hadn't considered with regards to ambiguity when you use enum. While I'm still pushing for $data, I think constant could stand on its own if it really needed to.

sceutre:
Maybe "const" instead of "constant".  Seems more natural/widely spread as a keyword, and trounces the "enum" form in character count ;-)

epoberezkin:
[...]

PR #139, opened 2016-11-08
Add "const" keyword for constant values.

handrews:
This addresses issue #58, which got quite a few upvotes and has
a "draft 6" milestone.

I used the shorter "const" that several people suggested rather
than "constant", but obviously that's easy to change if the longer
form is preferred.

The wording was pretty much directly lifted from "enum".

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -546,6 +546,16 @@
                 </t>
             </section>
 
+            <section title="const">
+                <t>
+                    The value of this keyword MAY be of any type, including null.
+                </t>
+                <t>
+                    An instance validates successfully against this keyword if its value is
+                    equal to the value of the keyword.
+                </t>
+            </section>
+
             <section title="type">
                 <t>
                     The value of this keyword MUST be either a string or an array. If it is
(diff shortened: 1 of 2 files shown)


awwright:
I was going to merge this right now but I just wondered if there's any demand for the short form.

Most keywords seem to prefer the verbose language (suggesting we use "constant") and @epoberezkin already seems to use "constant" too.

handrews:
@awwright one objection had been that `"constant": "foo"` was actually more characters than `"enum": ["foo"]` and "contst" was suggested as a shorter alternative, to which no one seemed to object.

I'm perfectly happy either way.  Let me know if you want me to change it, or feel free to change it yourself before accepting the merge (I think I flagged the thing that lets you do a further edit, and if all you do is change "const" to "constant" that doesn't really need input from me).  Also happy to do the edit myself.

awwright:
I realized, I guess "enum" is short for "enumeration".

And yeah, it looks like "const" would be more popular. This looks good then.

epoberezkin:
`const` is good :)

Issue #55, opened 2016-09-17
v5 validation: Clearly document validation principles


handrews:
**NOTE:**  This is a request for clarification in v5, and is **not** a proposal for changed behavior.
### The Problem

There are several underlying principles to validation which are currently poorly articulated, or even just implied.  Some of the more contentious arguments over feature proposals are due to unclear understanding of these principles.  Plainly stating these in the specification will help keep the evolution of JSON Schema focused and reduce feature debate noise.
### Terminology:  _indexing_ into a schema

You can index into JSON data by a property name or an array index.  This can be written in JavaScript access form, e.g.  A["foo"], A.foo, or A[0].

_Indexing_ into a schema by a property name or array index number will, within this issue, mean finding the schema that would validate a similarly indexed instance.  So if schema X validates instance A, then:

X.foo is the schema that is used to validate A.foo in the course of validating A with X.
X[5] is similarly the schema used to validate A[5]

Note that X.foo will in truth be one of:
X.properties.foo
X.patternProperties.patternThatMatchesFoo
X.additionalProperties   # if neither of the above and additionalProperties is a schema
{}   # the blank schema, if none of the above and additionalProperties is true

Similarly, X[5] will in truth be one of:
X.items[5]    # if items is an array with at least six members
X.additionalItems   # if items is an array with less than six members and addtionalItems is a schema
X.items   # if items is a schema rather than an array
{}   # if none of the above and additionalItems is true

"allOf"/"anyOf"/"oneOf"/"not" involve special considerations, which we will revisit within the principles below.  Here are the basics of how indexing applies to them:

if X is an "allOf" with two branches X1 and X2, then:
X.foo is {"allOf": [X1.foo, X2.foo]}

if X is an "anyOf" or "oneOf" with two branches X1 and X2, then X.foo must only take into account the schema(s) that validated A.  In the case of "anyOf" that may be both or just one, while in the case of "oneOf" it will _always_ be just one of the branches.

If X2 is the branch of "oneOf" that validates A, then X.foo is X2.foo
If both X1 and X2 validate A in an "anyOf", then X.foo is {"anyOf": [X1.foo, X2.foo]}

if X is a "not" schema {"not": Y}, then there is no meaningful index into X.  Depending on the rest of how Y is defined, Y.foo may or may not validate against A.foo, even though Y as a whole is guaranteed to fail validation with A due to the "not".
### Known or Suspected Principles

[...]

PR #143, opened 2016-11-15
Add general validation principles and examples.

handrews:
This addresses issue #55 plus concerns raised in the comments of
issue #101.

I replaced "linearity" with "independence" as I think it is
more general and intuitive.

The general considerations section has been reorganized
to start with the behavior of the empty schema, then explain
keyword independence, and finally cover container vs child
and type applicability, both of which flow directly from
keyword independence.

In draft 04, the wording obscured the connection between
keyword independence and container/child independence.
When we rewrote the array and object keywords to explicitly
classify each keyword as either validating the container
or the child, keyword independence became sufficient to
explain container/child independence.

The list of non-independent keywords has been updated, and
exceptions to the independence of parent and child schemas
have been documented.  Finally, I added a comprehensive example
of the frequently-confusing lack of connection between
type and other keywords.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -158,6 +158,58 @@
 
         <section title="General validation considerations">
 
+            <section title="Constraints and missing keywords">
+                <t>
+                    Each JSON Schema validation keyword adds constraints that
+                    an instance must satisfy in order to successfully validate.
+                </t>
+                <t>
+                    Validation keywords that are missing never restrict validation.
+                    In some cases, this no-op behavior is identical to a keyword that
+                    exists with certain values, and these values are noted where relevant.
+                </t>
+                <figure>
+                    <preamble>
+                        From this principle, it follows that all JSON values
+                        successfully validate against the empty schema:
+                    </preamble>
+                    <artwork>
+<![CDATA[
+{}
+]]>
+                    </artwork>
+                </figure>
+                <figure>
+                    <preamble>
+                        Similarly, it follows that no JSON value successfully
+                        validates against the empty schema's negation:
+                    </preamble>
+                    <artwork>
+<![CDATA[
+{
+    "not": {}
+}
+]]>
+                    </artwork>
+                </figure>
+            </section>
+
+            <section title="Keyword independence">
+                <t>
+                    Validation keywords typically operate independently, without
+                    affecting each other's outcomes.
+                </t>
+                <t>
+                    For schema author convenience, there are some exceptions:
+                    <list>
+	                    <t>"additionalProperties", whose behavior is defined in terms of "properties" and "patternProperties"</t>
+                        <t>"additionalItems", whose behavior is defined in terms of "items"</t>
+                        <t>"minimum" and "maximum", whose behavior may change for a special value of "exclusiveMinimum" and "exclusiveMaximum", respectively</t>
+                    </list>
+                    </list>
+                </t>
+            </section>
+
             <section title="Keywords and instance primitive types">
                 <t>
                     Most validation keywords only limit the range of values within a certain primitive type.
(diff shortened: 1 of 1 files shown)


awwright:
I just made the spelling fixes identified...

A brief description of some of the principles is OK: By identifying patterns to look for, this helps implementors comprehend the document. But I'd like to avoid turning it into a manifesto.

The "Constraints and missing keywords" section is fine, can you split that off into a commit?

"Keyword independence" is fine.

I'm not sure what the "Validation of primitive types and child values" is doing. There's not really a useful pattern here to describe.

"Keyword applicability to instance primitive types" is _really_ lengthy. I'm trying to make observations: "Most validation keywords..." sounds informative, "An important implication of keyword independence..." sounds more like normative text.

I would avoid using "incorrect" examples at all, like the one that lists `"type": "number"` inside a "not" block. Assume people are going to copy and paste examples without context.

handrews:
Thanks for the quick reply!  I will break this up and re-submit the parts you think are fine as-is.

"Validation of primitive types and child values" is currently section 4.2.  It's partially about the "context-free" nature of validation and I think it is important, but only if it's clear.  Stating it up front was originally about making the behavior of various items/properties keywords more clear, but I think when I rewrote all of those a while back I made them clear enough on their own.  You may be right that this is redundant now- I will just take it out and if I can think of anything that really needs to be said, I'll re-submit it separately.

I'll reply to the rest in a bit.

handrews:
For "Keyword applicability to instance primitive types", I added the example because we have had at least two people in the last few weeks file issues or bring up in comments that this principle baffled them, or that they thought it was wrong.  I used the "multipleOf" example in one issue and someone responded that it was really helpful.

I agree that it is long, so compressing it would be good.  This full example (with clear labeling of the "wrong" schema, which isn't actually wrong but just isn't useful) could move to the web site, but I really feel like some sort of example is called for rather than the one-liner that was there before that didn't really show the utility.

I'm trying to explain *why* type applicability is important.  Perhaps just show the right way (the last code block) and explain how it works?  That would be much more concise, and people could go look at the web site for more.

Let me know if you have other thoughts here, otherwise I will make a separate PR for this where we can discuss it further.

handrews:
@awwright see if this version suits you.  I retained the two sections you said were fine, and moved the unchanged section to the end as I think it flows better:

* First describe the general constraint concept and the default empty schema
* Then explain keyword independence, the most general concept
* Then explain type applicability (which actually follows from independence, but I left the wording completely unchanged- I'll come back to it in a separate PR)

handrews:
@awwright I stacked a commit for a much smaller change illustrating the applicability-to-types concept on top of this PR.  That extra commit can be seen as a PR in my fork: https://github.com/handrews/json-schema-spec/pull/1

awwright:
I don't think it's accurate to say it's a 'consequence of the "type" keyword validation operating independently'.

Independence isn't the same thing as saying we try to narrowly limit the range of instances that keywords reject, so that 

Do we have any cases of people misunderstanding the section _after reading it_?


Finally let me take a look at this schema:
```javascript
{
    "type": "number",
    "multipleOf": 1,
    "not": {
        "type": "number",
        "multipleOf": 2
    }
}
```

As far as I can tell, the addition of the second `type:"number"` doesn't do anything. All it's going to do is make the "not" successfully validate in cases already covered by the "type" keyword:

```javascript
null // fails first keyword, since it's not a number
0.5 // fails second keyword, since it's not a multiple of 2
2 // fails third keyword, since it _passes_ the provided schema (instance is both number and multiple of 2)
1 // passes: since it's a number, integer, and fails the "not" schema as expected (instance fails the multipleOf:2 keyword)
```

I might be wrong on this though, I nearly sprained a brain muscle thinking about the "not" logic too hard.

handrews:
> I don't think it's accurate to say it's a 'consequence of the "type" keyword validation operating independently'.

I'd like to get at least one more person's opinion on that.  To me it is exactly what happens and is the primary way that it makes sense to reason about this principle.  The fact that we don't need an extra fundamental principle to explain this is a _good_ thing.  It means the behavior is inherently consistent already.

> Independence isn't the same thing as saying we try to narrowly limit the range of instances that keywords reject, so that we can avoid keywords with overlapping behavior.

???

Independence is exactly the same as avoiding keywords with overlapping behavior.  That's what it means.

Limiting the range of types is one way that this works **but we do not have to add a principle for that**.  It is already true just because of independence.  

@Relequestual can I get a 2nd opinion here?  I think this is super-important.

handrews:
@awwright I removed the schema that confused you in handrews#1 .  And now you've made my head hurt with the "not"... ugh, you might be right about that :-P

I just woke up so I'll think through it again after a cup of tea, and of course if the example does not actually exemplify the problem I will drop it or change it.

Are you still OK with the commits that I kept here in #143 ?  I agree that the commit in handrews#1 is more questionable still.

handrews:
@awwright OK I think I see how you see this differently, in that we could have written into the rules that any value which does not explicitly satisfy the constraint will fail validation, no matter its type.  That technically wouldn't tie it to the validation of "type", although the effect is the same.

I still think there's something here worth noting to make the overall theoretical framework simpler and more consistent, but I can see that my current wording is not necessarily conveying it well.  Hopefully someone can help us converge on better wording (or validate one position or the other- I'm not dead set on it, of course, so if no one else who comments sees this my way I'll be happy to drop it).

handrews:
@awwright after further thought I'm setting aside the applicability change for now unless someone comes along and says they find it useful.

I've taken that out of this pull request entirely- are you OK with the remaining changes?  I'm fine with leaving the PR open for two weeks, I just want to make sure that the PR as it stands now, without the "applicability" change, works for you.

handrews:
@awwright - I deleted the applicability branch from my fork.  This PR is just about the default schema and independence now, so it should be what you want.  If you could confirm that would be great.

Relequestual:
Sorry for the assignment spam @awwright - the request review function seems to be broken right now =[

handrews:
Giving up.

PR #145, opened 2016-11-16
Correct number of primitive types from seven to six.

kw217:
The core spec lists six primitive types in section 4.2 (null, boolean, object, array, number, string), but this sentence claims there are seven. Correct it to match.

I'm not sure if this is the right fix - it's also possible that seven is correct, but the reference to the core spec is wrong. Presumably "integer" is meant to be allowed here too - but the type "integer" is nowhere defined - not in the core spec, or the validation spec.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -552,7 +552,7 @@
                     an array, elements of the array MUST be strings and MUST be unique.
                 </t>
                 <t>
-                    String values MUST be one of the seven primitive types defined by
+                    String values MUST be one of the six primitive types defined by
                     the core specification.
                 </t>
                 <t>


epoberezkin:
`integer` is used in validation spec, so rather than fixing the number it should say `6 primitive types or "integer"`

kw217:
Just adding `or "integer"` isn't enough, because the spec needs to say what `integer` means. I think that belongs in the core spec; after all, the core spec already talks about "integer JSON numbers" (5.3) and has an example that mentions `type: "integer"` (8.2.1).

jsonschema-validation.xml
```diff
@@ -552,7 +552,7 @@
                     an array, elements of the array MUST be strings and MUST be unique.
                 </t>
                 <t>
-                    String values MUST be one of the seven primitive types defined by
+                    String values MUST be one of the six primitive types defined by
```
handrews: Should the

Recall:  "number" includes "integer".

bit on line 560 also be removed?  Since there is no longer an "integer" primitive type, that sentence is a bit confusing.

handrews:
@epoberezkin "integer" was removed from the types in core, so if it's referenced anywhere else in validation that is also a bug and should be cleaned up.  Or we need to restore it to the list in core, but I think @awwright wanted to stay with the actual JSON primitive types instead.

epoberezkin:
What is the motivation to remove "integer" from validation types? When/where was it discussed?

epoberezkin:
@handrews @awwright The reason to have integer as a special validation type is that integers are more commonly used than any other number. Without "integer" you would have to use `{"type": "number", "multipleOf": 1}` which is very weird.

handrews:
@epoberezkin - @awwright removed it at some point, presumably before I was here as I don't recall any issue or discussion.

I'm in favor of "integer" as it is an extremely common case and annoying to do with multipleOf and/or a $ref.  I just had not gotten around to asking what happened to it.

If we are getting rid of integer, it's still used in an example in the core spec so that should be fixed, too.

epoberezkin:
All such changes will eventually lead to the fragmentation of what the standard means and as the result some other standard will be developed. None of the validators is going to drop the support of integers. And yet the standard will not have it. That's ridiculous even as proposal, leave alone as a change that happened without any discussion.

handrews:
@epoberezkin I'm with you on this

awwright:
I think the idea is there's six primitive types, as far as parsing data is concerned; however in the "type" keyword you can also list "integer" which is a subset of "number" (any number with a zero fractional part).

awwright:
@kw217 Does #147 work?

ramsey:
The removal of "integer" was a point of confusion for me. Can a line be added to the change log to note that "integer" was removed?

epoberezkin:
@ramsey integer was removed by mistake, #147 fixes it, this PR should be closed

ramsey:
Thanks for the clarification.

awwright:
#147 describes the intended behavior so I'll close this out. If there's any further problems please feel free to open a new issue!

Issue #146, opened 2016-11-16
"integer" type removed?


handrews:
As noted in PR #145 the "integer" type was removed in v5.

In issue #79 from last month, @awwright was still talking about "integer" as a type.  Neither @epoberezkin nor I recall any discussion of removing it.  @awwright , @Relequestual what was the rationale here?

Requiring the use of `{"type": "number", "multipleOf": 1}` for a very common use case seems burdensome.

awwright:
This was an oversight, see #145 or #147.

The change was supposed to resolve that `type:"integer"` definitively includes both `1.0` and `1.1e1` as "integers" for validation purposes. But it seems the "validation" spec got left out of this clarification.

Relequestual:
Yeah, removing it seems a bit much.

-sidebar
 I think a lot of people are feeling that rushing draft 5 was more dentrimental than any benifit (which I'm not sure anyone was really clear on either)
/-


The issue at hand. I think we need to consider what has actually happened here.
What was said about integer in draft 4?
What was removed and why?
What were the implications?

As far as I can tell, at they very least, it made draft 4 documents not forward compatible with draft 5 documents (which I think was originally an aim of moving forward with drafts?)

awwright:
At the end of the day, it's a draft. Its entire purpose is for collecting feedback just like this.

I don't see any reason this should break reverse compatibility. Originally, these were all defined in the "type" keyword. Later the validation keywords were moved out, but the seven type definitions were left in defined as "primitive" types, though it doesn't make much sense to keep "integer" in there if we're talking about how to interpret JSON, which makes no such distinction.

handrews:
@awwright:  @Relequestual and I have discussed your response and have quite a few concerns.  Most of the words below are mine, but I think I am capturing the main points that concern both of us (and I’m sure Ben can comment if I misrepresented something).

> At the end of the day, it's a draft. Its entire purpose is for collecting feedback just like this.

Just because it’s a draft as far as the IETF is concerned, doesn’t mean it can’t or won’t be used in production.  As such, we need to recognized the impact of drafts and behave appropriately.

It makes a difference that the draft was rushed out with many bugs and surprises.  At least four of us have expressed concern about how draft 5 was rushed out ( @Relequestual and I here, and two others in #50 , for instance), and #96 shows yet another person having been caught off guard by the lack of time to review the hyper-schema changes.  That’s nearly all of the most active participants in the JSON Schema project.

[...]

PR #147, opened 2016-11-17
Validation: Fix reference to seven primitive types

awwright:
Alternate fix for #145, #146


awwright:
I was not reading this close enough, I copied "true" and "false" instead of "array" and "object". Fixed in d751685.

Issue #142, opened 2016-11-14
Eliminate template pre-processing


handrews:
_**NOTE:** I'm splitting this out from #52 in order to get more attention on this specific point and the related pull request #129 .  I'll probably split other stuff out from #52 and then close it.  In hindsight, it was not a good idea to dump everything together there._

URI Template pre-processing is confusing, significantly complicates implementation, and does not address all limitations in the current rules for filling out templates with instance data.

The current approach tries to circumvent the URI Template spec's variable name limitations within the actual string used to express the URI Template.  A simpler yet more powerful approach is to make the URI Template strings normal URI Templates, and use a mapping object to translate legal URI Template names to expressions that can identify any part of the instance.  The proposed keyword for the map is `hrefVars`.

[Relative JSON Pointers](https://tools.ietf.org/html/draft-luff-relative-json-pointer-00) come closest to meeting the necessary requirements:  starting from any point in the instance (specifically, the point from which the LDO including the template is defined), they can identify nearly any other point in the instance (see #115 for limitations with respect to arrays).  See #126 for a discussion of whether Relative JSON Pointers should be a separate I-D or should start as part of JSON Schema.  Either would work fine for this proposal.

To preserve the current behavior when preprocessing is not needed, if a template variable "x" does not appear in `hrefVars`, it may be considered present with a relative pointer of "0/x".

### Examples based on current pre-processing features

Here is a subset of the table of examples for pre-processing, followed by a schema showing links using these variable names.

| Input | Output |
| --- | --- |
| "{(escape space)}" | "{escape%20space}" |
| "{(a (b)))}" | "{a%20%28b%29} |
| "{()}" | "{%65mpty} |
| "{+$*}" | "{+%73elf*} |

```JSON
{
    "links": [
        {
            "rel": "foo",
            "href": "/{(escape space)}/{(a (b))}/{()}"
        },
        {
            "rel": "bar",
            "href": "/{+$*}"
        }
    ]
}
```

Note that making use of $ in the "self" case with the "bar" link requires using the URI Template "+" operator to allow percent-encoded sequences, as the $ is replaced by a percent-encoded sequence during pre-processing.  This is particularly confusing since, without pre-processing, the "+" would make the "$" a literal dollar sign that did **not** need to be percent-encoded.

Given this instance:

```JSON
{
    "escape space": "x",
    "a (b)": "y",
    "": "z"
}
```

the "foo" link would expand to "/x/y/z"

[...]

Issue #109, opened 2016-10-25
format:  add uritemplate and jsonpointer


handrews:
We currently have uri and uriref for indicating an expectation of RFC 3986 conformance.  While I don't think we should have formats for all RFC-defined things, since hyper-schema itself relies on URI Templates ([RFC 6570](https://tools.ietf.org/html/rfc6570)) and JSON Pointer ([RFC 6901](https://tools.ietf.org/html/rfc6901)), it seems reasonable to include them.  It would make our meta-schema more clear as well.

handrews:
@Relequestual @awwright @epoberezkin any thoughts?  This should be a pretty straightforward yes or no, I think.  Just a question of "do we want enough defined formats to be able to identify the formats on which JSON Schema itself depends?"

erosb:
+1 for `jsonpointer` , I'm neutral about `uritemplate` .

epoberezkin:
@erosb same here

handrews:
@epoberezkin @erosb why less interested in uritemplate?  It is a distinct format from uri and urireference and one on which we are entirely dependent for hyper schema (currently in two different keywords, href and base).  Being able to express our own usage seems beneficial, particularly given the optional nature of format.

Relequestual:
@handrews Could you give examples of both? I'm not familiar with either, and therefore how this would look, which makes it difficult to know if I'm for or against.

handrews:
@Relequestual uritemplate is what we use for `href` and `base`.  jsonpointer is what we use with `$ref`.  So in the meta-schema:

```JSON
{
    "type": "object",
    "properties": {
        ...,
        "href": {
            "type": "string",
            "format": "uritemplate"
        },
        "base": {
            "type": "string",
            "format": "uritemplate"
        },
        ...
    },
    "definitions": {
        ...,
        "$ref": {
            "type": "string",
            "format": "jsonpointer"
        }
    }
}
```

handrews:
@Relequestual currently we don't have anything where we *only* use JSON Pointer, but JSON Pointer is the fragment syntax we use in `$ref`.  Some proposals use JSON Pointers directly instead of as part of URIs.

handrews:
Here is a better use of JSON Pointer.  This is a (probably not quite right) schema for the [JSON Patch](https://tools.ietf.org/html/rfc6902) media type:

```JSON
{
    "type": "array",
    "items": {
        "type": "object",
        "required": ["op", "path"],
        "properties": {
[...]

Issue #141, opened 2016-11-10
Proposal: "jsonpointer" type


erosb:
# Introduction

While the JSON Schema specification itself utilizes JSON Pointers, it has only a very basic support for letting schema authors specify a schema which expects JSON documents to use such JSON Pointers. Currently, when a schema author wants to describe a JSON Pointer, the only thing he/she can do is defining the schema with `"type":"string"` and probably adding a regex restriction which mandates the string to be a syntactically valid json pointer.

In practical usecases it is useful to be able to describe the schema of the referred JSON value. This proposal targets these usecases.

# Purpose

The purpose of this new type is to let schema authors expect that the document being validated contains a [JSON Pointer](https://tools.ietf.org/html/rfc6901), and this pointer denotes a value which conforms to the restrictions of a schema defined by the schema author.

# Syntax

Example schema:

```json
{
    "properties": {
        "ptrToNum": {
            "type": "jsonpointer",
            "referredSchema": {
                "type": "number"
            }
        }
    }
}
```

Example *valid* document against the above schema:
```json
{
    "ptrToNum" : "#/settings/width",
    "settings" : {
        "width" : 4
    }
}
```

Example *invalid* document against the above schema:
```json
{
    "ptrToNum" : "#/settings/width",
    "settings" : {
        "width" : [ 4 ]
    }
}
```

# Validation

If a schema instance has a `"type" : "jsonpointer"` property , then it must also have a `"referredSchema"` key and a  then an implementation should perform the following validation steps:
 * it should validate if the value is valid against the `{"type" : "string"}` schema
 * it should check if the string value is a valid JSON Pointer as specified by RFC 6901 (question: [string representation](https://tools.ietf.org/html/rfc6901#section-5) or [URI fragment identifier representation](https://tools.ietf.org/html/rfc6901#section-6) or both?)
 * it should [evaluate the JSON Pointer](https://tools.ietf.org/html/rfc6901#section-4). The result of the evaluation is a JSON value. If the evaluation fails then the validation is unsuccessful
 * it should validate the evaluated JSON value against the schema given by `"referredSchema"`

# Notes

 * I'm not sure about the `"referredSchema"` keyword , maybe something more expressive can be better
 * while writing the above proposal I didn't take the wording of the current spec into account so probably most of it must be re-phrased. I'm also not sure if it should go into the core or the validation spec.
 * we should specify what should be done by the implementation if it runs into infinite recursion while resolving JSON pointers
[...]

PR #150, opened 2016-11-20
"uritemplate" and "jsonpointer" formats

handrews:
Addresses issue #109 by adding "uritemplate" and "jsonpointer",
as defined by their RFCs, both of which already play roles
within JSON (Hyper-)Schema.

We use "uritemplate" in hyper-schema, but until the preprocessing
step is removed we technically can't put the format in the
meta-hyperschema.  This is being addressed elsewhere
(#142 and #129 propose one possible approach).

While we do not need "jsonpointer" for our meta-schema, it is
used for schema+json URI fragments, and in other JSON-based
media types such as JSON Patch.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -7,6 +7,8 @@
 <!ENTITY RFC3339 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3339.xml">
 <!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
 <!ENTITY RFC5322 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5322.xml">
+<!ENTITY RFC6570 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6570.xml">
+<!ENTITY RFC6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
 <!ENTITY RFC7159 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7159.xml">
 ]>
 <?rfc toc="yes"?>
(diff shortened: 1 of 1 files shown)


jsonschema-validation.xml
```diff
@@ -817,6 +819,25 @@
...
+                    </t>
+                </section>
+
+                <section title="jsonpointer">
+                    <t>
+                        This attribute applies to string instances.
+                    </t>
+                    <t>
+                        A string instance is valid against this attribute if it is a valid JSON Pointer,
+                        according to <xref target="RFC6901"/>
```
erosb: At this point it would be necessary to clarify if the JSON Pointer should be in its [string representation](https://tools.ietf.org/html/rfc6901#section-5) or [URI fragment identifier representation](https://tools.ietf.org/html/rfc6901#section-6) or is both permitted. This problem came up in #141

handrews:
No, RFC 6901 does **not** define a representation *of* a JSON Pointer *as* a URI fragment.  It defines how to encode a JSON Pointer *in* a URI fragment.  (specifically UTF8 + percent encoding).

That is just a URI fragment, and the interpretation of the fragment as a JSON Pointer is controlled by the media type of the document (which is always how fragments are interpreted).

If you want to use a URI with a JSON Pointer fragment, that is just a "uri" or "uriref" format.  It is **not**, by itself, a JSON Pointer.

epoberezkin:
My suggestion is to make it json-pointer and uri-template - to be consistent with the current name (e.g. date-time) and previous proposals about this format.

Also I would add relative-json-pointer (although we had a long conversation @handrews whether relative includes absolute, which I think it does not). Although it is a separate proposal probably...

handrews:
I was being consistent with "uriref"

handrews:
Relative JSON Pointer does not currently actually exist anywhere in the spec, so
1.  If it ever goes in, we can define it however we want and are not limited by the previous draft proposal
2.  If it ever goes in, we can add it to format then, and it will be more clear how that should look based on how we choose to use it in the spec.  If at all.

epoberezkin:
ok

awwright:
While we're going to be proliferating the number of "format"s we have I suppose this is fine, between JSON Patch and JSON Schema both of these formats are used...

But I can't help but think if there's a better way to avoid defining every possible "format" we could ever need.

handrews:
@awwright I agree.  There are quite a few open format-related issues, and I think we should pick either draft 7 or draft 8 to take a look at the whole thing (that wouldn't be the only thing in whichever draft, I'm just saying when we tackle format we should try to do all of the format issues together as they are kinda interrelated).

Issue #137, opened 2016-11-08
Need a contributor's guide


handrews:
As this project attracts more people now that it is active again, we could use some guidance on contributing.  I've personally stepped on @awwright 's toes about five times already and I'm sure it's getting just as tiresome for him as it is for me.  Things that should go in a contributor's guide:

* How do we resolve issues?
    * How do we indicate that one is accepted, and can move to PR?
    * How do we indicate that one is rejected
    * How do we resolve a conflict- it's clear that there are irreconcilable camps (e.g. $merge/$patch)
    * Closing "orphaned" issues after a "last call" period (two weeks?)'
    * Should implementors implement proposals to try them out, or should we wait until the issue is accepted to some degree?

* Pull requests
    * Any specific requirements?
    * Document that three approvals are required for (new keywords? changes to behavior?)
    * Who is eligible to give approvals?
    * Document that all PRs other than trivial bug fixes will be left open a minimum of 2 weeks
    * Should spec and meta-schema changes always be in the same PR? (I say yes)

* Release process
    * Are some things in-scope vs out-of-scope for the next draft?  How do we tell?
    * Is there any way for the community to escalate something in priority?  What are good reasons to do so?
    * How do we know when a draft is about to be published?
    * What, if any, additional rules exist during the final pre-publication period?

There's probably more, but these are things that have already come up.

Relequestual:
I'll hopefully be able to find time to properly reply to this issue next week!

awwright:
There's few hard-and-fast rules in the IETF model. We believe in rough consensus and running code.

Contributors should come from a diverse background with a variety of use cases. But everyone should try to be thinking about everyone's needs when they propose a change.

More formal implementations and comments should generally be made against the released I-D. At the same time, developers should feel free to work on prototypes and test proposals in running code.

Any issue can be revisited if there's a new reason to.

We use a few tools like GitHub, there's no special requirements, just use it like you would. To the extent GitHub offers a lot of different ways to use it, we're using it for consensus building. Approvals, emojis, and the like are to get rough ideas of the collective mood, and to make sure nobody has been forgotten.

I think the HTTPbis WG in particular should serve as a role model. They've made effective use of their website and GitHub in addition to the traditional mailing list, and have been very careful in writing drafts to avoid locking themselves into a particular design.

---

Anything else is really just a personal (and fairly arbitrary) process. Other people might do it differently, and that's totally valid.

[...]

PR #154, opened 2016-11-21
Change "id" to "$id", retain "id" as deprecated.

handrews:
This addresses issue #20.

"$id" matches "$schema" and "$ref", establishing a clear naming
pattern for all keywords defined in the core specification.  This
also reduces confusion in the very common case where objects
described by the schema have an "id" property.

The current plan is to continue to allow for "id" until a migration
tool can be provided.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -37,7 +37,12 @@
     },
     "type": ["object", "boolean"],
     "properties": {
+        "$id": {
+            "type": "string",
+            "format": "uriref"
+        },
         "id": {
+            "description": "This keyword has been deprecated in favor of \"$id\"",
             "type": "string",
             "format": "uriref"
         },
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -347,23 +347,23 @@
                 </t>
             </section>
 
-            <section title='The "id" keyword'>
+            <section title='The "$id" keyword'>
                 <t>
-                    The "id" keyword defines a URI for the schema,
+                    The "$id" keyword defines a URI for the schema,
                     and the base URI that other URI references within the schema are resolved against.
-                    The "id" keyword itself is resolved against the base URI that the object as a whole appears in.
+                    The "$id" keyword itself is resolved against the base URI that the object as a whole appears in.
                 </t>
                 <t>
                     If present, the value for this keyword MUST be a string, and MUST represent a valid <xref target="RFC3986">URI-reference</xref>.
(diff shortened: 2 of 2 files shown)


handrews:
Note:  If we end up going in the direction of #155 (which I posted to illustrate the approach- it's not a final proposal) then this PR becomes irrelevant.

epoberezkin:
I suggest not including it into v6, see https://github.com/json-schema-org/json-schema-spec/issues/137#issuecomment-262443996

handrews:
@epoberezkin having read #160 I see no reason to keep this small (and for v6, backwards-compatible) change out of v6.  #160 would be a radical departure whether we have "id" or "$id", so let's start being consistent about core keyword naming ("$schema", "$id", "$ref").  There is already at least one implementation that uses "$id" instead of "id", and at least one other that can't handle instance properties called "id" because of the naming overlap- that's a bug, but indicative of the problems of the current name.

Plus, we already managed to get broad agreement on this rename.

schema.json
```diff
@@ -29,7 +29,12 @@
     },
     "type": "object",
     "properties": {
+        "$id": {
+            "type": "string",
+            "format": "uriref"
+        },
         "id": {
+            "description": "This keyword has been deprecated in favor of \"$id\"",
```
epoberezkin: Maybe just drop it?

awwright: That's probably a good idea, though the patch does include a comment saying it'll likely get removed in the future.

And of course, just because it's removed doesn't mean that implementations can't still support it (i.e. silently upgrade it).

Lemme think about this some, unless you can elaborate more

handrews: I don't have strong feelings on this.  The deprecation option came in response to migration concerns, although as you note there are several easy options for handling the change.  If y'all decide to drop `id` entirely I'll be happy to update the PR.

epoberezkin: I think dropping from spec and optionally supporting in implementations addresses migration issues. I'd rather we keep it simple here, so new implementations don't have to carry this baggage...

handrews: @awwright @epoberezkin I've submitted PR #212 as a version of this that drops `"id"` entirely.

handrews:
See also #212 for an alternative that does not retain `"id"`.

handrews:
Since @awwright and @epoberezkin have both approved #212, I'm going to close this to reduce noise.
@Relequestual if you end up objecting to #212 I can either re-open this or re-start a new PR.

Issue #135, opened 2016-11-08
"additionalItems" should not be honored if "items" is an object.


SereneAnt:
As per [json-schema specification](http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.9):
> if "items" is not present, or its value is an object, validation of the instance always succeeds, regardless of the value of "additionalItems";

Currently (v 4.0.x), CollectionConstraint makes an extra checks for "additionalItems" if "items" is object and  fails validation.

SereneAnt:
Wrong repo.

awwright:
@SereneAnt Just a heads up that language is a tad bit confusing, the language seems to only apply for "items", "additionalItems" should still fail even if "items" isn't defined. If in doubt check the behavior against the [official test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite). If it's all too confusing, please let us know.

epoberezkin:
@awwright:

> "additionalItems" should still be able to fail even if "items" isn't defined

there is no test for such scenario in the test suite you are referring to. The draft says in clear terms that additionalItems are only taken into account when items is array.

handrews:
> The draft says in clear terms that additionalItems are only taken into account when items is array.

Also important is this line in "items":

> If absent, it can be considered present with an empty schema.

This means that "items" is effectively always defined, so "additionalItems" cannot be applied unless "items" is a schema.  The default empty schema value of "items" takes precedence.

epoberezkin:
> "additionalItems" cannot be applied unless "items" is a schema.

I'd say, array of schemas. If it is a schema, additionaltems is ignored.

handrews:
> I'd say, array of schemas. If it is a schema, additionaltems is ignored.

Oops, that was a typo, I'll edit it.

PR #156, opened 2016-11-21
Clarify "items"/"additionalItems" interaction.

handrews:
In #135 the question of whether additionalItems could fail validation
if items is not defined was raised.  Since items defines a default
value of {}, it is never undefined, which is clarified in this change.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -350,6 +350,11 @@
                     if every instance element at a position greater than the size
                     of "items" validates against "additionalItems".
                 </t>
+                <t>
+                    Otherwise, "additionalItems" MUST be ignored, as the "items"
+                    schema (possibly the default value of an empty schema) is
+                    applied to all elements.
+                </t>
             </section>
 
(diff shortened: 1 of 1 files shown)


awwright:
This is an improvement, but I'm thinking the entire section could just be rewritten to something like:

> If the instance is an array, then "additionalItems" is invalid if any item in the instance, that hasn't been validated with "items", is invalid

handrews:
From an implementation perspective, I am not tracking which things have and haven't been validated.  With additionalItems, I just look at the comparative lengths of items and the instance.  That's the rationale for the wording that's there right now.  I can therefore validate "additionalItems" without validating "items" (which is only important b/c I am implementing each keyword as a separate pure function and trying to make minimal use of other keywords in each function).

To me, your alternate wording suggests a different implementation approach that is not as useful of a design to me.

epoberezkin:
@handrews just to remind, additionalItems can be schema, if items is an array. I guess you mean it anyway.

Issue #108, opened 2016-10-24
Submitting both user and instance data via both the uri and message body


handrews:
[Edited for changes made between #159 and #179]
[Edited to replace `hrefVars` with `hrefSchema` based on later discussions]

The purpose of this issue is to explore the limitations of the current draft's usage of `method`, `schema`, and `href` (in terms of resolving template variables).  There are ideas here that could become a proposal, but I am not directly proposing anything in the example schema.  I'm really just trying to find a consensus on the limitations.

The main point here is the need to both specify URL parameters and a message body.  The API has events and invitations.  The invitation collection can be filtered from either the event perspective or (via email address) the invitee perspective.  I did not provide a schema for invitees as it made the example too long without adding anything.

Since the query string is as much a part of the identifier as the path components, this API treats `/invitations?event=1234` as the collection of invitations for the given event.  The API supports creating invitations for event 1234 by POSTing to that URI, just the same as it would for a URI of `/events/elements/1234/invitations`.  However, not all of the fields needed to create an invite can be a part of the URI, so a creation also takes a message body providing the remaining parameters.

The current `href` specification does not allow it to define user input.  But if `schema` is used for the body, there is no other mechanism for getting user input into the URI.  This example uses `hrefSchema` as a way to map the URI template parameters to a schema and/or a value from the instance.  This is to help illustrate the two routes of user input, not propose `hrefSchema` as necessarily being a solution.

`hrefSchema` is a schema which treats the template variables as if they were properties of a JSON object.  Each variable property's schema is used to validate "user" input (which may or may not be from an actual human user).  It can use `$data` with `default` to take a default value from the instance when no user input is supplied.  To implement the current behavior of only resolving from the instance, you would use a schema like this:

``` JSON
{
    "properties": {
        "email": {
            "enum": {"$data": "0/email"},
            "default": {"$data": "0/email"}
        }
    }
}
```

For convenience, the above can be specified by just giving the `"0/email"` relative JSON pointer instead of a schema (this is what you will see in the example).  [EDIT: originally `hrefSchema` was `hrefVars` and did not quite use regular schema syntax.  Now that `hrefSchema` is a regular schema, the shortcut might be a bad idea].  The default behavior for template variable `{foo}` is `"hrefSchema": {"properties": {"foo": "0/foo"}}` (which replicates the current behavior).

[...]

Issue #96, opened 2016-10-15
LDO "get" method is too ambiguous


jdesrosiers:
I'm just getting around to reviewing the new Hyper-Schema spec and I have a few issues.  Sorry, it's a bit late.  I'm doing the best I can to keep up with the lightning fast pace this project suddenly started moving at.

> If "method" is "get", the link identifies how a user can compute the
> URI of an arbitrary resource.  For example, how compute a link to a
> page of search results relating to the instance, for a user-selected
> query term.  Despite being named after GET, there is no constraint on
> the method or protocol used to interact with the remote resource.

The problem I have is the ambiguity.  Just like HTML is unambiguous enough for a web browser to generate a UI from HTML, a Hyper-Schema client should be capable of generating a UI (graphical or otherwise) from a hyper-json (JSON Document + Hyper-Schema).  Based on this wording, it seems like you can do almost anything you want.  What is a generic client supposed to generate when it encounters this LDO?

This is a preserving-the-spirit-of-hyper-schema level issue for me.  If I'm missing something, please explain and lets put a better explanation into the next draft of the spec.  If I haven't misunderstood, this change is a deal-breaker.  I can't use Hyper-Schema like this without assigning semantics to things that aren't in the spec.

jdesrosiers:
Same thing goes for "post", although it's not quite as bad.  The only time you send a request body to the server and it is ambiguous what you should be sending, is when you are doing a POST operation.  So, "post" should imply a POST operation.  But, there are too many hoops you have to just through to get to that conclusion.

> If "method" is "post", the link specifies how a user can construct a
> document to submit to the link target for evaluation.

handrews:
I just attempted to do something about this and went down in flames (the issues I opened are already closed and are not recommended reading).  I appreciate @awwright 's patience although I still share a lot of your concerns, @jdesrosiers .

My current understanding is this:

* All of this terminology is by analogy with HTML forms, and therefore incorporates how the HTML form element behavior has evolved as the web has expanded beyond HTTP/1.0, even though HTML5 still has pretty much the same form syntax.

* A link that takes input is a form.  There's nothing more to it than that.

* A "get" link/form encodes its input in the URI.  That's all that "get" means.  It has nothing to do with HTTP GET or any sort of more abstracted retrieval operation.  It does **not** mean that sending to the form is safe.

[...]

PR #159, opened 2016-11-23
Remove "method", URI input with "hrefVars"

handrews:
This implements a first step from the rather complex issue #108.

In the most recent draft, "method" simply controlled whether link
input was placed in the URI (for a value of "get") or in the request
body (for a valued of "post").  While correlating with HTML, this
was both confusing due to "get" and "post" not necessarily indicating
the HTTP methods of the same name, and limiting in that users/clients
could not submit data through both the URI and the request body
at the same time.

This introduces "hrefVars" and "baseVars", which provide schemas for
user input matching the "href" and "base" URI Template variables,
respectively.  It removes "method" and makes "schema" and "encType"
unambiguously apply to the request body in all cases.

Clients should choose when to use the request body submission based
on the rules of the protocol given by the URI scheme and the semantics
indicated by the link relation.

Additionally, the complex and apparently rarely if ever used
preprocessing rules have been removed, and both the resulting and
pre-existing limitations have been documented.  They will be the
subject of work for future drafts.

The meta-schemas have been updated accordingly, and the LDO schema has
been brought over from the web site repo and updated.  Improper
use of "$ref" has been fixed with "allOf".

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -3,7 +3,7 @@
     "id": "http://json-schema.org/draft/hyper-schema#",
     "title": "JSON Hyper-Schema",
     "allOf": [
-        {"$ref": "http://json-schema.org/draft-04/schema#"}
+        {"$ref": "http://json-schema.org/draft/schema#"}
     ],
     "properties": {
         "additionalItems": {
@@ -48,7 +48,8 @@
 
         "base": {
             "description": "URI Template resolved as for the 'href' keyword in the Link Description Object.  The resulting URI Reference is resolved against the current URI base and sets the new URI base for URI references within the instance.",
-            "type": "string"
+            "type": "string",
+            "format": "uritemplate"
         },
         "links": {
             "type": "array",
@@ -80,7 +81,13 @@
             "properties": {
                 "href": {
                     "description": "a URI template, as defined by RFC 6570, with the addition of the $, ( and ) characters for pre-processing",
-                    "type": "string"
+                    "type": "string",
+                    "format": "uritemplate"
+                },
+                "hrefVars": {
+                    "description": "an object where each property is a variable name from \"href\" and the value is a schema describing the data needed to fill in the variable.",
+                    "type": "object",
+                    "additionalProperties": {"$ref": "#"}
                 },
                 "rel": {
                     "description": "relation to the target resource of the link",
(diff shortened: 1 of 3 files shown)


handrews:
This is another PR that I have put up to facilitate comments as we now seem to be driving towards a rapid v6 release.  It implements @awwright 's idea of a first step towards solving issues documented in #108, while sidestepping more complex issues such as relative JSON pointers.  It is an alternative to the more controversial first step in #129, which would be deferred to v7 (if it is accepted at all).

Note that this assumes that PR #150 (uritemplate and jsonpointer formats) will be accepted and merged.

handrews:
This also addresses #96, although only in the sense of removing the confusing behavior rather than fully replacing the v4 "method" usage.

awwright:
The one blocker here is that "base" doesn't accept end-user input for variables - it's always deterministic. So "baseVars" will never support user input, and we can probably avoid introducing the property entirely for this iteration.

Also I think it's pretty clear the intention of URI Templates was to allow "." to identify a property in a map (people who want to identify a literal `.` can use the hex-escaped `%2E`). The behavior is fine as defined, but I would like to be prepared to consider building it out later.

handrews:
@awwright thanks for looking over this so quickly!

I will take "baseVars" out, that seems totally reasonable.

I also agree with considering dot-separated names later, I just wanted to make absolutely sure that people didn't rush into that quite yet.  Let me know if you want me to tweak the wording there, but this is an area I'd want to keep discussing while we review and publish v6.  Hopefully we can work that out and quickly follow v6 up with a v7 that has a full URI Templating solution.

awwright:
One more issue, I'm trying to figure out what impact this has:

>    Template variables from "href" which do not appear as properties can
>    be considered to be present with an empty schema.
>
>   Schema authors can prevent user input for a template variable by
>   setting its schema to false in "hrefVars".  Implementations MUST NOT
>   attempt to validate values resolved from instance data with
>   "hrefVars".

I'd think if you don't want to prompt a user about a value, you just don't include it as a property in "hrefVars". Any variable not explicitly listed (as a user field) in "hrefVars" is going to be data from the instance in some fashion.

handrews:
> I'd think if you don't want to prompt a user about a value, you just don't include it as a property in "hrefVars". Any variable not explicitly listed (as a user field) in "hrefVars" is going to be data from the instance in some fashion.

That's not consistent with the overall validation approach.  In general, if something is missing it is unconstrained.  To constrain it, you add keywords and set them to a value that specifies the constraint.

awwright:
@handrews Link templates aren't doing validation, though. We're defining user fields, not constraining ones that already exist.

handrews:
@awwright we are defining the validation rules for the user fields.  That is why they are associated with schemas.  And in general, the same principles should apply throughout the specifications.  I don't see any reason to flip it here just because this proposal, which involves validation, is not in the validation spec.

handrews:
@awwright @Relequestual I've come up with #179 which is both more powerful and more consistent/obvious than this approach.  It just uses a regular schema applied to the template variables as properties of a JSON object, and does not require special rules.  Just regular use of a validation schema.

handrews:
I'm going to close this one in favor of #179 which is a much better solution.  This PR is referenced from that one so if anyone wants to consider this option they can, and we'll re-open it as necessary.  But I would like the attention to be on #179.

Issue #161, opened 2016-11-23
core: Define how to apply subschemas to child instances


handrews:
In #157 @awwright noted that:

> Technically, "properties" and "items" are just keywords for validation. All they do is return true or false.
> People have been reading in that we can recursively apply the semantics, and that's fine too. Maybe we should make that more explicit.

Without ways to declare which subschemas apply to which child instances, JSON Schema as a _media type_ is very limited.

In theory, it should be possible to use some JSON Schema vocabulary _without_ validation, which is not possible at the moment.  We don't have such a vocabulary right now, but to demonstrate the idea, here is a hyper-schema that imposes no validation constraints:

```JSON
{
    "links": [
        {
            "rel": "self",
            "href": "/foos"
        }
    ],
    "items": {
        "links": [
            {
                "rel": "item",
                "href": "/foos/{id}"
            }
        ]
    }
}
```

This schema uses `items` from the validation vocabulary, but only to apply the "item" link to each element of an array instance.  The URI template implies that the elements of the array are objects with at least an "id" property, but this schema does not validate that at all.  Obviously, this is not the most realistic schema, but we don't have other vocabularies right now and I can't come up with one off the top of my head.

----

I recently rewrote the `items`, `additionalItems`, `properties`, `patternProperties`, and `additionalProperties` to make it clear that they do not directly validate the instance, but instead control which subschemas are applied to which child instances.  While I expressed that in terms of validation, reading back over the language shows that it is not necessary.

We could move those keywords into the core specification and simply reference their impact on validation in the validation specification.  There would not be any changes to the current functionality, but it would allow for much more flexible definitions of new vocabularies.

This is all a bit abstract, and probably hard to grasp from just eyeballing the current validation spec wording.  So I will write up a trial PR to demonstrate how that would look.

deuberger:
I second this issue. Reading over the latest schema at http://json-schema.org/latest/json-schema-validation.html is confusing to me now compared to what was there in early October. It almost seems like it would be valid to have invalid child instances since the specification doesn't require that child instances be validated it seems. Maybe I'm misunderstanding something, but it's definitely confusing to me one way or another.

handrews:
@deuberger I rewrote the sections for items, additionalItems, properties, patternProperties, and additionalProperties within the validation spec in PR #111 , could you take a look and see if that is easier to understand?  You can read it in the diff or check out the current master here and build the HTML form of the docs if you install xml2rfc.

handrews:
[...]

Issue #157, opened 2016-11-21
hyper-schema: LDOs and anyOf/allOf/oneOf/not/dependencies


handrews:
I think it would be a good idea to clarify the use of LDOs with the combinatoric keywords.  I *think* this is what should happen, but I'm not sure:

* An LDO may only be used with an instance if the instance is valid against the schema in which the LDO appears.
    * LDOs from the valid branch of a "oneOf" can be used, but not any of the others
    * LDOs from each of the valid branches of an "anyOf" can be used
    * LDOs from a schema value in "dependencies" can only be used if the property key controlling that schema is present in an object instance

* LDOs within a "not" cannot ever be used (I just can't figure out how using them would make any sense)

* What happens if LDOs in "allOf" branches or in valid "anyOf" branches define links with the same "rel" values?
    * Could just mean multiple links are available
    * What if the resulting URIs are different?  There is no way to tell which is intended, is there?

awwright:
This is my understanding as well.

Links never replace each other, there can be multiple links with the same relationship. The best example is the "stylesheet" relationship. Links can also have multiple "self" "author" "alternate" "tag" and "type" relations, among others.

handrews:
Sounds good.  I'm trying to figure out a situation in which an API client might want to follow a link with a particular relationship and not be able to decide among multiple links with that relationship, but I think I'm fabricating that problem out of nothing.  

Certainly for "stylesheet" and most of the others, if you have multiple, that just means you need to loop over them and look at each of them.

Multiple "self" links are a bit more confusing for me, although I may just be misunderstanding some aspects of "self".   I have only used "self" in a root schema.  Is the "context" which is identified by self the immediate containing schema or is it the schema document?  In other words, does it make sense to have a "self" link in a subschema?

awwright:
One case is if you're creating a front page of blog posts (e.g. list the 20 most recent blog posts in reverse chronological order), you have an array of JSON instances that each have their own URIs.

handrews:
Why would you use "self" instead of "items" in that case?

handrews:
I am assuming that by "array of JSON instances" you mean a single JSON document that validates against a particular schema, and the LDO with the "self" link is in the "items" (or "additionalItems") schema.

awwright:
An "items" link would be from some sort of collection of posts, with a link pointing to each blog post.

The blog post itself would also have a "self" link (to itself) that gives itself a URI.

handrews:
[...]

PR #163, opened 2016-11-23
Define subschemas -> child instances in core

handrews:
This is a trial balloon for #161.  While it seems like a radical change,
aside from using the "$" prefix for the keywords in question, there is no change
to any behavior.

This moves several keywords that control which subschemas apply
to which child instances from validation to core.  The only
actual validation that these keywords defined was that all child
instances need to validate against all applicable subschemas.

This simple child validation principle has been moved up to
the general considerations section of the validation spec.

The moved keywords are now prefixed with "$", which is what we've
discussed doing for all core keywords (Another potential core
keyword would be "$data", for instance, as it would be used in
multiple vocabularies depending on which proposals we accept,
if any.)

As with "id" vs "$id" and the boolean forms of
exclusiveMinimum/Maximum, the old forms are still allowed but
expected to be removed before RFC.  Migrating an old schema
to the new form is trivial, and was discussed before we decided
to rename "id".  It is equally applicable here.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -297,6 +297,127 @@
             </t>
         </section>
 
+        <section title="Applying subschemas to child instances">
+            <t>
+                The keywords in this section are used to control which subschemas
+                apply to child instances.  They are independent of any specific
+                schema vocabulary.  Vocabulary specifications SHOULD define any
+                vocabulary-specific effects of applying subschemas.
+            </t>
+            <section title="$items">
+                <t>
+                    The value of "$items" MUST be either an object or an array of objects.
+                    Each object MUST be a valid subschema.
+                </t>
+                <t>
+                    If absent, it can be considered present with an empty schema.
+                </t>
+                <t>
+                    This keyword applies only to array instances.
+                </t>
+                <t>
+                    If "$items" is a schema, that subschema applies to all elements
+                    in the instance array.
+                </t>
+                <t>
+                    If "$items" is an array of schemas, each subschema applies to
+                    the child instance element at the same position, if any.
+                </t>
+            </section>
+
+            <section title="$additionalItems">
+                <t>
+                    The value of "$additionalItems" MUST be valid subschema.
+                </t>
+                <t>
+                    If absent, it can be considered present with an empty schema.
+                </t>
+                <t>
+                    This keyword applies only to array instances.
+                </t>
+                <t>
+                    If "$items" is an array of schemas, the "$additionalItems"
+                    schema applies to every element of the instance array at
+                    a position greater than the size of "$items".
+                </t>
+                <t>
+                    Otherwise, "$additionalItems" MUST be ignored, as the "$items"
+                    schema (possibly the default value of an empty schema) is
+                    applied to all instance array elements.
+                </t>
+            </section>
+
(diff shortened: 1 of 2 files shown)


awwright:
I'm very skittish about expanding the meaning of core any beyond what's absolutely necessary for an application/schema+json parser to understand - namely, a way to name schemas, a way to reference other schemas, and a way to select a meta-schema.

One of the side effects this has is "allOf" is left behind and becomes harder to use.

I fully expect most meta-schemas to import JSON Schema Validation. Unless we can identify a problem with the approach that's been in use, how about we just formalize it and say "Metadata is applied in the course of validation to any instance successfully validated by the schema."

handrews:
@awwright I noticed the "allOf" problem as well and it bothered me but I couldn't easily figure out what to do with it.  Decided to see if anyone else said anything :-)

> how about we just formalize it and say "Metadata is applied in the course of validation to any instance successfully validated by the schema."

What's the exact definition of "metadata" here?  Does that just mean anything other than validation?

BTW I'm not 100% on board with my own change here, I just felt that I needed to write it up as a diff in order to really wrap my head around it.

I feel like there's something to the subschema application thing that would benefit from being called out more clearly.  The more I've separated (in my head) how these keywords work from how all of the other validation keywords work the more sense they (and associated issues around additionalProperties, etc.) make to me.

I don't have a real conclusion- I'm probably agreeing with you, actually, I just want to poke at the concept a bit more, if not at this approach.

epoberezkin:
I am strongly against this change. The difference between validation and controlling validation is perceptional, this is a change for the sake of change. It does not add any value

handrews:
@epoberezkin 

1.  I seriously doubt this is going in.  Maybe some ideas from it somewhere, but there's no need to get worked up about it.

2.  You keep talking about things as "perceptional".  I don't think you are using the same definition of that word that I am, because in no case has your usage made any sense to me.

epoberezkin:
Just to make the objection clearer. I don't mind changing the language about how items etc work and clarifying that they essentially apply subschemas - it's kind of obvious but improving the language may make it clearer for some users. I am only against changing keywords.

handrews:
> Just to make the objection clearer. I don't mind changing the language about how items etc work and clarifying that they essentially apply subschemas - it's kind of obvious but improving the language may make it clearer for some users. I am only against changing keywords.

Awesome.  If there's anything that I would want to take from this experiment, that would be it.  I'll probably play around with it tomorrow and see if there's any tweak that I think is a real improvement.  Not a requirement for v6, but if I come up with something everyone likes it may go in :-)

What I wanted to do with this was take Austin's comment in #157 to its logical extreme and see where that put us.  I've started labeling pull requests like this "trial balloons" ( #129 is another such trial balloon although I didn't call it that at the time).  I generally don't expect trial balloon PRs to get merged, at least not without substantial work and quite likely not ever.  I just find that making a concrete change to the spec can clarify things that are murky in issue discussions.

Sometimes these are even kind of "devil's advocate" proposals.  Not unlike `$combine` which I mostly wrote up to show how difficult it would be.

handrews:
I think this has gotten as much feedback as it needs.  Language improvements that were discussed are better submitted as a new PR which I will look at eventually.

Issue #77, opened 2016-10-10
Keywords for exclusive minimum/maximum


handrews:
This is somewhat related to the general idea of issue #55 (clearly document validation principles).  One such principle is that for the most part, each schema validation keyword can be evaluated against an instance independent of the presence, absence, or value of any other keyword.  This is a highly desirable property when refactoring or otherwise reasoning about complex schemas.

More formally, for nearly all validation keywords k1, k2, k3... the following holds true:

``` JSON
{
    "k1": "something",
    "k2": {},
    "k3": ["etc."]
}
```

is equivalent to

``` JSON
{
    "allOf": [
        {"k1": "something"},
        {"k2": {}},
        {"k3": ["etc."]}
    ]
}
```

There are only four exceptions:
- `minimum` is controlled by `exclusiveMinimum`'s value
- `maximum` is controlled by `exclusiveMaximum`'s value
- `additionalItems` is controlled by the value of `items`
- `additionalProperties` is controlled by the values of `properties` and `patternProperties`

The issues around `additionalItems` and `additionalProperties` are outside the scope of this issue, but I mention them as they (especially `additionalProperties`) are more obvious examples of the problems this sort of dependency causes.  In particular, the behavior when working with `allOf` can be counter-intuitive.

Getting back to numeric validation, here is how `minimum` and `maximum` behave:

``` JSON
{
    "type": "number",
    "minimum": 0,
    "exclusiveMinimum": true,
    "maximum": 1,
    "exclusiveMaximum": true
}
```

is equivalent to

``` JSON
{
    "allOf": [
        {"type": "number"},
        {
            "minimum": 0,
            "exclusiveMinimum": true
        },
        {
            "maximum": 1,
            "exclusiveMaximum": true
        }
    ]
}
```

but cannot be further separated.  This is particularly vexing since `exclusiveMinimum` and `exclusiveMaximum` could have been defined to take numbers like this:

``` JSON
{
    "type": "number",
    "exclusiveMinimum": 0,
    "exclusiveMaximum": 1
}
```

which now behaves nicely, being equivalent to this:

``` JSON
{
    "allOf": [
        {"type": "number"},
        {"exclusiveMinimum": 0},
        {"exclusiveMaximum": 1}
    ]
}
```

[...]

PR #168, opened 2016-11-29
Add "$ref", numeric exclusive*, default empty string arrays to meta-schemas, standardize meta-schema formats

handrews:
This is the corresponding change to json-schema-org/json-schema-org.github.io#57
Like that PR, is broken into several commits to make the diffs much easier to read.
It also has meta-schema changes for #77 numeric exclusives and #69 defaulting string arrays to empty arrays, which are the last bits to finish those two issues.

The commits in the sequence are:

* First standardize the formatting, title vs description usage, and order of major sections
* Remove descriptions from hyper-schema.json because they are too easy to get out of sync, and we do not have descriptions in schema.json
* Restore the links meta-schema
* Add JSON reference
* Reorganize the schemas to use either a normal object or a json reference, and never allow an object with a "$ref" property to validate as a non-reference.
* Update the exclusive numeric limits and default empty string arrays.

NOTE:  Various outstanding PRs will be reworked to put their meta-schema changes on top of this as soon as it is merged.  In particular this will simplify and clarify both #128 and #167 .

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -2,48 +2,46 @@
     "$schema": "http://json-schema.org/draft/hyper-schema#",
     "id": "http://json-schema.org/draft/hyper-schema#",
     "title": "JSON Hyper-Schema",
-    "allOf": [
-        {"$ref": "http://json-schema.org/draft-04/schema#"}
-    ],
+    "allOf": [ { "$ref": "http://json-schema.org/draft/schema#" } ],
     "properties": {
         "additionalItems": {
             "anyOf": [
-                {"type": "boolean"},
-                {"$ref": "#"}
+                { "type": "boolean" },
+                { "$ref": "#" }
             ]
         },
         "additionalProperties": {
             "anyOf": [
-                {"type": "boolean"},
-                {"$ref": "#"}
+                { "type": "boolean" },
+                { "$ref": "#" }
             ]
         },
         "dependencies": {
             "additionalProperties": {
                 "anyOf": [
-                    {"$ref": "#"},
-                    {"type": "array"}
+                    { "$ref": "#" },
+                    { "type": "array" }
                 ]
             }
         },
         "items": {
             "anyOf": [
-                {"$ref": "#"},
-                {"$ref": "#/definitions/schemaArray"}
+                { "$ref": "#" },
+                { "$ref": "#/definitions/schemaArray" }
             ]
         },
         "definitions": {
-            "additionalProperties": {"$ref": "#"}
+            "additionalProperties": { "$ref": "#" }
         },
         "patternProperties": {
-            "additionalProperties": {"$ref": "#"}
+            "additionalProperties": { "$ref": "#" }
         },
         "properties": {
-            "additionalProperties": {"$ref": "#"}
+            "additionalProperties": { "$ref": "#" }
         },
-        "allOf": {"$ref": "#/definitions/schemaArray"},
-        "anyOf": {"$ref": "#/definitions/schemaArray"},
-        "oneOf": {"$ref": "#/definitions/schemaArray"},
+        "allOf": { "$ref": "#/definitions/schemaArray" },
+        "anyOf": { "$ref": "#/definitions/schemaArray" },
+        "oneOf": { "$ref": "#/definitions/schemaArray" },
         "not": { "$ref": "#" },
 
         "base": {
@@ -52,7 +50,7 @@
         },
         "links": {
             "type": "array",
-            "items": {"$ref": "#/definitions/linkDescription"}
+            "items": { "$ref": "#/definitions/linkDescription" }
         },
         "media": {
             "type": "object",
@@ -66,17 +64,26 @@
                     "type": "string"
                 }
             }
+        },
+        "readOnly": {
+            "description": "If true, indicates that the value of this property is controlled by the server.",
+            "type": "boolean",
+            "default": "false"
         }
     },
     "definitions": {
-        "schemaArray": {
-            "type": "array",
-            "items": {"$ref": "#"}
+        "schemaArray": [
+            "allOf": [
+                { "$ref": "http://json-schema.org/drafts/schema" }
+                {
+                    "items": { "$ref": "#" }
+                }
+            ]
         },
         "linkDescription": {
             "title": "Link Description Object",
             "type": "object",
-            "required": ["href"],
+            "required": [ "href" ],
             "properties": {
                 "href": {
                     "description": "a URI template, as defined by RFC 6570, with the addition of the $, ( and ) characters for pre-processing",
@@ -92,7 +99,7 @@
                 },
                 "targetSchema": {
                     "description": "JSON Schema describing the link target",
-                    "$ref": "#"
+                    "allOf": [ { "$ref": "#" } ]
                 },
                 "mediaType": {
                     "description": "media type (as defined by RFC 2046) describing the link target",
@@ -109,7 +116,7 @@
                 },
                 "schema": {
                     "description": "Schema describing the data to submit along with the request",
-                    "$ref": "#"
+                    "allOf": [ { "$ref": "#" } ]
                 }
             }
         }
@@ -118,10 +125,6 @@
         {
             "rel": "self",
             "href": "{+id}"
-        },
-        {
-            "rel": "full",
-            "href": "{+($ref)}"
         }
     ]
 }
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -2,6 +2,56 @@
     "$schema": "http://json-schema.org/draft/hyper-schema#",
     "id": "http://json-schema.org/draft/hyper-schema#",
     "title": "JSON Hyper-Schema",
+    "definitions": {
+        "schemaArray": [
+            "allOf": [
+                { "$ref": "http://json-schema.org/drafts/schema" }
+                {
+                    "items": { "$ref": "#" }
+                }
+            ]
+        },
+        "linkDescription": {
+            "title": "Link Description Object",
+            "type": "object",
+            "required": [ "href" ],
+            "properties": {
+                "href": {
+                    "description": "a URI template, as defined by RFC 6570, with the addition of the $, ( and ) characters for pre-processing",
+                    "type": "string"
+                },
+                "rel": {
+                    "description": "relation to the target resource of the link",
+                    "type": "string"
+                },
+                "title": {
+                    "description": "a title for the link",
+                    "type": "string"
+                },
+                "targetSchema": {
+                    "description": "JSON Schema describing the link target",
+                    "allOf": [ { "$ref": "#" } ]
+                },
+                "mediaType": {
+                    "description": "media type (as defined by RFC 2046) describing the link target",
+                    "type": "string"
+                },
+                "method": {
+                    "description": "specifies that the client can construct a templated query (\"get\") or non-idempotent request (\"post\") to a resource.",
+                    "type": "string"
+                },
+                "encType": {
+                    "description": "The media type in which to submit data along with the request",
+                    "type": "string",
+                    "default": "application/json"
+                },
+                "schema": {
+                    "description": "Schema describing the data to submit along with the request",
+                    "allOf": [ { "$ref": "#" } ]
+                }
+            }
+        }
+    },
     "allOf": [ { "$ref": "http://json-schema.org/draft/schema#" } ],
     "properties": {
         "additionalItems": {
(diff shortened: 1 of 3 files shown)


schema.json
```diff
@@ -43,131 +43,142 @@
...
-        "format": { "type": "string" },
-        "allOf": { "$ref": "#/definitions/schemaArray" },
-        "anyOf": { "$ref": "#/definitions/schemaArray" },
-        "oneOf": { "$ref": "#/definitions/schemaArray" },
-        "not": { "$ref": "#" }
-    },
-    "dependencies": {
-        "exclusiveMaximum": [ "maximum" ],
-        "exclusiveMinimum": [ "minimum" ]
+            }
```
JanTvrdik: `"required": ["$ref"]` seems to be missing

JanTvrdik: missing `"required": ["$ref"]`

handrews: For actual JSON References it's on line 14 in the "jsonReference" definition.

In this case, all we need to happen is that if "$ref" is present at all, the validation of the "schemaObject" definition needs to fail, and the negated schema here is sufficient to do that without the need for "required".

The negated schema does not need to match the "jsonReference" definition, and in fact should not because a non-string "$ref" should fail both "jsonReference" and "schemaObject".

JanTvrdik: JSON Schema `{"properties": { "$ref": {} }}` is equal to `{}`, meaning it does not enforce anything. What you actually intended to write is `"not": { "required": ["$ref"] }`. Am I wrong?

handrews: "not": { "required": [ "$ref" ] }

also works but I don't like how it reads.  "cannot have a property named $ref" read better than "$ref is anti-required" to me.  Obviously preferences will vary.

handrews: @JanTvrdik wait, maybe I'm not sure... lemme think about it...  gah...

JanTvrdik: > So that negated schema successfully validates any object that has a "$ref" property. 

Yes, that is correct. It however also successfully validates any object without "$ref" property.

handrews: > Yes, that is correct. It however also successfully validates any object without "$ref" property.

Yeah, that's why I reversed myself :-P

epoberezkin: so it should be `"not": { "required": [ "$ref" ] }` here, correct?

schema.json
```diff
@@ -1,8 +1,18 @@
...
+                "$ref": {
+                    "type": "string",
+                    "format": "uriref"
+                }
+            },
+            "required": [ "$ref" ]
+        },
         "schemaArray": {
             "type": "array",
             "minItems": 1,
```
epoberezkin: @handrews I mean here re comment in the list of PRs

epoberezkin: minItems should be removed

handrews: Not quite sure I follow but I need to re-do this whole series of commits anyway so I will double check as I do that.

handrews: Oh, right- you meant in the unchanged part of the schema.  I was looking at the green-highlighted part and was confused.  Yes, I'll take a look at that.

handrews:
OK, this incorporates all feedback.
@epoberezkin I checked on the things that use schemaArray (items, anyOf, allOf, oneOf) and none of them have a minimum length requirement so I removed that.  None specify a default, so I did not add one.

Here are the individual commits (review it like this, **NOT** as one giant diff):

Formatting standardization, tie hyper-schema's "schemaArray" to validation's:
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/7cd05ab76783940d094b7c2902545495d7fd324b

Move definitions to top but do not change anything else (more standardization):
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/b8154b469ce553e398a7a5b0b5cd1109c5314b82

Add "$ref" for subschemas:
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/23b2fcc82defc1d1a49986d39baebe1da02580aa

Update for numeric exclusives, empty schema arrays, give string arrays a default of empty array:
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/a360b1777cc36389f7baabe0a271024cc4e72793

handrews:
@Relequestual notJsonReference has nothing to do with hyper-schema.  It implements this part of the core spec:

> All other properties in a "$ref" object MUST be ignored.

What that means is that if "$ref" is present, the object **cannot** be used as a regular JSON Schema.  So it should not validate as one.

To walk through it, let's start with the notJsonReference schema:

```JSON
"notJsonReference": {
    "not": {
        "required": [ "$ref" ]
    }
}
```

The inner schema, `{"required": [ "$ref" ] }` passes validation if a "$ref" property is present (or if the instance is not an object).  It doesn't matter what the value of "$ref" is, the only thing that matters is whether it is there.

So, when negated, this means that non-objects fail validation (because there is no "type" restriction in the inner schema) and objects with "$ref" fail validation, regardless of the value of "$ref" (because "$ref" being present at all causes the "required" to pass).

Now let's take a look at the oneOf/allOf construct:

```JSON
"subSchema": {
    "oneOf": [
        {   
            "allOf": [
                { "$ref": "#" },
                { "$ref": "#/definitions/notJsonReference" }
            ]   
        },  
        { "$ref": "#/definitions/jsonReference" }
    ]   
}   
```

This "oneOf" says that a subschema must either be a regular schema (the "allOf" branch) or a JSON Reference (the "#/definitions/jsonReference" branch) but _not_ both.

So consider this instance:

```JSON
{
    "type": "object",
    "description": "Lorem ipsum...",
    "$ref": "#/definitions/something"
}
```

This is both trying to be a regular schema (type and description) and a reference ($ref).  _Without_ "notJsonReference", this would validate as _both_ a regular schema _and_ a reference, which is wrong.

Even though there's nothing structural that would otherwise disqualify this as a schema, the spec says that if "$ref" is present, the other fields MUST be ignored.  So even if they look like a schema, they will not behave like one.  So this instance should **not** validate as a regular schema.  Without "notJsonReference", it would validate as a regular schema because JSON Schema does not forbid unknown properties.  So we need to specifcially forbid $ref, which is what "#/definitions/notJsonReference" does.

Instead, it _should_ validate as a reference, because the value of "$ref" is a URI Reference string.  And then the other fields are correctly ignored.

Finally, consider

```JSON
{
    "$ref": 42
}
```

This should fail validation entirely.  It will fail the "jsonReference" branch of the "oneOf" because 42 is not a string (much less a URI Reference).  But without "notJsonReference", this _would_ successfully validate as a regular schema, because unknown properties do not cause validation to fail.  So we need "notJsonReference" to ensure that this fails the regular schema branch of the oneOf as well as the JSON Reference branch.

With "notJsonReference" involved, this instance correctly fails validation entirely.

To summarize:  IF "$ref" is present, the instance must validate as a JSON Reference and not as anything else.  If it fails to validate as a JSON Reference, it must fail validation entirely.  ELSE ("$ref" is not present) it passes validation if it is otherwise a regular JSON Schema.

epoberezkin:
@handrews your metaschema is correct and explanation makes perfect sense. Do you see though how difficult it is for everybody to see the intention (even though @Relequestual has seen a couple of schemas) - although your schema achieves the intentiion you describe, it is very difficult to understand what it does. Also, I hope you notice the language you use to describe this intention (IF/THEN/ELSE). This schema would have been very easy for us to understand and wouldn't have raised the question if these keywords actually existed:

```json
{
  "if": { "required": ["$ref"] },
  "then": { "$ref": "#/definitions/jsonReference" },
  "else": { "$ref": "#" }
}
```

And it doesn't prescribe any order of execution any more than allOf/anyOf/oneOf do. Subschemas in the latter keywords can be evaluated in any order (although all validators I know evaluate them in the order of array items and short-circuit, unless the mode is to collect all possible errors). In the same way subschemas in if/then/else can be evaluated in any order - it will not affect the final result (although I can't see why validators will not do the sensible thing here as well).

This validation scenario is not constrained to meta-schema. It is very common in real world. And every time, without if/then/else you have to resort to the equivalent schema without if/then/else (see #180), as you did in meta-schema. The problem with that approach is that it is very difficult to see the intention behind this equivalent schema.

handrews:
@epoberezkin Please don't take over PRs with other issues.

epoberezkin:
@handrews ok. sorry. couldn't resist... Should've put the same comment in the issue.

handrews:
>  couldn't resist... Should've put the same comment in the issue.

No worries.  I read the issue first anyway and it all worked out as it was.

awwright:
@handrews This PR seems to introduce a few typos, mostly missing commas.

Can this get broken up into a smaller series of PRs?

There's a few things to note that would make our life easier, like I don't think boolean "exclusiveMinimum" actually depends on "minimum" existing (it just doesn't make any sense by itself).

handrews:
@awwright it's already broken into smaller commits.  Just review the individual commits.  If you tell me what you need changed I will apply the changes to each commit as needed.  GitHub doesn't do stacked PRs at all easily.

The dependency for the exclusives comes directly from the existing meta-schema and I don't want to change it.  "it doesn't make any sense by itself" is exactly what that is trying to show.  There's no need to take that out when it was fine before.

handrews:
Fixed various typos / missing comments /etc.

Since the commit hashes changed, I have updated the links to the individual commits in https://github.com/json-schema-org/json-schema-spec/pull/168#issuecomment-264674169

awwright:
@handrews This does a lot of things, I'm still reviewing and trying to figure out if GitHub is presenting this to me correctly. I pulled 7cee490207147d9d5f0002c8b65274accfb0bdca, what is that right?

Relequestual:
I don't see that commit in the commits tab at the top of this issue. I think you must be looking at old commits. May need to refresh your page before viewing the commits tab on the issue to get the updated commits.

awwright:
Maybe I'm on the wrong end of a caching server, GitHub is still showing me the presumably-old version with a missing comma.

handrews:
> Maybe I'm on the wrong end of a caching server, GitHub is still showing me the presumably-old version with a missing comma. It's not what I pulled, but it's not ce456a6 either.

It's not just you, I'm not sure what's going on.  I will take a look at it.

hyper-schema.json
```diff
@@ -47,52 +47,64 @@
...
+                    "allOf": [ { "$ref": "#/definitions/subSchema" } ]
                 }
             }
+        },
+        "subSchema": {
+            "oneOf": [
+                {
+                    "allOf": [
+                        { "$ref": "#" },
+                        { "$ref": "http://json-schema.org/draft/schema#/definitions/notJsonReference" }
```
Relequestual: Shouldn't this ref be a realative pointer? We wouldn't want to make this version of the meta-schema possible to be invalid by a change at a later date.

Relequestual: Same for line 62

handrews:
@awwright @Relequestual I don't know what the problem was but I amended the last commit (reworded the commit message) and force-pushed and it has the right commits now:

Formatting standardization, tie hyper-schema's "schemaArray" to validation's:
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/ef13c4963fc762cc4f30003e4461090b0b0411c8

Move definitions to top but do not change anything else (more standardization):
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/a8be3fd60cdbfb04ebf6e7a6b4305063d2be4d4b

Add "$ref" for subschemas:
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/018fe924ffc22c53b5807cb74c44a4ea890f6c79

Update for numeric exclusives, empty schema arrays, give string arrays a default of empty array:
https://github.com/json-schema-org/json-schema-spec/pull/168/commits/983e9d30cf1c9e0dedd058869a8a03d39424d36b

handrews:
As a reminder, the goal is to match the current specification and not change anything more than that.

* The $ref URIs in hyper-schema are not relative because they (specifically the "allOf") were not relative before.  This is not a change.
* The subschema approach is necessary because $ref is not allowed at the root.  We are considering changing that but that belongs in #174.
* Yes, exclusiveMinimum/exclusiveMaximum look weird.  That is how the spec is currently written- they are either numbers, _OR_ they are booleans and if so require the presence of minimum/maximum respectively (the dependencies usage is what the current meta-schema does).  Again, it's weird, but not a change from the spec.  Proposed changes such as dropping the boolean form entirely should get their own issue and not block this PR.

hyper-schema.json
handrews: References to the main schema have never been relative (see the Draft 4 schema) and I'd rather not change that as part of this PR.  I'm trying to do the minimal changes here.

Relequestual: @handrews when you see this, would be good to comment based on our discussion on irc, just so I'm clear why, if this can't change.

Relequestual:
Right. Thanks for explaining. That makes sense. None of these are changes, just fixes to bring the meta-schema in line with what the spec actually says. Great stuff. I'm totally happy with this! (Although we should make sure we capture the problems that exist, in new issues.)

hyper-schema.json
Relequestual: Right, that makes sense. Should be logged that it should be fixed though.

handrews: Logged #182 (boolean schemas, because #167 went in without meta-schema change because I thought this would go in first) and #138 (for the relative reference thing).

handrews:
> Although we should make sure we capture the problems that exist, in new issues.

#182 Add Boolean schemas to meta-schemas
#183 Relative references
The question of what to do about the exclusiveMinimum/Maximum can be continued in #77

awwright:
I keep reviewing this, but it's just too difficult to tease apart all of the things going on together, and that's holding up progress on this which is unfortunate. So I'll close this out, but we should deal with all of the issues presented here.

I filed #185 to deal with the case of "exclusive*" properties.

For formatting conserns, I'd like to see a particular rationale for why whitespace should be formatted the way it is. We should document these patterns in README.

I noticed anyOf/allOf/oneOf still requires at least one item, so we can't remove `"minItems": 1` from the meta-schema quite yet. Let's figure out how to deal with that in a new issue.

There's a lot going on here so I'm probably missing some stuff, I'd like to comment on these individually so please let me know if there was anything else being dealt with here.

handrews:
> I keep reviewing this, but it's just too difficult to tease apart all of the things going on together

Do you not understand how to review individual commits?  I would like a yes or no answer on this.

awwright:
@handrews I'm not even sure what to say. Yes, I see the feature to review commits. But there's multiple distinct issues here, that I would like to be able to merge individually, instead of having to merge everything at-once.

handrews:
@awwright Then go merge #186 and I will feed you the PRs one by one right now.

handrews:
@awwright You also could have tried telling me what these "multiple distinct issues" are.  With the exception of the point we're arguing in #185 I made every single change you asked for, immediately.  I went on IRC since you wouldn't do it here.  I walked you through everything repeatedly.  I answered questions, and I changed things and updated the stack appropriately.

I do all of that, and you just close it because of unspecified "issues" that you won't describe or give me a chance to fix.  Everyone else who reviewed this had no trouble articulating their concerns, which I also addressed.  But you just won't.

Even when closing the PR you're incapable of actually articulating the real problems, which we *still* don't know.

handrews:
@awwright I was about to split off the exclusive* bits since you opened #185, but you didn't even come back here and ask for that, you just closed it.  And even though you commented here after I opened #186 in an effort to accommodate you once you closed this PR, you didn't bother to comment on that at all.

Relequestual:
Should be noted that #188 fixes part of this.

Issue #70, opened 2016-10-03
v6 validation: propertyNames


epoberezkin:
**Definition**

The value of this keyword is a JSON-schema. The keyword only applies to objects. If the instance is not an object, validation succeeds. For the instance to be valid all property names of the object should be valid according to this schema. Possible variant: only apply it to property names not explicitly mentioned in `properties` keyword - there are pros and cons for it.

**Motivation**

Generally the current spec is more suited for the data stored in values with static keys. The only exception to that is `patternProperties`.

There are two disadvantages of patternProperties:
1. You cannot define any other restriction apart from, well, pattern
2. Even if pattern is sufficient but you need to use the same property names pattern in multiple objects (but with different schemas for values, so you can't use $ref), you have to repeat the same pattern over and over again. It is difficult to maintain and to ensure they are the same.

`propertyNames` addresses both issues. So instead of having:

``` json
{
  "patternProperties": {
    "^(\\d{1,3}\\.){3}(\\d{1,3})$": { "$ref": "hostData.json" }
  },
  "additionalProperties": false
}
```

you can have:

``` json
{
  "propertyNames": { "$ref": "#/definitions/host" },
  "additionalProperties": { "$ref": "hostData.json" },
  "definitions": {
    "host": { "pattern": "^(\\d{1,3}\\.){3}(\\d{1,3})$" }
  }
}
```

or even

``` json
{
  "propertyNames": { "$ref": "#/definitions/host" },
  "additionalProperties": { "$ref": "hostData.json" },
  "definitions": {
    "host": {
      "anyOf": [
        { "format": "hostname" },
        { "format": "ipv4" }
      ]
    }
  }
}
```

You can see how using `propertyNames` allows to have the schema for property names that is:
1. more advanced than simple pattern matching
2. can be reused and updated in a single place.

We use this keyword a lot, it is defined as a [custom keyword for Ajv](https://github.com/epoberezkin/ajv-keywords#propertynames). It seems to be so useful and simple that maybe it's worth considering adding it to the spec.

erosb:
I'm a bit confused about this proposal. I don't see the `propertyNames` keyword used in any of the examples you provided.
But I can see the `additionalProperties` , which already exists, and means exactly what you propose.

epoberezkin:
[...]

PR #172, opened 2016-11-30
Validation: Add "propertyNames" keyword

awwright:
Language for #70

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -548,6 +548,17 @@
                 </t>
             </section>
 
+            <section title="propertyNames">
+                <t>
+                    The value of "propertyNames" MUST be a valid JSON Schema.
+                </t>
+                <t>
+                    If the instance is an object, this keyword validates if every property name in the instance
+                    validates against the provided schema.
+                    Note the property name that the schema is testing will always be a string.
+                </t>
+            </section>
+
             <section title="enum">
                 <t>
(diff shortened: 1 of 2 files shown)


jsonschema-validation.xml
```diff
@@ -550,6 +550,20 @@
...
+                <t>
+                    The value of "propertyNames" MUST be a valid JSON Schema.
+                </t>
+                <t>
+                    If the instance is an object, this keyword validates if every property name in the instance
+                    validates against the provided schema.
+                    Note the property name that the schema is testing will always be a string.
+                </t>
+                <t>
+                    A missing keyword is the same as an empty schema, or a schema testing that the type is "string".
```
handrews: Same objection to this language as in PR #171, so this should be adjusted based on the resolution of that PR.

awwright: I can say "A missing keyword exhibits the same behavior as an empty schema" or something like that... The thing is it has to be informative, since it's noting something that's already true.

epoberezkin: @handrews @awwright I am ok with the current language or with "defaults to empty schema" (which will be consistent with meta-schema).

awwright:
I just removed the paragraph, we can figure out what to do in #171 regarding "default" language

epoberezkin:
at least something useful from me, not just the stream of disruption ;)

Issue #74, opened 2016-10-08
"deprecation" property


awwright:
Create a schema keyword that specifies an instance (property in an instance, usually) as deprecated -- it shouldn't be used, but exists only for internal use, or for reverse compatibility with obsolete or old products.

handrews:
+1 to this.  The point here, I assume, is for documentation and perhaps to allow implementations to issue warnings, rather than to affect validation.

Relequestual:
Interesting proposal.

I know that JSON Schema isn't specifically for API's, but we should consider how API's might version their endpoints. Could this proposal actually be part of a seperate doc about API driven information. For example, included in deprecated, how it's deprecated and if it's replaced, and possibly other upgrade path type data.
Very much a straw man comment, not well thought through.

handrews:
@Relequestual I'm not sure how far we want to chase the versioning idea here, but my approach to API evolution (rather than versioning) is to only change the version of the schema, and retain compatibility by (as much as possible) avoiding both `required` and `"additionalProperties": false`.  I use the following rules:

* Resources add new schema versions to their list of profiles, and do not remove old versions
* If an instance validates against a schema that is a valid profile for the resource, no matter what version that schema is, than the instance can be used according to that version of the schema.
    * So if your code was written for a specific version, as long as that still validates, you can keep using that without worrying about the new version.
    * This can allow for continuing to function in a scenario where some new instances may not validate against the old schema, but if the instances you work with still validate, you don't have to update to support the things you don't need.  For instance, a value was added to an enum, but you never use the new value.
* If you can't keep compatibility at all, make a new resource and handle the change through HATEOAS, use of 410 GONE, redirects, etc.
* There are no versions in the URI or primary media type (the schema versions likely show up in their profile URIs, since those need to be distinct somehow)

Relequestual:
Interesting. Most of the time our API (and others I've worked on) have needed "new versions" because of new fields or new allowed values. Sometimes that's backwards compatible, but often it isn't for me, as much as I'd like it to be.

handrews:
As long as you avoid setting additionalProperties to false, then new fields should be fine (unless you're generating strictly typed code off of the schemas without any provision for unrecognized fields, but I avoid that in favor of more generic client libraries).

Enumerations get tricky, but that's where the "if it still validates you can still use it" rule comes into play- technically the new version is not fully compatible but most instances will validate against either version.

[...]

PR #173, opened 2016-11-30
Hyper-schema: Add "deprecated" keyword

awwright:
Language for implementing #74 

Also adds meta-schema for "readOnly" that seems to have gone MIA

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -54,6 +54,12 @@
             "type": "array",
             "items": {"$ref": "#/definitions/linkDescription"}
         },
+        "deprecated": {
+            "type": "boolean"
+        },
+        "readOnly": {
+            "type": "boolean"
+        },
         "media": {
             "type": "object",
             "properties": {
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -334,6 +334,16 @@
                 </section>
             </section>
 
+            <section title="deprecated">
+                <t>
+                    If it has a value of boolean true, this keyword indicates that the value of the instance is deprecated, and intended for reverse compatability or internal use only, and may become missing in the future.
+                </t>
+                <t>
+                    The value of this keyword MUST be a boolean.
+                    The default value is false.
+                </t>
+            </section>
+
             <section title="readOnly">
                 <t>
                     If it has a value of boolean true, this keyword indicates that the value of the instance is managed exclusively by the server or the owning authority, and attempts by a user agent to modify the value of this property are expected to be ignored or rejected by a server.


hyper-schema.json
```diff
@@ -54,6 +54,12 @@
             "type": "array",
             "items": {"$ref": "#/definitions/linkDescription"}
         },
+        "deprecated": {
+            "type": "boolean"
+        },
+        "readOnly": {
+            "type": "boolean"
+        },
```
handrews: Should have "default": false

hyper-schema.json
```diff
@@ -54,6 +54,12 @@
             "type": "array",
             "items": {"$ref": "#/definitions/linkDescription"}
         },
+        "deprecated": {
+            "type": "boolean"
+        },
+        "readOnly": {
```
handrews: Should have "default": false

jsonschema-hyperschema.xml
```diff
@@ -334,6 +334,16 @@
                 </section>
             </section>
 
+            <section title="deprecated">
+                <t>
+                    If it has a value of boolean true, this keyword indicates that the value of the instance is deprecated, and intended for reverse compatability or internal use only, and may become missing in the future.
```
handrews: "value of the instance is deprecated" is strange phrasing.  Usually you deprecate a field, not a field's value, or you deprecate an entire resource.  So this could apply to any schema with a "self" link (which is basically what determines that the schema represents a resource), or to a field within an object schema.  Or perhaps a position in a tuple schema (when "items" is an array of schemas).

There's something a bit strange there where this is either about a resource or about the context in which the schema appears (the property name or pattern, for instance).

awwright: It is, isn't it, I'll fix that

Relequestual:
OK, so we agree the phrasing needs to be reworked and reviewed. Cool.

jsonschema-hyperschema.xml
handrews: So how does this work?

In a root schema with a "self" link is means that the resource itself is deprecated?  But it's still there currently?

In a root schema without a "self" link it means....?

In a property subschema (in any of the property keywords) it means that all matching properties are deprecated but will still be present?

In a schema applying to a single position in an array/tuple ("items" is an array) it means...?

In a schema applying to multiple array positions ("items" as one schema, or "additionalItems") it means...?

In the schema for "schema" or "targetSchema" it means...?

This concept is not ready for a PR.  There are far too many cases that need discussing and a coherent view of hypermedia resources that need to be developed before we can just toss this in.

handrews:
@awwright do you plan to re-work this exact PR or can we close this, finish the discussion in the issue, and start with a new one?  It's a bit confusing to me to have it open in two places, and there is a lot more context in the issue at this point.

awwright:
Yeah, let's re-start a discussion and close this... A lot of new ideas have
been brought up since that would be useful to start with.

On Aug 18, 2017 11:45 AM, "Henry Andrews" <notifications@github.com> wrote:

@awwright <https://github.com/awwright> do you plan to re-work this exact
PR or can we close this, finish the discussion in the issue, and start with
a new one? It's a bit confusing to me to have it open in two places, and
there is a lot more context in the issue at this point.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/json-schema-org/json-schema-spec/pull/173#issuecomment-323431814>,
or mute the thread

handrews:
Closing as agreed.

philsturgeon:
Just for directing people to the correct place, future deprecations conversations will be, for now, on https://github.com/json-schema-org/json-schema-spec/issues/74.

PR #181, opened 2016-12-06
fixing typo

dret:


awwright:
I'm not actually sure what the RFC Editor would prefer here... It probably should be "Hypermedia and Linking", but the other section titles are defining terms that might not necessarily be capitalized.

This PR is probably correct, but give me a little bit to get around to reviewing the document again.

PR #186, opened 2016-12-11
Standardize format of meta-schemas.

handrews:
This brings formatting and most of the initial field ordering
in line with the Draft 05 meta-schemas proposed in the web site
repo.

Also ensure that hyper-schema's definition of "schemaArray"
cannot get out of sync with schema's, and add the link schema.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -2,48 +2,46 @@
     "$schema": "http://json-schema.org/draft/hyper-schema#",
     "id": "http://json-schema.org/draft/hyper-schema#",
     "title": "JSON Hyper-Schema",
-    "allOf": [
-        {"$ref": "http://json-schema.org/draft-04/schema#"}
-    ],
+    "allOf": [ { "$ref": "http://json-schema.org/draft/schema#" } ],
     "properties": {
         "additionalItems": {
             "anyOf": [
-                {"type": "boolean"},
-                {"$ref": "#"}
+                { "type": "boolean" },
+                { "$ref": "#" }
             ]
         },
         "additionalProperties": {
             "anyOf": [
-                {"type": "boolean"},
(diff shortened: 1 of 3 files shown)


handrews:
@awwright This does not have any actual changes beyond formatting, please review and merge ASAP.

hyper-schema.json
```diff
@@ -66,17 +64,26 @@
                     "type": "string"
                 }
             }
+        },
+        "readOnly": {
```
Relequestual: I can see readOnly in draft-5 (aka latest), so am happy with this.

links.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema": "http://json-schema.org/draft/hyper-schema#",
```
Relequestual: When the draft 4 version had the URI of draft-4/hyper-schema, why do we not use the URI of draft-5/hyper-schema? (Or will that be a later PR?)

links.json
```diff
@@ -0,0 +1,42 @@
...
+    "id": "http://json-schema.org/draft/links#",
+    "title": "Link Description Object",
+    "type": "object",
+    "required": [ "href" ],
```
Relequestual: draft 4 version at https://github.com/json-schema-org/json-schema-org.github.io/blob/master/draft-04/links#L6 includes "rel" as required. Was this changed in draft 4? (Just checking)

links.json
handrews: Because this is for "master" so it's not a numbered draft (at least that's how @awwright set it up and I was trying to not change it- take a look at the "id" of the hyper-schema).

Relequestual:
As this doesn't change anything functionally, I see no reason to daley merging, which will enable further PRs.

PR #193, opened 2016-12-13
Improve child validation wording from feedback.

handrews:
This implements suggestions from @deuberger discussions in issue #161.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -179,15 +179,26 @@
 
             <section title="Validation of primitive types and child values">
                 <t>
-                    Two of the primitive types, array and object, allow for child values.  The validation of
-                    the primitive type is considered separately from the validation of child instances.
+                    Two of the primitive types, array and object, allow for child values.
+                    The validation of the primitive type is considered separately from
+                    the validation of child instances.
                 </t>
                 <t>
-                    For arrays, primitive type validation consists of validating restrictions on length.
+                    For arrays, primitive type validation consists of validating
+                    restrictions on length with "minItems" and "maxItems", while
+                    "items" and "additionalItems" determine which subschemas apply
+                    to which elements of the array.  In addition, "uniqueItems"
+                    and "contains" validate array contents as a whole.
                 </t>
                 <t>
-                    For objects, primitive type validation consists of validating restrictions on the presence
-                    or absence of property names.
+                    For objects, primitive type validation consists of validating
+                    restrictions on which and how many properties appear with
+                    "required", "minProperties", "maxProperties", "propertyNames",
+                    and the string array form of "dependencies", while "properties",
+                    "patternProperties", and "additionalProperties" determine which
+                    subschemas apply to which object property values.
+                    In addition, the schema form of "dependencies" validates the
+                    object as a whole based on the presence of specific property names.
                 </t>
             </section>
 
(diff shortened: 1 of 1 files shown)


jsonschema-validation.xml
```diff
@@ -172,15 +172,26 @@
...
+                    Two of the primitive types, array and object, allow for child values.
+                    The validation of the primitive type is considered separately from
+                    the validation of child instances.
                 </t>
                 <t>
-                    For arrays, primitive type validation consists of validating restrictions on length.
+                    For arrays, primitive type validation consists of validating
+                    restrictions on length with "minItems" and "maxItems", while
+                    "items" and "additionalItems" control which subschemas apply
+                    to which elements of the array.  In addition, "uniqueItems"
```
Relequestual: I find the phrasing
> ...control which subschemas apply to which elements of the array.

a little confusing. It may be just my lack of understanding though. Can you expand on this section please?

handrews: It's extremely important because implementing these is different from implementing other keywords.  There are basically four types of validation keywords:

* Immediate instance keywords (most of them)
* Apply-subschema-to-child-instance keywords (these)
* Logical combinations of schema results ("allOf"/"anyOf"/"oneOf"/"not"/"dependencies"/"if"-"then"-"else" assuming it eventually goes in)
* Array content keywords ("uniqueItems" and "contains") that can only be evaluated by looking at the entire array and all of its contents

handrews: It also highlights how these keywords (like "allOf"/"anyOf"/"oneOf"/"not") are important for more than just validation.  All other vocabularies, such as hyper-schema, are pretty much useless without these keywords as they are the only way to apply other vocabularies to complex instances.  The core vocabulary ("$schema", "$id", "$ref") does not provide any such mechanism, and vocabularies such as hyper-schema should not re-implement the process.  These keywords and the logical keywords have implications far beyond validation.

handrews:
@Relequestual I have changed the "controls" phrasing, replacing the word with "determines" and changing the sentence structure a bit to (hopefully) make it all more clear.  How does this look now?

handrews:
@awwright you assigned this to yourself a month ago, so it's blocked on you.  What is needed to get your approval here?

jsonschema-validation.xml
```diff
@@ -179,15 +179,26 @@
...
                 </t>
                 <t>
-                    For objects, primitive type validation consists of validating restrictions on the presence
-                    or absence of property names.
+                    For objects, primitive type validation consists of validating
+                    restrictions on which and how many properties appear with
+                    "required", "minProperties", "maxProperties", "propertyNames",
+                    and the string array form of "dependencies", while "properties",
+                    "patternProperties", and "additionalProperties" determine which
+                    subschemas apply to which object property values.
```
Relequestual: I'd consider adding "... and do not directly validate the instance itself.", after reading your explanation at https://github.com/json-schema-org/json-schema-spec/issues/161#issue-191384775 I found your phrasing there explaining...

> I recently rewrote the items, additionalItems, properties, patternProperties, and additionalProperties to make it clear that they do not directly validate the instance, but instead control which subschemas are applied to which child instances.

to actually be a clearer explanation that the changes you've made.

Relequestual:
Sorry for being a blocker. Thanks for calling me out.

jsonschema-validation.xml
handrews: @Relequestual I have made the changes you requested and replaced "validation always succeeds" with "does not directly validate the instance itself."

PR #202, opened 2016-12-21
Add Henry Andrews, author

awwright:
@handrews Can I get your sign-off on this PR's accuracy?

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -28,6 +28,13 @@
             </address>
         </author>
 
+        <author fullname="Henry Andrews" initials="H" surname="Andrews">
+            <organization>Cloudflare, Inc.</organization>
+            <address>
+                <email>henry@cloudflare.com</email>
+            </address>
+        </author>
+
         <date year="2016"/>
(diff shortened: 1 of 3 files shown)


handrews:
Should be "Cloudflare, Inc." (see https://tools.ietf.org/html/draft-ietf-regext-dnsoperator-to-rrr-protocol-01 for an example)

but otherwise looks great, thanks!

awwright:
@handrews Updated

Most documents seem to prefer the trade name over the legal name, but it doesn't seem to matter either way

handrews:
@awwright looking through the results at
https://www.google.com/#q=site:tools.ietf.org+cloudflare

all but one have the "Inc."  So it could go either way, I was just going with the obvious majority.  Although https://tools.ietf.org/html/rfc7710 does not have the "Inc."  So do whichever, I doubt our legal department is going to chase after me either way.

**NOTE**:  "Cloudflare" is correct, ignore the older ones with "CloudFlare".  We thankfully got rid of the capital F recently.

awwright:
Welcome aboard 🚢

Issue #209, opened 2016-12-28
additionalItems meta-schema dependency


epoberezkin:
At the moment the spec says:

> Otherwise [if items is not an array], "additionalItems" MUST be ignored, as the "items" schema (possibly the default value of an empty schema) is applied to all elements.

That causes a lot of confusion from users. The suggestion is to prohibit additionalItems (rather then require ignoring it) and enforce it via meta-schema unless items is present and it is an array.

The meta-schema will be inside dependencies, together with exclusiveMaximum and exclusiveMinimum (https://github.com/json-schema-org/json-schema-spec/pull/189#issuecomment-269455276)

```json
"dependencies": {
  "additionalItems": {
    "required": ["items"],
    "properties": { "items": { "type": "array" } }
  }
}
```

The spec also becomes simpler:

> Otherwise [if items is not an array], "additionalItems" MUST NOT be present

@awwright @handrews @Relequestual if that makes sense, I can make a PR

awwright:
Yeah, that makes sense.

I think it would make some sense to say "items" MUST be an array - that would make things more consistent, so that additionalItems/items can't serve the same purpose.

handrews:
@awwright additionalItems and items already can't serve the same purpose.   And @epoberezkin's proposed change does require items to be an array if additionalItems is present.  What do you want different?

awwright:
If I understand correctly, for schema values (when I want to specify what every item in the array must look like), I can use either "additionalItems" to "items" with no difference in behavior.

Specifying that "additionalItems" implies "items" MUST be an array seems reasonable.

Can we make "items" always be an array?

handrews:
> If I understand correctly, for schema values (when I want to specify what every item in the array must look like), I can use either "additionalItems" to "items" with no difference in behavior.

No, you cannot.  From the spec:

> If "items" is an array of schemas, ....
>
> Otherwise, "additionalItems" MUST be ignored, as the "items" schema (possibly the default value of an empty schema) is applied to all elements.

So the only case in which "additionalItems" is not ignore is when "items" is a schema.

Why should we break how "items" currently works?

handrews:
BTW I am not dead set against changing "items", and I agree that the current situation is kinda weird.  But it is a big change with a broad impact.  It would be easy enough to migrate with a script (which is true of our other recent incompatible changes as well), but it's not a change I'd want to rush into.

epoberezkin:
[...]

PR #210, opened 2016-12-28
Remove boolean form of exclusive* keywords.

handrews:
Yet another alternative to #185 and #189.  This just drops the boolean keywords and does the minimal changes necessary to implement that.

@epoberezkin, if you want to make it impossible for both `minimum` and `exclusiveMinimum` to appear together (and likewise for `maximum` and `exclusiveMaximum`) could you please file that as a separate issue like you did with the `items` and `additionalItems` dependencies?  All of those cases are ones that work without the dependencies but could be more helpful with, so let's discuss that whole concept on its own.  Maybe even just broaden #209 to cover all of these meta-schema dependency questions?  Then we could sort it all out at once and if it gets sorted before this gets merged (if it ever does) I will update this patch accordingly.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -213,7 +213,6 @@
                     <list>
                         <t>"additionalProperties", whose behavior is defined in terms of "properties" and "patternProperties";</t>
                         <t>"additionalItems", whose behavior is defined in terms of "items"; and</t>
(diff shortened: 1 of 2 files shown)


epoberezkin:
@handrews I think I'd rather make separate PR for all dependencies (exclusiveMaximum and additionalItems). This can be merged as is I think, I've approved.

handrews:
Thanks, @epoberezkin !

@awwright @Relequestual , over to you for final approval.

epoberezkin:
@awwright how can we stop them :)
Some still support required: true form from draft 3...

PR #212, opened 2016-12-28
Change "id" to "$id" to match other core keywords

handrews:
This addresses issue #20.  It is an alternative to PR #154.

"$id" matches "$schema" and "$ref", establishing a clear naming
pattern for all keywords defined in the core specification.  This
also reduces confusion in the very common case where objects
described by the schema have an "id" property.


epoberezkin:
Yep, simpler is better. Now implementations can decide what to look for based on $schema or even support both keywords regardless... But it's not spec's concern.

Relequestual:
Given the approval, I'll merge.

PR #219, opened 2017-01-03
Update jsonschema-hyperschema.xml

donaldpipowitch:
According to https://github.com/json-schema-org/json-schema-org.github.io/blob/master/draft-04/links it is not possible to have a `properties` field inside a link description object.


handrews:
Whoops- thanks!

PR #223, opened 2017-01-05
TRIVIAL: It's now 2017.  Keep xml2rfc from complaining.

handrews:

PR #224, opened 2017-01-06
It's now 2017.  Keep xml2rfc from complaining.

handrews:


handrews:
I previously created this as #223 and awwright approved it, and then I apparently hit "close" instead of "merge" and can't figure out how to re-open and merge it.  So this PR is just #223 done over so I can merge it.

Issue #207, opened 2016-12-27
Consistent formatting of "format" values


handrews:
As @epoberezkin noted, we're a bit inconsistent about format:

* date-time
* hostname (arguably one word these days)
* uriref
* uritemplate
* jsonpointer

In earlier drafts, there were more formats that used hyphens.  

Notably, as these are strings, we could use spaces.

Furthermore, as this is JSON, we could use camelCase, as we do for keywords.

Setting aside the question of whether "hostname" counts as two words, we should pick one of:

* date-time, uri-ref, uri-template, json-pointer
* datetime, uriref, uritemplate, jsonpointer
* dateTime, uriRef, uriTemplate, jsonPointer

Or we could decide that date-time is grandfathered in and stick with single words going forward?  

I doubt many people have implemented "uriref" yet, so perhaps changing all of them to hyphens is best.  I do *NOT* want to have "uriref" but "uri-template".  That is the one thing I will throw a fit about (it's why I added "uritemplate" spelled the way it is in the first place- I'm not otherwise dogmatic about this).

@awwright @Relequestual @Julian opinions?  Let's pick one and get it in Draft 6 since it's a trivial edit once we decide.

epoberezkin:
@handrews I prefer with hyphens, to be honest, also because at some point we may have "relativejsonpointer" and it breaks my head to read... And "hostname" indeed seems a single word to me :)

Relequestual:
I much prefer hyphens for readability. But that's my personal preference.

erosb:
+1 for using hypens

handrews:
Addressed by #225

PR #225, opened 2017-01-06
Standardize on hyphenated format names

handrews:
When we had more formats in earlier drafts, all compound names
(such as "date-time") were hyphenated.  Draft 05 added "uriref"
without a hyphen, which started looking weird when we added
"uritemplate" and "jsonpointer" without hyphens to match it.
This makes everything hyphenated (note "hostname" was deemed
to be a single word in this context).

Addresses issue #207.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -19,7 +19,7 @@
                 "href": {
                     "description": "a URI template, as defined by RFC 6570",
                     "type": "string",
-                    "format": "uritemplate"
+                    "format": "uri-template"
                 },
                 "hrefSchema": {
                     "description": "a schema for validating user input to the URI template, where the input is in the form of a JSON object with property names matching variable names in \"href\"",
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -849,7 +849,7 @@
                     </t>
                 </section>
 
-                <section title="uriref">
+                <section title="uri-reference">
                     <t>
                         This attribute applies to string instances.
                     </t>
@@ -859,7 +859,7 @@
                     </t>
                 </section>
 
-                <section title="uritemplate">
+                <section title="uri-template">
                     <t>
                         This attribute applies to string instances.
                     </t>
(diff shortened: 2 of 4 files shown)


handrews:
Paging @awwright as he had not weighed in on issue #207 before I made this PR.

handrews:
@awwright this has been out for almost two weeks and everyone else involved in the discussion has approved this approach either here or (two more people) in issue #207.  Any thoughts?

awwright:
@handrews Sorry for being so slow this week! I've been dealing with some server issues (if you stopped by the IRC channel you probably noticed I'm not there at the moment).

I'd like to look over how these terms are used in the wild, usually "uriref" is a single word, for example (formally "URI-Reference"); whereas an RFC3339 string is "date-time" and RFC6570 specifies "URI-Template". I think it makes sense to copy how they're defined in their respective specifications, which mostly seem to prefer the hyphen format.

I would vote to keep uriref a single term, but it doesn't matter too much, this is mostly OK.

handrews:
@awwright 
>  I've been dealing with some server issues (if you stopped by the IRC channel you probably noticed I'm not there at the moment).

Ugh, hope that clears up soon- I had looked on IRC to try to get you there before paging you on the PRs today.

I'm looking at [RFC 3986](https://tools.ietf.org/html/rfc3986) "uriref" anywhere.  Nor "uri-ref", just "URI-Reference".  Am I missing something?  If we're going by the spec, it should be "uri-reference" which seems reasonable to shorten to "uri-ref" (although I'd be fine with "uri-reference").

[RFC 6570](https://tools.ietf.org/html/rfc6570) uses "URI-Template" but never "uritemplate" (in any capitalization).

[RFC 6901](https://tools.ietf.org/html/rfc6901) uses "json-pointer" and never "jsonpointer".

So unless there's a source for "uriref" that I'm missing, I think all-hyphenated is clearly the way to go.  Let me know if you object, have another reference, or would prefer "uri-reference".  Otherwise I'll merge this after it reaches the 2-week mark on Friday since you're objection doesn't seem strong and there's otherwise consensus.

handrews:
From IRC: @awwright feels "uri-reference" is more consistent, as it is what is used in other RFCs.  I'm fine with this, it's still doing what the issue filers asked, so we're good to merge.

PR #229, opened 2017-01-13
Correct "mediaType" to "type" in example.

handrews:
"mediaType" is an LDO keyword.  The keyword in the "media" object
is just plain "type".


Relequestual:
This looks like a clear bug fix to me.

PR #236, opened 2017-01-27
Various language improvements

awwright:


jsonschema-validation.xml
```diff
@@ -167,8 +162,8 @@
 
             <section title="Keywords and instance primitive types">
                 <t>
-                    Most validation keywords only limit the range of values within a certain primitive type.
-                    When the primitive type of the instance is not of the type targeted by the keyword, the
+                    Most validation keywords only exclude a range of values within a certain primitive type.
```
handrews: What about

Most validation keywords only constrain values within a certain primitive type.

We've often talked about validation as a constraint system, and I think it feels more intuitive than talking about exclusion.

jsonschema-validation.xml
```diff
@@ -575,21 +559,23 @@
                     validates against the provided schema.
                     Note the property name that the schema is testing will always be a string.
                 </t>
+                <t>
+                    A missing keyword has the same behavior as an empty schema.
```
handrews: Should we specify this as a `true` schema instead of an empty schema?  Admittedly, the behavior is the same but `true` conveys the intent of permitting everything more clearly.

jsonschema-validation.xml
```diff
@@ -634,12 +620,12 @@
...
                 <t>
-                    This keyword's value MUST be an array. This array MUST have at least one
-                    element.
+                    This keyword's value MUST be an array.
+                    This array MUST have at least one element.
                 </t>
                 <t>
-                    Elements of the array MUST be objects. Each object MUST be a valid JSON
-                    Schema.
+                    Elements of the array MUST be objects.
```
handrews: I think we just want to say

Each element of the array MUST be a valid JSON Schema.

in order to allow boolean schemas as well.

jsonschema-validation.xml
```diff
@@ -649,12 +635,12 @@
...
```
handrews: Same thing:

Each element of the array MUST be a valid JSON Schema.

jsonschema-validation.xml
```diff
@@ -664,19 +650,21 @@
 
             <section title="not">
                 <t>
-                    This keyword's value MUST be an object. This object MUST be a valid JSON
-                    Schema.
+                    This keyword's value MUST be an object.
```
handrews: Same thing:

This keyword's value MUST be a valid JSON Schema.

Of course, `{"not": true}` doesn't make much sense, I think it should be syntactically valid.

jsonschema-validation.xml
awwright: I thought that's what we did at first and then it was changed back for some reason... idk. I'll do this though.

handrews: I think maybe I changed it like that in a PR that I ended up dropping?  The boolean schema thing went through a few variations before we settled on it.  Whatever happened, I think just saying "valid JSON Schema" is the best option- we already define what that means in an obvious place (the core spec).

jsonschema-validation.xml
awwright: The language is sort of tricky overall... revised

handrews:
@epoberezkin , @Relequestual are y'all OK with this?  I'd like to get these fixes merged if so.  There is no behavior change so it doesn't need to sit open for weeks.

Relequestual:
I need more time to review this PR. I feel in some places it's actually more confusing, and I have ideas how to fix that, but I can't do it right now.

jsonschema-validation.xml
epoberezkin: I think so too

awwright: "empty schema" is language that was used before, can we work on this in a new PR and call this one good?

handrews: If empty schema vs `true` is the only concern, I endorse doing that in a follow-up PR.  We should check for it in more places than just these lines, and that will be trivially easy to check up on on its own.

awwright: Yeah, that's the thing, it's used quite a few times

Relequestual: I find the phrasing "A missing key word...", does not make me think that the missing key word in question is THIS key word (as in, the key word in where the phrase is being used).
I would much rather prefer something like "Omitting this key word has the same effect as...". I feel the intent is far clearer.

If it's just me that feels this way, please say.

awwright: That's exactly the word I was looking for, actually

awwright:
@Relequestual How's it look now

Relequestual:
As such... I'm merging!

Issue #243, opened 2017-02-01
JSON pointer usage is not explained properly in core


epoberezkin:
I may be missing something, but all I could find is:

https://github.com/json-schema-org/json-schema-spec/blob/master/jsonschema-core.xml#L316
https://github.com/json-schema-org/json-schema-spec/blob/master/jsonschema-core.xml#L410

There seems to be nothing about:

- the fact that JSON pointers should be in hash-fragments
- them being URI-encoded

There are only some examples to suggest 1 and nothing to suggest 2...

Please do not publish it in such shape

@awwright @handrews @Relequestual

epoberezkin:
I think the problem here is that draft-04 is relying on JSON reference draft (https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03) so it was not necessary to explain the syntax and semantics of $ref in the core.

From all the discussions here it followed that the decision was to essentially include JSON reference in JSON schema, but it didn't happen. The reference to JSON reference was also lost. So now whoever reads the spec would have no way to understand the syntax and semantics of $ref unless they look in draft-04... /sigh

handrews:
@epoberezkin I have a change that I nearly finished a while ago to add the fragment syntax to the IANA media type declaration.  I'll tidy that up and post it.

We do not need to explain how to URI-encode JSON Pointers for fragments, as the JSON Pointer RFC already mandates that behavior.  We just reference the RFC.

awwright:
@epoberezkin Fragment handling is defined by the semantics media type. Of course, that means we need to define it for `application/schema+json`. I sort of thought `application/json` did this, but it turns out there's no such registration.

I don't think this is a roadblock, but it does need to happen before media type registration with the IANA.

epoberezkin:
@awwright @handrews thank you for hte answer.

> Fragment handling is defined by the semantics media type.
> I have a change that I nearly finished a while ago to add the fragment syntax to the IANA media type declaration.

It is a bit over my head (and probably a couple of other heads too)... In any case just saying something about it in the spec with the link(s) to the relevant document(s) or whatever would point people in the right direction rather than leaving them guessing what it is about and how it is connected.

Without some clarification I think there will be many other people who won't be able to join the dots in the same way I can't.

handrews:
I've made a first pass at the proper fragment definition.  I tried to follow all the best practices I could fine but it's probably not perfect.  I think as long as it's not egregiously wrong we should accept it and refine it further in the next draft.

[...]

Issue #144, opened 2016-11-16
Fragments for schema+json media type:  Need to add to IANA Considerations


handrews:
Currently the section on "id"  (or "$id" if the recent PR goes through) shows the keyword being used to define simple one-word fragment identifiers such as `{"id": "#bar"}`, while also showing the use of fragments with JSON Pointer such as `"#/definitions/B"`.

What are the official rules for fragments for the `application/schema+json` media type?  Do we want to standardize on JSON Pointers?  Or do we support both of this by saying that if the fragment begins with a `"/"` then it should be interpreted as a JSON Pointer, but otherwise it should be looked up as an `"id"`?  Defining an `"id"` that appears to be a JSON Pointer but does not point to the current schema would either have an undefined effect or produce an error, presumably.

@awwright I know we were talking about fragments for the media type but I don't recall exactly what, if anything, we decided to do.  I just remember being confused about whether JSON Pointers were generally the fragment type for `application/json` (for everyone else: no, they aren't, the JSON Pointer specification says so directly).

handrews:
Reading up on http://www.w3.org/TR/fragid-best-practices/ and looking at how application/xml and application/*+xml fragments are defined in https://tools.ietf.org/html/rfc7303 it seems clear that having both of these fragment identifier styles should be just fine (although I still need to do a bit more reading).

Assuming I don't find anything to the contrary, I will write up a PR adding a formal fragment identifier specification for application/schema+json.

awwright:
It's something we'll have to build out in the IANA Considerations section.

In the absence of a section describing fragment interpretation for `application/schema+json`, implementations should be reading the fragment as a JSON Pointer.

handrews:
Based on the best practices doc and what I see in the XML RFC and others, it shouldn't be too hard to write this up so I'll give it a shot.  We do not need to define any new fragid structures as we are just using plain name fragids in a way that is completely consistent with their use elsewhere, and we're using JSON Pointer which is already a standard which addresses its own use as a fragid structure.

handrews:
Addressed by PR #245

handrews:
PR #245 merged.  While there may be more details to address in the IANA registration, this gets us where we need to be for now.

PR #245, opened 2017-02-02
Clearly define schema+json fragid forms.

handrews:
This addresses issue #243. (edit: and #144)

Reference the JSON Pointer RFC for using JSON Pointers
as fragids (includes encoding requirements).

Remove the syntax description for how to write a fragment
URI reference (leading "#") as that is covered by RFC 3986.

Also fill in a bit more of the IANA considerations section.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3,12 +3,14 @@
 <!ENTITY RFC2119 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.2119.xml">
 <!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
 <!ENTITY RFC5988 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5988.xml">
+<!ENTITY RFC6839 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6839.xml">
 <!ENTITY RFC6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
 <!ENTITY RFC6906 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6906.xml">
 <!ENTITY RFC7049 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7049.xml">
 <!ENTITY RFC7159 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7159.xml">
 <!ENTITY RFC7231 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7231.xml">
 <!ENTITY ldp SYSTEM "https://xml2rfc.tools.ietf.org/public/rfc/bibxml4/reference.W3C.REC-ldp-20150226.xml">
+<!ENTITY fragid-best-practices SYSTEM "https://xml2rfc.tools.ietf.org/public/rfc/bibxml4/reference.W3C.WD-fragid-best-practices-20121025.xml">
 ]>
 <?rfc toc="yes"?>
 <?rfc symrefs="yes"?>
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -357,15 +392,26 @@
                     This value SHOULD be normalized, and SHOULD NOT be an empty fragment &lt;#&gt; or an empty string &lt;&gt;.
                 </t>
                 <t>
-                    The root schema of a JSON Schema document SHOULD contain an "$id" keyword with an absolute-URI (containing a scheme, but no fragment).
+                    The root schema of a JSON Schema document SHOULD contain an "$id" keyword with 
+                    a URI (containing a scheme).  This URI SHOULD either not have a fragment, or
+                    have one that is an empty string.
+                    <!-- All of the standard meta-schemas use an empty fragment in their id/$id values. -->
+                    <cref>
+                        How should an "$id" URI reference contianing a fragement with other components
```
epoberezkin: contianing -> containing

handrews: Fixed!  Thanks.

handrews:
@epoberezkin I fixed the typo, does this otherwise seem like a good step forward?  Even if it is not perfect.

epoberezkin:
@handrews yes, thank you

handrews:
@awwright ? @Relequestual ?

Relequestual:
I *THINK* I understand this change, but possibly not fully understand all of the possible implications. I'd really like @awwright to weigh in on this. However, if this is for draft-6, he can just review it as part of the final review process.

handrews:
@awwright could you please review/approve as @Relequestual asked?  @epoberezkin really wants to see this in Draft 06, and I support it (but would like to get it resolved- it's been open for more than two weeks).

handrews:
@awwright any comments?  This is the last PR currently open that is marked for Draft 06 and it is blocked awaiting your feedback.  For over 3 weeks now.

jsonschema-core.xml
```diff
@@ -363,15 +398,26 @@
...
-                    followed by a letter ([A-Za-z]), followed by any number of
-                    letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), or periods (".").
-                    <!-- This restriction is the same one defined by XML -->
+                    This is done by setting "$id" to a URI reference consisting only of a fragment.
+                    The fragment identifier MUST NOT match the syntax of a JSON Pointer fragment
+                    identifier.
+                </t>
+                <t>
+                    The effect of defining an "$id" that is appears to be a JSON Pointer that does not match
+                    the expected JSON Pointer for that schema is not defined.
```
awwright: I'm wary of changing this section (on the definition of the $id keyword) to liberalize what's acceptable. The specification should positively define what's acceptable in a regular grammar, without excluding a referenced set/relying on a compliment.

Otherwise, it might be the case that if the current RFC for JSON Pointer is superseded, that changes the behavior of this specification in a possibly incompatible way. The current behavior is the same restriction found in other documents like XML (that's also legal in a 7-bit URI), so the restriction shouldn't come as much of a surprise.

If JSON gets a fragment identifier defined, I don't want to assume it looks anything like JSON Pointer. I suspect it might operate like e.g. text/plain (defined in [RFC5147](https://tools.ietf.org/html/rfc5147) where it introduces a format like `#line=20` that's incompatible with e.g. XML's Name production.

Otherwise I like this PR. I'm trying to come up with some other comments, but this is turning into me re-reading XML and HTML.

handrews: @awwright I'm happy to restore the XML-derived restrictions on plain name fragments.  The possibility of JSON defining something different is a really good point.

handrews:
@awwright I've restored the exact restrictions that were previously present on plain name fragids, although I otherwise kept the new wording.  I think this should address your concern, while also still addressing @epoberezkin's concerns over how re-stating what a fragment is (leading "#") was confusing.  URI stuff references RFC 3986 which is correct, everything else is now explicitly defined either in the text or by referencing the JSON Pointer RFC.

PR #234, opened 2017-01-25
Another attempt at clarifying objects/arrays.

handrews:
**NOTE:**  I'm putting this up because @awwright and I both thought that my last attempt to clarify this still wasn't quite getting it.  Since this moves a lot of things around the diff looks **much** worse than the changes actually are.  It's better to grab this XML file and build it locally to see the changes.

While I like this approach better, I do not feel strongly enough about it to insist on it (or anything like it) going in for Draft 06.  I'm just posting it as an idea, and if it doesn't get support quickly I'll be happy to drop it.

----

This goes back to Draft 04's approach of grouping keywords,
but mostly retaining Draft 05's more concise wording and
per-keyword organization.

The general considerations section on validating primitive types
and child values has been simplified, and the details have been
moved down to each grouping of array or object keywords.  This
puts the information near where it is most relevant, which
allows removing some of the more confusing working in the
keywords that apply to child values.

PR #247, opened 2017-02-05
Validation: items/properties/patternProperties language changes

awwright:
This revision is intended to simplify/clarify the behavior for the definitions of several keywords. It removes references to "child validation" which is undefined and not any special kind of validation. It reworks the properties/patternProperties definitions to simplify language regarding which instances are processed/applied against which schemas.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -429,6 +429,10 @@
 
             <section title="properties">
                 <t>
+                    The "properties" keyword defines schemas to be validated against certain values in an object,
+                    based on an exact match to the property name.
+                </t>
+                <t>
                     The value of "properties" MUST be an object.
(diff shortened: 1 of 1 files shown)


awwright:
@handrews I'm really not sure which distinction you're making, can you elaborate some?

If we want to make it clear which parts of the keyword value are applied/mapped to which instances, well, we just have to describe how that works.

handrews:
> If we want to make it clear which parts of the keyword value are applied/mapped to which instances, well, we just have to describe how that works.

PR #234.

Relequestual:
I'd really much rather see https://github.com/json-schema-org/json-schema-spec/pull/234/ merged FIRST as I'm positive it's the right direction. Then we can look at this change after you've rebased if it still applies. I'm finding it hard to this change given #234

Issue #244, opened 2017-02-01
[draft-06 core] important - changes semantics of $schema


epoberezkin:
Draft 4 is saying:

> This keyword MUST be located at the root of a JSON Schema... It is RECOMMENDED that schema authors include this keyword in their schemas.

It means that it can’t be anywhere else (in subschemas), but it doesn’t mean it is required.

Draft 6 turns it upside down:

> The root schema of a JSON Schema document SHOULD use this keyword

That means that it can be located in subschema, which means different validation process for schemas against meta-schemas from the validation process of data against schema.

I don’t understand the motivation to allow $schema in subschemas. If one needs a schema with a different meta-schema it can be put in a different file and referenced. Mixing meta-schemas in one file is as bad as mixing javascript and typescript in the same file (if it were even possible) - why would anyone want to do it?

@awwright @handrews @Relequestual

epoberezkin:
Different validation process is not a trivial issue.

It makes meta-schema a thing of its own, essentially not a "schema" but something different. And if "meta-schema" is not a "schema" that instead of being called "meta-schema" (= "schema for schema") it should be called "schema definition file" - a different vocabulary altogether...

handrews:
Since the presence or absence is a RECOMMENDED, but the _location_ is a MUST, I agree with @epoberezkin that this is a change from Draft 04 that, AFAIK, was not discussed during Draft 05 so is reasonable to discuss now.

I have implemented $schema as it is now (allowed anywhere).  It does add some complexity and annoyance,  but orders of magnitude less than $id.

I only know of two plausible use cases for non-root $schemas:
* bundling schemas of multiple versions into the same file.
* selectively enabling hyperschema for some but not all of a schema.

As someone working primarily with hyperschema, I don't personally need that level of control- if I don't want hyperschema in part of a schema file, I just won't use its fields.

I have no specific need to mix versions either.  If someone does I would like to hear from them before we put a lot of effort into supporting it.

---

I also have concerns about how we are using $schema to indicate feature sets, which we also encourage people to extend.  That is not directly relevant to this issue, but it means that for future drafts, we need to discuss the role of $schema a bit more.  So I would support resetting $schema back to its Draft 04 usage for Draft 06, and taking a more thorough look at it later when we are not so close to publishing a much-awaited new version.

Relequestual:
[...]

Issue #151, opened 2016-11-21
Improve handling of profiles vs. sets


awwright:
JSON Schema is conflating two mutually exclusive ideas of class membership, and it's causing a handful of problems around "additionalProperties", especially when you have `"additionalProperties": false`

## Types of taxonomy relationships

Associating an instance as a member of a group is done one of two ways:

0. Sets - A JSON schema that identifies a set of JSON documents, where any JSON document that validates against the schema is a member of the set, even if the data has no semantic value or makes no sense.
0. Profiles - A JSON document has a profile only when explicitly indicated as such. A profile may additionally contain restrictions on what data the JSON document may have.

These forms of association exhibit different behaviors.

### Sets

Sets are motivated by the mathematical concept of a set and members of a set.

An instance in a set is uniquely identified by its contents.

JSON Schema can be used to create subsets - in which every member of a subset is by definition also a member of the superset.

If X and Y are sets described by schemas, where Y is a subset of X, then any instance of Y will also be an instance of X.

### Profiles

Profiles are similar in function to media types. A media type is a short, standardized string that associates with a string of octets to form a document. Profiles are somewhat more abstract as a concept.

Profiles are often used to tell people they can treat the document a certain way, that they couldn't otherwise, even if it happened to otherwise look like a valid document. For example, a media type or profile lets people know if the document is executable or not -- an important security measure to ensure that plain text documents can't get executed as a program! (See also "polyglot programming" where a program is valid in two different languages, possibly one benign and one malicious)

An instance of a profile can determine how it is uniquely identified; frequently a URI is used as an identifier, and there can be multiple instances with identical data.

A resource can be described my multiple profiles. Rules can be applied to profiles and implications made about their membership. If X and Y are profiles, where Y is a subclass of X, any instance of Y is also an instance of X.

Sometimes we want to use JSON to describe properties only when they're an instance of a profile, using this subclass logic. For example, instances of X have property A, and A is only found in instances of X; instances of Y have property B, and property B is only found in instances of Y.

How do we describe data like this? There's a few options:

0. Use one JSON document per profile per resource. If I have a resource Q that is an instance of Y, then have two JSON documents <Q.X.json> and <Q.Y.json>.

0. Allow additionalItems on a document, and ignore unknown properties. List the document having every profile, including superclasses (though it might sometimes be possible to omit superclasses or any profiles implied from other profiles). Properties across profiles must not overlap.

[...]

PR #248, opened 2017-02-05
Core: Restrict usage of $schema some

awwright:
PR for Issue #244

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -301,10 +301,16 @@
                 written for this particular version.
             </t>
             <t>
-                The root schema of a JSON Schema document SHOULD use this keyword.
                 The value of this keyword MUST be a <xref target="RFC3986">URI</xref>
                 (containing a scheme) and this URI MUST be normalized.
                 The current schema MUST be valid against the meta-schema identified by this URI.
+            </t>
+            <t>
+                The "$schema" keyword SHOULD be used in a root schema.
+                It MUST NOT appear in subschemas.
+            </t>
+            <t>
+                <cref>While this pattern is likely to remain best practice for schema authoring, implementation behavior is subject to be revised or liberalized in future drafts.</cref>
             </t>
             <t>
                 Values for this property are defined in other documents and by other parties.


seagreen:
Please ignore this if it's not helpful, I'm going to keep being a contrarian...

I don't believe in self-describing data like $schema. I think this information is better kept outside of the document itself.

To explain why, say you have a schema that's valid against draft 12. Then draft 13 comes out. Draft 13 has only very small changes, and you're schema is valid when used with both 12 and 13.

If the schema doesn't declare what draft it uses, then it can be referenced in the context of draft 12 and draft 13 validation perfectly fine. So what was gained by declaring `"$schema": "draft12"` in the first place? It would be like a piece of JSON data declaring exactly which JSON Schema document it matches -- this wouldn't be a good idea because there might be many different JSON Schema documents used to validate that instance over its lifetime of use.

So I think this should read: `The root schema of a JSON Schema document MAY use this keyword.`

awwright:
@seagreen This is perhaps a better discussion for a new Issue, however;

It has the same effect as declaring usage of a profile or media type, it is not merely asserting restrictions that the document must follow, it is also declaring which behaviors those keywords will have. #151 seems slightly relevant to this.

If a simple document can be described under two profiles (meta-schemas), then... it can be described under two profiles, and I don't really see how that's a bad thing.

I can see some arguments that this solution doesn't _need_ to rise to the level of being self-describing, but that's different from saying that's a negative property of a system. (And it very may well be a negative property, to be sure, look at Gödel's incompleteness theorem.)

Why the downgrade from `SHOULD` to `MAY`?

handrews:
@seagreen I have a number of concerns around $schema and how we indicate vocabularies, extension vocabularies, and versions of those things.  But that's a big discussion.  Can we please go forward with this for Draft 06?  I don't recall seeing any complaint about this in any issue in either repo up until now, which suggests to me that this is _not worth dragging out the next draft_ for.

I would love to track all of the ideas and concerns as their own separate issues which we can discuss for future drafts.  But revamping the entire versioning approach is not going to happen in the next week or so.

seagreen:
@handrews: Absolutely go ahead! Thanks for asking, you obviously didn't have to. Once you reassured me that these are actual drafts and not meant to be in prod for years like Draft 4 my concerns went away.

@awwright: Sorry I'm taking my time to get back to you, I'm really looking forward to thinking about #151 in depth, but haven't had the time yet. Once I do I'll open an actual issue for my concerns with $schema, so as far as I'm concerned this PR is good to go.

PR #256, opened 2017-02-16
Validation: s/child validation/validation/

awwright:


Relequestual:
> The "child validation" phrasing could be interpreted as indicating
> a separate validation process, which is incorrect.  Removed all
> remaining references to "child validation".
Extracted from https://github.com/json-schema-org/json-schema-spec/pull/234/commits/ac60d16f5cccb4928d4fa3761770f99bb3617a79 as rational by @handrews

Issue #260, opened 2017-03-01
What are 'defaultItems' ?


mspiegel:
https://github.com/json-schema-org/json-schema-spec/blob/master/schema.json#L36 a search in this GitHub repository and a quick Google search does not turn up any results.

handrews:
@mspiegel that is a typo :-)  Thanks for reporting it!

PR #262, opened 2017-03-01
"defaultItems" should be "default"

handrews:
...because "defaultItems" is not actually a thing.

Addresses #260, with thanks to @mspiegel


Relequestual:
Fixes #260

Issue #208, opened 2016-12-28
Update meta-schema URIs before publishing draft 6


handrews:
Currently the meta-schemas referenced in the validation and hyper-schema specs on master reference the Draft-04 meta-schemas (which are in fact the most recent published).

We need to update these before publishing these specs as draft 6.  Since we cannot update the draft after publication, that means we MUST do these things in order just before submitting:

1.  Put the draft-06 meta-schemas up on the web site
2. Update the specifications to the new meta-schema URIs
3. Submit specifications to IETF

handrews:
Done and published!

Issue #206, opened 2016-12-27
Draft 6 publication checklist


handrews:
This tracks all of the stuff around getting Draft 6 out the door other than actually submitting it to the IETF as a new personal draft.  It's mostly to help me remember what to review during the final review period.  Plus some stuff that's really obvious but felt like it should be on the list :-)  I don't expect anyone else to take action on this issue, but please comment if you know of something that does *not* make sense as an issue/PR against the spec that should get tracked.

- [x] Test cases for Draft 6 (and Draft 5 as needed)
- [x] Final review:  Verify that the meta-schema and specification match
- [x] Final review:  Overall spelling, grammar, and wording consistency check
- [x] Final review:  Verify that all examples make sense and match current specs
- [x] Final review:  Verify that all notable changes are in the changelog section
- [x] Submit drafts to IETF
- [x] Update the web site
- [x] Announce to mailing list

awwright:
@handrews The process I'll follow is listed on the wiki page, take a look at that and let me know if you think that's missing anything from here.

handrews:
@awwright the main point here is the final review stuff.  That is not in your process, and wikis don't track anything anyway.  You don't need to do anything for this issue, so if you don't find it useful please just ignore it.

Relequestual:
I agree with this list of requirements before rubber stamping the draft.

Relequestual:
@handrews Can any of these checks be checked? =]

Relequestual:
@epoberezkin How are the updates to the tests coming along?

epoberezkin:
@Relequestual They are coming ok, not too much left to do: https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/153

handrews:
@epoberezkin can you verify that the schema.json and hyper-schema.json currently on master validate themselves?  And the hyper-schema.json validates links.json?  If so we can copy them over to the web site repo and update the URIs in preparation for release.

epoberezkin:
@handrews there are the following problems:

- [ ] schema.json (meta-schema)
  - [ ] $id should be version specific, we have agreed not to use unversioned URIs I think
- [ ] hyperschema
  - [ ] $id should be version specific
  - [ ] unless I am missing something, inside /definitions/schemaArray should be anyOf and each subschema inside anyOf should be similar to schemaArray definition in meta-schema
[...]

Issue #165, opened 2016-11-24
Publication of meta-schema


awwright:
#164 adds a paragraph formalizing the meta-schema URI that identifies the semantics defined in the document.

However, it doesn't provide the meta-schema itself. If we want to normatively publish the schema, it would have to be included in the document text. This is exactly what's done for some XML/XSD schemas.

I'm thinking it's a better idea if we keep the meta-schema as an informative document - it's just a special value that validators recognize, and they can use the one provided informatively by us, or they can even write their own meta-schema, or hard-code the rules in logic... whichever option follows the behavior described in the respective documents.

handrews:
While I'm still learning the more subtle points of normative vs informative pieces, this sounds reasonable to me.  As long as we publish them clearly with the next draft and show that correlation so folks can keep track, I'll be happy.

Relequestual:
I think this is a great idea. We should totally do this.
It would need to apply with draft-6 onwards. I don't feel we can post release the meta-schema after draft-6. As you say, put it in the document itself.

Even if it doesn't end up in the document, it can be bundled in a github tagged release (which if we want, can have a DOI number, if we want to make sure it's kept safe externally.)

Relequestual:
@awwright I'm happy for it to not be part of the document if you feel it isn't required, however I would say it should at the very least be included as part of the release bundle available on github.

Relequestual:
I suggest we keep the schema documents as informative documents.

Because we have a chicken and egg situation with referencing the tagged "release" on github, I suggest in stead we say something like, "The meta-schema/s can be found in the Github tagged release at (link to repo releases)". Thoughts @awwright ?

Relequestual:
@awwright Your comments on this are needed. While this doesn't strictly prevent publication of the document, it does because we want to make the documents in a github release the ones that are uploaded to the IETF site. We need to resolve this issue before we seek publication.

Relequestual:
For now...
After a brief chat with @handrews , we agreed we can, for now at least, just copy over the meta-schema from this repo to the website repo, as needed. Changes to the meta-schema will still be managed from the spec repo.

handrews:
I think the publication of draft-06 with the meta-schemas and documented links as they are answers this, doesn't it?  Please re-open (and take out of the draft-06 milestone) if not.

PR #265, opened 2017-03-02
Update for new meta-schemas.

handrews:
I've asked @epoberezkin to verify that these validate themselves (see https://github.com/json-schema-org/json-schema-spec/issues/206#issuecomment-283751327 )

Once we are certain of that, we can copy them over to the web site repo and merge this change, as discussed in issue #165 

This PR directly addresses #208

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -210,10 +210,7 @@
 
         <section title="Meta-schema">
             <t>
-                The current URI for the JSON Schema Validation is &lt;http://json-schema.org/draft-04/hyper-schema#&gt;.
-            </t>
-            <t>
-                <cref>A revision describing newly added keywords will be added in the future.</cref>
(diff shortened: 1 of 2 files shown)


handrews:
Note that @epoberezkin validated the meta-schemas with Ajv, leading to fixes merged in PR #266.
PR json-schema-org/json-schema-org.github.io#82 puts the appropriate files in place in order for these URIs to work.

epoberezkin:
@awwright @handrews @Relequestual why is it not merged/approved? pre-release document still points to draft-04 schema:

https://github.com/json-schema-org/json-schema-spec/blob/20170309/jsonschema-validation.xml#L230

etc.

awwright:
Oh I'm sorry, I didn't see see this in between the prolific activity on the
test suite...

Yeah, this is good, the intention is to update this at the same time as we
copy it into archives/

(Feel free to update the PR with that, I'll make the change before
publishing if not.)

PR #266, opened 2017-03-02
Add meta-schema version numbers, fix reference.

handrews:
We are not using "latest"-style URIs anymore.

schemaArray in hyper-schema is intended to reference the schemaArray
in schema and just "allOf" the hyper-schema {"$ref": "#"} into the "items".

Addresses https://github.com/json-schema-org/json-schema-spec/issues/206#issuecomment-283775164

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -1,11 +1,11 @@
 {
-    "$schema": "http://json-schema.org/draft/hyper-schema#",
-    "$id": "http://json-schema.org/draft/hyper-schema#",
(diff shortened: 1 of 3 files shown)


handrews:
Paging @epoberezkin since I can't officially assign him as a reviewer (we should fix that).

epoberezkin:
lgtm, see https://github.com/json-schema-org/json-schema-spec/issues/206#issuecomment-283791054

handrews:
Thanks @epoberezkin !

Since this is purely a fix for bugs found by code, and was verified as fixed by that same code, I am going to go ahead and merge it.

PR #268, opened 2017-03-03
Fix hyper-schema meta-schema self link.

handrews:
"id" became "$id", and we handle "$" through percent-encoding

URI Templates reserve "$", but do not specify how percent-encoding
is to be used, except that implementations should be consistent about
it.  We document that we always percent-decode variables before
resolving them.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -129,7 +129,7 @@
     "links": [
         {
             "rel": "self",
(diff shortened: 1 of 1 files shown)


handrews:
@awwright yeah, definitely ugly, but we intend to revisit this sort of thing soon anyway.  This will get us through "draft-06"

PR #275, opened 2017-03-20
Fix missing </t>

handrews:
Not sure how this got in, but it is breaking the build now.


handrews:
I'm going to merge this because I can't make other pull requests without it.

PR #281, opened 2017-03-22
Add cref describing the situation with HTML forms

handrews:
The specification is currently in a slightly awkward state of both
drawing analogies with HTML forms and supporting a different, more
explicit paradigm with "hrefSchema".

Add a note acknowledging the confusion and providing some context
for the current state of things.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -847,6 +847,23 @@ GET /foo/
 
             <section title="Submission Form Properties">
                 <t>
+                    <cref>
+                        While this section curently draws explicit analogies to HTML forms,
+                        we expect significant changes to this analogy in future drafts.
+                        In particular, "hrefSchema" removes the need for "schema" to
+                        do double duty of either describing URL query parameters or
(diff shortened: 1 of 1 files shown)


handrews:
Paging @jdesrosiers to see if this explanation will help resolve #280.

jdesrosiers:
I have no objection.  I agree that we are moving away from LDO having direct parallels to a HTML Forms.  

But, I don't think the form analogy is the root problem with #280, even though I used it to express my thoughts at one point.  My latest comment on that PR I think better explains the problem and does it without the form analogy.

handrews:
@jdesrosiers agreed about this and #280 given our subsequent discussions.  I still think this would be a useful note, as long as @awwright and @Relequestual agree, so I'll leave it open.

handrews:
I'm going to close this because of #292 and #293, and the plan to add a third PR to address the `"form"` use case.  Something like this should probably be included in the `"form"` PR, but it will look different from this so let's get this out of the way to reduce noise.

PR #282, opened 2017-03-23
Add non-http "method": "post" LDO example

handrews:
**_This PR need not necessarily go into Draft 06_**

I'm posting this mostly to illustrate some of the arguments in #280, particularly around avoiding coupling of JSON Hyper-Schema and any particular URI scheme or protocol.

If we can't figure out a reasonable mapping for `mailto:` then we might be missing something.  Although that something will probably need to wait until Draft 07 for a fix.

I am aware that it's more common to use an HTTP request to ask the server to send mail, but I think this still makes a good example.  If the client for this resource runs in a known environment, it could be reasonable to expect the client to send mail if it wants to.

---

Add an example showing how a non-HTTP URI scheme can be used,
showing a form for constructing an email with two different
representations, as well as requiring a subject.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -918,8 +918,53 @@ GET /foo/
 ]]>
                             </artwork>
                         </figure>
-
+                    </t>
+                    <t>
                         If the method is "post", "application/json" is the default media type.
+                    </t>
+                    <t>
+                        As noted under <xref target="method">method</xref>, these fields
+                        are not restricted to HTTP URIs.
+
+                        <figure>
+                            <preamble>
+                                For example, this link indicates that if you want to
+                                send an email to the author of the context resource,
+                                your client needs to ask for both a plain text
+                                and an HTML representation.
+                            </preamble>
+                            <artwork>
+<![CDATA[{
+    "links": [{
+        "encType": "multipart/alternative; boundary=abc123",
+        "method": "post",
+        "rel": "author",
+        "href": "mailto:someone@example.com{?subject}",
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -918,9 +918,54 @@ GET /foo/
...
+                {
+                    "type": "string",
+                    "media": { "type": "text/plain; charset=utf8" }
+                },
+                {
+                    "type": "string",
+                    "media": { "type": "text/html" }
+                }
+            ],
+            "minItmes": 2
```
dlax: typo: minItems

handrews: heh.  Maybe I should remember to validate my examples against the meta-schema :-P

jdesrosiers:
I think this makes sense.  This is a lot like how you would do it with an HTML Form.

handrews:
@awwright if you agree with @jdesrosiers that this is reasonable, should we go ahead and add it to Draft 06?  It shouldn't change anything, only clarify.

awwright:
It's maybe a bit more verbose than _ought to_ be necessary, but still, a good illustration of this behavior is probably warranted.

handrews:
@awwright I tend to be verbose, unfortunately.  Anything in particular you'd trim?

handrews:
@awwright did that comment count as an approval?

awwright:
I think I just needed to sit on it another day. Looks good.

handrews:
@awwright thanks!  Totally reasonable, just wasn't quite clear :-)

Issue #288, opened 2017-03-30
Why does hrefSchema consider instance data last?


jdesrosiers:
> An "href" with at least one template variable and with an "hrefSchema" allows using external data to resolve the template, and falls back to resolving any remaining variables from the instance.

What is the reason for resolving instance variables last?

It seems to me that it should be the other way around.  I used to be able to resolve a hyper-schema against it's instance JSON and get a set of links that I can use to make requests with no dependency on instance data.  Now making a request has a dependency on instance data that I didn't used to have.

handrews:
@jdesrosiers I am re-writing this whole section b/c the changes around forms require a different overview.  What you are worried about is not what I meant (I'm pretty sure) so I will fix the wording- thanks for filing this!  We'll leave it open until you can verify the PR (I'll put this in the Draft 06 milestone).

The idea is:
* If there is no "hrefSchema", or if "hrefSchema" is `false` (or any other value that will never successfully validate anything), then client input is not allowed.  The behavior is identical to what was present before "hrefSchema" was introduced.

* For a given variable, if it corresponds to a `false` or otherwise never-validating subschema of "hrefSchema", then that variable does not allow client input.  It behaves as it would have in earlier drafts.

* For a given variable, if there is a subschema in "hrefSchema" that will successfully validate some instance(s), then **if** the client supplies input, it is used.  Otherwise it resolves from the instance.

* For a given variable, if there is a subschema allowing input, and the property matching that variable is marked as required in "hrefSchema", then client input MUST be used in order for the link to resolve.

Key observation:  Unless you make fields in "hrefSchema" required, you can *completely* ignore it and everything resolves from instance data by the same rules there were before.

Does that address your concern?

If you want to pre-populate input, but allow the client to override it, that gets into the use of "default", which is it's own otherwise unrelated morass, see #204.

jdesrosiers:
No, this doesn't change anything.

> if the client supplies input, it is used. Otherwise it resolves from the instance

I can't resolve variables from instance data until after I have client input.  Links shouldn't be dependent on instance data.  In Hyper-Schema, LDOs are more like link templates.  They must be resolved against instance data before they can be used.  Once they have been resolved, they can be used like regular links where the only dependency is client input.  That is no longer possible with the way `hrefSchema` is defined.

handrews:
> I can't resolve variables from instance data until after I have client input.

[...]

PR #291, opened 2017-03-31
Clarify options for resolving template variables

handrews:
Addresses #288 

If a variable can be resolved from either the instance or from
client data, the instance data may be used as a default when
asking for client data.

Also clarify how "hrefSchema" can handle a server-computed
URI path with client-supplied data string, as in traditional
HTML "get" forms.  This should address @awwright's
https://github.com/json-schema-org/json-schema-spec/issues/288#issuecomment-290522156

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -463,6 +463,15 @@
                             unresolved MUST be resolved from the resource instance data.
                         </t>
 
+                        <t>
+                            If a URI Template variable corresponds to both an instance value
+                            and a schema for external data, an application MAY choose to
+                            present the instance value to the user as a default value.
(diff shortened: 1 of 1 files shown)


jdesrosiers:
I might be missing something, but I don't see how this changes anything.  That dependency is still there.  If anything this just seems to acknowledge that the dependency exists.

handrews:
Given the state of things in #288, I am inclined to discard this unless @awwright or someone else thinks it is useful independent of the concerns in that issue.

handrews:
I've decided that this is probably just confusing, and given that we don't have consensus on this as a desired interpretation it's better to leave it open and see what sort of feedback we get.  In particular, we already have several discussions going around default values, and hopefully one of those will result in more philosophically consistent solutions.

PR #293, opened 2017-03-31
Rework the descriptions of using data with links

handrews:
***IGNORE the first commit, it is from PR #292***

Review this commit here: https://github.com/json-schema-org/json-schema-spec/pull/293/commits/af5ef3a9c98618ec34906d5e409562baa2cc9451

---

This replaces the rather lengthy discussion of hrefSchema
with a more concise and complete description of all of the
ways to use data with links.  This should now serve to set
the context for the various ways to use LDOs.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -399,7 +399,7 @@
                     There are several ways that a client can use data can with a link:
                     <list>
                         <t> URI Template variables resolved from server-supplied instance data </t>
-                        <t> URI Template variables resolved from client data </t>
+                        <t> URI Template variables resolved from user agent data </t>
                         <t> Replacing or modifying the target resource's representation </t>
                         <t> Submitting data for processing, where the data has no
                             inherent relation to the target resource's representation</t>
@@ -423,17 +423,17 @@
                         <xref target="hrefSchema">"hrefSchema"</xref> allows a link to specify
                         a schema for resolving template variables from client-supplied data.
                         Regular JSON Schema validation features can be used to require resolution
-                        from client data, forbid it, or allow client data while falling back to
-                        server-supplied instance data if no client data is provided.
+                        from user agent data, forbid it, or allow user agent data while falling back to
+                        server-supplied instance data if no user agent data is provided.
                     </t>
                     <t>
                         The common pattern of resolving a templated path component with
-                        server-supplied instance data while accepting client data to build
+                        server-supplied instance data while accepting user agent data to build
                         a query string can be implemented by setting the "hrefSchema" subschemas
                         for the path template variables to false, while giving the query string
                         template variables names that do not appear in the instance.  This ensures
                         that the path variables can only be resolved from the instance, and the
-                        query string variables can only be resolved from client data.
+                        query string variables can only be resolved from user agent data.
                         See the "hrefSchema" section for an example of this approach.
                     </t>
                 </section>
@@ -491,7 +491,7 @@
                     <section title="Values for substitution">
                         <t>
                             The URI Template is filled out using data from some combination of an external source and the instance.
-                            Where either instance data or external data may be used, this section will refer simply to "data" or to a "value".
+                            Where either instance data or user agent data may be used, this section will refer simply to "data" or to a "value".
                             When the source is important, it is specified explicitly.
 
                             To allow the use of any object property (including the empty string) or array index, the following rules are defined:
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -388,32 +387,81 @@
...
+                        that the path variables can only be resolved from the instance, and the
+                        query string variables can only be resolved from client data.
+                    </t>
+                </section>
+                <section title="Manipulating the target resource representation">
+                    <t>
+                        In JSON Hyper-Schema, <xref target="targetSchema">"targetSchema"</xref>
+                        supplies a non-authoritative description of the target resource's representation.
+                        A client can use "targetSchema" to structure input for replacing or
+                        modifying the representation.  Alternatively, if "targetScheam" is absent
```
jdesrosiers: typo: "targetScheam" => "targetSchema"

jsonschema-hyperschema.xml
```diff
@@ -863,81 +911,30 @@ GET /foo/
                 </section>
             </section>
 
-            <section title="Submission Form Properties">
+            <section title="submissionEncType" anchor="submissionEncType">
                 <t>
-                    The following properties also apply to Link Description Objects, and provide functionality analogous to <xref target="W3C.CR-html5-20140731">HTML forms</xref>, by providing a means for making a request with client- or user-selected information.
+                    If present, this property indicates the media type format the
+                    client should use to for the request payload described by
```
jdesrosiers: typo ?

handrews:
@jdesrosiers thanks for catching those!  Should be fixed now.

handrews:
@awwright this and #292 are awaiting your approval, at which point I think we are good to go with draft-wright-json-schema*-01

handrews:
@awwright this PR now only has the one commit- I've rebased it on top of the merged #292 and #296.  Should be more clear to review now.

awwright:
@handrews Thanks, you read my mind

jsonschema-hyperschema.xml
```diff
@@ -387,38 +387,87 @@
...
+                    </cref>
                 </t>
                 <t>
-                    A link with a "submissionSchema" allows submitting data for processing.
+                    Link Description Objects do not directly indicate what operations, such
+                    as HTTP methods, are supported by the target resource.  Instead, operations
+                    should be inferred primarily from link <xref target="rel">relation types</xref>
+                    and URI schemes.  Note, however, that a resource may always decline an operation at
+                    runtime, for instance due to application state that controls the operation's
+                    availability.
```
awwright: It seems like we're comparing the LDO against something (HTML forms?), but there's no indication of what that is.

I would move this paragraph down one so we lead with what it does, not with what it doesn't do.

jsonschema-hyperschema.xml
```diff
@@ -387,38 +387,87 @@
...
                 </t>
                 <t>
-                    See the individual keyword descriptions below for details related to each of these cases.
+                    There are several ways that a client can use data can with a link:
+                    <list>
+                        <t> URI Template variables resolved from server-supplied instance data </t>
+                        <t> URI Template variables resolved from client data </t>
```
awwright: These two lines might be too easy to conflate. I would say something like

* Plain links, where all variables in the URI Template (if any) are filled from instance data
* Link templates, where the URI Template contains unbound variables to be filled in by users

handrews: Your phrasing is not clear to me, and I don't have any intuition for "Plain links" vs "Link templates", especially because your plain links obvious involve template variables.

The distinction is not across the template, it's that there are different ways to use template variables, which might appear in any combination in a template.

jsonschema-hyperschema.xml
handrews: Good point, fixed.

jsonschema-hyperschema.xml
awwright: Oh I see. I suppose that works.

What about "URI Template variables resolved from user input"?

handrews: > What about "URI Template variables resolved from user input"?

This is a good point:  I've avoided "user input" in favor of "client data" because of situations (like autonomous Things in the Internet of Things sense) where there is no direct user input, but there is data coming through the client from somewhere.

So I don't want to make it sound like there has to be an actual user interaction involved, but I do want the purpose to be clear.

client input?  user-agent input?

awwright:
Yeah I see exactly what you mean... I figure most people know a "user" can
be automated.

Technically, it's always an agent making the request and filling in the
variables, how about "agent"? Iirc that's how HTML deals with it.

handrews:
@awwright "agent" works for me, and as you note there is precedent.  I'll look through the spec and try to make sure the language is consistent as I know I have "client data" in some places, and "external data" in others.  And probably still "user input" somewhere.

I'll update the PR tomorrow.

handrews:
@awwright the latest commit changes everything to "user agent data".  I looked through the HTML5 spec and RFC 7231 and both consistently user "user agent" (never "agent" on its own).  So that seems the safest option.

Previously, I'd used several different phrases all over the place, so I went ahead and cleaned it all up.  If you want look at this part on its own the commit diff is:

https://github.com/json-schema-org/json-schema-spec/pull/293/commits/6bf4061f8cca35cd0b20a38f05920cbf5d015791

PR #300, opened 2017-04-05
hyperschema: remove obsolete references to core terminology definitions

dlax:
At the beginning of hyperschema document, there's:

> The terms "schema", "instance", "property" and "item" are to be interpreted as defined in the JSON Schema core specification.

but core specification does no longer defines "property" and "item" (from 3cbfc8bcab58ce6c0397f205534d2a4cede76ce8 AFAICT). I guess the former is assumed by JSON terminology; the latter may be confusing w.r.t. to collection+item terminology that appears in sections about links. So removing both references.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -106,7 +106,7 @@
             </t>
 
             <t>
-                The terms "schema", "instance" and "property" are to be interpreted as defined in the <xref target="json-schema">JSON Schema core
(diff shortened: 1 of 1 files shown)


handrews:
@dlax feel free to update this to add yourself to the acknowledgement section

dlax:
> @dlax feel free to update this to add yourself to the acknowledgement section

@handrews done in 627801e.

Relequestual:
OK by me!

PR #304, opened 2017-04-11
Allow booleans in "contains" per schema.json

pdl:
The documentation for "contains" appears to contradict `schema.json`; it says:

> The value of this keyword MUST be an object.  This object MUST be a valid JSON Schema.

However, `schema.json` refers to the root schema, which may be object or boolean.

It seems simplest to say:

> The value of this keyword MUST be a valid JSON Schema.

While `"contains": true` seems to me to have little value, given the general approach that `true` is a schema equivalent to the empty schema `{}` and `false` is a schema equivalent to `{ "not":{} }`, making this an exception to the overall rule seems an unnecessary burden, and I suspect it is an oversight.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -412,8 +412,7 @@
 
             <section title="contains">
                 <t>
-                    The value of this keyword MUST be an object.  This object MUST be
-                    a valid JSON Schema.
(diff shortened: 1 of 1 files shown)


handrews:
BTW feel free to add yourself to the Acknowledgements section.

PR #308, opened 2017-04-19
Reformat XML documents to 100 columns length

dlax:
As mentioned in https://github.com/json-schema-org/json-schema-spec/issues/206#issuecomment-292872393, to hopefully make future review easier.

I choose 100 columns as a compromise because wrapping to 80 columns would lead to narrow blocks of text for deeply nested elements.

I could not find a tool to automate this that gives satisfactory results so I did that by hand. I occasionally left some text exceed the limit when it looked better not wrapped (e.g. to avoid breaking links).

`git diff --word-diff master` can be used to get a relevant diff.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -51,18 +51,21 @@
                 JSON Schema asserts what a JSON document must look like,
                 ways to extract information from it,
                 and how to interact with it,
-                ideal for annotating existing JSON APIs that would not otherwise have hypermedia controls or be machine-readable.
+                ideal for annotating existing JSON APIs that would not otherwise have hypermedia
+                controls or be machine-readable.
             </t>
         </abstract>
         <note title="Note to Readers">
             <t>
-                The issues list for this draft can be found at <eref target="https://github.com/json-schema-org/json-schema-spec/issues"/>.
+                The issues list for this draft can be found at
+                <eref target="https://github.com/json-schema-org/json-schema-spec/issues"/>.
             </t>
             <t>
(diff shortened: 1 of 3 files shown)


jsonschema-core.xml
```diff
@@ -551,15 +607,22 @@ Link: <http://example.com/my-hyper-schema#>; rel="describedby"
...
             <section title='Describing a profile of JSON' anchor="profile">
                 <t>
-                    Instances MAY specify a "profile" as described in <xref target="RFC6906">The 'profile' Link Relation</xref>.
-                    When used as a media-type parameter, HTTP servers gain the ability to perform Content-Type Negotiation based on profile.
-                    The media-type parameter MUST be a whitespace-separated list of URIs (i.e. relative references are invalid).
+                    Instances MAY specify a "profile" as described in
+                    <xref target="RFC6906">The 'profile' Link Relation</xref>.
+                    When used as a media-type parameter, HTTP servers gain the ability to perform
+                    Content-Type Negotiation based on profile.
+                    The media-type parameter MUST be a whitespace-separated list of URIs (i.e.
```
handrews: It seems that if a line ends in a period then the .txt generator automatically puts two spaces after that period.  It can't tell the difference between an abbreviation and the end of a sentence.

Could you please move the "(i.e." down to the beginning of the next line?

dlax: > Could you please move the "(i.e." down to the beginning of the next line?

Done, it's better now. Well spotted.

Issue #205, opened 2016-12-27
People can't find the historical drafts v3 and v4


agebhar1:
Maybe it's a good idea to keep historical drafts [v3](/kriszyp/json-schema/blob/master/draft-zyp-json-schema-03.xml) and [v4](/kriszyp/json-schema/blob/master/draft-zyp-json-schema-04.xml) from Kris Zyp et. al. within this repository to be near-complete. The current [archive](/json-schema-org/json-schema-spec/tree/master/archive) does only contains the (meta) schema.

handrews:
This is being tracked over at the web site repo which is where things are actually hosted (I'm not sure why we have those archive folders here, actually) json-schema-org/json-schema-org.github.io#59 is what you want.

agebhar1:
Thanks for the hint. It seems the schema within the archive folders here is not the same as in  json-schema-org/json-schema-org.github.io ... :confused: 

I wonder why the (historical) draft v3, v4 is contained in the website repository and not in spec (this) repository. Shouldn't the website repository only contains the representation and not the data from the drafts itself?

handrews:
You'll have to ask @Relequestual, who I think is away until January, as I don't quite understand the rationale for the exact split (it predates my involvement with the project).

agebhar1:
Thx.

Relequestual:
Sorry for the long delay. I'm not really sure why it happened that way. @awwright did the majority of the migration work.

I agree that the draft 3 and 4 versions should be in this repo... maybe we can add that to the backburner as an easy win. I'm going to re-open this issue, as I feel the XML used to build the documents should be in the archive in this repo, not just the meta-schemas.

awwright:
@agebhar1 Old drafts become obsoleted with a successor release and shouldn't be implemented. Otherwise for historical interest, the IETF Datatracker stores old drafts at their website.

Is there any particular use you have to host the old drafts on the website?

agebhar1:
@awwright some (at least 3) projects which support draft 4 pointing to 'latest' version (e.g. [fixed it](/everit-org/json-schema/commit/62c5712433fb77919bf0d0a770998482d68fe6b3)), which might be a bad idea.

[...]

PR #327, opened 2017-06-06
Restore missing history

adamvoss:
This PR makes not changes and requires no changes to process going forward, but it brings over all of the relevant commits from https://github.com/json-schema/json-schema.  As a result, anyone trying to connect history between the repositories can, it far easier for anyone looking for the historical sources to be able to find them, historical contributors attribution is restored, and there is redundancy in case IETF were to stop hosting expired drafts or otherwise be unavailable.

I think I made good decisions on how to merge things in history, but of course there were decisions to be made meaning other ways this could have been done.

Notable decisions include:

* I added new commits for the drafts 0 to 2 files that did not exist in either repository
* I added new commits, merging from the appropriate place in history, that organize the historical version in the structure of the current repository for easy comparison.
* I made an effort to allow linearizing of history if desired using `git replace`.  Specifically,

    ```bash
    git replace 314b5c3040c5af13f838165189df3e0585b35a26 a14190b51fdfbc6a2d66a29e86185cb183b5363e
    git replace 314b5c3040c5af13f838165189df3e0585b35a26 d94d409627bcc5b0bd0b33eaec49bce78398e7d3
    ```
    would accurately join the disparate branches at the proper places in historical.  (This does not include the commits I added that organize the old drafts, those only exist via merging). Once the commits hashes are known, one could then add an addition replace to hide this merge, if desired.  (Note: if the ability to hide these merges is desired, best results would come from creating an empty commit on master before and after this merge.)

Relevant issue #205


handrews:
This seems fantastic to me!  I'm a bit hesitant to unilaterally accept it:  @awwright and/or @Relequestual what do you think?  I don't see a downside here, and having a full history is definitely a good thing.

handrews:
Thanks, @epoberezkin, two reviews and I say let's merge (and hopefully no one objects b/c I already did it :-)

PR #333, opened 2017-06-20
validate property names in patternProperties

santhosh-tekuri:


handrews:
This is a significant change to the specification and should be submitted as an issue (on this repository) for discussion first.

Please look through the existing issues carefully as there are numerous object property related issues filed already.

Once it is ready for PR, the PR should go against the specification XML file here.  The meta-schema should be updated with it, but the XML file is the normative document.

Do not expect quick action- all of the major reviewers are pretty distracted this month, although we will get back to a more active mode soon.

santhosh-tekuri:
Thanks, raised issue #334 for discussion

PR #343, opened 2017-07-29
Create gentlejack

gentlejack:


Julian:
I don't know what this PR is, but it doesn't seem like it belongs here?

Closing, but if this is actually some magically amazing thing, please do explain what that is.

PR #344, opened 2017-08-04
Fix sentence fragment

rjmill:
This sentence does not have a main verb.

I didn't notice any contribution guidelines, so I apologize if this isn't the proper way to do this. Submitting a PR in this case was just easier than opening an issue describing the typo. Feel free to close this PR and make the change however you need to :slightly_smiling_face:

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -306,7 +306,7 @@
                 plain name fragment identifiers.
             </t>
             <t>
-                Defining and referencing a plain name fragment identifier
+                Defining and referencing a plain name fragment identifier are specified
                 in the <xref target="id-keyword">"$id" keyword</xref> section.
             </t>
             <t>


handrews:
@rjmill thanks for catching this!  It may just be a lack of caffeine but I'm having trouble deciding whether "is" or "are" is the correct verb here.  I think I tend to read "defining and referencing" as one thing which is why I want to say "is", but I'm probably wrong there.  Obviously you do those things in separate steps so I guess that means "are"?

Anyone else have thoughts?  I feel silly holding this up but I've been staring at the wording for way too long trying to convince myself one way or the other :-P

dlax:
@handrews, "are" looks correct to me

rjmill:
@handrews No worries about holding off on merging to make sure the change is correct :) At first glance, it looks like "is" might be correct because the words leading up to it, "a plain name fragment identifier," are singular, but the subject of the sentence is "Defining and referencing," making "are" the correct verb to use here. That was my thought process when choosing "are" instead of "is" (because I wasn't sure at first either.)

I'm pretty confident that "are" is the correct verb here, but I'm not offended at all if you're more comfortable waiting for more people to weigh in before committing :)

handrews:
Thanks, folks!  I just got confused staring at it too long and needed at least one more opinion :-P

Issue #140, opened 2016-11-09
Hyper-schema: Link subject/anchor


awwright:
JSON Schema should be able to define the subject (anchor, or source) of a link relation.

---

Suppose I define a link right here (HTML):

    <link rel="prev" href="139" />

We can express the information links convey in a format called N-Triples (a subset of the more general grammar Turtle). Since links by default have a subject of the current document, it would look like:

    <https://github.com/json-schema-org/json-schema-spec/issues/140> iana:prev <https://github.com/json-schema-org/json-schema-spec/issues/139> .

But it might be the case we want to use JSON Schema to let documents make links about other documents to different other documents. A subject that's not `<https://github.com/json-schema-org/json-schema-spec/issues/140>`

Well in that case we need to allow JSON Schema to define what that "subject"/"source"/"anchor" is.

Relequestual:
I don;t understand the purpose of this at all. Could you expand with a practical use case?

awwright:
Suppose I have a JSON document that specifies metadata about a blog post, or another image, or something. In that case, I don't want to specify a link from the _current_ document, I want to specify a link _about some third resource_.

So I have a document like:

```javascript
[ {
post: 34
next: 35
}, {
post: 35,
next: 36
} ]
```

and I want to generate the following links:

```
<34> iana:next <35> .
<35> iana:next <36> .
```

With the HTTP Link header, you'd specify it like this:

```
Link: <34>;anchor="35";rel="next"
Link: <35>;anchor="36";rel="next"
```

handrews:
@awwright @Relequestual @jdesrosiers this just clicked for me and I understand why it is essential.  I endorse the `"anchor"` keyword (for compatibility with RFC 5988) over the `"linkSource"` keyword that was proposed in #61, but see further down for concerns over relative positioning.

The simple example is that without this feature, collections with the "collection" and "item" relations simply don't work.  The key to understanding this is the following bit from the definition of `"rel"`:

> The relation to the target is interpreted as from the instance that the schema (or sub-schema) applies to, not any larger document that the instance may have been found in.

So if a link is attached to the sub-schema, then the context URI of the link (which is what is meant by "from") is the resource to which the sub-schema is attached, not the complete instance document (which corresponds to the complete schema document).  Consider this minimal collection schema:

```
{
    "title": "Generic collection with id references",
    "type": "array",
    "items": {
        "type": "object",
                "properties": {
                    "id": {"type": "integer", "minimum": 1}
[...]

Issue #351, opened 2017-08-19
application/schema-instance+json


handrews:
Using `application/json` as the instance data type imposes several limitations on JSON [Hyper-]Schema.  While it is essential that `application/schema+json` works with plain `application/json` instances, defining an optional instance media type to enable additional features would make several things much easier.  Specifically:

* Using URI fragments with instances (`application/json` does not define any fragment resolution semantics), which would help with `"anchor"` (#140).
* Using media type parameters to connect instances to schemas and conduct content negotiation based on the schema, e.g. choosing one of several versions of a given schema.

We have at times discussed proposing registering a media type relation for `application/json` as part of this project, and I'm sure the idea of JSON Pointer fragments for `application/json` have been raised several times.  But given the popularity of JSON, these things would probably have to clear a high bar, and would no doubt take quite a bit of time.

Defining an optional `application/instance+json` would let us manage these proposals within our own process.  If they seem likely to be accepted for `application/json` that could still be done.  There is also some precedent for this approach:  JSON-LD defines `application/ld+json` but also provides mechanisms to interpret `application/json` or other `+json` media types as JSON-LD via a link header.  We already do more or less the same thing, but as noted in #222 there are concerns with our existing approach.

Finally, my plan to get around these various issues was to define a custom vendor media type for my API.  I would rather not do that if possible.  And even if I have another reason to use a custom `+json` media type, having `application/schema-instance+json` [EDIT: thanks to @Relequestual for suggesting schema-instance instead of just instance) would allow me to reference it as the basis for the media type, and then only need to define additional / varying semantics.  This would further encourage consistency in design and implementation, particularly for hypermedia.

Tagging @dret, @dlax, @awwright

handrews:
Goals for `application/instance+json`

* Define clear and useful fragment semantics
* For content negotiation, asking for `application/instance+json` ensures a programmatically identifiable schema in the response
* Define a media type parameter for identifying that applicable schema(s), which can also be used in content negotation
* Define a media type parameter for identifying versions of relevant schema(s), also for content negotiation (this may be the same parameter, with versioning incorporated into the schema URI in some way defined by the schema/instance authors)

Goals may also include:
* Define a standard mechanism to reference schemas and schema versions from within the document
* Define a standard mechanism to associate plain name URI fragments with specific positions within the document

[...]

PR #352, opened 2017-08-20
Add "anchor" LDO keyword.

handrews:
Addresses the simpler half of #140.

This corresponds directly to the RFC 5988 "anchor" keyword.

I'm not sure whether all of the talk of the context URI and whether it is possible
to construct one belongs where I put it in the `"rel"` section or whether it should
be moved to `"anchor"`, or a bit of each.

I'm also not sure if there needs to be more explanation of `"anchor"`.  Since it is
adapted from RFC 5988 I went with less explanation.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -29,6 +29,11 @@
                     "description": "relation to the target resource of the link",
                     "type": "string"
                 },
+                "anchor": {
+                    "description": "the URI of the context resource",
+                    "type": "string",
+                    "format": "uri-reference"
+                },
                 "title": {
                     "description": "a title for the link",
                     "type": "string"
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -752,14 +752,28 @@
                 </t>
 
                 <t>
-                    The relation to the target is interpreted as from the instance that the schema
-                    (or sub-schema) applies to, not any larger document that the instance may have
-                    been found in.
+                    As defined by RFC 5988, a link connects a context resource
+                    to a target resource, where the nature of the connection
+                    is described by the link relation type.  The context
+                    resource is the instance to which the schema (or sub-schema)
+                    applies, rather than any larger document in which the
+                    instance may have been found.  The context may be changed
+                    with the <xref target="anchor">"anchor"</xref> property.
                 </t>
 
                 <t>
-                    Relationship definitions are not normally media type dependent, and users are
-                    encouraged to utilize existing accepted relation definitions.
+                    Depending on the media type of the instance, it may or may
+                    not be possible to assign a URI to the exact default context
+                    resource.  In particular, application/json does not define
+                    URI fragment resolution semantics, so properties or array
+                    elements within a plain JSON document cannot be identified
+                    by a URI.
+                </t>
+
+                <t>
+                    Relationship definitions are not normally media type
+                    dependent, and users are encouraged to utilize existing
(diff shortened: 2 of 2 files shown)


dlax:
> I'm not sure whether all of the talk of the context URI and whether it is possible
to construct one belongs where I put it in the "rel" section or whether it should
be moved to "anchor", or a bit of each.

I assume you're referring to the "[Depending on the media type of the instance, it it may or may not be possible to assign a URI to the exact default context resource](https://github.com/json-schema-org/json-schema-spec/pull/352/files#diff-6f6cc4b69f2e1ecfca964501e7a07e2eR765)" paragraph. I'm not sure it's needed at all for the simple case (URI) at stake here. Could you explain why you added it?
(Nit: having this in a dedicated commit would help review and understanding, as far as I am concerned.) 

> I'm also not sure if there needs to be more explanation of "anchor". Since it is
adapted from RFC 5988 I went with less explanation.

Looks enough to me.

handrews:
@dlax it's not in a dedicated commit because it's part of this logical change.  I would not make one without the other.  Without `"anchor"` there is no reason to think about what the context URI is.

As a general rule, I keep changes that interact with each other together in the same commit.  I would not merge these separately.

handrews:
> I assume you're referring to the "Depending on the media type of the instance, it it may or may not be possible to assign a URI to the exact default context resource" paragraph. I'm not sure it's needed at all for the simple case (URI) at stake here. Could you explain why you added it?

I thought I had answered this but apparently I didn't (or at least not very clearly).  My apologies.

I included that because:

 1.) not being able to produce a URI for the context resource feels strange to me, and therefore worth acknowledging.  Without it, I feel fairly certain that someone would file an issue requesting clarification.

2.) The fact that we can't guarantee a context URI is due to the split nature of JSON / JSON Schema hypermedia documents.  There are two resources involved: the schema and the instance (even if you never fetch the schema from a remote location).  Most hypermedia documents involve one media type, and can therefore declare a fragment syntax to fulfill all use cases.

JSON Schema needs to work with instances that are plain `application/json` which has no such syntax.  But people often think that the JSON Schema pointer fragment syntax can be applied to plain JSON.  While it would seem to make sense to use it that, way, it is in fact **explicitly forbidden** by RFC 6901:

>    Note that a given media type needs to specify JSON Pointer as its
   fragment identifier syntax explicitly (usually, in its registration
   [RFC6838]).  That is, just because a document is JSON does not imply
   that JSON Pointer can be used as its fragment identifier syntax.  **In
   particular, the fragment identifier syntax for application/json is
   not JSON Pointer.**

(added emphasis mine)

Many people (myself included once upon a time, before I understood media types and URI fragments as thoroughly as I do now) find this very counter-intuitive.  I want to avoid widespread incorrect implementations, so I documented this in the spec.  It's just too surprisng to ignore.

This should also help clarify the utility of the `application/instance+json` proposal, although discussion of that should stay over in #351.  But having an (optional) instance media type declared by this spec would allow people to reliably use a properly registered standard fragment syntax.  Is that worth making a new media type for?  I don't know yet, discuss it in #351 :-)

dlax:
@handrews Thanks for your detailed clarification. I agree it's worth mentioning.

handrews:
Last call: as of tomorrow this will have been up for 14 days and I will merge.  It is a near-universal link serialization feature and no one has really objected at any point (just discussed how to handle the non-URI-identifiable cases, which is independent of this keyword).

Issue #180, opened 2016-12-05
Validation: if/then/else


epoberezkin:
This can be seen as the extension of #31 and #64. I see it as the syntax sugar for existing boolean/compound keywords.

The validation process is very simple:
- if the schema in `if` is valid, than the schema in `then` should be validated and its outcome determines the instance validation.
- if the schema in `if` is invalid then the schema in `else` should be validated and its outcome determines the instance validation.
- if `if` is not present then the schema itself is invalid (metaschema validation will fail).
- if neither `then` nor `else` is present - the same as above

As I've written in some issue the schema with if/then/else: 

```json
{
  "if": {"$ref": "condition" },
  "then": {"$ref": "schema1"},
  "else": {"$ref": "schema2"}
}
```

is a boolean operation and it is equivalent to the schema below that is possible now:

```json
{
  "anyOf": [
    { "allOf": [ {"$ref": "condition" }, {"$ref": "schema1"} ] },
    { "allOf": [ {"not": {"$ref": "condition" } }, {"$ref": "schema2"} ] }
  ]
}
```

so if/then/else is as declarative as existing keywords but it provides a more convenient, clear and performance efficient alternative ("condition" will never be validated twice) for a quite common validation scenario.

Using if/then/else the problem in #31/#64 is also solved.

handrews:
> so if/then/else is as declarative as existing keywords

While Wikipedia is not always the best resource, this article is well-cited and the following definition is what I usually here for declarative programming:

> In computer science, declarative programming is a programming paradigm—a style of building the structure and elements of computer programs—that expresses the logic of a computation without describing its control flow.

So if/then/else is _by definition_ not declarative.  The distinction between declarative and imperative is control flow.  If/then/else is control flow.  Imperative vs declarative has nothing to do with whether the outcome is the same- all programming styles can produce the same output.  It's about how things are processed.

It is certainly worth discussing if we want to add this imperative construct to JSON Schema, but it is the very definition of an imperative construct.

epoberezkin:
I am not sure why implication is more imperative than and/or/negation - they are all operations from boolean algebra. The fact that the evaluation of implication can be short-circuited (i.e. when "if" is false then "then" doesn't need to be evaluated) desn't make it more imperative than "and" and "or" - they also can be short-circuited.

Also there are languages that use if/then/else as expression, not as a control flow statement. If you don't like the keywords, some other can be used.

[...]

PR #357, opened 2017-08-20
Split validation keywords into sections.

handrews:
**NOTE:** This diff should be viewed [with whitespace ignored](https://github.com/json-schema-org/json-schema-spec/pull/357/files?w=1)

This improves the readability of the spec, particularly
new readers trying to get a feel for the scope from the
table of contents.

A previous, more complex reorganization that was attempted
before got plenty of support, and this more minimal change
avoids the issues that bogged the previous approach down.

Finally, having sections will make the section grouping
"if", "then", and "else" (#180) make more sense.  Adding those
keywords without grouping them would seem odd.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -251,421 +251,433 @@
                 instance.
             </t>
 
-            <section title="multipleOf">
-                <t>
-                    The value of "multipleOf" MUST be a number, strictly greater than 0.
-                </t>
-                <t>
-                    A numeric instance is valid only if division by this keyword's value results in
-                    an integer.
-                </t>
+            <section title="Validation keywords for numeric instances (number and integer)">
+                <section title="multipleOf">
+                    <t>
+                        The value of "multipleOf" MUST be a number, strictly greater than 0.
+                    </t>
+                    <t>
+                        A numeric instance is valid only if division by this keyword's value results in
+                        an integer.
+                    </t>
+                </section>
+
+                <section title="maximum">
+                    <t>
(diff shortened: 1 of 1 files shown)


handrews:
I plan to merge this as soon as I'm ready to post the `if/then/else` PR which depends on this one, because:
* A far more complex change got four approvals during draft-06 work before bogging down on aspects that are not present here
* An active community member has approved this version (making, I think, a 5th relevant approval)
* There is no impact on JSON Schema syntax or semantics- it is purely a readability change.
* It is easily reversed if someone comes along with an objection that would have prevented it from merging otherwise.

Generally, we have not held this sort of change to the same 2-week comment window as changes that impact implementations.  This one is a bit more extensive than most such changes, but I think it's still reasonable to move forward.  `if/then/else` will require the 2-week comment window, and it is less than 8 weeks until draft-06 expires (a soft deadline, but one I'd like to hit).

I need to update https://github.com/json-schema-org/json-schema-spec/wiki with this sort of guidance as I can't even remember exactly what we agreed on before (hence the detailed explanation in this comment).

dlax:
One *unrelated* thing that strikes me while reading this PR is that keywords now in "Validation keywords for any instance type" subsection appear almost at the end of the "Validation" section. I tend to believe that having these before more specific ones (i.e. those applying to objects/array/etc.) would be more logical and make things easier to read. This is especially important for the ubiquitous `type` keyword. Thoughts?

handrews:
@dlax I almost moved it, but the difficult history of organizational PRs such as this one led me to do the absolute minimum possible.  I was planning to move it later :-)

Issue #197, opened 2016-12-15
"$comment" keyword


handrews:
[EDIT: I changed it from `"comment"` to `"$comment"` as explained in a later... um... comment]

JSON lacks comments.  Comments in complex schemas are good.  The `"description"` keyword, as a general annotation feature that is noted as being usable in interface display is not suitable for schema author-to-schema author comments.

Let's reserve `"$comment"` as a keyword.

Implementations MUST NOT use `"$comment"` in any sort of end user data presentation.

Implementations MAY choose to use the value of `"$comment"` in debugging messages.  Similar, schema editing tools MAY make use of or encourage the use of the comment keyword.

Implementations MAY opt to not store the comment field and its value in memory as an optimization.

There is no restriction on the comment's value.  While a string is the most obvious sort of comment, schema authors may decide to format comments in any way they wish.  Implementations that use comments for debugging or other purposes SHOULD support strings.  Schema authors inventing more complex comment forms MUST NOT assume that peer implementations will be able to understand those forms.

`"$comment"` would also be legal within a `"$ref"` object, and an implementation would still be able to use it for debugging.  All other keywords would still be in the "MUST be ignored" category should they appear in a `"$ref"` object.

[EDIT: originally this suggested the validation spec but as noted below it really belongs in core if we have it at all]

As an obvious use case, the reviews around "jsonReference" and "notJsonReference" in the meta-schema would have been much easier if a comment field is allowed.

erosb:
I'm not sure if it is a necessary feature. I can just hardly imagine any usecase where `"description"` is unsuitable for debugging and I need an other programmer-friendly `"comment"` too.

handrews:
Explaining complex usage of allOf/anyOf/oneOf/not/dependencies is very useful for schema maintainers, but should not be used as a description of anything user-visible.

erosb:
Ah-ha, now it is more understandable.

Well it an optional feature and not too complex to implement, so I don't mind it.

handrews:
> Well it an optional feature and not too complex to implement, so I don't mind it.

Yeah, the idea here is really just to reserve the keyword so that it is not co-opted by extensions.

handrews:
Should probably actually be `"$comment"` in the core spec, just as comments are part of the core syntax of programming languages rather than some sort of library or extension.  If a media type supports comments, that should be in the media type definition.

That will also make clear that it has nothing to do with validation or hypermedia.

handrews:
[...]

PR #359, opened 2017-08-21
Add "$comment" as a keyword

handrews:
This addresses issue #197.  The primary purpose is to reserve the
keyword and ensure that implementations do NOT do anything based
on its presence or contents.  This improves interoperability by
ensuring that comments are safe for use with any conforming system.

Without this keyword, two implementations may implement it with
differing semantics, potentially producing undesirable or
insecure behavior.

Effort has been made to ensure that existing implementations
that properly ignore unknown keywords will not need to make
any changes at all to be in conformance for this keyword.

I noticed while adding $comment that keywords are always quoted
in section headings unless the keyword is the entire heading.
This is true across all three specifications, so the trivial
second commit brings the only outlier ($ref) in line with everything else.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -573,6 +573,32 @@
             </section>
         </section>
 
+        <section title='Comments with "$comment"'>
+            <t>
+                This keyword is reserved for comments from schema authors to readers or
+                maintainers of the schema.
+
+                The value of this keyword MUST be a string. Implementations MUST NOT present this
+                string to end users.  Tools for editing schemas SHOULD support displaying and
+                editing this keyword.  The value of this keyword MAY be used in debug or error
+                output which is intended for developers making use of schemas.
+
+                Tools that translate other media types or programming languages
+                to and from application/schema+json MAY choose convert that media type or
+                programming language's native comments to or from "$comment" values.
+                The behavior of such translation when both native comments and "$comment"
+                properties are present is implementation-dependent.
+
+                Implementations SHOULD treat "$comment" identically to an unknown extension
+                keyword.  They MAY strip "$comment" values at any point during processing.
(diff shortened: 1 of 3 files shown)


jsonschema-core.xml
```diff
@@ -573,6 +573,38 @@
...
+                output which is intended for developers making use of schemas.
+
+                Schema vocabularies SHOULD allow "$comment" within any object containing
+                vocabulary keywords.  Implementations MAY assume "$comment" is allowed
+                unless the vocabulary specifically forbids it.  Vocabularies MUST NOT
+                specify any effect of "$comment" beyond what is described in this
+                specification.
+
+                Tools that translate other media types or programming languages
+                to and from application/schema+json MAY choose convert that media type or
```
dlax: MAY choose *to* convert

handrews: Fixed, thanks!

handrews:
Last call- this will get merged tomorrow at the 14 day mark barring a compelling reason to reconsider.

Julian:
I am not a huge fan of JSON Schema attempting to step in for JSON's shortcomings. But I won't pretend to have "compelling" reason not to add this, so I suppose I'm just a -0.

handrews:
@Julian yeah, I think that objection was pretty solidly covered in the issue.  The implementation requirements for this are literally "don't do anything" and many schema authors view it as very helpful (based on usage in teams that I have worked with, and/or reasons given for not working with JSON Schema unless allowed to use something other than JSON).

handrews:
It also does not attempt to be a generic comment mechanism for JSON, nor would I support it as such.  It's syntax and rationale are very specific to the structure of JSON Schema, and working with JSON Schema's extensible nature and the likelihood of problems caused by clashing extensions in this area.

PR #360, opened 2017-08-22
Set up a blank change log for the next draft.

handrews:
None of the open PRs have a change log, because they would
all conflict with each other.  This initializes empty change
logs and updates the draft numbering accordingly.

Note that there must be at least an empty <t></t> in the list
or the XML won't validate against its DTD.

Ensuring that the entire changelog is filled out is part of
the pre-publication checklist.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -20,7 +20,7 @@
 <?rfc rfcedstyle="yes"?>
 <?rfc comments="yes"?>
 <?rfc inline="yes" ?>
-<rfc category="info" docName="draft-wright-json-schema-01" ipr="trust200902">
+<rfc category="info" docName="draft-wright-json-schema-02" ipr="trust200902">
     <front>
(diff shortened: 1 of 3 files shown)


dlax:
I wonder if it wouldn't be worth using a tool to manage the changelog. Something like [towncrier](https://github.com/hawkowl/towncrier) is designed to managed release notes in a way to prevent merge conflicts as each PR would add a dedicated file. I don't know if maintaining the changelog is a burden, so maybe it's overkill. Just mentioning.

handrews:
@dlax my recollection from last draft was that the biggest effort in managing the change log was reassuring people that yes, we would audit the change log before publishing.  Actually writing it was easy (we did some things with the PRs, and I just tracked what else needed to happen with milestones and labels and went through it all at the end- super-easy).

handrews:
Also, the granularity of commits is waaaaaaaay too fine-grained.  Only a few things actually need to go in the change log.  Things that impact implementation, and major changes in how we explain or conceptualize things.  There's a lot of minor fixes that boil down to "made the words better" :-P

PR #362, opened 2017-08-22
Start a contributing guide from the wiki contents.

handrews:
**NOTE:  Since this wording is identical to our existing policy on the wiki, commentary on the contents will be ignored.  This is just about moving things into the file unchanged.  We can debate changing contents with additional PRs.  That is part of the point of moving this into the repo instead of the wiki.**

This file has exactly the contents of the front wiki page, except
for the I-D and meta-schema publishing processes which are not
relevant to most contributors.

I'm moving it into a CONTRIBUTING.md as it integrates with
GitHub's tools (e.g. auto-linking it to the create PR screen)
and will allow us to discuss changes as PRs.

I will change the front wiki page to link to this (but retain
the sections that have not been moved here).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,31 @@
+# JSON Schema
+
+## Issues
+
+Issues should identify an problem, enhancement, or use case; and propose some course of action for the draft. For alternate support channels, [see the json-schema.org website](http://json-schema.org/).
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
+# JSON Schema
```
dlax: As a minor change from the wiki page, the title could be more specific. Something like `JSON Schema contributor guidelines`.

PR #365, opened 2017-08-23
Reference validation from hyper-schema

handrews:
Despite having a whole section describing how hyper-schema
builds on validation, it was never referenced or listed
as a normative specifciation.

Also, we had labled the meta-schema link as a validation schema
in an apparent copy-paste error.

Finally, improve wording around building on the other specs in
the same way as was done for the validation spec referincing core.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1189,7 +1189,7 @@ GET /foo/
                     </author>
                     <date year="2016" month="October"/>
                 </front>
-                <seriesInfo name="Internet-Draft" value="draft-wright-json-schema-validation-01" />
+                <seriesInfo name="Internet-Draft" value="draft-wright-json-schema-01" />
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -1188,7 +1189,17 @@ GET /foo/
                     </author>
                     <date year="2016" month="October"/>
                 </front>
-                <seriesInfo name="Internet-Draft" value="draft-wright-json-schema-00" />
+                <seriesInfo name="Internet-Draft" value="draft-wright-json-schema-validation-01" />
```
dlax: This change looks wrong. The "json-schema" reference points to the "core" document not the "validation" one. Copy/paste error?

handrews: Yup, good catch, will fix.

PR #368, opened 2017-08-25
Create a branch with a linear, first-parent relationship between releases

adamvoss:
This _could_ be prepared to merge into `master`, but my recommendation is to push this as a new branch called "released" something like that.

Following the first parents on this branch will always take you to the previously released draft.  This would establish a new merging strategy so draft-07 would be merged into this branch when it is released.

The main advantage of this at the moment is that it creates new commits that represent draft-05 and draft-06 that are newer than the commits for the older drafts so the draft tags can be in-order on GitHub Releases.

NB.  The diff is meaningless here.  The important thing is the history.


dlax:
I don't understand the point of this branch. If I want to get to a particular draft, I'd just `git checkout draft-04` (assuming such a tag exists as you suggested in #369, which I full support).

handrews:
Perhaps it might be better to start with an issue laying out the goals and proposals and then come back and re-open this PR or open a new one?

I'd like to keep the focus on the spec, and we can tidy up process right after draft-07 goes out (like we did with @dlax's reformatting of the XML files after draft-06, IIRC).

dlax:
> Perhaps it might be better to start with an issue laying out the goals and proposals and then come back and re-open this PR or open a new one?

We can. Maybe a compromise would be to add tags as @adamvoss suggested in #369, I can't think of any drawback and it should be easy to do for anybody with write access to the repository.

adamvoss:
> The main advantage of this at the moment is that it creates new commits that represent draft-05 and draft-06 that are newer than the commits for the older drafts so the draft tags can be in-order on GitHub Releases.

Please see https://github.com/adamvoss/json-schema-spec/releases for a demonstration.  It shows the result if #369 is followed with the first set of commit ids:

image

Either as a merged branch or a separate branch these two commits would make all the drafts list in reverse-chronological order so draft-06 will appear above draft-05 and so on. If this [aesthetic] is not important, there is no strong purpose for these commits and I would recommend closing this and just following #369 as-is.

dlax:
@adamvoss So here the issue that github's release page lists tags sorted in reverse-chronological order not in alphabetical order (as `git tag` does). Arguably, that's a bug in github UI :) (On the other hand, there are inconsistencies in tag names in the repo...)

> If this [aesthetic] is not important, there is no strong purpose for these commits and I would recommend closing this and just following #369 as-is.

That'd be a good first step.

Julian:
Seems like just creating the tags using the commit dates is another reasonable way to solve this?

adamvoss:
> Seems like just creating the tags using the commit dates is another reasonable way to solve this?

The commit dates _are_ newer which is the "problem".

However, you did cause me to re-look at "annotated" tags (as opposed to the normal "lightweight" tags).  Annotated tags carry a commit date, GitHub Releases respects that date, and it is possible to spoof the date, so that would be a valid alternative.

handrews:
Since there are several competing proposals, please open an issue to discuss them and choose one to re-submit as a PR.

PR #370, opened 2017-08-25
Add Travis CI build

adamvoss:
This adds automatic running of `xml2rfc` on the spec files.  When a tag is pushed it has CD functionality to publish the build HTML and text files to GitHub Releases.

**IMPORTANT:** The `api_key` will need to be changed by someone that can generate a key for this repository.  The one there is for my fork, which I used for testing.  Additionally, a project member will need to enable this repository on Travis CI for this to do anything at all :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: python
+install:
+- pip install xml2rfc
+before_script:
+- TAG=$(git tag -l --points-at HEAD)
+script:
+- xml2rfc jsonschema-core.xml --basename=jsonschema-core-$TAG --text --html
+- xml2rfc jsonschema-validation.xml --basename=jsonschema-validation-$TAG --text --html
+- xml2rfc jsonschema-hyperschema.xml --basename=jsonschema-hyperschema-$TAG --text --html


dlax:
Great idea!
It's not clear to me if this only works for tags? I think it's worth having `xml2rfc` run for any PR.

handrews:
@dlax @adamvoss release process was really something @awwright owned, so I'd rather get his input, probably better done in issues rather than directly to PRs.

adamvoss:
@dlax The `script` section will run for all commits, so if any of the `xml2rfc` calls fail it will fail the build.  Publication to GitHub Releases is restricted to tags.

@handrews I understand this may need to sit for a while and be discussed. This automates step 2 from the existing [I-D Publish Process](https://github.com/json-schema-org/json-schema-spec/wiki#i-d-publish-process).  The only variation is that it does not `tar` the files making them easier to access (an no longer an inconvenience when uploading since the process is automatic).

dlax:
> The `script` section will run for all commits, so if any of the `xml2rfc` calls fail it will fail the build.

That's already a clear improvement in itself. If the release aspect needs more thinking/discussion, why not just activating the `script` section *now*? The other section could be moved to a separate PR and further discussed.

adamvoss:
@diax I have now split the commit so that will be easy to do if that is what is decided.

handrews:
@adamvoss I'd be happy to accept the simple build script commit now.  If you haven't already, please park the release idea in an issue and we'll revisit it when the right people are available.  Based on past outcomes in process and release management discussions, I am not willing to make changes on my own at this time.

handrews:
@adamvoss I have enabled the repo in Travis

PR #371, opened 2017-08-25
Utilize HOME.md from the wiki as CONTRIBUTING.md

adamvoss:
This largely duplicates #362 using a branch I made back in June, but includes a couple of sections excluded by #362 that I think are worth having visible and subject to PR.

Merging this also brings over all the history of the file from the wiki.  (unless this is pushed to `master` without a merge commit, #362 will still be the source for most `blame`/`annotate` algorithms because it will be the first-parent, but his is probably not important)


handrews:
I should have thought of this earlier, but while it's a good idea to have the other sections in the repo, I'd rather not have them in the CONTRIBUTING section as they are not relevant to most contributors.  I'll follow up with you over email on the backstory with this.

dlax:
> I should have thought of this earlier, but while it's a good idea to have the other sections in the repo, I'd rather not have them in the CONTRIBUTING section as they are not relevant to most contributors.

I tend to agree, but might be convinced otherwise.

adamvoss:
I really think this can be argued either way.   CONTRIBUTING.md may be read by someone opening a PR of Issue ([Issue/PR templates are also available for this case](https://github.com/blog/2111-issue-and-pull-request-templates)), or it may be read by someone wanting to know how the project "works" (and possibly considering greater involvement).  These sections are relevant to the later group of users.

Since they are short I don't see an issue with putting them at the bottom of CONTRIBUTING.md where they can easily be skipped by people who have already had their needs addressed at the top of the file rather than needing another file in the repository.

handrews:
@adamvoss I'd be happy to merge the commit history, but at this time I am not willing to put the release process into the contributing document, so the merged history would need to keep that out of master.

Feel free to file an issue for where and how to document the release process, but it needs discussion with other stakeholders, particularly @awwright who does the releasing, before going to PR.

Issue #64, opened 2016-09-19
conditional selection of alternate schemas (includes "switch" and other options)


handrews:
### The Problem (and current workarounds)

A common use case is to select an overall validation schema (or schemas) based on how a small subset of the instance validates.  In the simplest case, some property in the instance is checked against a set of literal values, and the overall validation schema(s) are chosen based on that literal value.  This is generally implemented with `oneOf` (or `anyOf`) and `enum` (although see also the `constant` proposal in issue #58 ).  Some forms of this problem may also be solved with `dependencies`.

**Note:** Throughout this proposal, the elements of the `oneOf`/`anyOf` lists are referred to as _branches_, as in this case they are being used as implicit (or in some options, explicit) conditionals.
#### Single selection with `oneOf`

This can be read as "if foo is firstValue, bar must be present and must be a list of number, otherwise if foo is secondValue, buzz must be present and a string that is at least 10 characters long":

``` JSON
{
    "type": "object",
    "oneOf": [
        {
            "properties": {
                "foo": {"enum": ["firstValue"]},
                "bar": {
                    "type": "array",
                    "items": {"type": "number"}
                }
            },
            "required": ["foo", "bar"]
        },
        {
            "properties": {
                "foo": {"enum": ["secondValue"]},
                "buzz": {
                    "type": "string",
                    "minLength": 10
                },
                "required": ["foo", "buzz"]
            }
        }
    ]
}
```

One difficulty with this approach is that the cause and effect are not clear.  This could just as easily be read as "If bar is present as a list of numbers, foo must be set to firstValue.  Otherwise if buzz is present and a string of at least ten characters, foo must be set to secondValue."

This is both a strength and a weakness.  The `oneOf` construct an capture complex alternatives, but does not clearly express the idea of one part of the schema being the determining factor.  In a small schema like this, either interpretation is easy to read, and it’s easy to spot the enum and guess that it is probably the determinant.  In a more complex schema, where maybe there are two values of the enum that select one branch and one in the other, that is much less easy to spot.
#### Multiple selection with `anyOf`

Here is a similar example using `anyOf` with, where foo being set to multiSelect can potentially validate against two branches (although is only required to validate against one or the other).  If foo is set to singleSelect it must validate against the lone schema that accepts that value.  Multiple matches are prominently addressed in one of the proposals so we will use this example of consider the options.

``` JSON
{
    "type": "object",
    "required": "foo",
    "anyOf": [
[...]

PR #375, opened 2017-08-27
Specify "if", "then", and "else"

handrews:
This addresses #180.

I have intentionally allowed for any combination of these
keywords to be present or absent.  While having a "then" or "else"
without "if" is pointless and/or nonsensical, so are many other
possible JSON Schema keywords.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -102,6 +102,9 @@
         "properties": {
             "additionalProperties": { "$ref": "#" }
         },
+        "if": {"$ref": "#"},
+        "then": {"$ref": "#"},
+        "else": {"$ref": "#"},
         "allOf": { "$ref": "#/definitions/schemaArray" },
         "anyOf": { "$ref": "#/definitions/schemaArray" },
         "oneOf": { "$ref": "#/definitions/schemaArray" },
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -218,9 +218,9 @@
                 <t>
                     Hyper-schemas MUST NOT be applied to an instance if the instance fails to
                     validate against the validation keywords within or containing the hyper-schema.
-                    Hyper-schema keywords in branches of an "anyOf" or "oneOf" that do not validate,
-                    or in a "dependencies" subschema that is not relevant to the instance, MUST be
-                    ignored.
+                    Hyper-schema keywords in branches of an "anyOf", "oneOf", or "if"/"then"/"else"
+                    that do not validate, or in a "dependencies" subschema that is not relevant
+                    to the instance, MUST be ignored.
                 </t>
                 <t>
                     Hyper-schema keywords in a subschema contained within a "not", at any depth,
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -635,7 +635,7 @@
                 </section>
             </section>
 
-            <section title="Applying subschemas conditionally">
+            <section title="Keywords for applying subschemas conditionally">
                 <t>
                     These keywords work together to implement conditional
                     application of a subschema based on the outcome of
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -635,6 +635,78 @@
                 </section>
             </section>
 
+            <section title="Applying subschemas conditionally">
+                <t>
+                    These keywords work together to implement conditional
+                    application of a subschema based on the outcome of
+                    another subschema.
+                </t>
+                <t>
+                    These keywords MUST NOT interact with each other across
+                    subschema boundaries.  In other words, an "if" in one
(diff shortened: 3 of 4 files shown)


handrews:
Paging @epoberezkin -- does this spec match your implementation?

epoberezkin:
@handrews I implemented it somewhat differently in terms of what is required.

`then` is required when `if` or `else` is present, it can be expressed as:

```json
"dependencies": {
  "if": ["then"],
  "else": ["then"]
}
```

Also I considered `if` to have a default value of `true`, meaning that when `if` is absent, then the instance should be valid against `then` and invalid against `else` (if present).

The reason for that is that it was how `switch` was specified - `then` was allowed without `if` and it was required that the instance is valid against `then` in this case.

Maybe it's better to say that if either `then` or `else` is present (or both), if MUST be present:

```json
"dependencies": {
  "then": ["if"],
  "else": ["if"]
}
```

That would make the paragraphs about ignoring `then`/`else` unnecessary. I think it's better to make the schema invalid, then ignore some parts of it in this case...

handrews:
@epoberezkin that all does make sense.

My rationale for not requiring any of them to be present based on the others is mostly because of re-use patterns.  I can see someone having a fairly generic `if` condition and a straightforward `else`, and then combining that with different `then` clauses.  Admittedly, I don't have a compelling use case for this, but JSON Schema in general allows weirdly incomplete and/or useless schema keyword combinations so it seemed like the most consistent approach.

epoberezkin:
@handrews that's ok as is i think, I simply wrote how it is different.

handrews:
I just rebased the original commit on top of the section reorganization @dlax did, but it is otherwise unchanged.  I also added a subsequent commit rewording the section title to be consistent with the one about allOf/anyOf/oneOf/not, in case anyone wants to see that.  But it does not change anything significant about the PR.

silkentrance:
@handrews This is imperative and and does not work for schemas that well. See XML Schema for example. There was a better approach using "inverse logic" but I am unable to find it right now.

Regardless of that, the condition schema for ``if`` should be made more clear. The existing examples do not include an express form of a condition but references to a definition of such a condition.

handrews:
@silkentrance Please take a look at issue #64 followed by #180 for an ***extremely*** extensive discussion of imperative vs declarative approaches, with more than 100 comments between the two issues.  (Note that while I filed #64, I did not write it- I imported it from the old proposal wiki).

Could you elaborate on what you want with the `if` schema?  It is intentionally unrestrained: validation provides a true/false outcome, which is all that is needed to decide between `then` and `else`.  I'm not sure what you mean by "an express form of a condition".

As for its suitability in schemas, @epoberezkin implemented it back in December and it has gotten a fair amount of use without any problems.  If you want me to look at XML Schema I'm going to need a specific link- there's way too much to slog through otherwise.  I avoid XML like the plague.

silkentrance:
@handrews Yeah, #64 and #180. I've already read a lot of the stuff that is in there...

Express form of a condition basically means to provide working examples in the specification. ATM and IMO the specification lacks proper examples. Compare for example the W3C's XML Schema specification and other specifications.

As for "slog", see the above two paragaphs 😀

handrews:
PRs welcome for examples.  Or maybe @epoberezkin can help as the primary advocate for this change?  I'm not trying to be a jerk but I've more than got my hands full with hyper-schema this round and no one ever likes my examples anyway.  It's much better if someone else writes them.

handrews:
@silkentrance in general the validation spec has very few examples.  I haven't had all that much to do with that spec, and I like it the way it is b/c I can quickly read the theory which is what I care about (this is why my examples are nearly always horrible).

In general, the idea is for there to be a lot more of a tutorial on the web site, rather than cluttering the spec with them: https://github.com/json-schema-org/json-schema-org.github.io

It might be worth an issue either here (if you want to argue for spec examples) or in the web site repo to help sort this out.  Most people aren't looking at closed PRs.

I'm mostly just nudging the validation spec along with non-controversial things for this draft.  My focus is really hyper-schema, and I'm hoping awwright or one of the other contributors (or someone new!) will have time to move validation forward more in the next draft after this one.  I admit I'm sucking up most available time from others with hyper-schema reviews right now.

silkentrance:
@handrews I am not pointing fingers at anybody, you are all doing a great job here! As for the PRs, I am looking into that and when I find the time, I will provide you with such.

handrews:
@silkentrance thanks!  My "I haven't had all that much to do with..." comment wasn't trying to direct blame, just trying to explain why I'm not jumping on this to the same degree that I am with Hyper-Schema concerns right now.

BTW, figuring out how to put in enough examples without totally breaking up the flow of the spec is something I'm wrestling with in Hyper-Schema as well.

Issue #296, opened 2017-04-03
Link usage hints


handrews:
JSON Hyper-Schema should not include keywords that come from specific protocols, and certainly not ones that do not generalize well enough to apply to any protocol.

This would suggest that `"allow"` as proposed by #73 would not be ideal as it is HTTP-specific, although the concept generalizes well, so it's not an unreasonable proposal either.

However, there are several things such as HTTP's "Allow", which would optimize and document how to ***use*** the protocol to interact with a link target, as opposed to the current LDO hints such as `"mediaType"` which describe the target resource itself, independent of how it is accessed.

A more clear example is the "Accept-Patch" header, which is critical for correctly constructing an HTTP PATCH request (in conjunction with either `"targetSchema"` or the schema linked in the response to a HEAD or GET to the target).  "Accept-Patch" is very specific to HTTP (or, really, an extension to HTTP).  No current Hyper-Schema keyword could accommodate it, and adding an `"acceptPatch"` field to the LDO would be inappropriately specific.

-----

One approach would be a `"usageHints"` field for hinting at protocol usage meta-data that could otherwise be discovered by interacting with the target resource.  This field would **not** encompass meta-data that describes the target resource independent of protocol.

The value of this keyword would be an object.  The exact possible values for the contents of the object are deferred to the protocol and the target resource, but Hyper-Schema would define general rules to provide for interoperability.

Here's a strawman proposal:

* keys would be the meta-data field names in whatever form is deemed canonical by the relevant standards, e.g. `"Accept-Patch"` and not a JSON-ized `"acceptPatch"`
* values would be as they would exist in the protocol, except that lists and named fields SHOULD be structured using JSON arrays and objects

So if we wanted to hint at the following authentication header (example taken directly from RFC 7235)

    WWW-Authenticate: Newauth realm="apps", type=1,
                           title="Login to \\"apps\\"", Basic realm="simple"

along with hinting at available HTTP methods (example taken directly from RFC 7231)

    Allow: GET, HEAD, PUT

the result would look like:

becomes:
```JSON
{
    "rel": "self",
    "href": "/foos/{id}",
    "usageHints": {
        "Allow": ["GET", "HEAD", "PUT"],
        "WWW-Authenticate": {
            "Newauth": {
                "realm": "apps",
                "type": 1,
                "title": "Login to \"apps\""
            },
            "Basic": {
                "realm": "simple"
            }
        }
    }
}

handrews:
[...]

PR #383, opened 2017-08-31
Add "targetHints"

handrews:
This is a very tentative first pass to get the discussion started.

This addresses half of #296, specifically advertising information
about the target that would likely be discovered through mechanism
such as the headers of an HTTP HEAD and/or OPTIONS response.
The half for describing request header (or analogous) data will
be posted separately after we see how this goes.

The goal for this draft is to specify this well enough to get
real-world implementations and feedback.  It seems best to
approach that goal by providing loose guidelines and encouraging
experimentation.

The lack of a standard way to advertise HTTP method support,
for instance, is a ***major barrier to adoption for many who have
expressed interest in Hyper-Schema***.

This draft needs to include an approach that will at least
easily fulfill that need while encouraging feedback on what
further guidance is needed for more comprehensive use.
Waiting for the next draft to work up a more detailed
proposal that might not hold up in the wild anyway will
only delay adoption and feedback.

This will be more useful for feedback than simply providing
an "allow" keyword, as it will encourage exploring other
use cases, which a narrowly targeted "allow" keyword would not.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1076,6 +1076,9 @@ GET /foo/
                 </t>
                 <section title='"targetHints" for HTTP'>
                     <t>
+                        <cref>It would be good to also include a section with CoAP examples.</cref>
+                    </t>
+                    <t>
                         JSON serializations of HTTP response header information SHOULD follow the
                         guidelines established by the work in progress
                         <xref target="I-D.reschke-http-jfv">"A JSON Encoding for HTTP Header Field Values"</xref>.
@@ -1101,11 +1104,6 @@ GET /foo/
 }]]>
                         </artwork>
                     </figure>
-                </section>
-                <section title='"targetHints for CoAP"'>
-                    <t>
-                        <cref>TBD</cref>
-                    </t>
                 </section>
             </section>
 
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -10,6 +10,7 @@
 <!ENTITY rfc5988 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5988.xml">
 <!ENTITY rfc6570 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6570.xml">
 <!ENTITY rfc7231 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7231.xml">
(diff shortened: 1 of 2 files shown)


jsonschema-hyperschema.xml
```diff
@@ -974,6 +975,86 @@ GET /foo/
...
+                    is expected to be discoverable through interacting with the target resource,
+                    typically in the form of protocol-specific control information or meta-data
+                    such as headers returned in response to an HTTP HEAD or OPTIONS request.
+                    The protocol is determined by the "href" URI scheme, although note that
+                    resources are not guaranteed to be accessible over such a protocol.
+                </t>
+                <t>
+                    The value of this property SHOULD be an object.  The keys to this object
+                    SHOULD be lower-cased forms of the control data field names.  Each value
+                    SHOULD be an array, in order to uniformly hanlde multi-valued fields.
```
dlax: typo: hanlde

handrews:
@dlax 
> I'm not sure we need the CoAP example (currently TBD) in a first version.

Well we don't **need** it.  But IoT folks use CBOR and CoAP and are an increasing presence.  I'm not going to delay anything for it, just wanted to remember to see if anyone would take a shot.

handrews:
@dlax I have fixed the typo (and another one- quote in the wrong place on targetHints for HTTP title) although I squashed them into the first commit (long-standing habit and I didn't think to leave the split for review, sorry).

I also moved the CoAP thing into a cref comment and out of the main text (this is it's own commit) and reworded the meta-schema commit message.

handrews:
Last call!  This will get merged tomorrow otherwise.  I expect it to evolve further based on feedback, and there are cref notes to that effect in the XML source.

The commit I just pushed adds the keyword to link.json, which I forgot.  It's a direct copy from hyper-schema.json so I'm not waiting on a separately review.  I loaded links.json through a JSON parser to make sure I didn't break the file.

handrews:
Updated on top of #406 which only affected links.json and hyper-schema.json

links.json
```diff
@@ -32,6 +32,9 @@
                 { "$ref": "http://json-schema.org/draft-06/hyper-schema#" }
             ]
         },
+        "targetHints": {
+            "description": "JSON representation of likely metadata such as response headers, with a protocol-dependent format"
```
dlax: Hm, didn't we just removed "description"?
(Will send a PR to fix that.)

handrews: Ugh, yes.  I've been rebasing and fixing these all evening.  Missed this one.

PR #384, opened 2017-08-31
Moving "media" into validation (as "contentEncoding"+"contentMediaType")

dlax:
Addresses the other part of #363 (see also #378) by moving the "media" keyword into the validation document and splitting it into "contentEncoding"+"contentMediaType" in a (new) dedicated section similar to the one about "format".

All references to "multimedia" are also removed from the Hyper-Schema document. I kept most of the original textual content and examples as a first step.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -121,8 +121,6 @@
             <t>
                 This document describes how JSON Schema can be used to define hyperlinks on instance
                 data.
-                It also defines how to provide additional information required to interpret JSON
-                data as rich multimedia documents.
             </t>
             <t>
                 As with all JSON Schema keywords, all the keywords described in the "Schema
@@ -131,8 +129,7 @@
             </t>
             <figure>
                 <preamble>
-                    Here is an example JSON Schema defining hyperlinks, and providing a multimedia
-                    interpretation for the "imgData" property:
+                    Here is an example JSON Schema defining hyperlinks:
                 </preamble>
                 <artwork>
 <![CDATA[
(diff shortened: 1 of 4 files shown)


jsonschema-hyperschema.xml
```diff
@@ -963,7 +873,7 @@ GET /foo/
         "properties": {
             "message": {
                 "description": "Re-interpret `message` as HTML",
-                "media": {
+                "contentMediaType": {
                     "type": "text/html"
```
handrews: You no longer need the sub-field "type" here :-)

handrews:
@dlax can you rebase and fix the conflicts?

@Relequestual does this change look good to you?  Similar to the readOnly move, but I think less controversial because semantic description of *content* (as opposed to usage) pretty clearly belongs with `format` in the validation spec.  It has nothing to do with hypermedia at all (unless you think media types are exclusive to hypermedia, which I do not, but that's the only reason I can think of that they were in hyper-schema at all).

jsonschema-validation.xml
```diff
@@ -993,6 +995,116 @@
...
+                <t>
+                    These properties provide additional information required to interpret JSON data
+                    as rich multimedia documents.
+                </t>
+            </section>
+
+            <section title="Implementation requirements">
+                <t>
+                    Implementations MAY support keywords defined in this section.
+                    Should they choose to do so, they SHOULD offer an option to disable validation
```
Relequestual: What is the reason / use case for this?

dlax: This is to mirror what's stated for "format" keyword (in "Implementation requirements" section). The reasoning is that "format" and "media*" keywords are semantically similar, so implementations should treat them similarly.

dlax:
> @dlax can you rebase and fix the conflicts?

Done.

PR #385, opened 2017-08-31
Add "anchorPointer" for in-instance contexts

handrews:
This will need a bit more work after "anchor" (#352) is merged in,
as the improved wording about what the default context is
is all in that other PR.  But the key points are well enough
developed for review.

In the interst of getting the main points out for review, I
referenced the long-expired Relative JSON Pointer I-D.  Maybe
that's OK since this is not even a working group document yet,
or we can talk about how to either move the relevant parts in,
or re-submit that I-D ourselves as well.

This goes along with "hrefPointers" ( #386 ), which adjusts template
variable resolution in the same way that this keyword adjusts
context location.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -9,6 +9,7 @@
 <!ENTITY rfc5988 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5988.xml">
 <!ENTITY rfc6570 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6570.xml">
 <!ENTITY rfc7231 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7231.xml">
+<!ENTITY I-D.luff-relative-json-pointer SYSTEM "https://xml2rfc.tools.ietf.org/public/rfc/bibxml3/reference.I-D.draft-luff-relative-json-pointer-00.xml">
 <!ENTITY I-D.reschke-http-jfv SYSTEM "https://xml2rfc.tools.ietf.org/public/rfc/bibxml3/reference.I-D.draft-reschke-http-jfv-06.xml">
 ]>
 <?rfc toc="yes"?>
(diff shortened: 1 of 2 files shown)


jsonschema-hyperschema.xml
```diff
@@ -883,6 +884,91 @@ GET /foo/
...
+                    <artwork>
+<![CDATA[{
+    "type": "object",
+    "properties": {
+        "theThing": {
+            "type": "object",
+            "properties": {
+                "name": {"type": "string"}
+            }
+        }},
```
dlax: extra `}`

jsonschema-hyperschema.xml
```diff
@@ -883,6 +884,91 @@ GET /foo/
...
+        "examples": {
+            "title": "The collection of example for the thing",
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "id": {"type": "integer"}
+                },
+                "links": [{
+                    "rel": "item"
```
dlax: missing trailing `,`

jsonschema-hyperschema.xml
```diff
@@ -883,6 +884,91 @@ GET /foo/
...
+                        Given this instance of media type application/json:
+                    </preamble>
+                    <artwork>
+<![CDATA[{
+    "theThing": {"name": "An actual thing"},
+    "examples": [{"id": 1234}, {"id": 5678}]
+}]]>
+                    </artwork>
+                    <postamble>
+                        The default context for the the link to "/example/1234" is the first
```
dlax: "the the"

dlax: Also, it should be `"/examples/1234"` I guess (same below).

handrews: yup, correct on all accounts.  Should be fixed now.

handrews:
The updated commit is just rebased to resolve conflicts after merging `"anchor"`, and squashed the typo fixes into the commit since they have been reviewed separately already.

There are no changes whatsoever to the overall contents of the PR.

handrews:
The additional commit pushed just now makes minor changes since "anchor" has now been merged, and this keyword builds on that concept.

dlax:
@handrews We should add the new keyword to the meta-schema, I guess?

handrews:
> We should add the new keyword to the meta-schema, I guess?

Oops- that was one of the post-"anchor"-merge fixups I was going to do, otherwise they would conflict as I'd put them in the same place.  Will update momentarily.

Issue #49, opened 2016-09-15
Criteria for building hyperlinks


awwright:
Right now, links that use variables that aren't defined are just cast as empty. There should be some way to specify a default value, or to depend on certain variables existing and skip the hyperlink if it doesn't exist.

For example, if there's a schema like

``` javascript
{ links: [
    { href:"/doc/{uuid}", rel:"self"  }
]
}
```

but multiple posts don't have a "uuid" property, then they all get the _same_ URI of <`/post/`> and all of a sudden we're saying multiple different posts are actually the "same". Oops!

This can sort of be done right now like so:

``` javascript
{ anyOf: [
   {
      required: ['uuid'],
      links: [ {href:'/doc/{uuid}', rel:'self'} ],
    }
]
}
```

... but this is bulky.

handrews:
I have thoughts on this but they really need to be part of a much more comprehensive discussion of hyper-schema.

Also, you use `doc

handrews:
#108 addresses this thoroughly and shows ways to handle both schema-supplied defaults and user-supplied data.

handrews:
@awwright does using an `"hrefSchema"` with `"default"` and/or `"required"` address this use case?

handrews:
I'm pretty sure that the following things together solve this problem:

* The `anyOf`/`required` approach for only allowing links under certain conditions.  This will become much more intuitive with `if`/`then`/`else`
* The ability to use `required` in `hrefschema` to require input.
* Both of these techniques can be used together to handle cases where you must have either instance data or client input, but need not have both.

While the required instance field technique is bulky, it is not complex, and I am not convinced the use case is common enough to warrant special syntax.  Especially with if/then/else helping things.

Barring a new argument or clarity on how widespread the usage really is, I plan to close this assuming if/then/else is merged.

handrews:
Note that we found another use case while working out #386, so more work is needed.

handrews:
After thinking this through a good bit, I don't think it's possible to solve this with `hrefSchema` at all.  It is designed to validate input independent of the instance (although the client MAY pre-populate the input with instance data if a variable appears in both the instance and `hrefSchema`).

Setting a property to `required` in `hrefSchema` means that it must be present in the input data set. 
But if you want to require something that can't be accepted as input, then what?

The spec says that giving a field a `false` schema indicates that it MUST NOT come from input data.  So I had a momentary thought of setting the field's schema to false but also making it required.

[...]

PR #386, opened 2017-08-31
Add "hrefPointers" for adjusting resolution

handrews:
This is a companion to "anchorPointer" (#385), as they use the same
mechanism for adjusting locations within the instance.  While
"anchorPointer" adjusts the context, "hrefPointers" adjust
how template variables are resolved.

"hrefPointers" replaces and is much more functional than the
preprocessing rules that were removed in draft wright-01.
Those rules only allowed adjusting to use the sub-instance
as a whole intead of individual properties or elements.
"hrefPointers" supports that as well as using data from
any containg or sub-instance.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -588,7 +588,7 @@
     "links": [
         {
             "rel": "self",
-            "href": "/trees{/root,parent,id}"
+            "href": "/trees{/root,parent,id}",
             "hrefPointers": {
                 "root": "/id",
                 "parent": "2/id"
@@ -597,17 +597,27 @@
     ]
 }]]>
                     </artwork>
-                    <postamble>
-                        The self links here should always include the root node id,
-                        followed by an immediate parent id if one exists, and always ending
-                        with the id of the node that is the context of the self link.
-                        The context node's id resolves with the standard process and
-                        does not appear in "hrefPointers".  The root id is located
-                        using an absolute JSON Pointer, while the parent node
-                        uses a Relative JSON Pointer, going up two levels (one level up
-                        is the array of children, two levels up is the whole parent node).
-                    </postamble>
                 </figure>
+                <t>
+                    The self links here should always include the root node id,
+                    followed by an immediate parent id if one exists, and always ending
+                    with the id of the node that is the context of the self link.
+                </t>
+                <t>
+                    The context node's id resolves with the standard process and
+                    does not appear in "hrefPointers".  The root id is located
+                    using an absolute JSON Pointer, while the parent node
+                    uses a Relative JSON Pointer, going up two levels (one level up
+                    is the array of children, two levels up is the whole parent node).
+                </t>
+                <t>
+                    This design never has more than three ids in the URI.  Nodes
+                    along the path between the root and the immediate parent,
+                    if any, are omitted.  This illustrates a limitation of JSON Schema:
+                    Producing a variable-length path URI to any arbitrary depth is
+                    not possible without redesigning the representation to make an
+                    array of path ids available in each node.
+                </t>
                 <figure>
                     <preamble>
                         Given the following instance:
@@ -621,7 +631,10 @@
       "id": 2,
       "children": [ {
         "id": 3
-} ] } ] } ] }]]>
+      } ]
+    } ]
+  } ]
+}]]>
                     </artwork>
                 </figure>
                 <t>
@@ -629,13 +642,15 @@
                     by JSON Pointers:
                     <list style="hanging">
                         <t hangText='"" (the root node)'>/trees/0/0</t>
-                        <t hangText='"/children/0/children/0"'>/trees/0/0/1</t>
-                        <t hangText='"/children/0/children/0/children/0"'>/trees/0/1/2</t>
-                        <t hangText='"/children/0/children/0/children/0/children/0"'>/trees/0/2/3</t>
+                        <t hangText='"/children/0"'>/trees/0/0/1</t>
+                        <t hangText='"/children/0/children/0"'>/trees/0/1/2</t>
+                        <t hangText='"/children/0/children/0/children/0"'>/trees/0/2/3</t>
                     </list>
                     This shows the use of an absolute JSON Pointer to always resolve
                     the root node from any depth, and a Relative JSON Pointer to always resolve
                     the immediate parent, which cannot be identified with an absolute pointer.
+                </t>
+                <t>
                     For the root node, there is no defined parent, so per URI Template resolution
                     rules, that variable and the path component that it would produce are simply
                     omitted, giving us the desired URI with two components after "/trees".
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -537,7 +537,7 @@
                         </section>
                     </section>
 
-                    <section title="Missing values">
+                    <section title="Missing values" anchor="missingValues">
                         <t>
                             Sometimes, the appropriate values will not be available.
                             For example, the template might specify the use of object properties,
@@ -574,8 +574,13 @@
                 </t>
                 <figure>
                     <preamble>
-                        Consider the following schema for an n-ary tree using the path expansion
-                        URI Template syntax:
+                        Recall that a Relative JSON Pointer which is not also a regular
+                        JSON Pointer begins with a number indicating how many levels
+                        up to move before applying the remainder of the pointer, if any, in the
+                        same manner as a regular JSON Pointer.  Consider the following schema
+                        for n-ary tree nodes, where a node is identified by its own id as well
+                        as the tree's root node id, and links are given for the IANA-registered
+                        "self" and "up" link relation types.
                     </preamble>
                     <artwork>
 <![CDATA[{
@@ -588,10 +593,17 @@
     "links": [
         {
             "rel": "self",
-            "href": "/trees{/root,parent,id}",
+            "href": "/trees/{rootId}/nodes/{id}",
             "hrefPointers": {
-                "root": "/id",
-                "parent": "2/id"
+                "rootId": "/id"
+            }
+        },
+        {
+            "rel": "up",
+            "href": "/trees/{rootId}/nodes/{parentId}",
+            "hrefPointers": {
+                "rootId": "/id",
+                "parentId": "2/id"
             }
         }
     ]
@@ -599,24 +611,12 @@
                     </artwork>
                 </figure>
                 <t>
-                    The self links here should always include the root node id,
-                    followed by an immediate parent id if one exists, and always ending
-                    with the id of the node that is the context of the self link.
-                </t>
-                <t>
                     The context node's id resolves with the standard process and
-                    does not appear in "hrefPointers".  The root id is located
+                    therefore does not appear in "hrefPointers".  The root id is located
                     using an absolute JSON Pointer, while the parent node
                     uses a Relative JSON Pointer, going up two levels (one level up
-                    is the array of children, two levels up is the whole parent node).
-                </t>
-                <t>
-                    This design never has more than three ids in the URI.  Nodes
-                    along the path between the root and the immediate parent,
-                    if any, are omitted.  This illustrates a limitation of JSON Schema:
-                    Producing a variable-length path URI to any arbitrary depth is
-                    not possible without redesigning the representation to make an
-                    array of path ids available in each node.
+                    is the array of children, two levels up is the whole parent node),
+                    and then looking at the "id" field from that point.
                 </t>
                 <figure>
                     <preamble>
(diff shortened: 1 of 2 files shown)


hyper-schema.json
```diff
@@ -21,6 +21,8 @@
                     "type": "string",
                     "format": "uri-template"
                 },
+                "hrefPointers": {
+                    "description": "a map of variable names to relative instance JSON Pointers, which adjust the instance location from which template variable resolution begins",
```
dlax: Missing closing `}`

jsonschema-hyperschema.xml
```diff
@@ -653,6 +662,84 @@
...
             </section>
 
+            <section title="hrefPointers" anchor="hrefPointers">
+                <t>
+                    The value of the "hrefPointers" link description property MUST be
+                    an object.  Each property value in the object MUST be a valid
+                    <xref target="I-D.luff-relative-json-pointer">Relative JSON Pointer</xref>,
+                    which is evaluated relative to the position in the instance from which
+                    <xref target="href">"href"</xref> template variable resolution would
+                    normolly begin.
```
dlax: typo: normolly

jsonschema-hyperschema.xml
```diff
@@ -653,6 +662,84 @@
...
+    "properties": {
+        "id": {"type": "integer", "minimum": 0},
+        "children": {"type": "array", "items": {"$ref": "#"}}
+    },
+    "links": [
+        {
+            "rel": "self",
+            "href": "/trees{/root,parent,id}"
+            "hrefPointers": {
+        }
```
dlax: I guess we'd expect something within `hrefPointers` :)

jsonschema-hyperschema.xml
```diff
@@ -653,6 +662,84 @@
...
+                <t>
+                    We get the following self links for each node in the instance as identifed
+                    by JSON Pointers:
+                    <list style="hanging">
+                        <t hangText='"" (the root node)'>/trees/0/0</t>
+                        <t hangText='"/children/0/children/0"'>/trees/0/0/1</t>
+                        <t hangText='"/children/0/children/0/children/0"'>/trees/0/1/2</t>
+                        <t hangText='"/children/0/children/0/children/0/children/0"'>/trees/2/3</t>
+                    </list>
+                    omitted, giving us the desired URI with two components after "/trees".  The
```
dlax: This line is weird and unfinished.

hyper-schema.json
dlax: Also maybe a "type" property.

handrews: I avoided a "type" on purpose.  While most will be objects, it would not be that odd to have an array.  I can't think of a situation where this sort of data would be best handled as a string or a number but I couldn't thing of a compelling reason to restrict it either.

We can always see how it gets used and tighten it in future drafts if it really seems necessary.  In general, the meta-schema errs on the side of permissiveness.

jsonschema-hyperschema.xml
handrews: left as an exercise for the reader :-P
(I'll fix it)

jsonschema-hyperschema.xml
handrews: Yeah... I thought I'd copied over the example from the issue, but I have no idea what actually happened here.

handrews:
@dlax the tree example should now be a lot easier to follow with actual pointers in the the schema, and without a stray line of explanation duplicated into a really confusing place.  That will teach me to post a PR without actual proofreading.

I would not be surprised if it's still a bit unclear, please do bring up those issues and I will improve the wording.

jsonschema-hyperschema.xml
```diff
@@ -653,6 +662,91 @@
...
+                    </artwork>
+                </figure>
+                <t>
+                    We get the following self links for each node in the instance as identifed
+                    by JSON Pointers:
+                    <list style="hanging">
+                        <t hangText='"" (the root node)'>/trees/0/0</t>
+                        <t hangText='"/children/0/children/0"'>/trees/0/0/1</t>
+                        <t hangText='"/children/0/children/0/children/0"'>/trees/0/1/2</t>
+                        <t hangText='"/children/0/children/0/children/0/children/0"'>/trees/0/2/3</t>
```
dlax: I'm trying to apply the explanation in `<postamble>` literally and can't get the last three examples working:
* "/children/0/children/0" points to `{"id": 2, "children": [{"id": 3}]}`, so it's URI should be `/trees/0/1/2` because it's parent has "id=1" and the node has "id=2"
* "/children/0/children/0/children/0" points to `{"id": 3}`, so it's URI should be `/trees/0/2/3` (parent's id=2, context node's id=3)
* "/children/0/children/0/children/0/children/0" does not resolve

jsonschema-hyperschema.xml
```diff
@@ -653,6 +662,91 @@
...
+    "type": "object",
+    "required": ["id"],
+    "properties": {
+        "id": {"type": "integer", "minimum": 0},
+        "children": {"type": "array", "items": {"$ref": "#"}}
+    },
+    "links": [
+        {
+            "rel": "self",
+            "href": "/trees{/root,parent,id}"
```
dlax: trailing `,` missing

jsonschema-hyperschema.xml
handrews: OMG I am so sorry, all of these have an extra /children/0 in them, no wonder none of this makes sense to you.  :-(

I'll update this ASAP, but you can see that your results work perfectly if you chop off that extra path component.  I think this was a copy-paste error from when I had more levels in the tree, and for some reason I cut the wrong part of the example when trimming it down.

handrews: > Similarly, I'd expect /trees/0/1/2/3 to be the URI of inner node ({"id": 3}), but the href model does not allow this.

This is where I always have trouble with examples.  Perhaps this is just because the current example literally makes no sense the way I messed it up.  But I find that so many people just want to redesign the example to do something else instead of reading the premise.

The intention was specifically *not* to do a normal tree layout.

The point is to show the need for both absolute and relative pointers.  I'm not attached to this example, it just (when written properly) shows it more concisely than anything else.  I try not to do this one with collections because inevitably, someone complains "I just wouldn't do collections that way."

I hate writing examples, and everyone always hates my examples :-P

handrews: Although you bring up an interesting point that it's not possible to do that "normal' tree layout with `hrefPointers` AFAIK.  Hmm... I don't think we absolutely need to solve every URI construction approach, but it's worth thinking about.

jsonschema-hyperschema.xml
```diff
@@ -653,6 +662,91 @@
...
+                    <artwork>
+<![CDATA[{
+  "id": 0,
+  "children": [ {
+    "id": 1,
+    "children": [ {
+      "id": 2,
+      "children": [ {
+        "id": 3
+} ] } ] } ] }]]>
```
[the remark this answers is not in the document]
handrews: Yes, because it takes up absurd amounts of space otherwise.  I can do it closer the normal way, though.  I'll still leave the "[ {" / "} ]" pairs together because otherwise it really takes up too much vertical space.  But I'll put each closing pair on its own line, indented.

handrews:
@dlax hopefully the example finally actually works now.  I am very sorry for posting such a mess :-(

I have added some wording around not being able to do the tree layout you expect, although I'm not sure if that makes things more or less confusing.  I am totally open to replacing this example with another one (or perhaps two instead of trying to show everything at once?)  Any ideas are welcome.

jsonschema-hyperschema.xml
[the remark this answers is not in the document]
dlax: Much better now!

handrews:
@dlax any further thoughts on this?  Does the current example work?  Would you prefer to see multiple examples illustrating things a piece at a time?  Any specific requests?  I'm happy to write them but I'm notoriously bad at choosing compelling examples, so any requirements you can provide will help a lot.

philsturgeon:
This looks really interesting, and I think I understand it, but it could do with a more simplistic example before getting into the tree example. I think it would help those with little-to-no context understand the feature.

handrews:
@philsturgeon thanks!  If there are any pointers you can give me about what individual topics you'd like to see that would help.  I have always struggled to write examples that others find compelling (I work from abstract to concrete, which is the opposite of most people and a recurring communication challenge for me).

dlax:
> @dlax any further thoughts on this? Does the current example work? Would you prefer to see multiple examples illustrating things a piece at a time? Any specific requests? I'm happy to write them but I'm notoriously bad at choosing compelling examples, so any requirements you can provide will help a lot.

Yes, the current example works, thanks for improving it! (The "abstract" part is also fine, by the way.)

One thing that makes the example difficult to follow is the URI structure. So if we try to keep resulting "href" flat (i.e. only one variable path segment), the reader would only need to focus on the `hrefPointers` value. Another thing would be to use several links. For instance:
```
{
    "type": "object",
    "required": ["id"],
    "properties": {
        "id": {"type": "integer", "minimum": 0},
        "children": {"type": "array", "items": {"$ref": "#"}}
    },
    "links": [
        {
            "rel": "self",
            "href": "/{id}"
        },
        {
            "rel": "top",
            "href": "/{rootId}",
            "hrefPointers": {
                "rootId": "/id"
            }
        },
        {
            "rel": "up",
            "href": "/{parentId}",
            "hrefPointers": {
                "parentId": "2/id"
            }
        }
    ]
}
```
Does this cover all the cases you wanted to highlight?

handrews:
@dlax yes, this is great!  Conveys the same information much more simply.  This is exactly what I mean when I say I struggle with examples.  My inclination is always to overload or go for the most pathological situation, and I have no idea how long it would have been before it occurred to me to simply do each case like that :-P

@philsturgeon is @dlax's version easier to follow?

jsonschema-hyperschema.xml
```diff
@@ -653,6 +662,106 @@
...
+        "id": {"type": "integer", "minimum": 0},
+        "children": {"type": "array", "items": {"$ref": "#"}}
+    },
+    "links": [
+        {
+            "rel": "self",
+            "href": "/trees{/root,parent,id}",
+            "hrefPointers": {
+                "root": "/id",
+                "parent": "2/id"
```
philsturgeon: This 2 is a bit of a magic number at this point, as explanation of IDs 1, 2, 3, 4 only comes after. on line 731.

handrews: Makes sense, I'll address that.

handrews:
This does create a bit of confusion as to what the "up" link should produce for the root node.  In the complex example, it still produced a valid URI (however convoluted it was to understand).

In this version, you'll just get `/` (or probably `/nodes/` since I feel a stable prefix makes the template more clear: `/nodes/{parentId}`).

This is actually a pretty good example of issue #49 about nonsensical template expansion results.

handrews:
What about
```JSON
{
    "type": "object",
    "required": ["id"],
    "properties": {
        "id": {"type": "integer", "minimum": 0},
        "children": {"type": "array", "items": {"$ref": "#"}}
    },
    "links": [
        {
            "rel": "self",
            "href": "/trees/{rootId}/{id}"
            "hrefPointers": {
                "rootId": "/id"
            }
        },
        {
            "rel": "up",
            "href": "/trees/{rootId}/{parentId}",
            "hrefPointers": {
                "parentId": "2/id",
                "rootId": "/id"
            }
        }
    ]
}
```

which would make the parent of the root be `/trees/{rootId}/` which becomes `/trees/0/` which, um... could be defined as the same as `/trees/0/0`?  Not sure how to handle that...

philsturgeon:
I like it!

handrews:
I think that I like the variation I did, with two links, a bit better- @philsturgeon are you OK with that one?  Not sure which of the two you were commenting on.

The pathological case of the root's parent will require a bit of explanation, but is ***no worse*** of a problem than it is without `hrefPointers` (as shown by #49), so I don't think it's worth holding this PR up for that.

handrews:
I am also happy to go with the three-link option if that is more clear.  As noted, I'm not really the best judge of examples.

dlax:
@handrews 
> I think that I like the variation I did, with two links
> I am also happy to go with the three-link option if that is more clear.

The point of having several different links is that of them illustrates a specific situation (standard context resolution without `hrefPointers` and then standard and relative JSON Pointer with `hrefPointers`), maybe two is enough (explanations might be less redundant).

> The pathological case of the root's parent will require a bit of explanation, but is no worse of a problem than it is without hrefPointers (as shown by #49)

Maybe changing the href to `/trees/{rootId}/nodes/{id}` and `/trees/{rootId}/nodes/{parentId}` would help? For the root node, both `rel="self"` and `rel="up"` links would point to `/trees/0/nodes/0` which arguably makes sense for a root node.

> so I don't think it's worth holding this PR up for that.

Yes, just bikeshedding about the example :)

handrews:
> maybe two is enough (explanations might be less redundant

Yes, this is what I meant.  "self" can show the use of absolute pointers, and "up" relative ones.  We don't need to re-display the standard process, just show how it looks in context (not present in `hrefPointers` even though "rootId" is).

> Maybe changing the href to `/trees/{rootId}/nodes/{id}` and `/trees/{rootId}/nodes/{parentId}` would help?

This is actually exactly what I have in my workspace, but the "up" link is still `/trees/0/nodes/` because a missing field in URI Templates is simply removed.

handrews:
@philsturgeon it's got 2 days of the comment period left, and since I'm about to push a new version of it, I'll probably give it a few days extra.  No obligation for further review, but you certainly don't need to get it done today.

handrews:
@dlax I accepted one of your commits (by rebasing it directly from your fork into my workspace) and took an alternate approach to the other one.  And fixed the links.json meta-schema which was underspecified for some reason.

PR #387, opened 2017-08-31
Clarify various usage scenarios for URI templates

handrews:
Format the one we have better, and also explain how using
both instance data and describing client input is analogous
to common interactive form usage.

This is a response to #288, in that it is an attempt to explain
why these features work as they do.  The discussion there
has not produced any alternative that solves the use case
added here (defining input conditions with schemas, but
allowing the existing data to remain unchanged if no
action is taken- identical to pre-populating HTML form
input widgets with "value").

If @jdesrosiers comes up with such a proposal, we will
definitely consider that.  Otherwise, I am asking reviewers
to approve this if they feel that the given use cases justify
the current behavior.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -504,14 +504,39 @@
                         to server-supplied instance data if no user agent data is provided.
                     </t>
                     <t>
-                        The common pattern of resolving a templated path component with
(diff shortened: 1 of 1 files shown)


handrews:
Anyone?  This is not a large PR, and does not change the spec behavior at all.

Issue #86, opened 2016-10-11
spec: remove/change extending meta-schemas recommendation


epoberezkin:
https://github.com/json-schema-org/json-schema-spec/blob/master/jsonschema-core.xml#L254

The suggestion to extend meta-schemas using allOf (assuming that one of subschemas is $ref) is not feasible because meta-schemas are recursive and they have references to the meta-schema itself (`"#"`). Therefore if meta-schema is extended using $ref to the original meta-schema the # ref in that meta-schema will not point to the extended one, it will still point to the original, so the schemas would not be properly validated.

The only working meta-schema extension mechanism at the moment is copy-pasting and adding keywords, suggestion to use allOf is misleading because it implies $ref.

We have several options:
1. Suggest copy-pasting with explanation why it is the only working option
2. Adopt $merge/$patch keywords proposed by @fge
3. Drop this suggestion from this version of the spec until we agree what is the right thing.

Happy to draft 1, can draft 2 as well...

handrews:
For any extension of JSON Schema, if an instance validates against the extended version, it should also validate against the regular JSON Schema.  So while you do need to re-declare the properties with child schemas so that those children reference the extended meta-schema, doing that within the `allOf` is sufficient to make everything work.

If your extension sets `additionalProperties` to `false` (which is not advised), then we're back in a specific re-use problem, and I have a very detailed proposal that I hope to post today or tomorrow (it's derived from use cases and vague proposals from the email thread on the google group a while back).  So, @epoberezkin , please wait before writing up $merge/$patch.  For one thing, it's already written up as issue #15 , even if not specifically for this case.

awwright:
Without thinking too much about it, I'd venture to guess that patching the official meta-schema to replace all meta-schema references references to your own meta-schema would be really painful (at least if absolute URIs were used exclusively)

epoberezkin:
> patching the official meta-schema to replace all meta-schema references references to your own meta-schema would be really painful

@awwright you don't need to. See the example below.

> doing that within the allOf is sufficient to make everything work.

@handrews if you limit the definition of "work" to "pass validation of valid schemas", allOf keyword would indeed be sufficient. Actually with this limitation you don't have to extend the original meta-schema. If you include into the definition of "work" the requirement "to fail validation of invalid schemas" then it won't work.

Consider the example: you add a new keyword `my` to the schema vocabulary that must be boolean. To validate your new keyword you extend the meta-schema in this way:

```json
{
[...]

PR #389, opened 2017-09-01
Highlight the challenges of extending meta-schemas

handrews:
This addresses the original question posed in #86, which
noted that simply saying that meta-schemas can be extended
with "allOf" is misleading and likely to lead to errors.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -356,6 +356,11 @@
                     This extended meta-schema SHOULD be referenced using the "$schema" keyword, to
                     allow tools to follow the correct behaviour.
                 </t>
+                <t>
+                    Note that the recursive nature of meta-schemas requires re-definining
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -356,6 +356,11 @@
                     This extended meta-schema SHOULD be referenced using the "$schema" keyword, to
                     allow tools to follow the correct behaviour.
                 </t>
+                <t>
+                    Note that the recursive nature of meta-schemas requires re-definining
+                    recursive keywords in the extended meta-schema, as can be seen in
+                    ith the JSON Hyper-Schema meta-schema.
```
dlax: typo: "ith"

handrews: Thanks!  I'll fix and then merge, as I don't think this fundamentally changes the approval and this has no impact on implementations at all.

PR #390, opened 2017-09-02
Add "headerSchema"

handrews:
This addresses the request half of #296, and is a companion
keyword to "targetHints" (#383).

Like "targetHints", this is a rough first pass to get the
discussion of specifics started.  I expect it to be somewhat
underspecified to encourage experimentation.

I intentionally chose an example that is both very complex
and very useful.  It might also be worth putting up a simpler
example.  Although when this appears right after "targetHints",
the simpler header value examples in that section will help
motivate what is going on here.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -10,6 +10,7 @@
 <!ENTITY rfc6570 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6570.xml">
 <!ENTITY rfc6906 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6906.xml">
 <!ENTITY rfc7231 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7231.xml">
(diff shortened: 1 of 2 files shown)


jsonschema-hyperschema.xml
```diff
@@ -1095,6 +1097,123 @@ GET /foo/
                 </section>
             </section>
 
+            <section title="headerSchema" anchor="headerSchema">
+                <t>
+                    <cref>
+                        As with "targetHints", this keyword is somewhat under-specified
+                        to encourage experiementation and feedback as we try to balance
```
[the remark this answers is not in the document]
handrews: It's extra-experimental and experiential!

links.json
```diff
@@ -38,6 +38,12 @@
         "mediaType": {
             "type": "string"
         },
+        "headerSchema": {
+            "description": "Schema for protocol-specific headers or analogous data, as signified by URI scheme",
```
dlax: Maybe drop "description" here as well?

handrews: srsly?  I swear I've pushed three times as many rebases/fixes as there are PRs tonight...

handrews:
Last call- barring objections or reasonable requests for more time to review, this will get merged tomorrow.

Issue #377, opened 2017-08-28
Emphasize context per RFC 5988bis


handrews:
The reworking of text to define things around the instance (or sub-instance) as the link context instead of simple "server-supplied" data was well received in trial balloon change #366.

This issue is for discussing the idea and ensuring we have the concept and wording right, after which point I will file a new PR.  If no one comments I'll just take a pass at it similar to the trial balloon but narrowly focusing on this one concept.

handrews:
Here's a possible reworked organization of JSON Hyper-Schema, both to frame things in 5988bis terms and to introduce the various concepts at the right time.  The sections involved should be either obvious from 5988bis or from the current draft.  Client input description is not part of 5988bis (unless considered a kind of target attribute, I suppose).  It seemed worth putting in its own subsection.

1. Introduction
1. Conventions and Terminology
1. Overview
1. Meta-schema
1. Schema keywords
    1. Interaction with validation
    1. base
    1. links
1. Link description objects
    1. Context
        1. anchor
        1. anchorPointer
    1. Relation type
        1. rel
        1. "self" links
        1. "collection" and "item" links
    1. Target
        1. href
        1. hrefPointers
    1. Target attributes
        1. title
        1. description
        1. targetSchema
        1. targetHints
        1. mediaType
    1. Client input description
        1. Links, operations, and data
        1. hrefSchema
        1. headerSchema
        1. submissionEncType
        1. submissionSchema

handrews:
I think I would like to add a section 7 which would be one or more examples that we would use to illustrate how all of the pieces fit together.  For instance, if the `hrefRequired` proposal from #49, or something like it, is added, then calculating the actual URI goes something like this:

1. Start by combining `base` and `href` to get the full URI template
1. Extract the variable names, which default to reading from fields in the context instance
1. Apply `hrefPointers` to adjust the location within the instance corresponding to each variable
1. If `hrefSchema` is present:
    1. Optionally pre-populate the resolution data set with any instance data matching those variables
    1. Ask the client for input for any variables that do not have an `hrefSchema` value of `false`
    1. Add the client input to the resolution data set, overriding any pre-populated values
    1. Validate the resolution data set against `hrefSchema` (if it fails validation, error out here)
1. For any variables in the resolution data set that still lack values, populate them from the instance as adjusted by `hrefPointers`
1. If `hrefRequired` is present, verify that all variables listed there have values (if not, error out here)
[...]

Issue #297, opened 2017-04-03
Behavior when "rel" is absent?


handrews:
In draft-wright-json-schema-hyperschema-00 we made `"rel"` optional.  Either I'm missing something, or there is no description of the behavior when `"rel"` is not present.  Is the link to be considered untyped?

I'm not proposing any particular behavior, it just seems ambiguous enough to warrant a one sentence note somewhere.

jdesrosiers:
> In draft-wright-json-schema-hyperschema-00 we made "rel" optional

Wow, I can't believe I didn't notice that.  In draft-04 it would have made sense for `rel` to be optional, but I'm not sure that is the case anymore.

handrews:
> In draft-04 it would have made sense for rel to be optional, but I'm not sure that is the case anymore.

Yeah, we'll have to wait for @awwright to comment as I remember him making the change but don't remember any particular discussions (or objections, for that matter).  I seem to recall it making sense at the time but now I don't recall why :-P

awwright:
It appears to be optional in [RFC 5988 HTTP Linking](https://tools.ietf.org/html/rfc5988) so having it be optional would make sense.

handrews:
@awwright it's definitely optional in the HTTP link header, but per RFC 5988's introduction:

>    To this end, this document defines a framework for typed links that isn't specific to a particular serialisation or application.

and then

> This document addresses [the non-standardness of the HTTP Link header] by re-specifying the Link header as one such serialisation, with updated but backwards-compatible syntax.

so I think that the main part of RFC 5988 doesn't say anything one way or the other about required-ness of "rel", while the HTTP link header is *a* serialization but not necessarily one that should be assumed as the template for all serializations.  It also doesn't seem overly clear on what happens in the absence of "rel".

I think a line such as "in the absence of a relation type, an implementation MUST (SHOULD?) NOT make any assumptions about the link's semantics" would clarify the situation without overly burdening the spec.

Anthropic:
Can I clarify my basic understanding please?

W3C html5 recommendation has ["A link element must have a rel attribute"](https://www.w3.org/TR/html5/document-metadata.html#attr-link-rel).

I'm assuming now that our links section is also for the equivalent of hyperlinks which the same doc says of links **without rel** ["*This implied hyperlink has no special meaning* (it has no link type) beyond linking the element's document to the resource given by the element's href attribute."](https://www.w3.org/TR/html5/links.html#hyperlink)

[...]

PR #393, opened 2017-09-03
Make "rel" required.

handrews:
Addresses #297 

There are two commits here: the first simply makes "rel" required, while
the second attempts to address concerns over having to create new URIs
for custom link relations by pointing out two simple approaches.

The `"rel"` section is getting large, but I will address that when I address
#377 in which I have suggested a new outline to more clearly organize
the spec.

RFC 5988bis lists the link relation type as one of the three
required aspects of a link.

Since we do not need to support the deprecated "rev" mechanisms
for specifying relation types, we can make "rel" required.

This will encourage hyper-schema authors to think more about
choosing relation types.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -645,6 +645,7 @@
                     resource. The value MUST be a registered link relation from the
                     <xref target="RFC5988">IANA Link Relation Type Registry established in RFC 5988</xref>,
                     or a normalized URI following the <xref target="RFC3986">URI production of RFC 3986</xref>.
+                    This property is required.
                 </t>
 
                 <t>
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -792,8 +792,8 @@
                     encouraged to mint their own extension relation types, as described in
                     <xref target="RFC5988">section 4.2 of RFC 5988</xref>.  The simplest
                     approaches for choosing link relation type URIs are to either use
-                    a URI scheme that the system is already using to identify its primary
-                    resources, or to use a human-readable, no-dereferenceable URI scheme
+                    a URI scheme that is already in use identify the system's primary
```
dlax: already in use *to* identify

handrews:
Last call!  Barring objections, this will get merged in the next day or two.

PR #397, opened 2017-09-07
Fix recursion in hyper-schema.json

levbishop:
- Fix recursion to cover `contains` (specifically mentioned in section 3.1 of the hyper-schema spec), and `propertyNames` (probably irrelevant, since none of the hyper-schema fields would apply, but feels like a better practise since the core spec refers to the hyper-schema definition for guidance on recursion when extending schemas (see also #86)) 
- Much more concise definition of recursion.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -3,14 +3,6 @@
     "$id": "http://json-schema.org/draft-06/hyper-schema#",
     "title": "JSON Hyper-Schema",
     "definitions": {
-        "schemaArray": {
-            "allOf": [
-                { "$ref": "http://json-schema.org/draft-06/schema#/definitions/schemaArray" },
-                {
-                    "items": { "$ref": "#" }
-                }
-            ]
-        },
         "linkDescription": {
             "title": "Link Description Object",
             "type": "object",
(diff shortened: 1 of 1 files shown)


levbishop:
Perhaps the recursion is a little *too* concise and the regexes should be expanded a bit from, eg, `"^(any|all|one)Of$"` to `"^(anyOf|allOf|oneOf)$"`? 

Or perhaps the original verbose form is clearest and only needs `contains` and `propertyNames` added?

handrews:
@levbishop thanks for submitting a PR!  It's great to see a new name in the PR list.

> Or perhaps the original verbose form is clearest and only needs contains and propertyNames added?

Yes, I think this is preferred.  Many people use the meta-schemas as shorthand documentation.  It's debatable as to how wise that is, but using patternProperties here is more than a little off-putting.  I rarely see it used at all across several large projects, and not everyone is comfortable with regexes.

While this PR is fine (especially since you are clearly open to re-working it), most of the time it's better to file an issue first until you've been through a draft cycle and have a feel for how things land with various team members.  But definitely no need for that here, if you change this to just at contains and propertyNames I'll be happy to merge it (I'm going to double check and see if there's a reason they were left out, but I'm pretty sure it was an accident)

levbishop:
Reworked to verbose format for recursion (still slightly simplified relative to original).

PR #403, opened 2017-09-13
Fix links.json being out of date.

handrews:
I always forget about links.json.  So does everyone else, it would seem.

Most of the recently added keywords were missing, and hrefSchema
had the wrong $ref target.


handrews:
No longer necessary due to #406.

PR #404, opened 2017-09-13
Control link applicability with "hrefRequired"

handrews:
***EDIT:*** _This PR has been rebased so that it only contains the new commits.  Originally it was posted piggybacking on #386, which has since been accepted._

-----

This addresses issue #49.

Add a keyword, "hrefRequired", that prevents a link from being
used if a variable named in that keyword's array cannot be resolved
to a value by any available mechanism.

Remove outdated language about non-object instances causing problems,
as this is addressed by "hrefPointers".

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -537,7 +537,7 @@
                     <section title="Missing values" anchor="missingValues">
                         <t>
                             Sometimes, the appropriate values will not be available.  In many
-                            cases, the URI Template behavior of simply removing varibles that
+                            cases, the URI Template behavior of simply removing variables that
                             do not have a value will be appropriate.  An example of this is
                             optional query parameters, the presence or absence of which does
                             not change the nature of the link relation type.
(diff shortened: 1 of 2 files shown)


jsonschema-hyperschema.xml
```diff
@@ -644,18 +638,23 @@
 
                     <section title="Missing values" anchor="missingValues">
                         <t>
-                            Sometimes, the appropriate values will not be available.
-                            For example, the template might specify the use of object properties,
-                            but no such data was provided (or "hrefSchema" is not present), and the
-                            instance is an array or a string.
+                            Sometimes, the appropriate values will not be available.  In many
+                            cases, the URI Template behavior of simply removing varibles that
```
dlax: typo "varibles"

handrews: I swear I actually do use a spell checker.  But so many technical words don't count as spelled right I sometimes miss real errors.

links.json
```diff
@@ -24,6 +24,12 @@
                 { "$ref": "http://json-schema.org/draft-06/hyper-schema#" }
             ]
         },
+        "hrefRequired": {
+            "type": "array",
+            "items": {
+                "type": "string"
+            }
```
dlax: We could have a `"uniqueItems": true` here I guess?

handrews:
Last call!  This goes in tomorrow barring substantial objections.

PR #399, opened 2017-09-08
Support date and time, and allow for more RFC 3339

handrews:
This addresses #199 by supporting the two least controversial and
most requested addtional date/time formats, and reserving the
remaining namespace while we decide whether or how to specify
additional RFC 3339 formats.

Issue #392, opened 2017-09-02
Add "regex" back as a format


handrews:
It's still being used in the meta-schema to describe `pattern`, FFS.  Probably a good indication of its usefulness.  Removing it was arguably a bug.

handrews:
Merged #398.

Issue #355, opened 2017-08-20
"iri" and "iri-reference" formats


handrews:
See also #59 about IRI support throughout the spec.

It would seem to me like we would want these for symmetry with "uri" and "uri-reference".

AFAIK there is no IRI Template spec so no "iri-template" format.

Putting this in the next milestone because it's easy and I've generally rounded up all of the format issues to address this time around.

handrews:
IRIs and URI Templates, from RFC 6570:

>   Although the URI syntax is used for the result, the template string
   is allowed to contain the broader set of characters that can be found
   in Internationalized Resource Identifier (IRI) references [RFC3987].
   Therefore, a URI Template is also an IRI template, and the result of
   template processing can be transformed to an IRI by following the
   process defined in Section 3.2 of [RFC3987].

handrews:
Merged #408.

Issue #305, opened 2017-04-12
Reference RFC 6531 in email format validation


Sintendo:
The "email" format explicitly references [RFC 5322](https://tools.ietf.org/html/rfc5322). We should consider referencing [RFC 6531](https://tools.ietf.org/html/rfc6531), so email addresses with non-ASCII characters can be recognized.

awwright:
This looks correct at first glance, but RFC 6531 doesn't obsolete RFC 5322 so this is probably a little more involved than just replacing the references.

epoberezkin:
I think there can be another format for such email - there may be applications that do not allow non-latin characters in emails, so they would need the current definition...

For the reference, there was the conversation here: https://github.com/epoberezkin/ajv/issues/434

handrews:
@awwright would you prefer a separate format value for RFC 6531?  @epoberezkin raised that in the Ajv issue that he linked to.

If I am reading RFC 6531, the part where it says that it does not modify RFC 5322 means that any email that is valid against RFC 5322 is also valid against RFC 6531.  This feels analogous to URI vs IRI, though, and I would probably rather add "iri" and "iri-reference" instead of changing the meaning of "uri" and "uri-reference" so I guess I'm in favor of a separate keyword?  (I'm going to file "iri[-reference]" if don't have that filed already...)

handrews:
@Sintendo or anyone else: any preferences on a name for the additional format?  "i18n-email"?  "smtp-utf8"?  Those are the first that come to mind after skimming a bit of [RFC 6531](https://tools.ietf.org/html/rfc6531)

If someone can name this thing, I'm happy to include it in this draft.  it fits our general desire to be all modern and stuff :-)

Sintendo:
Both options look good to me. I think I prefer a name that still contains the text 'email' (that way it's more obvious that this new format is somehow related to the existing "email" format), so "i18n-email" gets my vote.

Julian:
Can we just axe the email format instead :)

handrews:
@Julian in all seriousness I think it would make sense to give implementations more flexibility about how formats are implemented.  Saying "if you implement format at all, you have to implement all formats" sets a really high bar for formats.  I feel like we don't have a good story for why format behaves the way it does.

[...]

PR #408, opened 2017-09-15
Internationalized formats

handrews:
This addresses #305 (email) and #355 (IRIs), and also adds internationalized domain/host names as they are what the internationalized email addresses are based on.

It's a multi-commit omnibus but if any are controversial they can be split.

I also put ipv4 and ipv6 together in a section as it just reads better like that.  Alongside the "Dates and times" section introduced in PR #399, these changes result in most of the formats being grouped into sections by topic.  Only `uri-template`, `json-pointer`, and `regex` are on their own (the latter of those being introduced in PR #392).

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -7,6 +7,7 @@
 <!ENTITY RFC2673 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.2673.xml">
 <!ENTITY RFC3339 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3339.xml">
 <!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
+<!ENTITY RFC3987 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3987.xml">
 <!ENTITY RFC4291 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.4291.xml">
 <!ENTITY RFC5322 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5322.xml">
 <!ENTITY RFC6570 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6570.xml">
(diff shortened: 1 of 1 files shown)


handrews:
Bueller?  @epoberezkin? @Julian? @awwright?

epoberezkin:
Format names look ok, should  I also double check the RFCs?

handrews:
@epoberezkin checking the RFC's is always helpful.  But mostly I just want to know if you think these formats make sense to have as an implementation author.

epoberezkin:
I guess they do, I've been chased to update the existing formats to include international chars, so I definitely prefer separate format names.

The only one that seems inconsistent is "idn"... Maybe "idn-hostname"? I kind of understand why you didn't do it this way as well, but hostname is a wider term than domain name, isn't it? so maybe there should be both "idn" and "idn-hostname"? Or should users use anyOf in case they need "ind-hostname"?

epoberezkin:
On another hand we don't have "domain", so why should we have "idn"?

handrews:
From RFC 5890:

>    The base DNS specifications [RFC1034] [RFC1035] discuss "domain
   names" and "hostnames", but many people use the terms
   interchangeably, as do sections of these specifications.  Lack of
   clarity about that terminology has contributed to confusion about
   intent in some cases.  These documents generally use the term "domain
   name".  When they refer to, e.g., hostname syntax restrictions, they
   explicitly cite the relevant defining documents.

Hostname vs domain name is one of those endlessly debatable things (we certainly debate it endlessly at Cloudflare, which is probably why we just call things "Zones" which is less intuitive than either other option).

Technically "foo.com" can be a hostname, although it's probably a domain name and "a.foo.com", "b.foo.com", etc. are hostnames.  Except that they can be "subdomains", etc. etc. etc.  TLDs can't be hostnames but if there are other restrictions I can't be bothered to look them up right now.

So I'd be OK with "idn-hostname" for symmetry, but I definitely don't want separate "[idn]-domain" ***and*** "[idn]-hostname" because I don't think they would be used properly.  Whatever "properly" would even mean.

handrews:
@epoberezkin I'm leaning towards changing "idn" to "idn-hostname".

Given the tendency of people to use "hostname" and "domain name" interchangeably, should we clarify whether we mean strictly hostnames (which imposes some restrictions on what characters are legal in the first component) or whether both hostnames and domain names are considered valid with these formats.

Note that it is apparently not unusual for host names to be configured that violate the hostname restrictions but would be legal domain names.

epoberezkin:
Looks good.

Julian:
@handrews this all looks both sensical and easily implementable to me. +1.

Issue #409, opened 2017-09-15
Standardize media type field names


handrews:
@philsturgeon observed that the `submissionEncType` is really a media type, unlike `contentEncoding` (formerly`binaryEncoding` within `media`)  which is actually an encoding type.

Let's change it to `submissionMediaType` for consistency with `contentMediaType`, and while we're at it change `mediaType` to `targetMediaType` so that it aligns with `targetSchema` and `targetHints`.

`submissionEncType` is left over from HTML's `encType`, which is completely irrelevant here as a naming convention.  The rationale for `encType` as a name in HTML is itself rather convoluted.  In any event, `submissionEncType` was only introduced in draft-06 so changing it is low-impact.

At different points in earlier drafts, `mediaType` has been used for both what is now `contentMediaType` and what is now `mediaType`, so let's stop using the unqualified name and make the existing one `targetMediaType` so that it is unambiguous.  This is a higher-impact change but a good one for consistency.  And (to the extent that Hyper-Schema has been adopted at all beyond API doc generation), `mediaType` is not one of the most widely used fields anyway.

philsturgeon:
That sounds good to me! So long as things are consistent. With a PHP background I've become really careful about this sort of thing.

handrews:
> With a PHP background I've become really careful about this sort of thing.

LOLOLOLOLOLOL

handrews:
Merged #411.

PR #411, opened 2017-09-15
Name media type keywords consistently.

handrews:
This addresses issue #409.  Which I just filed, so if the whole thing
gets rejected, no worries.  I can also remove the `targetMediaType`
part if that is more controversial.

"submissionEncType" is actually a media type, not an encoding,
so name it accordingly.  Rename "mediaType" to "targetMediaType"
to make it unambiguous and fit with "target" and "targetHints".

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -435,7 +435,7 @@
                 <section title="Submitting data for processing">
                     <t>
                         The <xref target="submissionSchema">"submissionSchema"</xref> and
-                        <xref target="submissionEncType">"submissionEncType"</xref> keywords
+                        <xref target="submissionMediaType">"submissionMediaType"</xref> keywords
                         describe the domain of the processing function implemented by the target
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -425,7 +425,7 @@
                 <section title="Submitting data for processing">
                     <t>
                         The <xref target="submissionSchema">"submissionSchema"</xref> and
-                        <xref target="submissionEncType">"submissionEncType"</xref> keywords
+                        <xref target="submissionMediaType">"submissionMediaType"</xref> keywords
                         describe the domain of the processing function implemented by the target
                         resource. Otherwise, as noted above, the submission schema and encoding are
```
dlax: The line above still talks about "encoding". Maybe change to media type. (This is the only place in the document where "encoding" is used, AFAICT.)

handrews: Good catch.  I'd mostly done search-and-replace with a bit of skimming, it did not occur to me to search for "encoding" for some reason.

Issue #222, opened 2017-01-04
"profile" media type parameter usage


handrews:
In mnot/I-D#175 @dret states:

> profile was not intended to refer to a schema. it is intended to refer
to a profile, which is defined as extending and/or constraining the
original media type. the spec gives examples.

yet in issue #26 filed here by @dret it looked like we had received his approval on how we are wording our specification around the use of "profile"

@dret can you help us out here?  Is what we have acceptable or not?  To me it seems like JSON Schema is doing exactly this- further constraining the `application/json` media type by declaring that only a subset of possible documents are valid rather than all possible expressions of the JSON syntax.  But it sounds like you do not view it this way.

Also in the same issue in mnot/I-D:

> a profile media type parameter has to be allowed by the media type.
very few media types do this. so while i think that it's possible (and
maybe even better) to use a media type profile parameter, if that
mechanism is required, in many cases the media type simply will not
allow such a parameter to be used.

Since `application/json` does not explicitly specify `profile` as a media type parameter, does that mean that even if "profile" is correct we cannot use it with `application/json`?  Do we need to specify an `application/instance+json` media type in order to add a parameter (whether it is "profile" or not?)  Honestly my entire API evolution strategy relies on having schema URIs in a media type parameter so it's a huge problem for me if we can't do this somehow.

The balkanization of IETF standards is immensely frustrating.  "Look, here are these two great ideas that do not conflict in any way and perfectly complement each other, but you cannot use them together because the two RFC authors did not coordinate and/or wrote their RFCs at different times!"

Finally:

> in my mind, "profile", "type", and "describedBy" are distinct, but i
have heard world views where they are more similar and do have quite
some overlap.

@dret if aspects of this distinction are relevant to JSON Schema and not covered in #26 could you please elaborate?

awwright:
It's my understanding we can update the JSON media type registration to add the 'profile' parameter -- with the same semantics as expected, nothing JSON Schema specific.

Let's keep investigating this.

handrews:
@awwright oh cool, glad you've thought of this.  Yeah, let's keep this issue open to track that dependency.

Would it make sense to create an "RFC" milestone to group issues that need to be handled before RFC but are not directly part of the draft iteration process?

dret:
On 2017-01-15 12:25, Austin Wright wrote:
> It's my understanding we can update the JSON media type registration to
> add the 'profile' parameter -- with the same semantics as expected,
[...]

PR #414, opened 2017-09-19
Minimal change from "profile" to "schema"

handrews:
This change removes "profile", since as noted in issue #222 the
usage is not correct.  It replaces it with "schema", with the
minimal possible change.

We may wish to also remove the "describedby" link relation and
instead mandate using "schema", or a similarly named
extension relation, and allow downloading schemas based on the
parameter.  That is outside of the scope of this change.

This also continues to address #351 `application/schema-instance+json`

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -5,7 +5,6 @@
 <!ENTITY RFC5988 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5988.xml">
 <!ENTITY RFC6839 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6839.xml">
 <!ENTITY RFC6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
-<!ENTITY RFC6906 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6906.xml">
 <!ENTITY RFC7049 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7049.xml">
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -724,34 +725,39 @@ Link: <http://example.com/my-hyper-schema#>; rel="describedby"
...
 ]]>
                     </artwork>
                 </figure>
 
                 <t>
-                    HTTP can also send the "profile" in a Link, though this may impact media-type
+                    <cref>
+                        This paragraph assumes that we can register a "schema" link relation.
+                        Should we instead specify something like "tag:json-schema.org,2017:schema"
+                        for now?
```
dlax: Or a URL like `http://json-schema.org/<something here>` ? Like JSON-LD does with `http://www.w3.org/ns/json-ld#context`.

handrews: I really like using non-dereferenceable URI schemes like `tag:` when there is nothing to dereference.  While it is certainly valid to use `http://` or `https://` for such a purpose, it is counter-intuitive and leads to a.) people blindly dumping it in a browser and being annoyed that it doesn't work and b.) awkward paragraphs in specifications about how even though the URI looks dereferenceable, implementations MUST NOT dereference it.

Just make it non-dereferenceable.  It solves all the UX problems of apparently-dereferenceable-URIs-that-are-not-actually-dereferenceable.

handrews:
This hits 2 weeks tomorrow, but I will be leaving it open until @awwright has had a chance to spend time on this issue and/or offer an alternate proposal (either is very welcome!).

Comments from others still also welcome!

handrews:
Final review deadline is Friday, Oct. 20th.  Barring actionable objections, this will be merged then to kick off the final review period.

PR #417, opened 2017-09-23
Add a relative-json-pointer format

handrews:
We have generally included formats necessary for describing
JSON Schema vocabularies, and JSON Hyper-Schema now uses
Relative JSON Pointers.

Also use it in the links meta-schema (as a separate commit).

I did not bother to file an issue for this as it seems pretty straightforward
and in line with existing precedent.  Also, it was faster to write a PR
so even if rejected this was less work :-)

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -12,6 +12,7 @@
 <!ENTITY RFC6570 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6570.xml">
 <!ENTITY RFC6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
 <!ENTITY RFC7159 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7159.xml">
+<!ENTITY I-D.luff-relative-json-pointer SYSTEM "https://xml2rfc.tools.ietf.org/public/rfc/bibxml3/reference.I-D.draft-luff-relative-json-pointer-00.xml">
 ]>
 <?rfc toc="yes"?>
 <?rfc symrefs="yes"?>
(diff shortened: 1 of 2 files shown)


awwright:
We might have to pull in the text of "draft-luff-relative-json-pointer", or figure out how to advance that draft.

And in my opinion, I'm still not really sure if relative pointers are actually ever necessary.

handrews:
> And in my opinion, I'm still not really sure if relative pointers are actually ever necessary.

How would you handle #385 and #386 without them?  And without making up some other arbitrary new format, because there is no reason to introduce an entirely different syntax when there is one that is compatible with the syntax (JSON Pointer) that we already use extensively.

handrews:
@awwright I emailed Geraint and he's fine with me/us re-submitting the I-D or pulling it in.  His thoughts:

>  I had originally separated it out into its own document (as opposed to adding it to the JSON Schema draft) because I thought it worked well as a standalone spec - I was already using the syntax in some of my own unrelated code because I felt it was a natural extension of JSON Pointer.  It was also a clearer and stronger way to explain all the details than a forum post or GitHub comment.
>
> If JSON Schema references it then it's a dependency for standardisation, but since the Relative JSON Pointer spec is much smaller I imagine it could be reviewed and finalised faster, and leave the JSON Schema spec slightly simpler.  So, I don't know which will get you to your goal faster - from a practical perspective I don't mind, because if it ends up folded into the JSON Schema spec it can still be cited by referencing a particular subsection.

I'm inclined to just rebuild the XML (Geraint could not find the sources) and re-submit it as a separate I-D myself leaving Geraint as the author and putting myself as editor on the grounds of doing the mechanics to get it going again (he and I discussed this in email).  It hardly matters at this stage, and that's just a lot easier than figuring out the best way to merge it into JSON Schema.

handrews:
@awwright and I chatted off-github and (I think) agreed that `anchorPointer` and `hrefPointers`/`templatePointers` are sufficient justification for this, as neither of us have an alternative proposal for implementing them.  @awwright would you be OK with signing off on this now?  If we end up finding a different solution and dropping Relative JSON Pointer from Hyper-Schema, I'd assume we could drop this as well.

handrews:
@awwright this is coming up on 2 weeks in a couple of days.  Given our discussion offlist I'm going to assume you've dropped your objections to this unless you comment otherwise.

PR #418, opened 2017-09-23
Clarify how base is used with templates.

handrews:
While trying to implement this, I realized it was a bit unclear.
RFC 6570 addresses this, but it is helpful to make a note of
it in the section on "base".

This PR does not change any requirements, it just makes
an existing requirement more explicit as otherwise you
have to dig through the other RFC a bit.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -50,7 +50,8 @@
         "propertyNames": { "$ref": "#" },
 
         "base": {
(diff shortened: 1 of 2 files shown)


handrews:
On second thought, I'm reworking the whole URI Templating algorithm description, and this will get handled naturally as part of that.  No sense in adding yet another PR to the count.

Issue #423, opened 2017-09-25
"Applicability", annotations, & hyper-schema


handrews:
While JSON Schema validation is fundamentally built on the idea that each pairing of a (sub-)schema and (sub-)instance can be evaluated in isolation, this idea breaks down for features built on the idea of associating information with the (sub-)instance as a result of validation outcomes.

In JSON Hyper-Schema, the `links` keyword attaches Link Description Objects (LDOs) to each (sub-)instance from every (sub-)schema against which it successfully validates in the process of validating the complete instance against the complete schema document.

------

For the purpose of this issue, a (sub-)schema is said to be ***applicable*** to a (sub-)instance if it validates in this way.  When processing a particular application of a (sub-)schema to a (sub-)instance, the particular (sub-)instance is said to be the point or location of ***attachment***, and may be identified by a JSON Pointer (other identifiers may be possible or even more appropriate, for this issue we will assume JSON Pointers).  Applicable (sub-)schemas may be identified either by their `$id` or a JSON Pointer within a given schema document.  

Calculating applicability and attachment can be thought of as a function taking an instance document and a schema document and producing the many-to-many map of schemas to their attachment points and instances to their applicable schemas.  Of course, it is possible to generate only specific portions of the mapping as needed rather than producing it all at once.

------

It seems clear to me that implementing the concept of applicability is a MUST requirement for a JSON Hyper-Schema implementation.  Fundamentally, you can only do anything in hyper-schema if you can find all of the links that apply to a given location.

It also seems clear that it is a requirement for properly implementing the annotation keywords as well (`title`, `description`, `examples`, `default`, `readOnly`, and whatever we do for deprecation).  To make use of any of these, you start from a (sub-)instance and need to know what value(s) of whichever keyword you are considering apply.

Some of the usage challenges of the annotation keywords come from this requirement.  The semantics of finding multiple applicable `links` in hyper-schema are clear: just put all of their LDOs in a set together, it doesn't matter which (sub-)schema provided a given LDO.

A similar approach should work for `examples`.

For `readOnly`, it seems logical that if it is `true` in any applicable schema, then it is `true` overall.  Deprecation would probably work something like this, although depending on how much information is involved in the concept it may be more complex (see #74 for extensive discussion).

`title` and `description` are less clear.  One could reasonably just use all descriptions, although the result might be somewhat incoherent to read.  Applying multiple titles is less reasonable; generally a title is supposed to be concise and to the point.

[...]

Issue #124, opened 2016-11-03
Hyper-schema: Alternative forms for listing link relationships


awwright:
For a variety of cases, it's often simpler to use:

```javascript
{
links: { "self": { href:"/document/{id}" } }
}
```

or:

```javascript
{
links: { "stylesheet": [ {href:"style.css"}, {href:"base.css"} ] }
}
```

HAL and other JSON-based hypermedia formats follow a similar convention.

handrews:
I'm definitely in favor of this.  I always have to translate the links into something like this inside code anyway because searching a list for the right rel value is really awkward.

slurmulon:
I only see one minor downside to this, and it's consistency in a different sense - when `links` can only be an `Array`, you don't have to check for special cases (like is it an `Object` instead). This should be generally considered in the user's Hyper-Schema library that parses LDOs and the user should not have to worry about it at all, but I honestly don't even know of a Hyper-Schema library that does this (for instance, I had to write my own LDO parser because I couldn't even find a Hyper-Schema library for looking up or parsing LDOs)

slurmulon:
My minor concern also applies to `rel` - in one case it's the LDO's key, in another case it will be defined in `rel`. This makes writing parsers more difficult and bug prone, but I think it's just a consideration that people should be aware of. Until is a more robust JSON Hyper-Schema consumer library that provides a simple API for finding and resolving LDOs is available, the user will always have to consider these complications themselves.

handrews:
> My minor concern also applies to rel - in one case it's the LDO's key, in another case it will be defined in rel. This makes writing parsers more difficult and bug prone

No, rel is always the key.  It's never in the LDO as written out.  The only cases to handle are whether there is one LDO attached to rel (the value is an object) or several (the value is an array of LDO objects).  This can be normalized as the hyper-schema is read in.

handrews:
Having given this some thought and looked at other media types more, I think I am reluctantly against this.

* RFC 5988 allows multiple space-separated "rel"s per link.  While we could use keys with space-separated URI references, that's much more awkward than allowing a JSON array for a "rel" keyword in the LDO
* URI templating for "rel" as requested in #56 is a really useful feature that would be much easier to manage if "rel" is kept separate (this allows us to potentially use a "relVars" in the manner of "hrefVars" and "baseVars" from #142 if it is accepted)

[...]

PR #424, opened 2017-09-25
Define "applicability" and "attachment"

handrews:
This is a bit more groundwork for the rewrite, and is relevant
to issue #423.

***NOTE:** This has been updated from its original approach to move the concept into the validation specification!*

The commits have been broken up with individual messages into useful chunks:

1. [Just move the metadata section](https://github.com/json-schema-org/json-schema-spec/pull/424/commits/dd0b873a0227e5a7a5bafbe35ae3302fa5c6ba8a)
1. [Split out and re-word "definitions"](https://github.com/json-schema-org/json-schema-spec/pull/424/commits/4f9e4e3ccdd110e054899592ff88f40de56c77b9)
1. [Generalize and move applicability from hyper-schema to validation/annotation](https://github.com/json-schema-org/json-schema-spec/pull/424/commits/c55678a7e8845ba7eb9ee334d56bc1dbaeb7c376)
1. [Group annotation keywords & re-indent without changing](https://github.com/json-schema-org/json-schema-spec/pull/424/commits/c2eccf96355eb7797f628ddbd76fcddf5c2a9ef5)
1. [Update annotation keywords with multi-value behavior](https://github.com/json-schema-org/json-schema-spec/pull/424/commits/d9c913a98a6b8589c4f2bb1e1868261c39ef3be2)
 
-----

Rework the whole applicability/attachement concept from the 
hyper-schema specification into the validations specification,
and establish it as the basis for working with annotation keywords.

Update the annotation kewyords with clarifications on how to
handle values across multiple applicable schemas.

Update the hyper-schema specification to reference the now more
generic description of validation interaction in the validation
spec itself, with additional clarifications about how to handle
combinign keywords from multiple applicable schemas.

Split the "Metadata" section into a re-use section for just the 
"definitions" keyword, and a section on annotation and extensibility
which defines applicability/attachement and defines the simple
annotation vocabulary that was already present in the validation
spec.

Move all of that after "format" and "String-encoding non-JSON data"
as those sections are both more closely aligned with validation
than with either re-use or annotations.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -175,32 +175,6 @@
                     relations for the instance, with URIs incorporating values from the instance.
                 </postamble>
             </figure>
-
-            <section title="Interaction with validation">
-                <t>
-                    Hyper-schema keywords can be applied when the instance is valid against
-                    the schema that includes those keywords.
-                </t>
-                <t>
-                    Hyper-schemas MUST NOT be applied to an instance if the instance fails to
-                    validate against the validation keywords within or containing the hyper-schema.
-                    Hyper-schema keywords in branches of an "anyOf", "oneOf", or "if"/"then"/"else"
-                    that do not validate, or in a "dependencies" subschema that is not relevant
-                    to the instance, MUST be ignored.
-                </t>
-                <t>
-                    Hyper-schema keywords in a subschema contained within a "not", at any depth,
-                    including any number of intervening additional "not" subschemas, MUST be
-                    ignored.
-                </t>
-                <t>
-                    If the subschema for a "contains" keyword contains hyper-schema keywords they
-                    MUST be applied to all array elements that validate against the schema.  While
-                    finding a single validating element is sufficient to determine the validation
-                    outcome, when hyper-schema keywords are present, the subschema MUST be evaluated
-                    against all array elements.
-                </t>
-            </section>
         </section>
 
         <section title="Meta-schema">
@@ -211,6 +185,17 @@
         </section>
 
         <section title="Schema keywords">
+            <t>
+                Hyper-schema keywords can be applied when the instance is valid against
+                the schema that includes those keywords, as outlined in
+                <xref target="json-schema-validation">Section 10.1 of JSON Schema validation</xref>.
+            </t>
+            <t>
+                When multiple subschemas are applicable to a given sub-instance, all "link"
+                arrays MUST be concatenated, in any order, into a single array.  Each object
+                in the resulting array MUST retain its own list of applicable "base" values,
+                in resolution order, from the same schema and any parent schemas.
+            </t>
             <section title="base">
                 <t>
                     If present, this keyword is resolved against the current URI base that the
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -1160,7 +1160,7 @@
                     keywords.  Implementation of this feature MAY instead be done separately.
                 </t>
                 <t>
-                    Since many subschemas can be applicable to any single sub-isntance, each
+                    Since many subschemas can be applicable to any single sub-instance, each
                     annotation keyword or vocabulary needs to specify how to handle multiple
                     occurrences with different values.  In the absence of keyword-specific
                     handling rules, an implementation MUST collect all values and make them
@@ -1217,7 +1217,9 @@
 
                 <section title='"default"'>
                     <t>
-                        There are no restrictions placed on the value of this keyword.
+                        There are no restrictions placed on the value of this keyword.  When
+                        multiple occurrences of this keyword are applicable to a single
+                        sub-instance, implementations SHOULD remove duplicates.
                     </t>
                     <t>
                         This keyword can be used to supply a default JSON value associated with a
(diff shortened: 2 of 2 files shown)


jsonschema-hyperschema.xml
```diff
@@ -210,6 +184,17 @@
...
+            <t>
+                Hyper-schema keywords can be applied when the instance is valid against
+                the schema that includes those keywords, as outlined in
+                <xref target="json-schema-validation">Section 10.1 of JSON Schema validation</xref>.
+            </t>
+            <t>
+                When multiple subschemas are applicable to a given sub-instance, all "link"
+                arrays MUST be concatenated, in any order, into a single array.  Each object
+                in the resulting array MUST retain its own list of applicable "base" values
+                from the same schema and any parent schemas.
```
dlax: This paragraph makes me think back about #124. So if links was an object, what does concatenation would mean?
Might be an argument in favor of keeping the array...

jsonschema-validation.xml
```diff
@@ -1098,6 +1098,66 @@
...
+                    the schema that includes the annotation keywords.  For any given location
+                    in the instance (referred to as a sub-instance), each (sub)schema against
+                    which it successfully validates is considered to be <spanx>applicable</spanx>
+                    to that sub-instance.  The (sub)schema is said to be <spanx>attached</spanx>
+                    to each sub-instance to which it applies.
+                </t>
+                <t>
+                    A validation implementation MAY choose to implement determining subschema
+                    applicability and providing access to the value(s) of applicable annotation
+                    kewords.  Implementation of this feature MAY instead be done separately.
```
dlax: typo: kewords

jsonschema-validation.xml
```diff
@@ -1098,6 +1098,66 @@
...
+                </t>
+                <t>
+                    Since many subschemas can be applicable to any single sub-isntance, each
```
dlax: typo: isntance

jsonschema-validation.xml
```diff
@@ -1098,6 +1098,66 @@
...
+                    annotation keyword or vocabulary needs to specify how to handle multiple
+                    occurrences with different values.  In the absence of keyword-specific
+                    handling rules, an implementation MUST collect all values and make them
+                    available as an array in which order is not significant and items need not
+                    be unique.
+                </t>
+                <section title="Combinatoric and conditoinal schemas">
```
dlax: typo: conditoinal

dlax:
@handrews 
> The commits have been broken up with individual messages into useful chunks

Thanks for this, it really makes reviewing pleasant as far as I am concerned.

handrews:
> > The commits have been broken up with individual messages into useful chunks

> Thanks for this, it really makes reviewing pleasant as far as I am concerned.

You're welcome!  I do try to make it easy when I see how it can be split up (I just tend to see that criteria a bit differently sometimes, it seems).

BTW, I finally figured out why I keep having so much trouble with spelling typos:  the vim spell-checker ignores text within elements in XML.  It will check text in attributes, but not element contents.  I DO NOT UNDERSTAND WHY.  But at least I finally know why they keep getting through.  (although apparently I forgot to turn it on at all this time as it would have caught the "conditional" at least).

awwright:
I'm still looking at this.
I don't see any obvious problems, but maybe some cases where language could be clarified. Maybe I just need to read the paragraphs in context.

handrews:
@awwright language suggestions welcome.  I am also working on a test suite format for this to add to the test suite repo, so I think that will help clarify as well.

jsonschema-hyperschema.xml
handrews: If `links` becomes an object, then it's an object of arrays, so for each relation that appears in multiple subschemas, you would concatenate the arrays for that relation to produce the array in the resulting combined value.

handrews:
@dlax I think I fixed all the typos.  I merged them into the existing commits because then they are less distracting for other people reading for the first time.

jsonschema-hyperschema.xml
dlax: True. I realized about that after writing the comment.

jsonschema-validation.xml
dlax: @handrews That one is still there.

handrews: dammit... I distinctly remember searching for this, but I must have done it in hyperschema and forgotten about validation.  Fixed now.  For reals.  I think.

Relequestual:
Just to make sure I understand the applicability and attachment element...

It's mainly to do with https://github.com/json-schema-org/json-schema-spec/pull/424/commits/c55678a7e8845ba7eb9ee334d56bc1dbaeb7c376#diff-2391a2ef934e9aed11cbf87a65eb61c9R1172

In that, annotation key words are only attached to a sub-instance if the sub-schema validates successfully for the sub-instance?

handrews:
@Relequestual yup, that's correct

handrews:
@awwright @Relequestual please let me know if your concerns have been addressed, and if not, what actionable objections you have.

This is a fundamental piece of defining how JSON Schema works as a multi-vocabulary media type, and in particular how validation serves as a base for applying other vocabularies.

handrews:
> I'm still looking at this.
I don't see any obvious problems, but maybe some cases where language could be clarified. Maybe I just need to read the paragraphs in context.

@awwright has this cleared up with more reading?  Or do you have actionable feedback such as indicating the unclear language?  It's been two weeks since this comment and I'm not sure how to move forward.

handrews:
Given that this has two approvals, I am going to go ahead and merge.  @awwright I am happy to accept further issues on this to work on it more, but it's been nearly 3 weeks since your "I'm thinking" comment and neither mentioning you here nor asking offline has produced anything actionable.  We have plenty of time before anything will be published so this is not irreversible.

PR #425, opened 2017-09-25
URI Template resolution as pseudocode

handrews:
_NOTE:  This is a chunk of work that I have done as part of the rewrite.  This probably isn't quite the right way to specify it, but seems like the best way to get some feedback on the approach and changes before putting a lot of work into new prose.  Please assume that perfect examples are present wherever you want them to be :-)  I have not yet consolidated the examples into a more focused set so they're not included in this PR._

_This is one of the two most complex things an implementation must do in order to implement hyper-schema (the other thing being finding all applicable schemas per #424)._

------
This consolidates all of the URI Template resolution sections into
one section.  The new section is at the top level, as it applies
to both LDO keywords ("href", "anchor", and all of the algorithm
modifiers) and to a schema keyword ("base").

As much of the algorithm as possible is now given in pseudocode,
although whether this is the most readable form is debatable.
It is certainly more concise, and forced detailed consideration
of some steps which had been under-specified, particularly with
respect to "hrefSchema".

The main changes are that setting any applicable subschema for
a template variable to "false" in "hrefSchema" excludes that
variable from accepting input, and that the input data for
"hrefSchema" is prepopulated only with instance data that
validates against all applicable subschemas within "hrefSchema".
Expressing this properly requires the "applicability" concept
from PR #424.

Previously, the only mention of how to exclude a field from being
used to pre-populate input was the advice to set the "properties"
subschema to "false".  However, this does not hold up once you
start using more complex schemas.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1441,13 +1441,15 @@ GET /foo/
                 <preamble>
                     This is the high-level algorithm as pseudocode.  "T" comes from either
                     "href" or "anchor" within the LDO, or from "base" in a containing schema.
-                    Pseudocode for each step follows.
+                    Pseudocode for each step follows.  "anchorOrHref" indicates which of the
+                    two started the process (since "base" is always resolved in order to finish
+                    resolving one or the other of those keywords).
                 </preamble>
                 <artwork>
 <![CDATA[
 templateData = populateDataFromInstance(T, ldo, instance)
 
-if resolving "href" or "base" for "href" and ldo.hrefSchema exists:
+if anchorOrHref == "href" and ldo.hrefSchema exists:
     inputData = acceptInput(ldo, instance, templateData)
     for varname in inputData:
         templateData[varname] = inputData[varname]
@@ -1464,13 +1466,20 @@ R = rfc6570ResolutionAlgorithm(T, templateData)
 
             <section title="Populating template data from the instance">
                 <t>
-                    This step looks at various data sources in the instance:
+                    This step looks at various locations in the instance for variable values.
+                    For each variable:
                     <list>
-                        <t>Use "templatePointers" to find values where appropriate</t>
-                        <t>Look in the immediate instance for remaining values</t>
                         <t>
-                            Either way, if there is a value there, put it in the
-                            template resolution data set
+                            Use "templatePointers" to find a value if the variable
+                            appears in that keyword's value
+                        </t>
+                        <t>
+                            Otherwise, look for a property names matching the variable in
+                            the instance location to which the link is attached
+                        </t>
+                        <t>
+                            In either case, if there is a value at the location, put it in
+                            the template resolution data set
                         </t>
                     </list>
                 </t>
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1428,10 +1428,19 @@ GET /foo/
                 as it is not always possible to tell whether a template will resolve
                 to a full URI or a URI-reference.
             </t>
+            <t>
+                In English, the high-level algorithm is:
+                <list>
+                    <t>Populate template variable data from the instance</t>
+                    <t>If input is desired, accept input</t>
+                    <t>Check that all required variables have a value</t>
+                    <t>Encode values into strings and fill out the template</t>
+                </list>
+            </t>
             <figure>
                 <preamble>
-                    This is the high-level algorithm.  "T" comes from either "href"
-                    or "anchor" within the LDO, or from "base" in a containing schema.
+                    This is the high-level algorithm as pseudocode.  "T" comes from either
+                    "href" or "anchor" within the LDO, or from "base" in a containing schema.
                     Pseudocode for each step follows.
                 </preamble>
                 <artwork>
@@ -1454,6 +1463,17 @@ R = rfc6570ResolutionAlgorithm(T, templateData)
             </figure>
 
             <section title="Populating template data from the instance">
+                <t>
+                    This step looks at various data sources in the instance:
+                    <list>
+                        <t>Use "templatePointers" to find values where appropriate</t>
+                        <t>Look in the immediate instance for remaining values</t>
+                        <t>
+                            Either way, if there is a value there, put it in the
+                            template resolution data set
+                        </t>
+                    </list>
+                </t>
                 <figure>
                     <artwork>
 <![CDATA[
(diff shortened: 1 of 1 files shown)


handrews:
An alternative approach here would be to say that an implementation MUST provide the from-instance template data set to callers alongside of `hrefSchema`, and let the caller decide whether and how to pre-populate the input form.

If we went that route, I would still like to include this pre-population approach as an option that an implementation SHOULD support directly as a convenience, as I think it will be the most common and interoperable approach.  But I could see requiring a more flexible option alongside of it.

dlax:
While I've been reading the algorithms a couple of times, I'm still having hard time following... So it's hard for me to figure out the scope of the changes of this PR and make an opinion.

One issue with the "pseudocode" form is that it's part of a normative section of the spec (as opposed to in an appendix for example), so it can hardly be skipped. I have the impression that such extensive use of pseudo-code in a normative section is not an usual practice. I can't really tell how this would look like in another form: certainly be more verbose, but maybe easier to follow?

For instance, would something like the [Implementation Hints](https://tools.ietf.org/html/rfc6570#appendix-A) in RFC6570 work here?

handrews:
@dlax it's mostly that I needed to write the pseudocode in order to work out the steps that needed to be done for it to actually work.

I'm not at all attached to it as the form of explanation.  I just posted this for feedback on the direction.

handrews:
@dlax I added some English descriptions of the various steps, see if that makes it easier to comprehend.  I added the same commit to the main rewrite PR as well.

dlax:
As reading the rewrite PR, I haven't seen any obvious problem with this approach of presenting the URI template resolution consolidated into a dedicated section.

jsonschema-hyperschema.xml
```diff
@@ -1539,6 +1389,212 @@ GET /foo/
...
+                <preamble>
+                    This is the high-level algorithm as pseudocode.  "T" comes from either
+                    "href" or "anchor" within the LDO, or from "base" in a containing schema.
+                    Pseudocode for each step follows.
+                </preamble>
+                <artwork>
+<![CDATA[
+templateData = populateDataFromInstance(T, ldo, instance)
+
+if resolving "href" or "base" for "href" and ldo.hrefSchema exists:
```
dlax: I don't understand this line. Perhaps `resolving` needs to be explain or expressed as a function? Then it's not clear how things come together, perhaps adding some parentheses would help.

jsonschema-hyperschema.xml
```diff
@@ -1539,6 +1389,212 @@ GET /foo/
...
+                    This step looks at various data sources in the instance:
+                    <list>
+                        <t>Use "templatePointers" to find values where appropriate</t>
+                        <t>Look in the immediate instance for remaining values</t>
+                        <t>
+                            Either way, if there is a value there, put it in the
+                            template resolution data set
+                        </t>
+                    </list>
+                </t>
```
dlax: This paragraph (the whole `<t>`) does not look correct to me, or I'm not understanding it. It seems to me that it does not match with the pseudocode below. Specifically, I don't get the "Look in the immediate instance for remaining values".
Also I wouldn't say that "various data sources in the instance" are used (or, again, I may be missing what you mean by "various data sources"); to me the instance is just one source.

handrews:
@dlax I've reworked some wording based on your feedback, please take a look and see if that's more clear.

dlax:
@handrews It's more clear, thanks.

jsonschema-hyperschema.xml
```diff
@@ -1539,6 +1389,221 @@ GET /foo/
...
+                    <t>If input is desired, accept input</t>
+                    <t>Check that all required variables have a value</t>
+                    <t>Encode values into strings and fill out the template</t>
+                </list>
+            </t>
+            <figure>
+                <preamble>
+                    This is the high-level algorithm as pseudocode.  "T" comes from either
+                    "href" or "anchor" within the LDO, or from "base" in a containing schema.
+                    Pseudocode for each step follows.  "anchorOrHref" indicates which of the
```
dlax: I'd suggest something like `templateKind` instead of `anchorOrHref`.

jsonschema-hyperschema.xml
```diff
@@ -1539,6 +1389,221 @@ GET /foo/
...
+                "base", "anchor", and "href".  Each are resolved separately to URI-references,
+                and then the anchor or href URI-reference is resolved against the base (which
+                is itself resolved against earlier bases as needed, each of which was first
+                resolved from a URI Template to a URI-reference).
+            </t>
+            <t>
+                All three keywords share the same algorithm for resolving variables from
+                instance data, which makes use of the "templatePointers" and "templateRequired"
+                keywords.  When resolving "href", both it and any "base" templates
+                needed for resolution to an absolute URI, the algorithm is modfied to
```
dlax: typo: modfied

jsonschema-hyperschema.xml
```diff
@@ -1539,6 +1389,221 @@ GET /foo/
...
+                <t>
+                    This step looks at various locations in the instance for variable values.
+                    For each variable:
+                    <list>
+                        <t>
+                            Use "templatePointers" to find a value if the variable
+                            appears in that keyword's value
+                        </t>
+                        <t>
+                            Otherwise, look for a property names matching the variable in
```
dlax: typo: names -> name

jsonschema-hyperschema.xml
```diff
@@ -1539,6 +1389,221 @@ GET /foo/
...
+                            Put the input in the template resolution data set, overriding
+                            any existing values
+                        </t>
+                    </list>
+                </t>
+                <figure>
+                    <preamble>
+                        "InputForm" represents whatevers sort of input mechanism is appropriate.
+                        This may be a literal web form, or may be a more programmatic construct
+                        such as a callback funciton accepting specific fields and data types,
```
dlax: typo: funciton

jsonschema-hyperschema.xml
```diff
@@ -1539,6 +1389,221 @@ GET /foo/
...
+
+    value = instance.valueAt(valuePointer)
+    if value is defined:
+        templateData[varname] = value
+]]>
+                    </artwork>
+                </figure>
+            </section>
+
+            <section title="Accepting input for template data">
```
dlax: Might be worth mentioning that this step corresponds to the `acceptInput` call in this code block above.

handrews: Really?  I have to say that "Accepting input" == "acceptInput"?  Even when it's in the right order and it's the same words already?  Let me think on if there is a better way to present this if it's that confusing as is.

dlax: It's certainly not *confusing* now that you mention it. Perhaps I just read it without enough concentration... Let's leave this comment out for the moment.

Issue #431, opened 2017-10-02
Why is `$schema` restricted to root schemas?


handrews:
In PR #248 we forbade the use of `$schema` in subschemas.  I can't remember why I approved this.

One use case we've noted over and over is that of "packing" multiple schema files into one.  It's the main justification for the base-uri-changing functionality of `"$id"`.  What that means is that the only time something being a root schema matters is if it is the root schema of the entry point file for processing.

Once you are using multiple files, then whether you `$ref` to another file (with no fragment or "#" as the fragment), or whether you pre-process and "pack" that file into the original file, the result is the same.  But in the former, the referenced schema is a root schema.  In the latter case, it is a subschema.

That means that using `$ref`, you can reference a draft-04 schema from a draft-06 schema.  But if you pack it, suddenly that is illegal, because you can't use `$schema` in the packed subschema to switch the processing rules.

This seems very wrong.

@awwright you wrote the PR- do you remember why it seemed correct?  What am I missing?

I know @epoberezkin had some concerns about implementation, but I don't recall why that was compelling- the first thing you do on processing a schema is check `$schema` to set the rules for processing the rest of the schema (this is how my embryonic implementation handled things before I decided it was probably best to leave that to other libraries).  Perhaps @Julian has thoughts on this as well?

handrews:
I suppose we could redefine "root" schema as "a schema that sets a new base URI" or something, but that seems confusing.

epoberezkin:
@handrews we had a very long conversation on the issue, both about $ref and inclusion is not the same and that meta-schema is actually a JSON-schema that cannot be changed during validation.

I can point you to the relevant comments somewhat later but let's please keep it as it is.

Referencing draft-04 schema from draft schema is fine, because $ref is not schema inclusion. Having $schema in the middle of the schema is not fine, as there is no validation process defined that allows to change schema (in this case, meta-schema) on the fly.

epoberezkin:
Root schema is the top level of a separate JSON instance, not any inner schema that changes base URI. So we can't redefine what root schema means.

epoberezkin:
@handrewd I am happy to have this discussion again, as long as it stays the same for draft-07, it's not seen as either critical or bug, and you re-read our previous conversation on the subject.

Let's get draft-07 out as is and then we can talk again about it.

epoberezkin:
With the current spec, you can pack multiple schemas into a single file that is a collection of schemas but not a JSON schema, in the general case.

[...]

Issue #314, opened 2017-04-30
Understanding extended meta-schemas


handrews:
Custom meta-schemas are problematic for interoperability, specifically when they extend standardized meta-schemas.  This is because `"$schema"` is doing two things:

1.  Indicating how the schema itself should be validated
2.  Indicating the purpose and semantics of the schema (e.g. plain validation, hypermedia, or something else)

So if I have custom keywords to add to the regular validation or hyper-schema meta-schemas, which do not change their behavior in incompatible ways:

* I want to define a custom meta-schema to validate my custom extension keywords and allow my own tools to recognize that they are being used.
* I want other tools to be able to continue to recognize the underlying validation or hyper-schema vocabulary and ignore my extensions.

Note that none of this is a concern if my custom meta-schema indicates a completely different vocabulary, or adds modifications that break the expectations of the standard meta-schema(s).

We currently hide this problem- all implementations that are aware of hyper-schema at all just hard-wire that it is an extension of validation.  But this kind of hard-wiring does not scale and is not generally interoperable.

This is somewhat related to #86 (extending meta-schemas), since the difficulty of extending recursion comes into play when considering solutions here.  For instance, one solution to this issue would be to allow `"$schema"` to take an array indicating all supported vocabularies.  Implementations could then ignore unrecognized vocabularies.  The spec would state that schema authors MUST NOT declare contradictory semantics, or else the behavior is undefined.

Ideally, you could do this without also having to mess with the recursion problems with something like `"$root"` ( https://github.com/json-schema-org/json-schema-spec/issues/86#issuecomment-263423192 ) although as noted in that comment, I'm not entirely sure I even like my own idea there.

awwright:
Probably related: #151

handrews:
Another approach would be to define some sort of "extends" or "isCompatibleWIth" link relation type.  This would **not** impact the processing of the actual schema- the schema would still need to use `"allOf"` or some other mechanism to implement the extension.

But the link relation could be followed so that a user agent knows that if it does not understand the meta-schema it was given, but does understand a compatible/extension base meta-schema, that it can still make use of the schema to the extent described by the base meta-schema.

handrews:
This issue would be solved by PR #432.

handrews:
[...]

PR #432, opened 2017-10-04
Add $vocabularies, clarify $schema and $ref

handrews:
This addresses issues #314 and #431.

`"$schema"` is now explicitly intended for meta-schema declaration.
An explanation for why it MUST only be present in root schemas
has been added so we (I) don't forget about it and freak out (again).

The concept of vocabularies is now introduced explicitly, and
the `"$vocabularies"` keyword is introduced to declare vocabulary
support.  For compatibility (and simplicity in the case of
a single standard meta-schema conveying sufficient information),
omitting `"$vocabularies"` in the root schema causes it to behave
as if the `"$schema"` value is listed as a vocabulary.  This preserves
all existing behavior.

Finally, a paragraph is added to `"$ref"` clarifying the conceptual
model, specifically with respect to `"$schema"` and meta-schema
validation.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -1,53 +1,63 @@
 {
     "$schema": "http://json-schema.org/draft-06/hyper-schema#",
     "$id": "http://json-schema.org/draft-06/hyper-schema#",
+    "$vocabularies": [
+        "http://json-schema.org/draft-06/schema#",
+        "http://json-schema.org/draft-06/hyper-schema#"
+    ],
     "title": "JSON Hyper-Schema",
     "definitions": {
         "schemaArray": {
             "allOf": [
                 { "$ref": "http://json-schema.org/draft-06/schema#/definitions/schemaArray" },
                 {
-                    "items": { "$ref": "#" }
+                    "items": { "$ref": "#/definitions/subschema" }
                 }
+            ]
+        },
+        "subschema": {
+            "allOf": [
+                { "$ref": "#" },
+                { "properties": { "$schema": false } }
             ]
         }
     },
     "allOf": [ { "$ref": "http://json-schema.org/draft-06/schema#" } ],
     "properties": {
-        "additionalItems": { "$ref": "#" },
-        "additionalProperties": { "$ref": "#"},
+        "additionalItems": { "$ref": "#/definitions/subschema" },
+        "additionalProperties": { "$ref": "#/definitions/subschema"},
         "dependencies": {
             "additionalProperties": {
                 "anyOf": [
-                    { "$ref": "#" },
+                    { "$ref": "#/definitions/subschema" },
                     { "type": "array" }
                 ]
             }
         },
         "items": {
             "anyOf": [
-                { "$ref": "#" },
+                { "$ref": "#/definitions/subschema" },
                 { "$ref": "#/definitions/schemaArray" }
             ]
         },
         "definitions": {
-            "additionalProperties": { "$ref": "#" }
+            "additionalProperties": { "$ref": "#/definitions/subschema" }
         },
         "patternProperties": {
-            "additionalProperties": { "$ref": "#" }
+            "additionalProperties": { "$ref": "#/definitions/subschema" }
         },
         "properties": {
-            "additionalProperties": { "$ref": "#" }
+            "additionalProperties": { "$ref": "#/definitions/subschema" }
         },
-        "if": {"$ref": "#"},
-        "then": {"$ref": "#"},
-        "else": {"$ref": "#"},
+        "if": {"$ref": "#/definitions/subschema"},
+        "then": {"$ref": "#/definitions/subschema"},
+        "else": {"$ref": "#/definitions/subschema"},
         "allOf": { "$ref": "#/definitions/schemaArray" },
         "anyOf": { "$ref": "#/definitions/schemaArray" },
         "oneOf": { "$ref": "#/definitions/schemaArray" },
-        "not": { "$ref": "#" },
-        "contains": { "$ref": "#" },
-        "propertyNames": { "$ref": "#" },
+        "not": { "$ref": "#/definitions/subschema" },
+        "contains": { "$ref": "#/definitions/subschema" },
+        "propertyNames": { "$ref": "#/definitions/subschema" },
 
         "base": {
             "type": "string"
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -365,36 +365,113 @@
 
         </section>
 
-        <section title='The "$schema" keyword'>
-            <!-- TODO a custom $schema keyword might also be used to enforce minimum required functionality of a validator -->
+        <section title="Schema vocabularies">
             <t>
-                The "$schema" keyword is both used as a JSON Schema version identifier and the
-                location of a resource which is itself a JSON Schema, which describes any schema
-                written for this particular version.
+                JSON Schema is a media type, and as described by this specification offers
+                minimal functionality.  Schema authors will need to use at least one vocabulary,
+                such as 
+                <xref target="json-schema-validation">validation</xref> or
+                <xref target="json-hyper-schema">hyper-schema</xref>, to create meaningful
+                schema documents.
             </t>
             <t>
-                The value of this keyword MUST be a <xref target="RFC3986">URI</xref>
-                (containing a scheme) and this URI MUST be normalized.
-                The current schema MUST be valid against the meta-schema identified by this URI.
+                Schemas can themselves be described by a schema, referred to as a meta-schema.
+                This process, as with other instance documents, applies a single meta-schema
+                across the entire schema document.  A meta-schema is declared with "$schema".
             </t>
             <t>
-                The "$schema" keyword SHOULD be used in a root schema.
-                It MUST NOT appear in subschemas.
+                Schema vocabularies, however, can be mixed within a single document, at the
+                schema object level.  While a schema vocabulary is identified by a meta-schema,
+                vocabularies used within a schema document are not necessarily capable of
+                validating the entire document.  Vocabularies are declared with "$vocabularies".
             </t>
             <t>
-                <cref>
-                    While this pattern is likely to remain best practice for schema authoring,
-                    implementation behavior is subject to be revised or liberalized in future
-                    drafts.
-                </cref>
+                Either the "$schema" or "$vocabularies" keyword SHOULD be present in a non-empty root
+                schema, or else the interpretation of the schema is undefined.  Implementations MAY
+                make assumptions based on other input or on the schema's contents, but MUST NOT assume
+                identical behavior in peer implementations.
             </t>
             <t>
-                Values for this property are defined in other documents and by other parties.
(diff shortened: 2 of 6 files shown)


handrews:
> I believe $vocabulary should always be a single string, rather than array, so multiple vocabularies should not be allowed in the schemas. I also believe that $vocabulary should only be allowed in the root schema. I will elaborate in a separate comment.

The entire purpose of `$vocabularies` is to clearly support multiple vocabularies, for instance validation + hyper-schema + UI generation + some custom vocabulary.

The nature of vocabularies is that they build on each other, so in order for a general system to tell whether their is a usable vocabulary, they need to know each vocabulary in the stack.

You are essentially saying that you want `$vocabularies` to be exactly like `$schema`, which is pointless.  `$vocabularies` was specifically designed to not impose any file-wide requirements on schema processing.

handrews:
@epoberezkin I could see an alternative where meta-schemas declare which vocabularies (plural) they describe.  In this approach, implementations would examine the meta-schema to look for recognizable vocabularies, and instances would declare `$schema` exactly as in draft-06.

I would be just as happy with that approach.

In both approaches, a conforming implementation that only wants to support validation with no extensions can almost entirely avoid `"$vocabularies"` as the meta-schema is sufficiently well-known.

epoberezkin:
@handrews, I definitely appreciate the progress, there quite a few things in this PR we agree on. I will try to summarise what we agree on, please correct me if I am wrong:

- $schema has a purpose to define meta-schema that can be used to validate the schema. Historically it's used to define the vocabulary as well, but given that the meta-schema can be extended (even though the mechanism is very verbose due to unsolved questions with adding properties in recursive cases), it is in general unfit for the purpose of defining vocabulary.
- $schema must only appear in the root schema, as it defines the JSON schema that should be used to validate the current schema as JSON instance
- $ref should not be seen as inclusion as the included fragment can belong to another schema file (=JSON instance) and therefore it can require a different meta-schema to validate.
- $vocabulary keyword should be added to define the vocabulary of the schema.

Things we do not agree on (again, please correct me I am wrong):

@handrews:
- schema can define multiple vocabularies ($vocabulary is array of strings)
- $vocabulary can be used in subschemas
- $vocabulary is URI (of what? spec or meta-schema? if the latter will it be the same as $schema in most cases?)

@epoberezkin:
- schema needs to support only one $vocabulary (it is a string); but a vocabulary can be extending another vocabulary (in a way hyper-schema extends validation vocabulary)
- $vocabulary can be used only in root schema
- $vocabulary is a descriptive identifier defined in the respective I-D (e.g "validation", "hyper-schema", "ui-schema", etc.)

Arguments to allow separate vocabularies in subschemas and allow multiple vocabularies
- you can package multiple schemas in the same file conveniently
- you can have schemas that both generate code and UI (e.g.)
- anything else?

Arguments to NOT allow vocabularies in subschemas:
- validation concern. Each vocabulary assumes a specific meta-schema for validating the schema, whether the standard or extended meta-schema for this vocabulary is used. If the subschema defines a different vocabulary from the root schema, then it is either not validated (in case $schema allows additional properties) or fails validation (in case $schema prohibits additional properties). @handrews, do you have a solution to this problem?
- processing concern. In practice, different vocabularies are implemented by different libraries. If the vocabulary is defined only in the root schema, then it is very easy for application to determine which library should be used for a given file. If the vocabulary can be re-defined (or even worse, extended) in any sub-schema) than one of the following should happen:
  1. application should perform schema traversal, which is not trivial (see json-schema-traversal I had to abstract from ajv)
  2. each (!) library becomes responsible for processing this keyword and somehow notifying the application that some part of the schema should be processed by another library.
  3. a special routing library should be used
  4. Some library that implements all vocabularies should be used - I find it highly unlikely that there will be such a library.
  5. @handrews any workable idea you have in mind? All above seem too complex for practical purposes...

Arguments to NOT allow multiple vocabularies
- validation concern. I really hope we will eventually be in a place that to use any custom keyword one MUST use extended meta-schema. It will prevent hours spent debugging schemas with mis-spelled keyword. Once we have an effective, elegant and agreed mechanism for schema extension it will be very easy to additionalProperties: false to all meta-schemas. Multiple vocabularies makes this pragmatic approach impossible
- processing concern. same as above, it makes it much more difficult for the application to determine how a given schema file should be processed.

So I may be missing something, but I really don't see how a questionable convenience of being able to package multiple schemas in a single file (particularly given that a simple alternative approach is possible - just define a package of schemas in the spec) can outweigh the above concerns.

epoberezkin:
@handrew please hold on - writing more, answers to your questions :)

epoberezkin:
> The nature of vocabularies is that they build on each other

I completely agree with that statement

> , so in order for a general system to tell whether their is a usable vocabulary, they need to know each vocabulary in the stack.

Not necessarily, as the extended vocabulary knows which vocabulary it extends. hyper-schema extends validation. ui-schema may extend validation. Mixing ui-schema and hyper-schema - really? Do you have a use-case?

> You are essentially saying that you want $vocabularies to be exactly like $schema, which is pointless. $vocabularies was specifically designed to not impose any file-wide requirements on schema processing.

No, I don't say that. meta-schema can be extended. The whole purpose of using vocabulary is to define which library should be used.

> I could see an alternative where meta-schemas declare which vocabularies (plural) they describe. In this approach, implementations would examine the meta-schema to look for recognizable vocabularies, and instances would declare $schema exactly as in draft-06.

That works for me too, as they are linked. Probably it is even better. But it means that the vocabulary cannot be changed in subschema.

epoberezkin:
> Once we have an effective, elegant and agreed mechanism for schema extension it will be very easy to add `additionalProperties: false` to all meta-schemas

By the way, quite a few people asked how they can prohibit additional properties in the meta-schema.

epoberezkin:
> I could see an alternative where meta-schemas declare which vocabularies (plural) they describe. In this approach, implementations would examine the meta-schema to look for recognizable vocabularies, and instances would declare $schema exactly as in draft-06.

The only problem for that approach is that the meta-schema is a schema that should validate itself. So we could say, schema MAY include $vocabulary, but in this case it MUST have $schema and the $vocabulary in the schema should be the same as in the meta-schema. The meta-schema for any vocabulary would then look like this:

```json
{
  "$schema": "some_uri",
  "$vocabulary": "whatever",
  "type": ["object", "boolean"],
  "properties": {
    "$vocabulary": {"const": "whatever"},
    "etc.": {}
  },
  "dependencies": {
    "$vocabulary": ["$schema"],
    "etc.": []
  }
}
```

epoberezkin:
One more though.

I actually agree that in the future we may need to be able to mix multiple vocabularies that can be used both separately and together. I hope that by then we will be in a position when we have a mechanism that allows to define as separate meta-schema by mixing individual meta-schemas. When (and if) we find ourselves in such predicament, nothing would stop us to allow a $vocabulary to be an array of strings as well.

I just don't think we are there now and we may never get there, as at the moment we only have two standardised vocabularies: "validation" and "hyper-schema", the latter already inherits from "validation", it cannot be used WITHOUT it.

Once we have mixable vocabularies together with meta-schema extension mechanism I would be very happy to support $vocabulary as array. At the moment we have neither. So why don't we keep things simple for now?

handrews:
@epoberezkin yeah, I think we can sort this out :-)

As you observed from the proposal to move `$vocabularies` off into the meta-schema, I can drop the per-schema aspect.  I want to think on it a bit more, but right now I doubt I have a compelling enough use case for it to push it.  The only time you wouldn't just declare all vocabularies in a schema in the root (or meta-schema) is because you have vocabularies that somehow conflict.  And... just don't do that.

Pushing the conflicting bits out into separate files and `$ref`-ing them is an acceptable workaround.

> Not necessarily, as the extended vocabulary knows which vocabulary it extends.

I'm not worried about an extended implementation knowing its base.  It has to (if only to delegate it to another library).  What I need is for a base implementation to understand an extended vocabulary.

If I have a "handrews-hyper-schema" vocabulary, what I need is for my implementation to be able to recognize my own private extensions.  However, I MUST NOT expect peer implementations to support them.  What I need, then, is the principle of graceful degradation:  If the standard hyper-schema vocabulary is explicitly declared (in the schema or meta-schema), then a standard hyper-schema library can make use of my extended schemas.

I cannot think of any way to provide this without explicitly listing all of the schemas.  We can explicitly list them in some form other than a flat list, but that is by far the easiest to handle.  applications don't always validate or even have a copy of the meta-schema, and may not even be able to download one.

> Mixing ui-schema and hyper-schema - really? Do you have a use-case?

Um... yes.  HATEOAS-driven UI.  It's one of the primary use cases I have for hyper-schema.  If there is no standard UI vocabulary, I'll do it with a custom one.

> when we have a mechanism that allows to define as separate meta-schema by mixing individual meta-schemas

I am only willing to accept this as a reason if you have a workable proposal now.  I have not heard anything.  `$vocabularies` solves the problem, and your objection (aside from the root schema only thing, which I'm fine changing) is basically that you don't need it.  You aren't doing a lot with vocabularies, but I am.

So I have a solution to this problem and you don't.  Why should I throw away my solution for your vaporware that may or may not ever materialize?

> I just don't think we are there now and we may never get there, as at the moment we only have two standardised vocabularies: "validation" and "hyper-schema", the latter already inherits from "validation", it cannot be used WITHOUT it.

No, *you* only see those.  We also have three proposals, at least one of which has a de-facto implementation (json-schema-form).  You also have no idea what I might be planning to do with schema vocabularies that I'm not proposing a standard vocabulary for (because not all vocabularies need to be standard).

-----

With all this in mind, I prefer that `$vocabularies` be present in the root schema of the (for lack of a better term) instance schema.  We should not require the meta-schema to be accessible.

epoberezkin:
All you problems can be either solved with vocabulary extension (in which case you don't need multiple vocabularies) or with allOf, where different subschemas can be references to other files.

I am well aware about the progress of other vocabularies, but none of them is a published draft at this point, so adding features to the core without having them published is premature.

You are ignoring the main question - how the meta-schema combining multiple vocabularies should be constructed. Once we agree on $merge or any other option from your vote-a-rama this problem will be solved.

I see absolutely no problem with you using multiple vocabularies even if only one is specified in the spec - there are many people using $data, for example, without any problem. I keep saying that the usage practice should precede the spec. So at this point we really only need $vocabulary, singular. Later we can allow $vocabulary to be plural by allowing an array of strings (same as with "type").

The argument "I need it, you have no idea what I might be planning to do with schema vocabularies, and therefore it should be added to the spec" is neither proper nor convincing. Firstly, these plans should be explained and discussed to a wider community. Also, I would like to see substantially more support for mixing multiple vocabularies at this point, before we have more than 2 vocabularies standardised, when the second is the extension of the first (so no mixing is needed).

To summarise, I am categorically against supporting mixing schema vocabulary at this point of JSON schema evolution - there is no proven need for it.

handrews:
@epoberezkin I am going to talk with the other active project members before continuing.

handrews:
I'm also just going to close this for now.  I don't think the conversation is productive for others to read at this point, and I want people to focus on the hyperschema rewrite.

Issue #249, opened 2017-02-11
Add boolean isSecret metadata attribute?


chrisdostert:
It would be fantastic to have an optional "isSecret" metadata attribute similar to "description" and "title", which  was a boolean. 

The intent for the instance of that schema would then be captured by the schema and consumers of said schema could then handle this in their own way.

chrisdostert:
just had an interesting though; this attribute along with "default" are somewhat useless depending on use case. 

If the schema is being used for object validation they don't apply; if the schema is being used for object creation or presentation (such as form display) they do apply.

awwright:
Can you please provide a description and use case?

chrisdostert:
One example use case is in the swagger spec.
They have defined a [custom type](http://swagger.io/specification/#data-types-12) "password" to "Hint UIs the input needs to be obscured".

possible description:
This keyword can be used to indicate instances of the schema should be kept secret. It is RECOMMENDED that implementations keep the instance secret, the meaning of which will be application specific. As a common example, a UI may hide characters of a string as a user types it.

handrews:
@chrisdostert if this is primarily presentation-oriented (hiding the characters of the string) then I think it would belong with #67 JSON UI Schema.

Another possible use would be in hyper-schema to indicate that an `"isSecret": true` field should only be transmitted via a secure protocol.  But then, the URI scheme usually handles that so it's not clear to me that it's really useful in hyper-schema.

I'd probably lean towards putting this in the proposed UI Schema vocabulary.

As for `"default"`, see #204.

handrews:
I've warmed to this proposal given how we've reorganized the annotation keywords during this draft.

I'd probably call it `secret` to match `readOnly` (because it's not `isReadOnly`)

This keyword, in a hyper-schema context, would generally mean that while it is usable in requests, it would never be present in responses.  That, plus the UI usage, provides use cases in two separate vocabularies, which I think is a good way to answer the question, "is this useful?"

handrews:
Merged #433 adding `writeOnly`, and json-schema-org/json-schema-vocabularies#6 tracks the remainder of the `secret` proposal.

PR #433, opened 2017-10-08
Add a "writeOnly" (formerly "secret") annotation

handrews:
This is for passwords and other senstive or write-only fields.

This addresses #249.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -1408,6 +1408,7 @@
                             <t>Warn of possibly removing "dependencies" in the future</t>
                             <t>Grouped validation keywords into sub-sections for readability</t>
                             <t>Moved "readOnly" from hyper-schema to validation meta-data</t>
+                            <t>Added "writeOnly"</t>
                             <t>Added string-encoded media section, with former hyper-schema "media" keywords</t>
                             <t>Restored "regex" format (removal was unintentional)</t>
                             <t>Added "date" and "time" formats, and reserved additional RFC 3339 format names</t>
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -1263,6 +1263,27 @@
                     </t>
                 </section>
 
+                <section title='"secret"'>
+                    <t>
+                        The value of this keyword MUST be a boolean.  When multiple occurrences
+                        of this keyword are applicable to a single sub-instance, the resulting
+                        value MUST be true if any occurrence specifies a true value, and MUST
+                        be false otherwise.
+                    </t>
+                    <t>
+                        If it has value of boolean true, this keyword indicates that the value
+                        of the instance is sensitive in nature and SHOULD NOT be displayed
+                        except while being created.  A system MAY implement this keyword as a
+                        write-only value.
+                    </t>
+                    <t>
+                        This keyword can be used to assist in user interface instance generation.
+                    </t>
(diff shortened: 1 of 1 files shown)


handrews:
Last call-ish on this one!

Given it's simplicity, and the fact that I'd like to move into the feature-freeze editorial review on Friday, I will probably merge this on Friday even though it will be one day early.

philsturgeon:
I'm not sure if secret is the way to advertise the feature, even if it's what the keyword is being used for. I think writeOnly would be a tad more consistent? 

Writing things but not being able to see them after is what they're asking for, no?

handrews:
@philsturgeon I'm open to `writeOnly`, the same thought had occurred to me but I figured I'd see if this got any reaction at all :-)

@dlax and/or @chrisdostert how do you feel about it?

Thinking a bit more, I can see `writeOnly` as being a general annotation here, which UIs may or may not interpret as using a "password" widget in the UI.  The UI Schema proposal would then including a more detailed indicator of what kind of widget to use.

Does that seem like a good distinction?

handrews:
I reworked this as `writeOnly` and I like it a lot.  Not sure if it needs to sit longer for consideration?  I am still inclined to merge it on Friday and include it in final review but definitely speak up if you would be concerned about that.  I could add a CREF note about "secret" as an alternative if there are concerns about people missing this discussion.

Another use case for `writeOnly` is that I often find that triggering actions in a RESTful system is done using state manipulation fields that are often logically nulled out as soon as they are processed (not a literal JSON `null`, but removed entirely).  They have no state representation appearance, they are purely buttons/triggers that can be activated.  So that would be a `writeOnly` field that is not a `secret` field.  It's not removed from responses because of sensitivity, just because it is not useful in responses at all.

philsturgeon:
Fantastic.

chrisdostert:
@philsturgeon @handrews I think both are useful & have specific semantics & use cases. 

Knowing something is write only is not the same as knowing it's secret. 

For example, think of logging; write-only instances might be included in logs whereas secret instances might not.

handrews:
@chrisdostert yes, agreed they are both useful.  The proposal is that `writeOnly` is useful in general, and `secret` is useful in presentation (logging would be a presentation concern- I should be clear that I did not just mean "UI" as in "web form")

There is a similar separation for `readOnly`, where a read-only field might be managed as hidden form field (you can't change it so don't show it), or it might be managed by displaying the value as a non-editable field (you can't change it, but you should know about it).

So...
generic usage:  `readOnly`, `writeOnly`
presentation:  `hidden`, `secret`

handrews:
This spec is for generic usage, presentation ideas go in json-schema-org/json-schema-vocabularies#2

handrews:
Also, you can easily treat `writeOnly` as if it meant "secret" in your application without violating either the letter or spirit of the specification.  The reverse is not true.

handrews:
@chrisdostert if you want to argue that we should **not** move forward with `writeOnly` please do so by tomorrow.  Otherwise I will assume that you want to see `secret` added somewhere but are OK with that being tracked as json-schema-org/json-schema-vocabularies#6 which I filed for this as a presentation description request.

handrews:
Of course if you come back next week with new objections we can continue to discuss changes, I would just like to include this in the "final review" announcement for this weekend.

PR #434, opened 2017-10-09
Rough draft of a few examples

handrews:
This is just to get the conversation started about this approach
to examples.

Note that the diff is relative to the rewrite.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -8,6 +8,7 @@
 <!--<!ENTITY rfc5226 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5226.xml">-->
 <!ENTITY rfc5789 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5789.xml">
 <!ENTITY rfc5988 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5988.xml">
+<!ENTITY rfc6068 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6068.xml">
 <!ENTITY rfc6570 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6570.xml">
 <!ENTITY rfc6573 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6573.xml">
 <!ENTITY rfc6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
@@ -1468,7 +1469,7 @@ Link: <https://schema.example.com/entry> rel=describedBy
     "$schema": "http://json-schema.org/draft-07-wip/hyper-schema#",
     "base": "http://api.example.com",
     "type": "object",
-    "required": "data",
+    "required": ["data"],
     "properties": {
         "id": {"$ref": "#/definitions/id"},
         "data": true
@@ -1477,7 +1478,7 @@ Link: <https://schema.example.com/entry> rel=describedBy
         {
             "rel": "self",
             "href": "things/{id}",
-            "hrefRequired": ["id"],
+            "templateRequired": ["id"],
             "targetSchema": {"$ref": "#"}
         }
     ],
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -1137,6 +1137,209 @@ for varname in templateData:
...
+                    The linked hyper-schema defines the API's base URI and provides
+                    two links:  an "about" link to API documentation, and a "self"
+                    link indicating that this is a schema for the base URI.  In this
+                    case the base URI is also the entry point URI.
+                </t>
+                <figure>
+                    <artwork>
+<![CDATA[
+{
+    "$id": "https://schemaexample.com/entry",
```
dlax: missing `.`

jsonschema-hyperschema.xml
```diff
@@ -1137,6 +1137,209 @@ for varname in templateData:
...
+            "href": "things/{id}",
+            "hrefRequired": "id",
+            "targetSchema": {"$ref": "#"}
+        }
+    ],
+    "definitions": {
+            "type": "integer",
+            "minimum": 1,
+            "readOnly": true
+        }
```
dlax: missing an `id` key in "definitions"

jsonschema-hyperschema.xml
```diff
@@ -1137,6 +1137,209 @@ for varname in templateData:
...
+                    <artwork>
+<![CDATA[{
+    "rel": "tag:rel.example.com,2017:thing",
+    "href": "things/{id}",
+    "hrefSchema": {
+        "required": ["id"],
+        "properties": {
+            "id": {"$ref": "thing#/definitions/id"}
+                "type": "integer",
+                "minimum": 1
```
dlax: `"type"` and `"minimum"` should not be there I guess?

jsonschema-hyperschema.xml
```diff
@@ -1137,6 +1137,209 @@ for varname in templateData:
...
+    "required": "data",
+    "properties": {
+        "id": {"$ref": "#/definitions/id"},
+        "data": true
+    },
+    "links": [
+        {
+            "rel": "self",
+            "href": "things/{id}",
+            "hrefRequired": "id",
```
dlax: `hrefRequired` must be an array.

jsonschema-hyperschema.xml
```diff
@@ -1137,6 +1137,209 @@ for varname in templateData:
...
+}]]>
+                    </artwork>
+                </figure>
+                <t>
+                    The "href" value here is the same, but everything else is different.
+                    Recall that the instance is an empty object, so "id" cannot be resolved
+                    from instance data.  Instead it is required as client input.  This LDO
+                    could also have used "hrefRequired" but with "required" in "hrefSchema"
+                    it is not strictly necessary.  Providing "hrefRequired" without marking
+                    "id" as requried in "hrefSchema" would lead to errors, as client input
```
dlax: typo: requried

jsonschema-hyperschema.xml
handrews: This is what I get for deciding, fairly late at night, "meh, this is far enough along for feedback" and just pushing it with only a cursory proofread.  Sorry about that :-/

philsturgeon:
Very simple. Start with the most common usages and leaving room to get super technical later. 👍

dlax:
One thing that'd be nice ultimately would be that all examples in the spec share the same "theme". For instance, in OpenAPI, most examples concern a pet store; in JSON API, they're all about articles/author/comments.
I think this helps the reader because they can focus on the specification vocabulary more easily as they get familiar with the underlying theme and they don't have to get through a new theme on each example.

@handrews Just mentioning the idea before you engage into more work on this side.

handrews:
@dlax yeah, that has occurred to me.  Unfortunately that's the part of writing examples that I'm particularly bad at, hence the "thing" schema.  Concrete examples usually feel too contrived to me (plus I find that if there's a concrete example, someone inevitably complains that this isn't how they would have done it, seeming to fail to get the point that an example is just an example and not an endorsement of perfection or whatever).

I really, really hate designing examples.

I think the best way to handle this is for me to work through writing them all the best way that I can manage, which is semi-abstract.  Afterwards maybe someone can come up with a storyline for them, in which case I will be more than happy to rework the examples as needed.

I agree that that is the best way to do it, I just have no clue what will work as my notion of what makes a compelling example has repeatedly been proven to be out-of-step with the vast majority of readers.

handrews:
@dlax @philsturgeon I have added another example about processing multiple kinds of input.  It also covers using an alternate URI scheme (`mailto:`) and media type (`multipart/alternative`, with `text/plain` and `text/html` alternatives modeled as an array, using `contentMediaType` to describe each alternative's format).

I went with this next rather than adding the Collections section as it turns out the the Collections schema basically uses nearly every feature together.  It's probably best done last, although the narrative may be less clear that way.

Again, I'm going to work through this with my own preferred examples, because if I try to come up with a concrete narrative before putting up a PR we're going to be here until 2020.  So I'll go through the rest of the cases and do the collection at the end, and we'll see if we can build a narrative for that once the whole sequence is up.

I'm planning to keep adding commits to this PR rather than stack a whole bunch of PRs, unless someone objects.

dlax:
@handrews 
> I think the best way to handle this is for me to work through writing them all the best way that I can manage, which is semi-abstract. Afterwards maybe someone can come up with a storyline for them, in which case I will be more than happy to rework the examples as needed.

Sure, that seems eminently reasonable!

jsonschema-hyperschema.xml
```diff
@@ -1384,37 +1385,196 @@ Link: <https://schema.example.com/entry> rel=describedBy
...
-                    Forbidding input and requiring instance data.
-                    Requiring input and forbidding instance data.
-                </cref></t>
+            <section title="Submitting a payload and accepting URI input">
+                <t>
+                    This example covers using the "submission" fields for non-representation
+                    input, as well as using them alongside of resolving the URI Template with
+                    input.  Unlike HTML forms, which require either constructing a URI or
+                    sending a payload, but do not allow not both at once, JSON Hyper-Schema can
+                    describe both sorts of input to for the same operation on the same link.
```
dlax: typo: "to for"

jsonschema-hyperschema.xml
```diff
@@ -1384,37 +1385,196 @@ Link: <https://schema.example.com/entry> rel=describedBy
...
+                    payload, as seen in the <xref target="HTTP">appendix on HTTP usage</xref>.
+                </t>
+                <t>
+                    In this case, we use a "mailto:" URI, which, per
+                    <xref target="RFC6068">RFC 6068, Section 3"</xref>, does not provide any
+                    operation for retrieving a resource.  It can only be used to construct
+                    a message for sending.  Since there is no concept of a retrievable,
+                    replaceable, or deletable target resource, "targetSchema" and
+                    "targetMediaType" are not used.  Non-representation payloads are
+                    described by "submissionSchema" and "submission MediaType".
```
dlax: typo: "submission MediaType".

Also, the last sentence seems superfluous with respect to the next paragraph (and also the first sentence of this section).

jsonschema-hyperschema.xml
```diff
@@ -1384,37 +1385,196 @@ Link: <https://schema.example.com/entry> rel=describedBy
...
+                    "title": {
+                        "type": "string"
+                    },
+                    "cc": {
+                        "type": "string",
+                        "format": "email"
+                    },
+                    "email": false
+                }
+            }
```
dlax: missing `,`

jsonschema-hyperschema.xml
```diff
@@ -1384,37 +1385,196 @@ Link: <https://schema.example.com/entry> rel=describedBy
...
+                </figure>
+                <t>
+                    So, given the following instance retrieved from "https://api.example.com/stuff":
+                </t>
+                <figure>
+                    <artwork>
+<![CDATA[{
+    "title": "The Awesome Thing",
+    "stuffWorthEmailingAbout": "Lots of text here...",
+    "email": "someone@exapmle.com"
```
dlax: typo: exapmle

handrews:
I'm going to update this with fixes to the last batch of errors @dlax spotted but otherwise this will get merged on Friday to kick off the final review period.  I will open new PRs for further examples.

PR #438, opened 2017-10-12
Reconstruct the Relative JSON Pointer I-D XML

handrews:
This is the best I can manage for reproducing the previous
draft's XML (but adding myself as editor and updating the draft name).
A few of the xref's may be in slightly different styles, as I've
never entirely figured out all of the options for that.  But they
are all equally readable compared to the past generated document
as far as I can tell.

Also add to the Makefile and .gitignore.

### DIFF
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -1,269 +1,282 @@
-Internet Engineering Task Force                             G. Luff, Ed.
-Internet-Draft
-Intended status: Informational                             July 01, 2013
-Expires: January 02, 2014
+<?xml version="1.0" encoding="US-ASCII"?>
+<!DOCTYPE rfc SYSTEM "rfc2629.dtd" [
+<!ENTITY RFC2119 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.2119.xml">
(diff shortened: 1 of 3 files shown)


handrews:
BTW I know there are numerous other things to update (for instance, moving to the most recent JSON RFC.)  I will work on those things as subsequent PRs.  This PR is just a "hey, does this look like a decent starting point?" sanity check.

handrews:
Also, I used "handrews" in the draft name instead of "andrews" as there have already been drafts submitted by another "andrews" and it seemed preferable to avoid confusion.

The official guidance on naming is:

> Individual: A string related to the name of one of the authors in some way. There are no mechanical rules for this string but objectionable or misleading strings are subject to change or removal at the discretion of the Secretariat.

so this would seem to be fine.

relative-json-pointer.xml
```diff
@@ -0,0 +1,282 @@
...
+                    </t>
+                </list>
+            </t>
+        </section>
+
+        <section title="JSON String Representation">
+            <t>
+                The concerns surrounding JSON String representation of a Relative
+                JSON Pointer are identical to those laid out in
+                <xref target="RFC6901">RFC 6901, Section 5</xref>
```
dlax: missing `.`

handrews:
I'm going to go ahead and merge this as it is literally an import of existing text, and I would like to make some minor changes in separate PRs.

Issue #442, opened 2017-10-14
Removing dependencies?


epoberezkin:
@handrews could you please point me to the discussion about adding the note about the possibility of removing dependencies?

If it wasn't discussed, I would propose to remove this note from the spec until such possibility is agreed.

Without "dependencies", a concise schema:

```json
{
  "dependencies": {
    "foo": ["bar"],
    "baz": ["quux"]
  }
}
```

will have to be replaced with:

```json
{
  "allOf": [
    {
      "if": {"required": ["foo"]},
      "then": {"required": ["bar"]},
    },
    {
      "if": {"required": ["baz"]},
      "then": {"required": ["quux"]},
    }
  ]
}
```

which seems both more verbose and more difficult to understand.

So I am not sure what could be the motivation to remove it.

handrews:
You were the one who OK'd it:
https://github.com/json-schema-org/json-schema-spec/issues/180#issuecomment-266986089

The reason for putting the comment in is to provoke reactions from people who are actually using the keyword and want to keep it.  The note is doing what it is intended to do.  It is not in any way a commitment to removing the keyword.

I'm closing this not because your objections aren't useful (they are, and I've noted the point), but because I want people other than the usual suspects to file objections.  That works better if they can't find one that's already open :-)

handrews:
Thinking on this a bit more, it's really the schema form of `dependencies` that I would prefer to remove.  I agree that the property name form is a nice shortcut.  But I don't think that `"if": {"required": ["foo"]}` on its own, with an arbitrary `then`, is worth having a shortcut for.

Or if it is, I *really* don't like having the same keyword used for the string and schema forms.  They're not really the same thing and it's hard to reason about the keyword as it can be either a subschema keyword or a value keyword which is weird.

handrews:
`type` has two forms, but both are value forms (single string vs array of strings).  `items` has two forms, but both are subschema forms (single subschema vs array of subschemas).

`dependencies` has one value form and one subschema form.  I think the value form is worth keeping, but I'd like to encourage removal of the subschema form in favor of `if`.  Perhaps I could clarify the note in this manner?

epoberezkin:
I agree about schema form, this one was a bit confusing indeed.

pipobscure:
I find the schema form **very useful**.  Please let's not remove it. If anything let's remove the values form. 

`{ "foo": { "required": [ "bar" ] } }`

isn't that much worse than

[...]

PR #444, opened 2017-10-15
Clarify dependencies deprecation

handrews:
It is focused on the subschema form, not the string array form.

This addresses #442.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -608,10 +608,13 @@
                     <t>
                         <cref>
                             Now that "if", "then", and "else" are keywords, it is not clear whether
-                            there is any benefit in keeping dependencies.  It is frequently
-                            misunderstood and seems to be rarely used.  Depending on feedback
-                            with "if", "then", and "else", this keyword may well be removed in a
-                            future draft.
+                            there is any benefit in keeping the schema form of "dependencies".
+                            It is frequently misunderstood, and having a form that takes a subschema
+                            as well as a form that takes a primitive value is particularly
+                            confusing.  And it seems to be rarely used.  Depending on feedback
+                            with "if", "then", and "else", the schema form of this keyword may
+                            well be removed in a future draft.  Current thought is that the string
+                            form (giving property names in an array) is a useful shortcut.
                         </cref>
                     </t>
                     <t>


jsonschema-validation.xml
```diff
@@ -608,10 +608,13 @@
                     <t>
                         <cref>
                             Now that "if", "then", and "else" are keywords, it is not clear whether
-                            there is any benefit in keeping dependencies.  It is frequently
-                            misunderstood and seems to be rarely used.  Depending on feedback
-                            with "if", "then", and "else", this keyword may well be removed in a
-                            future draft.
+                            there is any benefit in keeping the schem form of dependencies.  It is
```
Anthropic: typo 'schem'

handrews: Thanks!  Fixed.

pipobscure:
I find the schema form **very useful**.  Please let's not remove it. If anything let's remove the values form. 

`{ "foo": { "required": [ "bar" ] } }`

isn't that much worse than

`{ "foo": [ "bar" ] }`

but replacing the schema version with a combination of `allOf/anyOf/oneOf/if/then/else` is much worse.

It's much clearer than `if`/`then`/`else`.

Side-jibe at `if`/`then`/`else`: These are a bad idea to begin with, but since they are not obligatory to use, I'm fine keeping them. One reason why they are bad is that they mutate JSON-Schema from a schema to a programming language.

epoberezkin:
I don't mind keeping both.

> Side-jibe at if/then/else: These are a bad idea to begin with, but since they are not obligatory to use, I'm fine keeping them. One reason why they are bad is that they mutate JSON-Schema from a schema to a programming language.

We had a long discussion and agreed that if/then/else is boolean implication (that can be expressed via and/or/not) rather than imperative construct. There are quite a few languages that use if/then/else syntax to mean the same as ternary in JS, and not a control flow statement.

handrews:
@pipobscure yes this objection (raised primarily by me) was covered in excessive detail in #180.
This warning will stay in, as the point is to have the warning in the published draft for comment, but your objection is noted on the issue.

handrews:
I'm going to merge this so that people who want to talk about why to remove or not remove `dependencies` will comment in the issue and not here.  I'll leave the issue open though- I put it in the "future" milestone.

PR #445, opened 2017-10-17
New overview based on applicability and assertions

handrews:
***NOTE:*** _No implementation requirements are changed by this PR!_

This reworks the "General validation considerations" section into
an overview that incorporates the applicability concept that was
recently added to the annotation section.  Some of the content
from that section was removed as it became duplicate content
with the new Overview.

The new Overview also talks about assertions, which is a word
used in the existing Abstract and Introduction sections.  It
helps by providing a name for the non-annotation keyword behavior,
other than "validation" which is used for the complete
specification.

The very awkward "Validation of primitive types and child values"
is covered much more clearly when written in terms of applicability
vs assertions.  These concepts map naturally to the distinction
that was being made in that section, and make it less of a weird
exception that needs explaining.

The other old sections are also covered in the new wording,
but are less dramatically affected.  And in some cases, kept
nearly word-for-word in new locations.

Also, move Interop after Overview.  It is too much low-level technical
detail to be that early in the document.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -259,7 +259,11 @@
                     A JSON Schema MUST be an object or a boolean.
                 </t>
                 <t>
-                    Boolean values are equivalent to the following behaviors:
+                    JSON Schema vocabularies can include assertions, which return a boolean result
+                    when applied to an instance.  The boolean schema values "true" and "false"
+                    are trivial assertions that always return themselves regardless of the
+                    instance value.  As an example, in terms of the validation vocabulary,
+                    boolean schemas are equivalent to the following behaviors:
                     <list style="hanging">
                         <t hangText="true">Always passes validation, as if the empty schema {}</t>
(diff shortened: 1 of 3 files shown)


handrews:
@Anthropic @dlax we talked a bit about these changes offline, could y'all give an official stamp of approval if you are still OK with this approach?

Everyone: As this does not change implementation requirements but does make an already-introduced concept (applicability) fit into the overall wording better, I would like for it to be merged on Friday for the start of the final review period.  Further feedback welcome, of course, and if mjor objections arise after that we can revert it.  But I want this wording to be part of the broader review period.

dlax:
I'm still OK with the approach. I can do a more detailed review, but not before October 30.

handrews:
@dlax are you OK enough with this being merged and doing your detailed review after that as part of the overall month-long final review period?

dlax:
> @dlax are you OK enough with this being merged and doing your detailed review after that as part of the overall month-long final review period?

Yes.

Issue #421, opened 2017-09-23
"self" links, "hrefSchema", and collections


handrews:
While working on the rewrite, it occurred to me that using `hrefSchema` with a "self" link doesn't make much sense.  So I wrote up a bit on how `hrefSchema` MUST NOT be used with a "self" link, and `hrefRequired` SHOULD be used to ensure that the self link can always fully resolve from instance data (or if i cannot, it cannot be used).

Then I thought about collections, with filtering and pagination.  What does "self" mean in such a case?  Strictly speaking, a page of a filtered collection is **not** the same resource as an unfiltered, unpaginated collection.  Or a different page from the same filter, etc. etc.  Similar concepts apply to other resources that are logically "modified" or made into a sub-resource by query parameters.

However, there is no inherent distinction between this use of query parameters and a use that produces entirely different resources which are only distinguished by non-hierarchical identifiers.

So if a "self" link MUST NOT use `hrefSchema` (which is a debatable point), how does a collection or similar resource expose its pagination/filtering/whatever interface for users?

While there are numerous pagination link relations, they should not take user input for the page parameters either.  A "first" link must always go to the first page, a "next" link must always go to the next page, whatever that means.

Using a "collection" link alongside of "self" to specify `hrefSchema` for pagination/filtering/etc doesn't seem to work, as setting the context of "collection" correctly is more or less impossible (if there are no items in the collection, then it is entirely impossible).

There is a "search" link relation that has as it's target a description of an interface to search the context.  This seems close to what we want, although the target would ideally be the `hrefSchema` of... um... something?  Ugh.. brain hurts.  Anyone have ideas on this?

handrews:
Related:  Is it possible for a resource to have more than one "self" link?  I mean where the resource is the context of the link, not the situation where there are "self" links attached to, for instance, individual array items within a larger resource.  In that case, each "self" link has a different context so there is still only one per context.

It's definitely possible for the same abstract resource to be accessible at multiple URIs, but would those other URIs also be "self" links or would at most one be "canonical" and the others be "alternate"?

MetaAbstract:
Hi, i am working on project . It's based on json and json schema. In practice href depends from server which keep document. For example json document   has been sent from one server to another. It's will be different hyper schema elements for one document and ever different id and related schema's $ref.  Really i was inspired by hyper schema first. Especially by media types. I think hyper schema is movement to graphql technology.

handrews:
[...]

PR #448, opened 2017-10-19
Add most of a collection example.

handrews:
There are some things that are not addressed here, and most
importantly this highlights the problems that I wrote about
in issue #421.  Hopefully they are more clear when laid out
like this and noted in CREFs.

I feel like it is a prerequisite of a hypermedia format that
claims to be feature-complete to be able to recognize
self-describing collections and make use of them without having
specific knowledge of the data type of their contents.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -440,7 +440,7 @@
                 <section title='"self" links' anchor="self">
                     <t>
                         A hyper-schema implementation MUST recognize that a link with relation
-                        type "self" that is applicable to the entire instance document describes    
+                        type "self" that is applicable to the entire instance document describes
                         how a user agent can interact with the resource represented by that
                         instance document.  A "self" link MUST be resolvable from the instance,
                         and therefore "hrefSchema" MUST NOT be present.
@@ -1467,7 +1467,7 @@ Link: <https://schema.example.com/entry> rel=describedBy
 <![CDATA[{
     "$id": "https://schema.example.com/thing",
     "$schema": "http://json-schema.org/draft-07-wip/hyper-schema#",
-    "base": "http://api.example.com",
+    "base": "https://api.example.com",
     "type": "object",
     "required": ["data"],
     "properties": {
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -1722,6 +1722,302 @@ Link: <https://schema.example.com/entry> rel=describedBy
...
                 </cref></t>
+                <t>
+                    In many systems, individual resources are grouped into collections.  Those
+                    collections also often provide a way to create individual item resources with
+                    server-assigned identifiers.
+                </t>
+                <t>
+                    This schema describes a collection where each item representation is
+                    identical to the individual resource item representation, and there
+                    is enough metadata included in the collection representtion to
```
philsturgeon: `representtion`

philsturgeon:
This is a great direction. A little typo but great content. Seeing as a _lot_ of what JSON Schema is going to be used for is CRUDy REST(ish) APIs, this example is really going to drive the point home to people doing that.

jsonschema-hyperschema.xml
handrews: Thanks- fixed with a rebased commit that only changed this one word.

handrews:
Given the significance of this example and the one enthusiastic approval, I'm going to go ahead and merge this to get it into the "final review" announcement.  Obviously further comment is very much desired and welcome, and merging this does not set it in stone at all.

dlax:
I agree with @philsturgeon that this example demonstrates the power of JSON Hyper Schema, however, it is *very* complicated and thus I suspect many people will get lost quite quickly. (I find myself lost, to be honest.)

One thing that'd help would be to leave the pagination part aside and only focus on collection+item links (plus POST semantics). Hence, the collection schema could be a simple array.

A more complete example with pagination could be in the form of a tutorial on the website.

handrews:
@dlax I don't find an array example to be very compelling.  For me, that's not a real-world example.  Collections are filterable or paginated/scrolling are unmanageable under hyper-schema without meta-data, and showing a "collection" that can't be filtered or paginated is not realistic.

While I don't want to make the standard a tutorial, this is ***THE*** design pattern of REST and experience tells me that few people will get it right on the first try with Hyper-Schema.  I certainly did not.  It uses nearly all of the LDO features in ways that are not always obvious but are crucially important.

I would definitely be interested in suggestions on how to build up to it in more steps, organize it better, or explain it better.  But I think a fully functional collection (with at least one of pagination or filtering, and pagination is easier) is the most important example to have.

This gets back to the difference in how you and I see examples.  I dislike small examples all over the place that break up the flow of text.  What I really want to see is how you use everything together for something of real-world complexity.  That is what is interesting, because it is _not obvious_ how to do this from just reading all of the features.

@philsturgeon do you have any thoughts on this?

I would love to find a way to reach this level of example even if we break it up into more subsections.

handrews:
@dlax also if you want to get into an involved discussion on how to improve this example, maybe file an issue for it?  I just realized this is a closed PR.

philsturgeon:
@handrews in my opinion @dlax makes a great point. The confusion between "items" in the server response and "items" as a JSON Schema keyword is really going to throw a lot of people, but as you say documenting a bare array is going to confuse people further. 

People do not generally return bare arrays for collections due to whispers of "its not secure", which is mostly [a tiny vector in outdated browsers](https://haacked.com/archive/2008/11/20/anatomy-of-a-subtle-json-vulnerability.aspx/) that only exists if your API supports sessions, which... ugh what.

So, we have a confusing example, with a confusion option suggested. How do we resolve this?

Renaming "items" might be a good start?

handrews:
@philsturgeon we can definitely rename items.  Sometimes I call it "elements", depending on whether I think the symmetry or lack of symmetry is more confusing.  ¯\_(ツ)_/¯

I'm also open to other names.  "contents"?  "theActualStuff"?  :-P

Should this be broken into two examples?  I could do it without pagination, and then add pagination.

philsturgeon:
Right on! 

Some folks call it data which is pretty darn vague, or... `<foo>s`

handrews:
I think I'll avoid "data", "results", or anything else that might look like a generic response envelope rather than something specific (so "contents" is probably out as well).

I go back and forth with "foos".  If there's a well-defined domain vocabulary with rules for plurals, that works well.  If not, it's better to have something standard-ish ("items" or "elements").  Of course I also don't want the examples to be a rathole of stylistic debates on API design, either.

handrews:
@dlax @philsturgeon posted PR #464 to attempt to address this.

dlax:
@handrews 
> PR #464 to attempt to address this.

That's a good step forward.

> @dlax also if you want to get into an involved discussion on how to improve this example, maybe file an issue for it? I just realized this is a closed PR.

Opened #465 to discuss about pagination more specifically.

PR #449, opened 2017-10-19
Add request/response implementation requirements

handrews:
This mostly involves explaining that hyper-schema is not a system
for executing requests and processing response.  It informs a
user agent how to construct various requests, and is used to
interpret whatever response (expected, error, whatever) is returned.

Also define "operation" up front as I keep forgetting where it's defined
and not being sure whether it was defined recently enough in the spec
for a reader to have noticed and remembered.

This is groundwork for better advice on HTTP, and is intended to provide
context around how hyper-schema systems should work with requests
and responses that is non-HTTP-specific.  The HTTP section will then
apply HTTP methods, headers, etc. to this conceptual framework
to illustrate how HTTP should be used with hyper-schema LDOs.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -207,6 +207,14 @@
                             command-line arguments, configuration files, or hardcoded values
                             in source code.
                         </t>
+                        <t hangText="operation">
+                            A specific use of a hyperlink, such as making a network request
+                            (for a URI with a scheme such as "http://" that indicates a protocol)
+                            or otherwise taking action based on a link (reading data from a
+                            "data:" URI, or constructing an email message based on a "mailto:"
+                            link).  For protocols such as HTTP that support multiple methods,
+                            each method is considered to be a separate operation on the same link.
+                        </t>
                     </list>
                 </t>
             </section>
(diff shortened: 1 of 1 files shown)


handrews:
@vasilakisfil here is another PR on top of the rewrite that is probably relevant to your interests, although may make more sense once I submit the companion PR updating the HTTP section to match this.

jsonschema-hyperschema.xml
```diff
@@ -1183,24 +1191,81 @@ for varname in templateData:
...
+                    Target URI construction rules, including "hrefSchema" for accepting input,
+                    are identical for all possible requests.
+                </t>
+                <t>
+                    Requests that do not carry a body payload do not require additional keyword
+                    support.
+                </t>
+                <t>
+                    Requests that take a target representation as a payload SHOULD use the
+                    "targetSchema" and "targetMediaTYpe" keywords for input description and
```
vasilakisfil: Small typo here on "targetMediaTYpe"

vasilakisfil:
Thanks! This is very helpful section actually that clearly explains the uses of `targetSchema` and `submissionSchema`!

handrews:
Based on the positive feedback so far, and given that this does not change implementation requirements, I'm going to go ahead and merge this so it can be part of the "final review" announcement.

This does not in any way lock in these changes, and feedback and suggestions for a better approach are still very much wanted and appreciated.

jsonschema-hyperschema.xml
```diff
@@ -1186,24 +1194,81 @@ for varname in templateData:
...
+                    Requests that take a payload that is not derived from the target resource's
+                    representation SHOULD use the "submissionSchema" and "submissionMediaType"
+                    keywords for input description and payload validation.  Protocols used in
+                    hypermedia generally only support one such non-representation operation
+                    per link.
+                </t>
+                <t>
+                    RPC systems that pipe many application operations with arbitrarily different
+                    request structures through a single hypermedia protocol operation are outside
+                    of the scope of a hypermedia format such as JSON Hyper-Schema.
```
dlax: I find this paragraph is more confusing than helpful.

jsonschema-hyperschema.xml
```diff
@@ -1186,24 +1194,81 @@ for varname in templateData:
...
+                    processing status representations SHOULD also link to their own schemas
+                    and use appropriate media types
+                    (e.g. <xref target="RFC7807">"application/problem+json"</xref> for errors).
+                    Certain errors might not link a schema due to being generated by an
+                    intermediary that is not aware of hyper-schema, rather than by the origin.
+                </t>
+                <t>
+                    User agents are expected to understand protocol status codes and response
+                    media types well enough to handle common situations, and provide enough
+                    information to client applications to handle domain-specific responses.
```
dlax: This paragraph also seems superfluous.

PR #450, opened 2017-10-20
Changelog for hyper-schema

handrews:
Lots of change.


handrews:
Since this is mechanical, and since I'm sure someone will object if I send out a "final review" notice without a change log, I'm going to go ahead and merge this.  Obviously trivial to fix if someone notices as problem.

Issue #221, opened 2017-01-04
instance equality - trailing zeroes are insignificant?


asgs:
Hello! The spec says that [trailing zeroes are insignificant](http://json-schema.org/latest/json-schema-core.html#rfc.section.4.2) when it comes to instance equality.

Was it meant as leading zeroes instead? Also, I'm assuming this is referring to the fields of type "number". Please clarify.

Relequestual:
That's actually a really good point... No idea how that skipped through the net. I believe you are correct.

@awwright @handrews could either of you confirm this?

Relequestual:
The intention may have been trailing zeroes AFTER the decimal place, but I agree it's somewhat ambigious.

awwright:
Yeah that can probably be better worded. It's just an informative sentence,
but maybe we could update it to say "insignificant zeros" (leading or
excess in a decimal part)

Relequestual:
OK. I'll self assign to make a PR on this one.

awwright:
Cool.

Also, j/k on my leading zeros part, I recalled the JSON ABNF doesn't permit
that.

handrews:
This just intends to say that differences in trailing zeros do not impact equality because they are not mathematically significant, right?  1, 1.0, and 1.00000 are all mathematically equivalent, so as instances they are equal.  This is different from JSON equality as discussed in RFC 7159 which just discusses equality in terms of encoded strings.

awwright:
Here's some examples as I understand the concern and behavior of JSON:

`01` is invalid JSON
`1.0`==`1.00`
But `10`!=`100` (I.e. the spec should say leading zeros in the fractional
part)

handrews:
@awwright I don't think that JSON Schema needs to clutter itself with explanations that 10 and 100 are different numbers.

awwright:
I mean *trailing zeros in the fractional part

On Jan 5, 2017 16:05, "Austin William Wright" <aaa@bzfx.net> wrote:

Here's some examples as I understand the concern and behavior of JSON:

`01` is invalid JSON
`1.0`==`1.00`
But `10`!=`100` (I.e. the spec should say leading zeros in the fractional
part)

awwright:
Can we just say "excess zeros" here instead?

seagreen:
> Here's some examples as I understand the concern and behavior of JSON:
> ...
> `1.0`==`1.00`

[...]

PR #455, opened 2017-10-25
Clarify data model vs unicode characters

handrews:
This addresses #221.

JSON Schema does not care about unicode characters, only the
data model.  Vocabularies SHOULD account for this by working
with data model strings instead of the original unicode
representation of numbers.

For example, define a value for the "format" validation keyword
that indicates that a string should be interpreted as a fixed-point
decimal if you wish to retain precision information in your data.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -188,7 +188,13 @@
                         </list>
                     </t>
                     <t>
-                        Whitespace and formatting concerns are thus outside the scope of JSON Schema.
+                        Whitespace and formatting concerns, including different lexical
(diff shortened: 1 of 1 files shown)


seagreen:
Look 👌 to me!

If you wanted to you could add an example (perhaps about how this means JSON Schema doesn't distinguish between `1.0` and `1`), but that's a judgement call.

handrews:
@seagreen thanks!  I think I'm going to leave that out for now, mainly because there are a couple of issues filed that may address how to work with such distinctions using `format`.  So if people are interested I'd rather they took a look at the GitHub issues than saw something quite that definitive and left it alone.

PR #458, opened 2017-10-27
Example using "anchor", "base", "templatePointers"

handrews:
***NOTE:* This is not the same tree example as before the rewrite**

It also shows why "rev" is not necessary (it is deprecated in
other link specifications, so it is omitted here), and how
to use both absolute and relative JSON Pointers in
"templatePointers".

Also fix typos and margins and stuff.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1802,10 +1802,106 @@ Link: <https://schema.example.com/entry> rel=describedBy
                     </list>
                 </t>
             </section>
-            <section title='"anchor" and "base" as URI Templates'>
-                <t><cref>
-                    "base" used as a template with both "anchor" and "href" templates.
-                </cref></t>
+            <section title='"anchor", "base" and URI Template resolution'>
+                <t>
+                    A link is a typed connection from a context resource to a target resource.
+                    Older link serializations support a "rev" keyword that takes a link relation
+                    type as "rel" does, but reverses the semantics.  This has long been deprecated,
+                    so JSON Hyper-Schema does not support it.  Instead, "anchor"'s ability to
+                    change the context URI can be used to reverse the direction of a link.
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -1802,10 +1802,105 @@ Link: <https://schema.example.com/entry> rel=describedBy
...
+                        link is easily detectable by tools when a "self" link is also present.
+                    </postamble>
+                </figure>
+
+                <figure>
+                    <preamble>
+                        The following hyper-schema, applied to the instance in the response
+                        above, would produce the same "self" link and "up" link with "anchor".
+                        It also shows the use of a templated "base" URI, plus both absolute
+                        and relative JSON Pointers in "tempatePointers".
```
dlax: typo: `tempatePointers`

jsonschema-hyperschema.xml
```diff
@@ -1802,10 +1802,105 @@ Link: <https://schema.example.com/entry> rel=describedBy
...
+                    </preamble>
+                    <artwork>
+<![CDATA[
+    "base": "trees/{treeId}",
+    "properties": {
+        "id": {"type": "integer"},
+        "treeId": {"type": "integer"},
+        "childIds": {
+            "items": {
```
dlax: missing `"type": "array"` in `childIds`, I think.

jsonschema-hyperschema.xml
handrews: Blerg.  Spell check does not understand camelCase.  Thanks.

jsonschema-hyperschema.xml
handrews: Yeah, technically necessary for it to behave in an intuitive way, although I think folks can get the idea without it.  But better to avoid confusion.

PR #459, opened 2017-10-28
Remove partial archive of meta-schemas

handrews:
We're not keeping this up-to-date and we don't use it on the web
site either, so let's get rid of it.  We've already had one bit
of confusion over differences in formatting between theses and
the published ones.


awwright:
Is there anything to keep up-to-date? The meta-schemas never change.

The idea is to have this be the canonical definition of meta-schemas (unless the canon one published prior to this repository was different, in which case we should fix the mistake).

I suppose we don't strictly need this, but I'd like to see something lay out what is supposed to be defined where.

handrews:
We have fixed mistakes in 4 meta-schema files across 3 draft revisions in the last month.

All of the meta-schema versions are published through the web site repo, so I don't see the point of having them here.  We have gotten at least one confused person on the mailing list b/c they got out of sync.

handrews:
> I'd like to see something lay out what is supposed to be defined where.

@awwright could you explain what this means exactly?

The canonical versions are git submodule'd in to the web site repo based on a branch or tag.  Currently I have branches for draft-03, draft-04, and draft-06 because they had bugs, but I was planning on replacing those with tags once it seems pretty likely that we have the bugs sorted.

handrews:
No reply after a week, and the comment was not an outright objection, so I'm going to merge this.  Whatever it is that you meant by "lay out what is supposed to be defined where", please file it as a new issue.

PR #467, opened 2017-11-03
Add back examples in "JSON Hyper-Schema and HTTP" section

dlax:
Restore examples about `targetHints` and `headerSchema` as they were before #427.

These keywords being new and explicitly meant for experimentation and feedback (as noted in CREF), I think it's worth having examples in the specification so that people get clear idea of what they are for.
Also, the introduction of "Examples" section currently states that the section about HTTP contains example, and there are currently none.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -13,6 +13,7 @@
 <!ENTITY rfc6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
 <!ENTITY rfc7230 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7230.xml">
 <!ENTITY rfc7231 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7231.xml">
+<!ENTITY rfc7240 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7240.xml">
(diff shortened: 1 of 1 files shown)


handrews:
> Also, the introduction of "Examples" section currently states that the section about HTTP contains example, and there are currently none.

That's because I intended to move these or something like them to that section.  I would still prefer that.

dlax:
>> Also, the introduction of "Examples" section currently states that the section about HTTP contains example, and there are currently none.
> 
> That's because I intended to move these or something like them to that section.  I would still prefer that.
> 

Since I have no idea what you intent to do, I'm afraid I cannot help...

Again, gathering other opinions on this "examples" issue would really be 
beneficial.

handrews:
> Again, gathering other opinions on this "examples" issue would really be beneficial.

I think part of how we do that is we publish the document with the examples done a certain way and see what the response is.  From past experience, draft publication definitely does flush out a new round of comments.

I would like to see how a larger audience reacts to an approach where examples are concentrated in one or two sections.

I definitely dropped the ball on this and just forgot to cover it.  So you're reacting to a real problem here and I would not expect you to guess my intentions.  Let's leave this open for the moment and I'll think about how it should work.  I'm not dead set against examples in these sections to the same degree as the introduction, so maybe they will work best here.

handrews:
Let's fold this into #469 as well, as it is all related in some way to the section that addresses.

PR #468, opened 2017-11-03
Move reference to RFC 7231 above sentence about "targetSchema" purpose

dlax:
Reading the sentence about RFC 7231 (relationship between resource
representation and HTTP requests/responses) feels quite abrupt in this
section about "targetSchema and HTTP". So first restate what
"targetSchema" is meant for and then add the reference to protocol
guidelines.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1330,13 +1330,13 @@ for varname in templateData:
             </section>
             <section title='"targetSchema" and HTTP' anchor="targetHTTP">
                 <t>
+                    "targetSchema" suggests what a client can expect for the response to an HTTP
+                    GET or any response for which the "Content-Location" header is equal to the
+                    request URI, and what a client should send if it replaces the resource in an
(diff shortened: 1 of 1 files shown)


handrews:
I wrote it in that order for a reason, and it seemed to be fine in draft-06.  It's more important to me to motivate it from the standard document as many people claim various things about what HTTP means.

I'm open to reworking this section, but the paragraphs read very strangely to me in this order.  The now-first paragraph has irrelevant minor details, and the second one doesn't connect well to the first.

dlax:
The actual problem is that I don't get the point of the section as a whole. So I filled issue #469 for that but then felt that with the change suggested here, I might understand it better. Apparently, I'm not even going the right direction, so let's forget about this and discuss on the issue.

handrews:
@dlax thanks, I saw #469 and didn't comment b/c I was too tired last night to come up with anything (had just walked in the door from a long drive back from vacation), but I'll think it through today and comment.  It's an important section so I'm glad you are raising these issues.

PR #470, opened 2017-11-03
A couple of edits/fixes in hyperschema spec

dlax:


jsonschema-hyperschema.xml
```diff
@@ -1378,7 +1378,7 @@ for varname in templateData:
                     similarly structured headers wherever possible.
                 </t>
                 <t>
-                    No distinction is made between headers that may appear in responses to
+                    No distinction is made with headers that may appear in responses to
```
handrews: "with" is not the right word, but really this whole sentence is a mess so I'm not surprised you want to change something here.  I'll look at it tomorrow unless you feel like a more thorough rewording.

All the rest of these changes look great

dlax:
> "with" is not the right word, but really this whole sentence is a mess 
> so I'm not surprised you want to change something here. I'll look at it 
> tomorrow unless you feel like a more thorough rewording.

Ok, I dropped the first commit and let you work on this.

PR #472, opened 2017-11-03
Move up the JSON Hyper Schema in "mailto:" example

dlax:
While reading this section, I felt that the example came too late after
a lot of explanation on it. So moving it up just after its introduction
paragraph.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1653,7 +1653,7 @@ Link: <https://schema.example.com/entry> rel=describedBy
                     described by "submissionSchema" and "submission MediaType".
                 </t>
                 <t>
-                    Therefore, we use "submissionMediaType" to indicate a multipart/alternative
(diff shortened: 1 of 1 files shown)


handrews:
I generally like this, although I think some aspect of explaining why/how `multipart/form-data` is modeled in JSON Hyper-Schema needs to be explained before the example.  It's just to weird to make sense of without a bit of lead-in.

dlax:
> I think some aspect of explaining why/how multipart/form-data is modeled in JSON Hyper-Schema needs to be explained before the example.

Makes sense. As we cannot put all explanations before, we cannot either put them all after I guess.
I can move back the first (and maybe second) paragraph before, what do you think?

handrews:
Yes, I think moving the first two paragraphs back would work nicely.

dlax:
> Yes, I think moving the first two paragraphs back would work nicely.

Ok :)

I amended the commit and just force-pushed.

Issue #469, opened 2017-11-03
feedback about section '"targetSchema" and HTTP'


dlax:
The intent of section ["targetSchema" and HTTP]( http://json-schema.org/work-in-progress/WIP-jsonschema-hyperschema.html#rfc.section.8.2) is not quite clear currently. In particular, the discussion about "Accept" header in the second paragraph seems out of place. Same for the part on `submissionSchema` at the end. At least, this needs clarification and rewording.

I just sent #468 to slightly improve the introduction, but that's not enough and I'm not sure what to do next without having a better idea of the original intent.

Probably related, there's also a paragraph in section [Content negotiation and schema evolution](http://json-schema.org/work-in-progress/WIP-jsonschema-hyperschema.html#rfc.section.8.6) about "Accept" header that I don't get the point of. Especially, "Accept" being a request header, I find it surprising that it could appear in hyper-schema links which (as far as I understand) either advertise responses or submission requests.

handrews:
For "Accept", yes, it's a request header, and the `headerSchema` is for describing valid request header values.  So I don't quite understand the surprise.  Could you elaborate a bit more?

Also, could you please give more detail on your problem with `submissionSchema`?  I don't understand what is unclear about it.  In HTTP the "submission" fields are for POST requests.  What is confusing about that?

dlax:
> For "Accept", yes, it's a request header, and the headerSchema is for describing valid request header values. So I don't quite understand the surprise. Could you elaborate a bit more?

I think it was a matter of wordings. In particular the paragraph:
> The media type of the representation is given by the "targetMediaType" field.  Alternatively, possible available media types MAY be advertised for HTTP-accessible resources using the "accept" field in "headerSchema", corresponding to the HTTP "Accept" header.

Admittedly, I never seriously thought about putting an `Accept` in `headerSchema` but maybe this is appropriate and there are use-cases for this. So fine.

The rework in #476 makes this clearer anyways, thanks!

dlax:
> Also, could you please give more detail on your problem with submissionSchema? I don't understand what is unclear about it. In HTTP the "submission" fields are for POST requests. What is confusing about that?

I don't find the paragraph confusing *per se*. It's just that it appears within a section named `"targetSchema" and HTTP` and hence seems unrelated with the beginning of the section. That's why I mentioned the "content negociation" section. Maybe it's a matter of re-organizing contents.

handrews:
[...]

PR #476, opened 2017-11-07
Rework '"targetSchema" and HTTP'

handrews:
This is an attempt to introduce concepts in a better order
and generally improve the flow.  Examples relevant to this section
will be submitted separately once decisions are made as to where
they should go.

This hopefully addresses #469.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1330,29 +1330,38 @@ for varname in templateData:
             </section>
             <section title='"targetSchema" and HTTP' anchor="targetHTTP">
                 <t>
-                    The relationship between a resource's representation and HTTP requests and
-                    responses is determined by <xref target="RFC7231">RFC 7231, section 4.3.1 -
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -1330,29 +1330,38 @@ for varname in templateData:
             </section>
             <section title='"targetSchema" and HTTP' anchor="targetHTTP">
                 <t>
-                    The relationship between a resource's representation and HTTP requests and
-                    responses is determined by <xref target="RFC7231">RFC 7231, section 4.3.1 -
-                    "GET", section 4.3.4 "PUT", and section 3.1.4.2, "Content-Location"</xref>.
+                    "targetSchema" describes the resource on the other end of the link, while
```
awwright: "other end" might be ambiguous here (the first think I actually thought of was "link" as in "satellite link" not "Web link"). I might say "target end" (or in semweb contexts, "object end").

handrews: Yeah I kind of worried about that.  Thanks for the feedback, it helps and I'll update this.

PR #481, opened 2017-11-13
Add "examples" to the draft-06 meta-schema

stuartpb:
"examples" is described in the draft-06 spec, but is not defined in the meta-schema.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -57,6 +57,10 @@
             "type": "string"
         },
         "default": {},
+        "examples": {
+            "type": "array",
+            "items": {}
+        },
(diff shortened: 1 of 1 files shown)


stuartpb:
Note that the CI failure here is due to the draft-06 branch not having a .travis.yml; the "Build only commits with .travis.yml file" option should be set for this repository, but it is not: https://blog.travis-ci.com/2014-03-05-repository-settings/#new-settings

handrews:
I just updated the web site and pushed it, and verified that the correct version downloads now.

Issue #478, opened 2017-11-07
make collection+item example more self-contained


dlax:
_Follow-up on review comments in PR #464 (https://github.com/json-schema-org/json-schema-spec/pull/464/files#r148458346 and https://github.com/json-schema-org/json-schema-spec/pull/464/files#r148459558 in particular)._

Currently the [collection+item example section](http://json-schema.org/work-in-progress/WIP-jsonschema-hyperschema.html#rfc.section.9.5) is not self-contained as it implicitly refers to a JSON Hyper Schema defined three sections above (the "thing" schema in [section 9.2](http://json-schema.org/work-in-progress/WIP-jsonschema-hyperschema.html#rfc.section.9.2)). I'd suggest this "thing" schema to be repeated in the collections example, or inlined in the "thing-collection" example so that one can read and understand the section as a self-contained example. Alternatively, we might add a reference to another example, but this would arguably be more disruptive.

handrews:
@dlax I've been thinking about this and part of the decisions should be based on whether the typical reader (if there is such a thing) is likely to be reading the whole thing or using it as a reference.

If using as a reference, it should definitely be more self-contained.  To some degree I want to say that the spec is not intended to primarily be a reference (rather, it is a guide for implementors), but realistically that is how people use them.  I certainly use RFCs that way a lot myself.

So I am becoming more in favor of duplicating info even if it makes the document longer.  Still thinking on the best way to accomplish this, and whether some things would work fine as links, and if so where the example that is linked to fits best.

No real conclusion here, just an update on my thought process to show that I haven't forgotten :-)

handrews:
Also, if anyone has a better entity to use than "thing" that they think would be more clear, do speak up.  I made the examples about "things" because otherwise I'd still be sitting with a blank screen trying to come up with the perfect concrete example.

handrews:
@dlax It seems easy enough to copy the individual thing schema exactly as it appears earlier to be right above the thing collection schema.  I'd be fine with that duplication.

I also considered moving collection/pagination up so the individual thing is not so far away, but it seems to me like the other examples are needed to explain techniques used in collection and pagination.  Let me know if you think a different order of examples would work better, though.  I am definitely open to rearranging them.

If we want to just copy the individual thing down, is there anywhere else where these later examples feel incomplete?  I'm sold on the idea of them being self-contained at this point.

dlax:
@handrews Yes, a plain copy would fine and enough to solve this issue.

[...]

PR #490, opened 2017-11-16
Rework collection example for clarity

handrews:
Addresses #478.

The first commit is a few trivial typo fixes outside of the collection example.

It's best not to review the 2nd commit directly given the extent of the reworking.
Just build it and read it as an entire new example.  It was not reworked in anything
remotely resembling discrete steps, and I wouldn't even know where to start to split it.

@dlax note that several of the output URIs in the main example were flat-out wrong.
As we saw with the older tree example before the big rewrite, if you can't understand
an example it's quite likely not just badly organized but wrong.  I'll have to remember
to consider that more closely.  This is part of the problem of a spec with no implementation;
I can't just test it out.

Anyway, here are the main changes:

* Restate the item resource, including the newly added "collection"
  link.
* Remove some of the more complex discussions that weren't
  needed to introduce the key patterns and challenges.
* Make pagination and first-item-creation subsections, and put
  pagination first.
* Write creation guidance as if it generally works, and leave the
  doubts to the crefs.
* Generally improve the wording and flow.  Hopefully.

Note the references to issue #421, which I have to update with some of the things
I figured out while doing this.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1505,7 +1505,7 @@ for varname in templateData:
                 in <xref target="HTTP">an Appendix</xref>.
             </t>
 
-            <section title="Entry point links, no templates">
+            <section title="Entry point links, no templates" anchor="entryPoint">
                 <t>
                     For this example, we will assume an example API with a documented
                     entry point URI of https://example.com, which is an empty JSON object
@@ -2128,7 +2128,7 @@ Link: <https://api.example.com/trees/1/nodes/456> rev=up
                 </t>
                 <t>
                     There are two "item" links, one for each item in the "elements" array.
-                    Unilke the "self" links, these are defined only in the collection schema.
+                    Unlike the "self" links, these are defined only in the collection schema.
                     Each of them have the same target URI as the corresponding "self" link that
                     shares the same attachment pointer.  However, each has a different context
                     pointer.  The context of the "self" link is the entry in "elements",
@@ -2297,7 +2297,8 @@ Link: <https://api.example.com/trees/1/nodes/456> rev=up
                         </cref>
                     </t>
                     <t>
-                        Let's add a link for this collection to the entry point schema, including
+                        Let's add a link for this collection to the entry point schema
+                        (<xref target="entryPoint"/>), including
                         pagination input in order to allow client applications to jump directly
                         to a specific page.  Recall that the entry point schema consists only
                         of links, therefore we only show the newly added link:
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -2248,43 +2278,32 @@ Link: <https://api.example.com/trees/1/nodes/456> rev=up
...
+                            It's not clear how pagination should work with the link from
+                            the "collection" links in the individual "thing" schema.
+                            Technically, a link from an item to a paginated or filtered
+                            collection should go to a page/filter that contains the item
+                            (in this case the "thing") that is the link context.
+                            See GitHub issue #421 for more discussion.
+                        </cref>
+                    </t>
+                    <t>
+                        Let's add a link for this collection to the entry point schema, including
```
dlax: Might be worth adding a reference to the "entry point schema" which is currently defined in section 9.1 (we are in section 9.5.1).

dlax:
@handrews See also https://github.com/handrews/json-schema-spec/pull/4 spotted while reading again these examples. (Now that they are clearer, I pay more attention to details!)

jsonschema-hyperschema.xml
```diff
@@ -2029,104 +2093,72 @@ Link: <https://api.example.com/trees/1/nodes/456> rev=up
...
-                        but how do we recognize it?  We can't use a "collection" link relation
-                        here, because there is no identifiable "thing" to serve as the context
-                        resource.  We could look at the submission schema, notice that it
-                        has a "collection" link and is therefore an item, and notice that
-                        its "collection" link produces the same(-ish?) URI, but that seems
-                        overly complicated.  And gets more so once pagination and filtering
-                        are introduced.
-                    </cref>
+                    There are two "item" links, one for each item in the "elements" array.
+                    Unilke the "self" links, these are defined only in the collection schema.
```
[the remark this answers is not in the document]
handrews: I swear I ran the spell checker!  Repeatedly!  Argh... I need to find an intelligent spell-checker for XML in vim that I can just leave on instead of the convoluted filetype switching that I do to get the regular one to work at all.

jsonschema-hyperschema.xml
handrews: Good call.

handrews:
I'm going to go ahead and merge this b/c of the approval plus I accepted the suggested PR, and the remaining fixes are trivial and build passed.  And more importantly, anyone reading the pre-publication review materials has no doubt been confused.

PR #492, opened 2017-11-16
Typo fixes

dlax:
@handrews You earlier comment about Vim's spellcheck for XML files bugged me and I found that `:syntax spell toplevel` (in addition to `:set spell`) makes spellchecking work on tag's text.


handrews:
Awesome!  I kept meaning to look into it b/c obviously it must be a solved problem but just never got around to it.  I would just switch to text mode and turn on spell check and then turn it off and switch back.  Do the same thing switching to JSON mode to catch syntax errors.  But that's error-prone (and seriously why was I making myself do that anyway?)

Issue #473, opened 2017-11-03
example in overview of hyperschema spec


dlax:
Prior to the rewrite in #427, there was a simple yet illustrative example in the overview of hyperschema specification. See http://json-schema.org/latest/json-schema-hypermedia.html#rfc.section.3

Should we restore it?
I think so and can submit a PR if there's an agreement.

handrews:
I always hated that example.  It's a bunch of stuff that is not explained or motivated and puts the emphasis on the wrong things.  And breaks up the flow way too much.  I would really rather not put it back.

If we must have an example in the overview (and I'm actually pretty strongly agains that) then it needs to have a very clear, concise, and specific purpose.  What is it trying to illustrate?  "Just show a hyper-schema" is not sufficient.  People can jump to the "Examples" section if that's how they want to start.  There has to be a really compelling reason to break up the overview.  Keeping the overview high-level is very important.

handrews:
@dlax I can perhaps see the need to just show what the basic structure of a hyper-schema is in the overview.  I think what I really dislike about the example in draft-06 is how long it is, and how much it clutters hyper-schema with other things.  However, a minimal example such as:

```JSON
    {
       "links": [
           {
               "rel": "self",
               "href": "article/{id}"
           }
        ]
    }
```
(or perhaps with `$schema` and `$id` included to make it a proper schema document?)

could work fine.  This would show the minimum useful hyper-schema (ignoring that `{}` is trivially a valid hyper-schema, which is covered elsewhere I think).

I considered putting something like that in §5 Schema Keywords, but that didn't feel quite right (too much that was not a schema keyword).

But the above example illustrates how to connect a minimally templated link with a standard link relation type to an unconstrained schema, using the instance's request URI as the base URI for `href`.

@dlax would that work for you?

dlax:
@handrews Thank you taking time to answer here and try to move forward.

I'm not sure I get your proposal right, do you mean to include the schema above with only `links` and nothing else (especially no data description like `type`, `properties`, etc)? If so, it'd certainly be minimal but I'm afraid it wouldn't be very understandable and readers would then wonder where `{id}` comes from.

My point about having a simple but demonstrative example upfront is that readers should be able to infer how JSON Hyper-Schema can work by just reading the example and practically without reading the spec (which I think is possible with the previous example).

[...]

PR #494, opened 2017-11-16
Add minimal example to the hyper-schema Overview

handrews:
This should be enough to give a feel for the basics without
needing a detailed explanation or taking up too much space
during the introductory sections.

Addresses #473.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -149,11 +149,9 @@
             </t>
             <figure>
                 <preamble>
-                    The empty schema is a valid hyper-schema, as all schema keywords are optional,
-                    and simply has no effect.  No hyperlinks are applied by default.  The following
-                    fairly minimal non-trivial hyper-schema shows a single link, with the
-                    IANA-registered link relation type "self", that is build from the lone object
-                    field in the instance:
+                    Here is an example hyper-schema that adds a single link, with the
+                    IANA-registered link relation type "self", that is built from an instance
+                    with one known object field named "id":
                 </preamble>
                 <artwork>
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -147,6 +147,38 @@
                 other link serializations or representation formats at runtime, or pre-emptively
                 follow links to facilitate server push usage.
             </t>
+            <figure>
+                <preamble>
+                    The empty schema is a valid hyper-schema, as all schema keywords are optional,
+                    and simply has no effect.  No hyperlinks are applied by default.  The following
```
philsturgeon: Maybe this should be a new paragraph, to show we've moved on from "no links is fine.

Here are some links"

handrews: I went back and forth on that a couple of time.  Let's see what @dlax (or anyone else who wants to jump in) says :-)

Reading your comment I'm leaning towards splitting it, but like I said, back and forth...

dlax: I think I'm missing the purpose of the first two sentences. And I'm not sure I get what you want to split off either :(

Why not simply have the third sentence phrased as "Here is an example of JSON Hyper-Schema with a single link, ..."? No need to mention it's minimal or non-trivial, I think.

(Apart from this, this looks good to me.)

handrews: @dlax I like the simplification idea.  The empty schema is covered elsewhere, and you're right that the non-trivial/minimal language wasn't adding anything important.

I've updated this with a shorter caption now.  @philsturgeon does this look OK to you?

Issue #471, opened 2017-11-03
Audit normative vs informative references


dlax:
[Currently](http://json-schema.org/work-in-progress/WIP-jsonschema-hyperschema.html#rfc.references), RFC8288 on "Web Linking" appears in "Informative References". Is this intended? At a first glance, I would have expected it to be normative.

handrews:
You don't need to *implement* RFC8288 for hyperschema.  We're referencing its philosophical model and terminology.  A lot of the "normative" references should be "informative", actually.

handrews:
@dlax I hope you don't mind but I've generalized this to track an overall audit of the references.  As I noted above, I'm pretty sure it's a jumbled mess.  I'll update this issue with a better explanation before going into a PR.

handrews:
@awwright I notice that all of the standards that are referenced as providing the syntax for values for `format` and `content*` are listed as informative.  Is that correct?  The IESG statement on normative vs informative says:

> Normative references specify documents that must be read to understand or implement the technology in the new RFC, or whose technology must be present for the technology in the new RFC to work.

In some sense, to implement format validation, one would need at least a partial implementation of such standards.

Compare this with listing the Linked Data Platform as a normative reference just because it defines "describedby" as a link relation.  Given that JSON Schema Core defines its own refinement of "describedby", which is itself documented in the IANA registry, is LDP really a normative reference here?

@dlax depending on the answers to these, RFC 8288 may indeed be normative for hyper-schema.  You definitely need to understand its concepts in section 2 (even though you would not re-implement the HTTP `Link` header from section 3).

awwright:
@handrews I would think should be normative references.

My understanding is things like examples ("compare this to how XXX specification works") are informative, references to behavior defined elsewhere need to be normative.

Mere references to values defined elsewhere I suspect would be normative too. It might depend on how it's worded. For example, you don't need to import every single possible value for a link relation, importing Web Linking is sufficient.

handrews:
@awwright sounds good.  So giving CBOR as an example of another media type that could fit the data model is informative, but the essential syntax for a value, or a closed enumerated set, would probably be normative.  An open enumerated set (such as link relations) would only need to reference where the concept and/or registry is defined normatively.

handrews:
Merged #496

PR #496, opened 2017-11-17
Fix normative vs informative references

handrews:
Addresses #471.  Note that no changes were needed in the core spec, and the commented-out RFC reference that was left in hyper-schema will be addressed by the next PR.

I left the HTTP RFCs (including PATCH) as informative because there is guidance on how they are used but I don't think anything rises to the level of a testable requirement, even an optional one.  Implementations, for instance, are not obligated to look for "accept-patch" in "targetHints", and if they do, the general directive that the meta-data is treated according to whatever protocol is in use is sufficient.  Additional information is present only to reduce confusion.

The one way in which this might not be correct is the requirement that API implementors MUST NOT define POST semantics for a collection other than collection item creation semantics.  That hinges on the use of POST.  But it's not actually a requirement for Hyper-Schema implementations, it's a requirement for hyper-schema users.  I'm not sure how that's supposed to work (or if that's improperly written to start with).


awwright:
This looks good at a glance

Issue #485, opened 2017-11-13
equivalent to or a sub-path of the request URI?


handrews:
From the Hyper-Schema Security Considerations:

> When link relation of "self" is used to denote a full representation of an
                    object, the user agent SHOULD NOT consider the representation to be the
                    authoritative representation of the resource denoted by the target URI if
                    the target URI is not equivalent to or a sub-path of the URI used to request
                    the resource representation which contains the target URI with the "self"
                    link.

Does anyone understand the "sub-path" part of this?  It seems a bit related to the old [`pathStart`](https://tools.ietf.org/html/draft-luff-json-hyper-schema-00#section-4.5) keyword.  I can't find anything in RFCs 3986, 4287 (where "self" was first defined), 7230, or 7231 that indicates anything special about URI "sub-path"s, with respect to "self" links or otherwise.

@awwright do you have any idea what this was about?  @Julian?

awwright:
I'm not sure where that came from either. It probably intends to refer to the origin instead, following the Web origin security model.

I would expect the definition of the "self" link relation to give us all the necessary security concerns, though... what about it poses a unique threat to Hyper-schema?

handrews:
> what about it poses a unique threat to Hyper-schema?

No clue.  Should we just remove this subsection?

handrews:
@awwright Reading through this again, I think that the case this is addressing is the one in the collection example (that I just reworked yesterday), where each element in the collection representation is in fact a full representation of the item, and therefore each element has both a "self" link and an "item" link.

The issue seems to be the question of whether you can safely assume that the context representation of those "self" links can be used in place of fetching a representation from the "self" link's target, saving a network operation.

In this case, the sub-path stuff seems to line up.  In that example, the "URI used to request
the resource representation which contains the target URI with the "self"
link" is `/things`, while the target URIs of the "self" links of the two collection elements are `/things/1234` and `/things/6789` (or something like that).

Those are sub-paths of the collection URI, which means that according to this guideline, you can treat the embedded element representation within the collection representation as an authoritative full representation, and skip fetching individual elements.

A corollary of this is that "self" links SHOULD NOT (or even MUST NOT) be used when the context representation is **not** a valid full representation (defined as validating against the full representation's schema, I suppose, which does leave some leeway assuming optional fields).

I see the point of this, but it's explained in a very convoluted way and I have no idea if it is grounded in any larger theory or security framework.

awwright:
[...]

PR #497, opened 2017-11-17
xref for "self" rel type, improve the wording for usage guidance, security concerns

handrews:
For some reason the reference to the Atom RFC, where "self" was
originally defined, was commented out.  Restore the reference
and cite it.

The text for "self" was confusing in that it was unclear which
parts applied in general and which parts applied to only "self"
links for the entire instance document.  Rearrange some sentences
to make that more clear.

Also partially addresses issue #485 by improving CREF text about
the "self" security concerns.


handrews:
The force-push is just a rebase to resolve conflicts with the merged change that re-arranged the normative vs informative reference lists.

handrews:
And also updated with a better paragraph order.

PR #508, opened 2017-11-25
Various hyper-schema example fixes

handrews:
What is it about publishing/releasing something that makes
errors jump out of text that you (and others) went over so
many times during final review?

* "base" URIs need to end in "/" for the "href" values to work,
  although when "base" has no path component it probably works
  either way
* "-wip" got left in the example meta-schema URIs :-(
* A cref about "-wip" being a placeholder got left in, although
  not anywhere near the examples where the "-wip" was otherwise
  apparent
* Tree node example lacked "$id", "$schema", and opening {
* There was no text in between the collection schema and instance

These are small but potentially quite confusing.  I plan to add an errata
list to the web site, but after some time for other obvious errors to be
spotted it might be worth publishing a purely bugfix update as
draft-handrews-json-schema-hyperschema-01 (still with draft-07 meta-schema).

Apparently I did remember to remove all of the "-wip" from the validation spec.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -285,9 +285,6 @@
 
         <section title="Meta-Schemas and Output Schema">
             <t>
-                <cref>The "draft-07-wip" is a placeholder.</cref>
-            </t>
-            <t>
                 The current URI for the JSON Hyper-Schema meta-schema is
                 <eref target="http://json-schema.org/draft-07/hyper-schema#"/>.
             </t>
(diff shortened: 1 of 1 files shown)


handrews:
I think what I'm going to do is merge PRs like this to both the `master` and `draft-07` branches.  Then if we decide that a bugfix draft is needed (meaning no specification changes, only corrections to align with the specification intended to be described by the draft-07 meta-schemas), we can publish that off of the `draft-07` branch.

This would not be all that unusual btw.  The reason that you don't see a draft-zyp-json-schema-00 on the [IETF tracker chart](https://datatracker.ietf.org/doc/draft-zyp-json-schema/) is that it had an error and draft-zyp-json-schema-01 was published later that same day.

PR #509, opened 2017-11-27
Fix several early changelog oddities

handrews:
The core spec's change log seemed to have been spliced across
draft-04 and draft-00.  This dates back to draft-04.  Move the
obviously draft-04-related stuff up to that part of the changelog,
and add a line about splitting out hypermedia (analogous to the
existing line about splitting out validation).

It did not seem worthwhile to import the old changelogs from
draft-01, -02, and -03, although we can do that if necessary.
This change just fixes things that really did not make sense.

Also, in both the validation and hyper-schema specs, the oldest
draft was erroneously numbered -01 when it should be -00.

_**NOTE:** I'll probably also merge this to the draft-07 branch,
since if we do a bugfix this would be fine to include._

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1116,16 +1116,18 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
                     </t>
                     <t hangText="draft-zyp-json-schema-04">
                         <list style="symbols">
-                            <t>Split validation keywords into separate document</t>
+                            <t>Salvaged from draft v3.</t>
+                            <t>Split validation keywords into separate document.</t>
+                            <t>Split hypermedia keywords into separate document.</t>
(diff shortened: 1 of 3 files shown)


handrews:
@awwright it's really not clear.  It's like this in the first commit when you set up this repo.  Looking at the imported commits from the old repo, the tagged commit for that draft shows just the "draft-00" entry in the changelog, and git blame shows it as fge doing it.  Which means the tag is not actually on the commit that was published but I am not even going to start trying to sort that out.  It's close enough.

My guess is that fge was originally going to publish a unified draft-fge-json-schema-00 document, and then they decided to split that into draft-zyp-json-schema-04, draft-fg-json-schema-validation-00, and draft-luff-json-hyper-schema-00.  So viewed that way the arrangement makes sense.  The draft-04 bit was a last minute insertion.  But it doesn't make sense to read, so I think we're better off with this fix applied.

Issue #441, opened 2017-10-14
Keywords minContains/maxContains


epoberezkin:
These keywords will complement the existing keyword `contains` to define minimum and/or maximum number of items that are valid against the schema in `contains` keyword (in the same schema object).

Example:

```json
{
  "$comment": "validate that the array contains >= 2 and <= 4 boolean true values",
  "type": "array",
  "items": {"type": "boolean"},
  "contains": {"const": true},
  "minContains": 2,
  "maxContains": 4
}
```

Meta-schema:

```json
{
  "properties": {
    "minContains": {"type": "number", "minimum": 0, "default": 1},
    "maxContains": {"type": "number"}
  },
  "dependencies": {
    "minContains": ["contains"],
    "maxContains": ["contains"]
  }
}
```

`"minContains": 0` would allow limiting the maximum number of matching elements without requiring one matching element.

handrews:
Removed the [proposal] prefix as I've been removing all of them since we have control over the labels now.  Proposals are labeled with "Enhancement".

handrews:
This came up in https://stackoverflow.com/questions/47348506/json-schema-require-a-specific-array-element

handrews:
This proposal seems pretty reasonable to me.  Any objections?  I'm going to tentatively put it in the draft-08 milestone.  Anyone who feels like submitting a PR is welcome to, otherwise I'll get around to it eventually.

mchastain-turner:
It seems that `minContains` and `maxContains` could allow for schema definitions that would have a count validator. Otherwise you would have to do an `allOf` which seems like it would be unwieldy. 
```
{
  "$comment": "validate that the array has at least one headline and multiple paragraphs",
  "type": "array",
  "contains": {"const": true},
  "minContains": [{"const": "headline", "count":1}, {"const":"paragraph", "count":1}],
  "maxContains": [{"const": "headline", "count":1}]
}
```

Relequestual:
I feel `allOf` would be a more natural way to express the constraint you describe.
Given you can already do this, and it isn't exactly a "shorthand", I don't see any reason to change what's already there to what you describe.

mchastain-turner:
It does seem odd though that `minContains` and `maxContains` are applied as constraints to `contains`.  It seems like you could define the constraints directly in the `contains` schemas:

```
{
  "type": "array",
[...]

PR #525, opened 2017-12-28
Add minContains and maxContains

handrews:
Addresses #441

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -42,7 +42,7 @@
             </address>
         </author>
 
-        <date year="2017"/>
+        <date year="2018"/>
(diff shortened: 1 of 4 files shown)


handrews:
Two-day warning!  If anyone else wants to comment on this you have two days before we hit the two week threshold, at which point I will merge it (since there is one approval already.)

epoberezkin:
@handrews do you think it is worth adding the note about `minContains: 0` that it essentially makes contains doing nothing unless maxContains is present?

handrews:
@epoberezkin thanks, updated.

handrews:
Note that this is temporarily on hold pending decisions on whether to do an interim bugfix release.  It is not abandoned!

PR #535, opened 2018-01-18
Prep for handrews-*-01 (Hyper-Schema and Rel JSON Pointer **ONLY**)

handrews:
This is just for Hyper-Schema and Relative JSON Pointer, both
of which need some bug fixing.  In particular, the hyper-schema
meta-schema link and some aspects of examples were blatantly
wrong and would not work.

***If folks think I should push all three instead, we could clarify
some wording and do that- there's been at least one wording
complaint on if/then/else, for instance.  But I don't want to
hold things up for that, or add new keywords or change behavior.***

This is against "newyear" which is against "master".  Once these
are merged to "master", I will merge "master" to "draft-07" and
tag new releases of these two specs.  The only commits on master
since the handrews-*-00 tags are:

* actual bug fixes to hyperschema (already on draft-07 and master)
* fixing the munged old changelogs (useful to include for -01)
* actual bug fixe to relative json pointer
* update year to 2018
* some README and CONTRIBUTING stuff which is irrelevant to the actual drafts

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -24,7 +24,7 @@
 <?rfc rfcedstyle="yes"?>
 <?rfc comments="yes"?>
 <?rfc inline="yes" ?>
-<rfc category="info" docName="draft-handrews-json-schema-hyperschema-00" ipr="trust200902">
+<rfc category="info" docName="draft-handrews-json-schema-hyperschema-01" ipr="trust200902">
     <front>
         <title abbrev="JSON Hyper-Schema">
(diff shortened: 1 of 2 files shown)


handrews:
No clue why the build is failing- it works locally after a make clean, and the error is a connection time-out, presumably while the system tries to resolve a remote document.  Since it looks like an external failure, I'm just going to ignore it and see if it works tomorrow.

handrews:
... and now the build passes.  We're just going to pretend that whole failure thing didn't happen :-D

Issue #545, opened 2018-02-07
Why do we need a special "Internal References" section?


handrews:
I've lost track of the reason for [Section 9.2.1 Internal References](https://tools.ietf.org/html/draft-handrews-json-schema-00#section-9.2.1).  Is this intended to produce any difference in actual behavior from just relying on the natural behavior of defining a plain name fragment with `$id`?

The only thing that I can see that might have any effect is (emphasis added):

> which is understood as the schema defined elsewhere ***in the same document*** without needing to resolve the fragment against the base URI.

So given this schema, which modifies the 9.2.1 example to have two nested "items", the outer of which sets its own absolute URI `$id`:

```JSON
{
    "$id": "http://example.net/root.json",
    "items": {
        "$id": "http://example.net/other.json",
        "items": {
            "type": "array",
            "items": { "$ref": "#item" }
        }
    },
    "definitions": {
        "single": {
            "$id": "#item",
            "type": "integer"
        }
    }
}
```

If we go by pure RFC 3986 rules, then the `$ref` resolves to `http://example.net/other.json#item` which does not appear to exist, as the fully resolve URI of the schema at `#/definitions/single` is `http://example.net/root.json#item`.

But the "in the same file" wording could be read to mean "forget base URIs, if there's a plain name fragment in the same physical file, you MUST ignore RFC 3986 and resolve to that fragment even if the fully resolved URIs of the `$ref` and `$id` in question are different.

So my questions are:

1.  Is this difference in behavior what is intended?
2.  If so, ***WHY???????***
3.  If we don't have a really, really, really compelling reason for the behavior, can we get rid of it and just rely on the usual interpretation of fragments and URI reference resolution per RFC 3986?  `$id` is hard enough for people to get right without special-casing it.

Paging @awwright @Relequestual @johandorland

handrews:
I think the original motivation for this was [draft-04 §7.2.4 Inline dereferencing and fragments](https://tools.ietf.org/html/draft-zyp-json-schema-04#section-7.2.4), which made a point of saying that resolving plain-name fragments defined by `id` (as `$id` was then called) was an optional behavior (because "inline dereferencing" was said to be optional, while "canonical dereferencing" was required, but why such a fragment wouldn't be usable via canonical dereferencing was never clear to me- it's a fragment, with resolution rules specified by the media type, it should work just fine).

The closest thing to that concept that we've kept is "which is understood... without needing" bit in:

[...]

PR #550, opened 2018-02-28
Remove internal/external reference terminology, expand use cases and examples

handrews:
***NOTE:* This PR is broken into several commits for easier review.  Some commits relocate large blocks of text, others rework and expand existing content.  No commits do both.  It is probably worth looking at individual commits for diffs, but building the branch locally to look at the complete flow of the reworked section to get a feel for how it reads as a whole.*

This addresses #545.

The "Internal References" concept and explanation was confusing, as noted in #545.  The "External References" was really just a discussion of schema identification uniqueness and whether or not you can load remote schemas.

It felt odd to me to introduce `$ref` before `$id`, as you need to understand identifying schemas before you can reference them by identifiers.  So I re-ordered that.

In the `$id` section, made subsections for the different use cases and expanded on the implications, in particular for how JSON Pointer fragments can be used with any base from any parent in the same schema document (this was another aspect brought up in #545 or related discussions).  I expanded the example to show this clearly.

Finally, I switched the order of the external and internal sections, and made them about loading and resolving references, respectively.  The same example is present, but rather than talking about exceptions to RFC 3986 and "in the same document", it phrases things in terms of whether the implementation has seen the identified schema or not.  This matches the discussion in the loading section (formerly external references).

Finally, added a cref note to get us to consider guidance on previously unknown schema identifiers.  In the case of Hyper-Schema in particular, forbidding network dereferencing makes the entire concept unusable with dynamic, evolving hypermedia.  Some further guidance is required (although the actual guidance should be developed separately- I will file an issue for it).

Also paging @johandorland and @spenced in addition to the requested reviewers.

*This may go into a bugfix update before we move to a draft-08 meta-schema, as it is not intended to change any behavior, and definitely does not involve a meta-schema change.*

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -578,7 +578,7 @@
                     <t>
                         Using JSON Pointer fragments requires knowledge of the structure of the schema.
                         When writing schema documents with the intention to provide re-usable
-                        schemas, it is preferable to use a plain name fragment that is not tied to
+                        schemas, it may be preferable to use a plain name fragment that is not tied to
                         any particular structural location.  This allows a subschema to be relocated
                         without requiring JSON Pointer references to be updated.
                     </t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -554,36 +554,54 @@
                     This value SHOULD be normalized, and SHOULD NOT be an empty fragment &lt;#&gt;
                     or an empty string &lt;&gt;.
                 </t>
-                <t>
-                    The root schema of a JSON Schema document SHOULD contain an "$id" keyword with
-                    a URI (containing a scheme).  This URI SHOULD either not have a fragment, or
-                    have one that is an empty string.
-                    <!-- All of the standard meta-schemas use an empty fragment in their id/$id values. -->
-                    <cref>
-                        How should an "$id" URI reference containing a fragment with other components
-                        be interpreted?  There are two cases:  when the other components match
-                        the current base URI and when they change the base URI.
-                    </cref>
-                </t>
-                <t>
-                    To name subschemas in a JSON Schema document,
-                    subschemas can use "$id" to give themselves a document-local identifier.
-                    This is done by setting "$id" to a URI reference consisting
-                    only of a plain name fragment (not a JSON Pointer fragment).
-                    The fragment identifier MUST begin with a letter ([A-Za-z]), followed by
-                    any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons
-                    (":"), or periods (".").
-                </t>
-                <t>
-                    Providing a plain name fragment enables a subschema to be
-                    relocated within a schema without requiring that JSON
-                    Pointer references are updated.
-                </t>
-                <t>
-                    The effect of defining a fragment-only "$id" URI reference that neither
-                    matches the above requirements nor is a valid JSON pointer
-                    is not defined.
-                </t>
+                <section title="Identifying the root schema">
+                    <t>
+                        The root schema of a JSON Schema document SHOULD contain an "$id" keyword with
+                        a URI (containing a scheme).  This URI SHOULD either not have a fragment, or
+                        have one that is an empty string.
+                        <!-- All of the standard meta-schemas use an empty fragment in their id/$id values. -->
+                    </t>
+                </section>
+                <section title="Changing the base URI within a schema file">
+                    <t>
+                        When an "$id" sets the base URI, the object containing that "$id" and all of
+                        its subschemas can be identified by using a JSON Pointer fragment starting
+                        from that location.  This is true even of subschemas that further change the
+                        base URI.  Therefore, a single subschema may be accessible by multiple URIs,
+                        each consisting of base URI declared in the subschema or a parent, along with
+                        a JSON Pointer fragment identifying the path from the schema object that
+                        declares the base to the subschema being identified.  Examples of this are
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -635,33 +630,120 @@
...
+                    </t>
+                </section>
+                <section title="Dereferencing">
                     <t>
                         Schemas can be identified by any URI that has been given to them, including
-                        a JSON Pointer or their URI given directly by "$id".
+                        a JSON Pointer or their URI given directly by "$id".  In all cases,
+                        dereferencing a "$ref" reference involves first resolving its value as a
+                        URI reference against the current base URI per
+                        <xref target="RFC3986">RFC 3985</xref>.
```
philsturgeon: which is it? 3985 or 3986

handrews: Thanks, fixed.

jsonschema-core.xml
```diff
@@ -581,39 +554,61 @@
...
+                    <t>
+                        Using JSON Pointer fragments requires knowledge of the structure of the schema.
+                        When writing schema documents with the intention to provide re-usable
+                        schemas, it is preferable to use a plain name fragment that is not tied to
+                        any particular structural location.  This allows a subschema to be relocated
+                        without requiring JSON Pointer references to be updated.
+                    </t>
+                    <t>
+                        To name subschemas in a JSON Schema document,
+                        subschemas can use "$id" to give themselves a document-local identifier.
```
awwright: Rereading this language I know I wrote, it may be clearer to add a qualifier to '"$id"' like

> To name subschemas in a JSON Schema document,
> subschemas can use a bare fragment for "$id" to give themselves a document-local identifier.

Maybe I can PR for this after merge.

jsonschema-core.xml
```diff
@@ -581,39 +554,61 @@
...
+                        This is done by setting "$id" to a URI reference consisting
+                        only of a plain name fragment (not a JSON Pointer fragment).
+                        The fragment identifier MUST begin with a letter ([A-Za-z]), followed by
```
awwright: Also rereading this language, it might not be clear that the property value as a whole still begins with `#`.

I can PR for this after merge if you like.

jsonschema-core.xml
```diff
@@ -581,39 +554,61 @@
...
+                        any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons
+                        (":"), or periods (".").
+                    </t>
```
awwright: Also I thought I had a comment in here explaining I just copy-pasted the criteria from XML element IDs.

jsonschema-core.xml
```diff
@@ -581,39 +554,61 @@
...
+                        its subschemas can be identified by using a JSON Pointer fragment starting
+                        from that location.  This is true even of subschemas that further change the
+                        base URI.  Therefore, a single subschema may be accessible by multiple URIs,
+                        each consisting of base URI declared in the subschema or a parent, along with
+                        a JSON Pointer fragment identifying the path from the schema object that
+                        declares the base to the subschema being identified.  Examples of this are
+                        shown in section <xref target="idExamples" format="counter"></xref>.
+                    </t>
+                </section>
+                <section title="Location-independent identifiers">
```
awwright: I just realized RFC3986 has a term for what we're doing, "Same-Document Reference"

But it's not _quite_ the same concept, and I think I like this title better. (Nice pick.)

jsonschema-core.xml
```diff
@@ -581,39 +554,61 @@
...
-                    matches the above requirements nor is a valid JSON pointer
-                    is not defined.
-                </t>
-                <t>
-                    For example:
+                <section title="Identifying the root schema">
+                    <t>
+                        The root schema of a JSON Schema document SHOULD contain an "$id" keyword with
+                        a URI (containing a scheme).  This URI SHOULD either not have a fragment, or
+                        have one that is an empty string.
```
awwright: This is probably better suited for a follow-up PR, but this is sort of confusing to parse, maybe

> This URI SHOULD have no fragment, or an empty fragment.

Maybe we simplify the entire paragraph down to:

> The root schema of a JSON Schema document SHOULD contain an "$id" keyword with a base URI [RFC3986] (a full URI with a scheme but no fragment), or a full URI with an empty fragment.

jsonschema-core.xml
```diff
@@ -581,39 +554,61 @@
...
+                        a JSON Pointer fragment identifying the path from the schema object that
+                        declares the base to the subschema being identified.  Examples of this are
+                        shown in section <xref target="idExamples" format="counter"></xref>.
+                    </t>
+                </section>
+                <section title="Location-independent identifiers">
+                    <t>
+                        Using JSON Pointer fragments requires knowledge of the structure of the schema.
+                        When writing schema documents with the intention to provide re-usable
+                        schemas, it is preferable to use a plain name fragment that is not tied to
```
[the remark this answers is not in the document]
handrews: For now I'll use "may be preferable" and we can think about the SHOULD.

handrews:
@awwright thanks!  I'll make the one fix and merge since @philsturgeon approved and you seem generally fine with it.  If you can go ahead and write a PR for the other things soon that would be great- I would like to make a "bugfix" release of core (and validation) like I did for hyper-schema within the next few days to a week.

handrews:
@johandorland and @spenced please feel free to comment further if you spot anything of concern, or perhaps better yet open new issues and we will keep working on this if needed.  Since @awwright has some further ideas I did not want to block on this step.

PR #551, opened 2018-02-28
Core: Improve language for schema references section

awwright:


jsonschema-core.xml
```diff
@@ -583,18 +583,16 @@
...
                     <t>
-                        To name subschemas in a JSON Schema document,
-                        subschemas can use "$id" to give themselves a document-local identifier.
-                        This is done by setting "$id" to a URI reference consisting
-                        only of a plain name fragment (not a JSON Pointer fragment).
-                        The fragment identifier MUST begin with a letter ([A-Za-z]), followed by
-                        any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons
-                        (":"), or periods (".").
+                        To specify such a subschema identifier,
+                        the "$id" keyword is set to a URI reference with a name fragment (not a JSON Pointer fragment).
```
handrews: I was using "plain name" rather than just "name" because that is the terminology used in the W3C fragment best practices document that we cite.  i think we should be consistent with that.

jsonschema-core.xml
```diff
@@ -709,11 +714,14 @@
...
-                        identified by URI. As specified above, this does not necessarily mean
-                        anything is downloaded, but instead JSON Schema implementations SHOULD
-                        understand ahead of time which schemas they will be using, and the URIs that
-                        identify them.
+                        The use of URIs to identify remote schemas does not necessarily mean anything is downloaded,
+                        but instead JSON Schema implementations SHOULD understand ahead of time which schemas they will be using,
+                        and the URIs that identify them.
+                    </t>
+                    <t>
+                        When schemas are downloaded,
```
handrews: I don't follow this paragraph.  Taking the dependent clause out I read "When schemas are downloaded... see Usage over HTTP" which is a bit unclear.

Could this be rephrased as something like:

"For use cases when it is desirable to download schemas, such as a hypermedia system where schemas are discovered at runtime, see <xref>Usage for Hypermedia</xref>"

Note I'm proposing linking to section 11 as a whole rather than 11.3, because the actual linking of a downloadable schema is covered in 11.1, and I think that is needed to fully understand the issue (along with 11.3).

awwright: Oh, I totally missed that. Yes, I was looking for that section.

PR #554, opened 2018-03-02
Fix missing "if present" for "else"

handrews:
Also other minor wording clarifications for if/then/else such as
standardizing the "regardless of the validation outcome against
the subschema" language.  That language is important as it
clarifies that subschema validation can fail (causing annotations
to be dropped) without causing the containing schema to fail.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -746,9 +746,22 @@
                     branch of an "allOf" MUST NOT have an impact on a "then"
                     or "else" in another branch.
                 </t>
+                <t>
+                    There is no default behavior for any of these keywords
+                    when they are not present.  In particular, they MUST NOT
+                    be treated as if present with an empty schema, and when
+                    "if" is not present, both "then" and "else" MUST be
+                    entirely ignored.
+                </t>
                 <section title="if">
                     <t>
                         This keyword's value MUST be a valid JSON Schema.
+                    </t>
+                    <t>
+                        This validation outcome of this keyword's subschema
+                        has no direct effect on the overall validation
+                        result.  Rather, it controls which of the "then"
+                        or "else" keywords are evaluated.
                     </t>
                     <t>
                         Instances that successfully validate against this
@@ -763,9 +776,10 @@
                         present.
                     </t>
                     <t>
-                        Validation of the instance against this keyword
-                        on its own always succeeds, regardless of the
-                        validation outcome of the instance against its subschema.
+                        If <xref target="annotations">annotations</xref>
+                        are being collected, they are collected from this
+                        keyword's subschema in the usual way, including when
+                        the keyword is present without either "then" or "else".
                     </t>
                 </section>
                 <section title="then">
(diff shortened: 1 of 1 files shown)


jsonschema-validation.xml
```diff
@@ -799,8 +801,9 @@
                     <t>
                         When "if" is absent, or the instance successfully
                         validates against its subschema, validation against
-                        this keyword always succeeds.  Implementations
-                        SHOULD avoid attempting to validate against
+                        this keyword always succeeds, regardless of the
```
Relequestual: I know this isn't your change, but I find the phrasing...
"When "if" is absent, or the instance successfully validates against its subschema, validation against this keyword always succeeds..."

This reads to me as, when the subschema of `if` successfully validates, `else` always succeeds [validation], regardless of the actual validation result of the instance against it's subschema.

That can't be the intent...

jsonschema-validation.xml
```diff
@@ -780,8 +781,9 @@
                     <t>
                         When "if" is absent, or the instance fails to
                         validate against its subschema, validation against
-                        this keyword always succeeds.  Implementations
-                        SHOULD avoid attempting to validate against
+                        this keyword always succeeds, regardless of the
```
Relequestual: See comment below

jsonschema-validation.xml
handrews: Great feedback, I'll play around with this some more.
The main point of bug-fixing the validation spec is to clarify `if`/`then`/`else` so please highlight everything that's awkward.

handrews:
Rewrote "if" etc. to be more explicit about how the keywords 
interact with each other and with both validation and annotation
collection.

In particular, an "if" on its own collects annotations but 
never affects validation, annotations are only collected from
"them" or "else" if their use is indicated by the validation
outcome of an adjacent "if", and there is no default behavior
for any of the keywords.

Technically, "then" could be assigned an empty schema default
without causing problems, but that felt weirdly unbalanced.

If "if" has a default of true/empty schema, then "then" on
its own would be active, which is counter-intuitive.  If "else"
had a default it would need to be false rather than true, and 
saying it has no default seemed simpler, and matches how "not"
is handled.  "not" is the only other keyword that would otherwise
need a false schema as a default value.

Note that the meta-schema always puts in a default, but this was
already true (and incorrect) for "not", so I'm not worrying about it
now.  We can fix it in draft-08 if necessary.

jsonschema-validation.xml
```diff
@@ -746,10 +746,23 @@
                     branch of an "allOf" MUST NOT have an impact on a "then"
                     or "else" in another branch.
                 </t>
+                <t>
+                    There is no default behavior for any of these keywords
+                    when they are not present.  In particular, they MUST NOT
+                    be treated as if present with an empty schema, and when
+                    "if" is not present, both "then" and "else" MUST be
```
philsturgeon: :+1:

jsonschema-validation.xml
```diff
@@ -746,10 +746,23 @@
...
+                </t>
                 <section title="if">
                     <t>
                         This keyword's value MUST be a valid JSON Schema.
                     </t>
+                    <t>
+                        This validation outcome of this keyword's subschema
+                        has no direct effect on the overall validation
+                        result.  Rather, it controls which of the "then"
+                        or "else" keywords are evaluated.
```
Relequestual: /evaluated/are applicable/ ? Thinking we should use applicability termonology where possible.

PR #555, opened 2018-03-03
handrews-*-01 bugfix for core & validation

handrews:
This makes the draft numbering change and adds the change log for doing a clarification / bugfix draft of core and validation.  The changes include:

* Clarifications on `$id` and `$ref` (previously merged: #550 and #551)
* Clarifications/bugfix on `if`/`then`/`else` (#554)
* Minor list formatting improvements (first commit of this PR)
* Updates to CREF for proposals regarding `dependencies` (second commit of this PR)

The usefulness of the `$id`/`$ref` clarifications in core should be readily apparent.

For validation, the `if` change seems worth publishing because a.) I've received two separate complaints/questions about it, and b.) the validation spec would otherwise be the only one still on handrews-00 so we might as well publish an update even though it's very small.  It does address a real-world point of confusion.

As with the draft-*-handrews-01 for hyper-schema and relative JSON pointer, there are no functional changes here, and no changes to the meta-schema.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -19,7 +19,7 @@
 <?rfc rfcedstyle="yes"?>
 <?rfc comments="yes"?>
 <?rfc inline="yes" ?>
-<rfc category="info" docName="draft-handrews-json-schema-00" ipr="trust200902">
+<rfc category="info" docName="draft-handrews-json-schema-01" ipr="trust200902">
     <front>
         <title abbrev="JSON Schema">JSON Schema: A Media Type for Describing JSON Documents</title>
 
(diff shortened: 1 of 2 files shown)


awwright:
Oh I'm surprised I didn't see this.

This is mostly fine except afaict the colon isn't supposed to be provided in labels. I think that's just the style of RFCs, maybe there's an argument to affect the display? Doesn't matter too much. The RFC Editor can yell at us once we that far.

handrews:
@awwright ugh it's confusingly unreadable without them in most cases- it just looks like badly done spacing.  We can take them out when we have to.

PR #573, opened 2018-03-23
Split "requiredDependencies" out of "dependencies"

handrews:
"dependencies", with the schema behavior, moves to the validation
spec with the other applicators.  "requiredDependencies", with
the property array behavior, stays in validation as an assertion
alongside "required"

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -452,15 +452,19 @@
 
                 <section title="requiredDependencies">
                     <t>
-                        This keyword specifies property requirements that are conditional
-                        based on the presence of other properties.
+                        The value of this keyword MUST be an object.  Properties in
+                        this object, if any, MUST be arrays.  Elements in each array,
+                        if any, MUST be strings, and MUST be unique.
                     </t>
                     <t>
-                        This keyword's value MUST be an object. Each property specifies
-                        a dependent requirement, which MUST be an array.  Each element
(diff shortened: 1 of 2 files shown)


jsonschema-validation.xml
```diff
@@ -454,36 +450,17 @@
...
-                            an array of property names rather than a subschema getting a new
-                            name.  The dual behavior is confusing and relatively difficult to
-                            implement.  In the previous draft, we proposed dropping the keyword
-                            altogether, or dropping one of its forms, but we received feedback
-                            in support of keeping it.  See issues #442 and #528 at
-                            &lt;https://github.com/json-schema-org/json-schema-spec/issues&gt;
-                            for further discussion.  Further feedback is encouraged.
-                        </cref>
+                        This keyword specifies property requirements that are conditional
+                        based on the presence of other properties.
```
dlax: I'm not sure this sentence is useful considering the one that follows (in fact, I cannot easily read it (but maybe it's because I'm not native English)). Also, most keyword's description start out by "The value of this keyword MUST be ..." so this one would be (is already) an exception with this sentence. Not a big deal, though :)

handrews: Good points, let me see if I can rework it to be more consistent and clear.

handrews:
@dlax how does this version look?

dlax:
Looks good :+1:

Issue #558, opened 2018-03-06
Recursive schema composition


handrews:
## TL;DR:

Re-using recursive schemas is a challenge.  

* `$recurse` is a specialized version of `$ref` with a context-dependent target
* The target is the root schema of the document where schema processing began
* Processing can be either static schema walking or dynamic evaluation with an instance
* The value of `$recurse` is always `true` (discussed in the "alternatives" section)
* This is based on a keyword we have long used in [Doca](https://github.com/cloudflare/doca)

## Example

_APPARENTLY MANDATORY DISCLAIMER: This is a minimal contrived example, please do not point out all of the ways in which it is unrealistic or fails to be a convincing use case because you can refactor it.  It's just showing the mechanism._

foo-schema:
```JSON
{
    "$id": "http://example.com/foo-schema",
    "properties": {
        "type": "object",
        "foo": {"$recurse": true}
    }
}
```
bar-schema:
```JSON
{
    "$id": "http://example.com/bar-schema",
    "allOf": [{"$ref": "http://example.com/foo"}],
    "required": ["bar"],
    "properties": {"bar": {"type": "boolean"}}
}
```

The instance:
```JSON
{
    "bar": true,
    "foo": {
        "bar": false,
        "foo": {
            "foo": {}
        }
    }
}
```
is valid against the first schema, but not the second.

It is valid against foo-schema because the `"$recurse": true` is in foo-schema, which is the same document that we started processing.  Therefore it behaves exactly like `"$ref": "#"`.  The recursive "foo" works as you'd expect with `"$ref": "#"`, and foo-schema doesn't care about "bar" being there (additional properties are not forbidden).

However, it is not valid against bar-schema because in that case, the `"$recurse": true` in foo-schema behaves like `"$ref": "http://example.com/bar-schema"`, as bar-schema is the document that we started processing.  Taking this step by step from the top down:

* Processing the root of the instance, we have the "bar" property required by bar-schema; we got this directly from the root schema of bar-schema, without `$recurse` being involved
* Looking inside "foo", processing follows the `allOf` and `$ref` to foo-schema.  The top-level instance is an object, so we pass the `type` constraint
* Still processing foo-schema, for the contents of the "foo" property, we have `"$recurse": true.  Since we started processing with bar-schema, this is the equivalent of `"$ref": "bar-schema"
[...]

PR #589, opened 2018-05-24
Add "$recurse" for extending recursive schemas

handrews:
Addresses #558.

The 3rd and 2nd to last two commits rework `$ref` and `$recurse` as adjacent subsections in the "Schema Referencing" section, as most of their behavior is identical (particularly once #585 is complete).

The third-to-last commit just moves things around with no content changes other than obvious section titles, while the second-to-last commit updates the text, including minor updates to `$ref`'s subsection.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -571,9 +571,10 @@
                 </t>
                 <t>
                     Authors of extensions to JSON Schema are encouraged to write their own
-                    meta-schemas, which extend the existing meta-schemas using "allOf".
+                    meta-schemas, which MAY extend the existing meta-schemas using "allOf".
                     This extended meta-schema SHOULD be referenced using the "$schema" keyword, to
-                    allow tools to follow the correct behaviour.
+                    allow tools to follow the correct behaviour.  The "$recurse" keyword is
+                    provided to facilitate this usage.
                 </t>
                 <t>
                     Note that the recursive nature of meta-schemas requires re-defining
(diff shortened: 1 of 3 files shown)


jsonschema-core.xml
```diff
@@ -904,6 +905,86 @@
...
+                        root, and a regular fragment URI reference does not provide the
+                        correct semantics.  Should other values be added in the future,
+                        it is expected that a boolean true value will remain an alias for
+                        this original use case.
+                    </cref>
+                </t>
+                <t>
+                    The presence of this keyword with a boolean true value indicates that,
+                    during processing, it MUST be treated as a reference to the schema document
+                    where processing was initiated.
```
philsturgeon: > it MUST be treated as a reference to the schema document where processing was initiated.

vs 

>  Note that even if processing began at a subschema within a document, the "$recurse" target MUST be the root of the document.

The first seems like its saying if we ref into a document at some lower level, it should recurse to THAT level, but the 2nd says it must be the root? Confusing.

jsonschema-core.xml
```diff
@@ -904,6 +905,86 @@
...
+            "$ref": "#",
+        },
+        "recursive": {
+            "$recurse": true
+        }
+    }
+}
+
+{
+    "$schema": "
```
philsturgeon: Oops?

handrews: er... yeah, thanks!

jsonschema-core.xml
handrews: Yeah, the key here is the wording "schema document" which is defined in [section 4.3](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-4.3) in both draft-07 and the current HEAD commit.  A schema document is a document, not a JSON object.  So saying that it _MUST be treated as a reference to the **schema document** where processing was initiated_ means exactly what the clarification says.

But I'm open to suggestion on how to word it better, preferably without entirely re-stating 4.3.  Maybe just an `<xref />` to it?  I suspect the complete list of people who understand the distinction without having to look it consists of is @awwright and me.

awwright:
How would I refer to extended schemas from schemas of my own?

For example, suppose I have a media type that tries to represent HTML forms. It's an array of objects, each describing a form field. One of the properties is a schema that can describe a schema, and I want to 

```javascript
{
  type: "array",
  items: {
    type: "object",
    properties: {
      "name": { type: "string" },
      "schema": { $ref: "http://json-schema.org/draft-nnn/hyper-schema" }
    }
  }
}
```

where `http://json-schema.org/draft-nnn/hyper-schema` is some future schema for Hyper-schema that uses this feature.

What happens when a JSON Schema validator hits `$recurse: true`?

handrews:
@awwright that is an excellent question, and definitely a problem for this proposal as it stands.

Embedding the meta-schema is a different use case than "extending" a schema (I hate to use the word "extending" as it makes people think of OOP, but JSON Schema is a constraint system which is very different- not sure what else to call it, though).  But there is no way in this proposal to distinguish the two, so your example would break.  And in a non-obvious way.

The simplest fix would be to add something alongside the `$ref` that controls whether recursion takes the containing schema into account.  This draws from your other idea in #558 of changing how a specific `$ref` is involved.  I'm worried about the added complexity, but `$recurse` as proposed here can't work so we need some sort of fix.

The thing I want to preserve about `$recurse` is that it's target is dynamic, while `$ref`'s target is static.  I'm open to any proposal that keeps that distinction.

handrews:
I'm going to move this discussion back to the issue and make some revised proposals there.

handrews:
Closing per https://github.com/json-schema-org/json-schema-spec/issues/558#issuecomment-396015607 due to the problems noted by @awwright.  I hope to revisit this, but as noted in the issue, it will be a while and we can get by without it if we must.

PR #595, opened 2018-05-26
Create an Applicators section, group into a Keyword Behaviors section

handrews:
_**NOTE:** The second commit in this PR indents for a new section without changing the text, so it's best to look at each commit separately._

This is groundwork for more of the vocabulary and annotation
collection process work.

Move the general description of applicators from the applicator
vocabulary to a section alongside the Assertion and Annotation
sections.  Rework the text a bit, and also clarify an aspect
of annotation collection related to applicators.

Group the Annotation, Assertion, and Application sections
as Keyword Behaviors.

Arguably, this has overgrown the "Overview" section, but since
the "Vocabulary" part of the overview is also about to be
expanded, let's wait until then to figure out how to refactor it.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -126,10 +126,24 @@
             <section title="Keyword Behaviors">
                 <t>
                     JSON Schema keywords fall into several general behavior categories.
-                    Assertions validate that an instance satisfies constraints, annotations
-                    attach information that applications may use in any way they see fit,
-                    and applicators allow for building more complex schemas than a single
-                    schema object would allow.
+                    Assertions validate that an instance satisfies constraints.  Annotations
+                    attach information that applications may use in any way they see fit.
+                    Applicators apply subschemas to parts of the instance and combine
+                    their results.
+                </t>
+                <t>
+                    Evaluating an instance against a schema involves processing all of the
+                    keywords in the schema against the appropriate locations within the instance.
+                    Typically, applicator keywords are processed until a schema object with no
+                    applicators (and therefore no subschemas) is reached.  The appropriate
+                    location in the instance is evaluated against the assertion and
+                    annotation keywords in the schema object, and their results are gathered
+                    into the parent schema according to the rules of the applicator.
+                </t>
+                <t>
+                    Evaluation of a parent schema object can complete once all of its
+                    subschemas have been evaluated, although in some circumstance evaluation
+                    may be short-circuited due to assertion results.
                 </t>
                 <t>
                     Extension keywords SHOULD stay within these categories, keeping in mind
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -123,63 +123,109 @@
...
             </t>
 
-            <section title="Assertions">
+            <section title="Keyword Behaviors">
                 <t>
-                    JSON Schema can be used to assert constraints on a JSON document, which
-                    either passes or fails the assertions.  This approach can be used to validate
-                    conformance with the constraints, or document what is needed to satisfy them.
+                    JSON Schema keywords fall into several general behavior categories.
+                    Assertions validate that an instance satisfies constraints, annotations
```
awwright: I'd break this into three sentences.

jsonschema-core.xml
```diff
@@ -123,63 +123,109 @@
...
+                    attach information that applications may use in any way they see fit,
+                    and applicators allow for building more complex schemas than a single
```
awwright: Would it be accurate to say something like "Applicators allow a sub-schema to be applied recursively to some part of the instance, or the entire instance." ?
The current description wasn't clear to me.

jsonschema-core.xml
```diff
@@ -123,63 +123,109 @@
...
-                    These default behaviors MUST NOT result in assertion failures.
-                </t>
-                <section title="Assertions and Instance Primitive Types">
+                <section title="Applicators" anchor="applicators">
                     <t>
-                        Most assertions only constrain values within a certain
-                        primitive type.  When the type of the instance is not of the type
-                        targeted by the keyword, the instance is considered to conform
-                        to the assertion.
+                        Evaluation of an instance against a
```
awwright: Is there a difference between "evaluation" and "application"?

awwright:
Also, you can now change the diff settings within GitHub to ignore whitespace \o/

handrews:
> Also this section would be a good place to remark on the existence of "argument keywords" (for lack of a better term), properties in the schema that do nothing by themselves, but are used by other keywords, or affect the behavior of other keywords. The old boolean form of "exclusiveMinimum"/"exclusiveMaximum" would be the best example of this.

Take a look at PR #600.  That kind of effect can be defined in terms of annotation behavior, which keeps it all within a simpler theoretical framework.

I also think that modifiers like the boolean `exclusive*` approach are a really bad idea and don't want to encourage them by codifying such an approach.  After being phrased in terms of annotations, the `additional*` keywords stop seeming like such weird exceptions.  Although, for performance reasons, I expect they will often be implemented as weird exceptions.

jsonschema-core.xml
handrews: Throughout these PRs (and earlier changes) I classify applicators as "in-place" or "child" applicators.  I would not say "the entire instance" because people will take that to mean the whole instance document, which may be a much larger chunk of JSON than the current location.

Just saying "Applicators allow a sub-schema to be applied to some part of the instance." might be fine, though?  We'd just leave the enumeration of the different locations to the more detailed explanation.

jsonschema-core.xml
handrews: Hmm... yes that's unclear.  There is a difference, I'll add some text in the Keyword Behaviors intro section about it.

jsonschema-core.xml
handrews: Updated- I went with sentences for now.

handrews:
@awwright could you see if my changes address your concerns and let me know?  A lot of other work is stacked up on this PR so it would be great to get it merged.

Issue #557, opened 2018-03-05
"unevaluatedItems" for arrays


handrews:
Issue #556 proposes an `unevaluatedProperties` keyword, together with the mechanisms for implementing it.  _This issue will **only** be considered if #556 is accepted._

Those same mechanisms (which should be debated in #556, and **not** here) could be used to define an `unevaluatedItems` keyword, based on an **"evaluatedItems"** annotation that is updated whenever "items" or "additionalItems" are successfully applied to an index.

Successful application of schema-form `items` or `additionalItems` (alongside the relevant array-form "items")  would indicate that all array elements have been evaluated, and the "evaluatedItems" annotation ought to be able to convey that concisely.

`additionalItems` is used for things like positional function arguments where the first N arguments have specific types or structures, and an indefinite number of further positional arguments are allowed, all with the same type/structure.  It's not exactly a common use case, but it is clear and has obvious parallels in numerous programing languages.

`unevaluatedItems` would be used in some sort of situation where, due to a `oneOf` or similar construct you will only know at runtime how long the prefix of individually typed elements will be.

```JSON
{
  "type": "array",
  "oneOf": [
    {
      "items": [{"type": "object"}]
    },
    {
      "items": [{"type": "number"}, {"type": "boolean"}]
    }
  ],
  "unevaluatedItems": {"type": "string"}
}
```

As with `unevaluatedProperties`, `unevaluatedItems` is mainly useful in very large schemas, or schemas assembled out of many pieces.  It is obviously trivial to refactor the schema above to use `additionalProperties` instead.

We don't have a lot of data on `additionalItems` and how (or how often) it is used, so it's hard to say how useful `unevaluatedItems` would be.  But, once you're already implementing `unevaluatedProperties` it's not hard to add this, and the symmetry is appealing.

It would be interesting to find an example of schemas for systems involving complex sequential data.

handrews:
Done with #656

Issue #556, opened 2018-03-04
"unevaluatedProperties" to facilitate re-use and better schema organization


handrews:
## TL;DR

`unevaluatedProperties` is like `additionalProperties`, except that it can "see through" `$ref` and "see inside" `allOf`, `anyOf`, `oneOf`, `if`, `then`, `else`, and whatever else I'm forgetting right now.

This is a problem that becomes important at large scale.  In small examples, you can easily refactor your way out of it, but it becomes very cumbersome at scale.  This is why there's a lot of discussion about OpenAPI's schema for their spec.

Figuring out how to make this work turns out to be complicated, which accounts for the length of the rest of the issue.

-----

_This is a distillation of the results of 230+ comments on #515, not to mention the 300+ comments spread across several other older issues that fed into that one.  I know it's long.  Please don't complain unless you can offer a shorter write-up. :-)_

-----

_**CRITICALLY IMPORTANT NOTE:** This issue is for discussing the use case given in the next section, and the `unevaluatedProperties` proposal to solve it.  If you want to discuss a different use case or a different proposal, you **MUST file your own issue** to do so.  Any comments attempting to revive other lines of discussion from #515, introduce new problems or solutions, or otherwise derail this discussion will be deleted to keep the focus clear.  Please file a new issue and link back to this one instead._

## The use case

Many JSON Schema authors want validation to fail if any properties that are not explicitly matched to a successfully validating schema object are present.

An example is the [OpenAPI Specification (OAS)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md), which uses a regular expression, `^x-`, for all extension keywords.  Any keyword that is neither a standard keyword nor begins with "x-" is in violation of the specification and should cause validation of the OpenAPI document to fail.

OAS has many conditional keywords or allowed values.  The most complex scenario is the [Parameter Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#parameterObject), which can describe parameters in four different values for "in" (path, query, header, or cookie), each of which works with a different subset of another enumerated field, "style", and in the case of "path", restricts the behavior of "required".  Additionally, the object can use one of several mutually exclusive schema and example approaches to describe the parameter.

[...]

PR #600, opened 2018-05-28
Keywords can depend on subschema or adjacent keyword annotation results

handrews:
## Keywords can depend on subschema or adjacent keyword annotation results

_The base of this PR is PR #595, not the current master_

This is the critical conceptual change that lets us implement #556 `unevaluatedProperties` and #557 `unevaluatedItems`.

There are three commits- two significant one and one mechanical one removing outdated text.  These are grouped together to ensure that a real-world use of the new concepts, with the familiar `additionalProperties` and `additionalItems` keywords, is included with the more abstract changes.

The [first commit](https://github.com/json-schema-org/json-schema-spec/commit/b0c208b378e3f540e215dfe19bf152cf7ecf1062) describes the newly allowed behavior and its implications for all types of keywords.  In particular, this required working out how default keyword behavior works with respect to annotations.

The [second commit](https://github.com/json-schema-org/json-schema-spec/commit/c80954647441fe24289a75e612da7e6b40d39bf5) just removes outdated boilerplate that I meant to remove when I moved the applicators from validation to core.  Now that all applicators are grouped into a vocabulary, we don't need to re-explain that concept on each keyword.

The [third commit](https://github.com/json-schema-org/json-schema-spec/commit/0942a62fc5e585d6754421e9e06e5e65951ea245) reworks the `*item` and `*properties` keywords to use the new approach of depending on annotation behavior.  Implementations are **not** required to code this approach exactly, as the current ways of implementing these keywords are faster (and implementations that do not opt-in to annotation collection still need to support these keywords!)

`items`, `properties`, and `patternProperties` now produce annotation results indicating what indexes / property names they cover, and `additionalItems` and `additionalProperties` now define their behavior in terms of those annotation results.  They, too, produce annotation results about what they cover- those results will be used by `unevaluatedItems` and `unevaluatedProperties`.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1264,9 +1264,10 @@
                             index of the instance, such as when "items" is a schema.
                         </t>
                         <t>
-                            Annotation results from multiple "items" keyword are combined
+                            Annotation results for "items" keywords from multiple
+                            schemas applied to the same instance location are combined
                             by setting the combined result to true if any of the values
-                            are true, and otherwise retaining the larges numerical value.
+                            are true, and otherwise retaining the largest numerical value.
                         </t>
                         <t>
                             Omitting this keyword has the same assertion behavior as
@@ -1293,14 +1294,17 @@
                             If the "additionalItems" subschema is applied to any
                             positions within the instance array, it produces an
                             annotation result of boolean true, analogous to the
-                            single schema behavior of "items".
+                            single schema behavior of "items".  If any "additionalItems"
+                            keyword from any subschema applied to the same instance
+                            location produces an annotation value of true, then
+                            the combined result from these keywords is also true.
                         </t>
                         <t>
                             Omitting this keyword has the same assertion behavior as
                             an empty schema.
                         </t>
                         <t>
-                            Implementation MAY choose to implement or optimize this keyword
+                            Implementations MAY choose to implement or optimize this keyword
                             in another way that produces the same effect, such as by directly
                             checking for the presence and size of an "items" array.
                             Implementations that do not support annotation collection MUST do so.
@@ -1341,8 +1345,9 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names matched by this keyword.  Multiple annotation
-                            results for "properties" are combined by taking the union
+                            property names matched by this keyword.  Annotation results
+                            for "properties" keywords from multiple schemas applied to
+                            the same instance location are combined by taking the union
                             of the sets.
                         </t>
                         <t>
@@ -1366,8 +1371,9 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names matched by this keyword.  Multiple annotation
-                            results for "patternProperties" are combined by taking the union
+                            property names matched by this keyword.  Annotation results
+                            for "patternProperties" keywords from multiple schemas applied to
+                            the same instance location are combined by taking the union
                             of the sets.
                         </t>
                         <t>
@@ -1395,6 +1401,9 @@
                         <t>
                             The annotation result of this keyword is the set of instance
                             property names validated by this keyword's subschema.
+                            Annotation results for "additionalProperties" keywords from
+                            multiple schemas applied to the same instance location are combined
+                            by taking the union of the sets.
                         </t>
                         <t>
                             Omitting this keyword has the same assertion behavior as
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -126,8 +126,9 @@
             <section title="Keyword Behaviors">
                 <t>
                     JSON Schema keywords fall into several general behavior categories.
-                    Assertions validate that an instance satisfies constraints.  Annotations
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -1240,26 +1243,52 @@
                             same position, if any.
                         </t>
                         <t>
-                            Omitting this keyword has the same behavior as an empty schema.
+                            This keyword produces an annotation value which is the largest
+                            index to which this keyword applied a subschema.  The value
```
dlax: Why is the result the index of the last applied item rather than the list of indices of applied items? This looks unsymmetrical with how it works for "properties". (If this something that got discussed before, I probably missed it, so any pointer or clarification would be welcome.)

jsonschema-core.xml
```diff
@@ -1240,26 +1243,52 @@
...
+                            MAY be a boolean true if a subschema was applied to every
+                            index of the instance, such as when "items" is a schema.
+                        </t>
+                        <t>
+                            Annotation results from multiple "items" keyword are combined
```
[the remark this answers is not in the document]
handrews: Yes, need to improve the language here, thanks.

What's a good shorthand for "several X keywords present in elements of `allOf`/`anyOf`/`if`/etc."?

```JSON
{
  "anyOf": [
    {
      "items": [{"type": "integer", "multipleOf": 2}, {...}, {...}]
    },
    {
      "items": [{"type": "integer", "multipleOf": 3}, {...}, {...}, {...}]
    },
    {
      "items": {"type": "string"}
    }
  ]
}
```
The rolled-up annotation from "items" might end up as 3 (if it's a multiple of 2 but not of 3) or 4 (if it's a multiple of 3, regardless of whether it's a multiple of 2).  Or `true` if the first element is a string.  Or there wouldn't be an "items" annotation if the first element is neither a string nor an integer that is divisible by either 2 or 3.

Does that help?

jsonschema-core.xml
handrews: Because it's not possible to have gaps in the list.  There could be if `contains` were used, but as noted in that section, `contains` is not used.  And if it were, then its annotation would be a list of indices, but it still wouldn't be possible for `items` to ever produce a gap, so why bother?  Particularly if the list is very long.

jsonschema-core.xml
```diff
@@ -1240,26 +1243,52 @@
...
+                            by setting the combined result to true if any of the values
+                            are true, and otherwise retaining the larges numerical value.
```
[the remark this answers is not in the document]
handrews: I swear I run a spell-checker...
thanks

jsonschema-core.xml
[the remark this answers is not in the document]
dlax: > What's a good shorthand for "several X keywords present in elements of allOf/anyOf/if/etc."?

Maybe "several X keywords collected by application of a combination keyword", where "combination keyword" would link to section [Keywords for Applying Subschemas in Place](https://github.com/json-schema-org/json-schema-spec/blob/3a517363ebe16441f4c0ba74a0737b742aa76132/jsonschema-core.xml#L1008)? Just picked-up the "combination" term, as I think we don't have a word yet for this group, maybe there's a better one...

> Does that help?

Yes, thanks.

jsonschema-core.xml
dlax: Ok, makes sense, thanks!

handrews:
> I don't think I see anything wrong with this, but I'm trying to think if there's a way this can be made more compact.

Good luck and let me know if you can think of anything.  This is where I got after editing stuff down a lot, but I'm sure it could be improved.

jsonschema-core.xml
```diff
@@ -1279,42 +1343,54 @@
...
-                            Omitting this keyword has the same behavior as an empty schema.
+                            The annotation result of this keyword is the set of instance
+                            property names validated by this keyword's subschema.
+                        </t>
+                        <t>
+                            Omitting this keyword has the same assertion behavior as
+                            an empty schema.
+                        </t>
+                        <t>
+                            Implementation MAY choose to implement or optimize this keyword
```
handrews: Note to self: typo -"Implementations"

handrews:
@awwright I should note that I expect to do a big editing pass towards the end of the draft work.  Once we get all of the new material into the core spec, the optimal organization will be more clear.  And that will help me edit the wording down.

Right now I'm not worrying too much about it, which is why "Overview" has become rather enormous.  But reorganizing and condensing things is not something that I can do piecemeal, I kind of need all of the content to be present before I can really see the patterns and use them to reduce the text.

So while I'm definitely interested in improvements, I'd rather not hold this up just for verbosity.  We can file an issue to track doing that editing pass to make sure I don't forget, if you'd like.

handrews:
OK, @dlax I tried to improve the wording around the multiple annotations per keyword thing.

@awwright that made it even more verbose, but as noted I'd like to settle for clarity at this stage and then focus on brevity later.  I suspect that as we fill out and organize the front matter, some of the things that currently need to get re-explained for each keyword will be more clear overall, so we should be able to trim the per-keyword text.

Issue #594, opened 2018-05-26
Expand "contentEncoding" to include RFC 4648


handrews:
[RFC 4648](https://tools.ietf.org/html/rfc4648) defines several encodings, including "base64url", which should be usable with `contentEncoding`.  Currently, that keyword only references [RFC 2045 §6.1](https://tools.ietf.org/html/rfc2045#section-6.1), which (with some persistence) leads to the [IANA registry for Transfer Encodings](https://www.iana.org/assignments/transfer-encodings/transfer-encodings.xhtml).

Notably, this registry only includes the 2045 encodings and not the 4648 ones, or any others.  It is controlled by [RFC 4289](https://tools.ietf.org/html/rfc4289), which is specifically about MIME transfer encodings.

From what I can make out of 4648, it's arguably more relevant to modern web APIs.  Also, it explicitly sets out to make "base64" more consistent and usable, so we probably want 4648's "base64" over 2045's.  I think?

handrews:
Merged

PR #604, opened 2018-06-13
Allow RFC 4648 encodings for "contentEncoding"

handrews:
Addresses #594 

[RFC 4648](https://tools.ietf.org/html/rfc4648) is a more modern standard for encodings, and addresses
ambiguities and application-specific limits of the "base64"
encoding from [RFC 2045](https://tools.ietf.org/html/rfc2045).  It also adds several commonly used
variations, such as "base64url", that are relevant to modern systems.

RFC 2045 values are still allowed, although for "base64", RFC 4648
now takes precedence.


handrews:
@awwright @philsturgeon @Relequestual last call on this- I'll be merging it on Wednesday given @dlax's approval barring serious objections from anyone else.

PR #609, opened 2018-06-17
Explain referenced/referring schemas

handrews:
For several upcoming features, we need better language to
talk about the schemas on either side of a "$ref" or similar
linking keyword.

In particular, annotation collection (#530) interacts with references,
and calling referenced schemas "subschemas" is confusing and
technically inaccurate.

Issue #530, opened 2018-01-09
Need more details of annotation collection


handrews:
There should be pseudocode and a sample/recommend output format, as there is with Hyper-Schema.  This should include guidance on how annotation collection can be implemented as part of the same evaluation process as validation.

* Should explain that annotations are dropped whenever a validation check fails; the lack of annotations from `not` and non-validating branches of `oneOf`, `if`, etc. is a natural consequence of that
* Should cover how to indicate the schema from which the annotation was contributed- see example below for concerns around `$ref`
* Should note that there are both static annotations (all of the ones in the validation spec) and dynamic annotations that change based on the instance's contents (`links` and `base` in the hyper-schema spec)
* Should perhaps give some guidance around complex object annotations (e.g. `links`): how do they fit into the overall JSON Schema model?  Are there any particular limitations?
* Should provide guidance on whether keywords can only register annotations under the keyword name or can define different annotation names.  The use case for different names would include defining an annotation to which multiple keywords contribute, and wanting to avoid associating the annotation with one specific keyword

Some of the above may get split out into their own issues.

The following example shows the motivation for indicating where we were in the schema when the annotation was contributed.  A common request is to be able to figure out which `title` or `description` to use, with a typical desired heuristic is that values in parent schemas are preferred over values in child schemas.  Note that "child" in this cases is what me might call the logical child, following `$ref`s, instead of the physical child (a nested subschema within the current schema object).

So let's have a schema:

```JSON
{
    "type": "object",
    "properties": {
        "foo": {
            "title": "Title adjacent to reference",
            "description": "Lots of text",
            "readOnly": true,
            "$ref": "#/definitions/fooDef"
        }
    },
    "definitions": {
        "fooDef": {
            "title": "Title in reference target",
            "description": "Even more text"
        }
    }
}
```
With an instance of:
```JSON
{
    "foo": "whatever"
}
```
The annotation results might look like this (I literally made up how `$ref` appears in the JSON Pointer-ish thing while typing this example, so don't take that as set in stone):

```JSON
{
    "/foo": {
        "title": {
            "/properties/foo/$ref/title": "Title in reference target",
            "/properties/foo/title": "Title adjacent to reference"
        },
        "description": {
            "/properties/foo/$ref/description": "Even more text",
            "/properties/foo/description": "Lots of text"
        },
        "readOnly": true
    }
}
```

`"/foo"` is the instance pointer- this is the place in the instance to which the annotations apply.

[...]

Issue #396, opened 2017-09-05
Standard schema for results, including errors


trajano:
The schema provides a way of validating the input, but there should be a standard for the validation results as well.

handrews:
@trajano could you elaborate on this a bit more?  Are you talking about input to a specific validator?  Or input/results from some other system that you want to use validation with?

trajano:
I was thinking that we should have something along the lines of a validation result structure akin to https://docs.oracle.com/javaee/7/api/javax/validation/ConstraintViolation.html

For example:
```
{
   "valid": true | false,
   "violations": [
      {
         "property-path" : "$/path/to[10]/bad/element",
         "invalid-value": "who's bad" | { "some": [ "complex", "object" ] }
         "rule-path": "$/path/in/schema/that/triggered/violation",
          // these may be optionally added but should not be part of the spec
         "message-template": "{0} is who is bad",
         "message": "who's bad is who is bad",
     } ,
      {
         "property-path" : "$/path/to[10]/bad/element",
         "invalid-value": "who's bad" | { "some": [ "complex", "object" ] }
         "rule-path": "$/path/in/schema/that/triggered/violation",
          // these may be optionally added but should not be part of the spec
         "message-template": "{0} is who is bad",
         "message": "who's bad is who is bad",
     } 
    ]
}
```

trajano:
That way each application need not develop their own validation error standard messages

handrews:
@trajano I like the idea, in some form or another.

RFCs generally do not constrain implementations in this way, due to the wide variety of implementation languages, environments, and requirements such as performance vs usability.  This idea helps interoperability among users of validators, rather than interoperability in schema processing and outcome.  So I don't think this proposal belongs in the spec repo (at least not with its current scope).

I see two possible approaches:

1.  Write a schema for the desired output and perhaps we can discuss it with implementation maintainers and publish it as a recommendation on the web site (which is a separate GitHub repo).  Depending on how broadly it is adopted maybe there will be a clear need to formalize it further.

2.  Approach it like the `"application/problem+json"` specification.  That is an error-reporting format that is defined independent of the system producing the errors.  You would obviously want specific schema error-oriented fields, and would not be going for the completely generic approach of that spec, but separating it encourages its use in implementations that can support it, without burdening more constrained implementations with support.

erayd:
I really like this:
[...]

PR #610, opened 2018-06-17
Initial description of annotation collection

handrews:
***NOTE:*** _This PR uses language added in PR #609, which was recently merged._

This covers the basics of annotation collection (#530), and its interaction
with assertions and applicators.  It does not get into the exact
output structure for annotation results, which is intentional.
The goal is to establish the process, and then nail down the
format once the process is solid.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1140,6 +1140,94 @@
             </t>
         </section>
 
+        <section title="Collecting Annotations">
+            <t>
+                <cref>
+                    The exact structure and format of the information collected is TBD,
+                    but will be defined before the next draft.  Some details of this
+                    section may change as a result, but the overall process is expected
+                    to remain the same.  See GitHub issue #396 to track progress.
+                </cref>
+            </t>
+            <t>
+                Annotations are collected by keywords that explicitly define
+                annotation-collecting behavior.  Note that boolean schemas cannot
+                produce annotations as they do not make use of keywords.
+            </t>
+            <t>
+                A collected annotation MUST include the following information:
+                <list>
+                    <t>
+                        The name of the annotation, which MUST be identical to the keyword
+                        that defines that annotation.
+                    </t>
+                    <t>
+                        The instance location to which it is attached, as a JSON Pointer
+                    </t>
+                    <t>
+                        The schema location of the attaching keyword, as a list of URIs
+                        constructed as described below
+                    </t>
+                    <t>
+                        The attached value(s)
+                    </t>
+                </list>
+            </t>
+            <t>
+                If multiple schema locations attach values to the same instance location,
+                and the annotation defines a process for combining such values, then the
+                combined value MUST also be associated with the instance location.
+            </t>
+            <t>
+                Applications MAY make decisions on which of multiple annotation values
+                to use based on the schema location that contributed the value.
+                This is intended to allow flexible usage.  Collecting the schema location
+                facilites such usage.
+            </t>
+            <t>
+                For example, one application may consider a "description" annotation that is
+                in the same schema object as a "$ref" to override any "description" in the
+                reference's target schema.  A different application may prefer to concatenate
+                all "description" annotations based on whatever ordering it defines.
+            </t>
+            <section title="Annotations and Assertions">
+                <t>
+                    If any keyword in a schema object produces a false assertion
+                    result, then all annotations from all keywords in that schema
+                    object, and any of its subschemas or referenced schemas, MUST
+                    be discarded.
+                </t>
+                <t>
+                    This may be due to an assertion keyword directly producing
+                    a false result, or an applicator keyword producing a false
+                    assertion result as the aggregate assertion result of
+                    its subschema(s).
+                </t>
+            </section>
+            <section title="Annotations and Applicators">
+                <t>
+                    In addition to possibly defining annotation results of their own,
+                    applicator keywords aggregate the annotations collected in their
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -1066,6 +1066,94 @@
...
+                    The exact structure and format of the information collected is TBD,
+                    but will be defined before the next draft.  Some details of this
+                    section may change as a result, but the overall process is expected
+                    to remain the same.  See GitHub issue #396 to track progress.
+                </cref>
+            </t>
+            <t>
+                Annotations are collected by keywords that explicitly define
+                annotation-collecting behavior.  Note that boolean schemas cannot
+                produce annotations as they do not make use of keywords.
```
awwright: Is this not obvious enough?

jsonschema-core.xml
```diff
@@ -1066,6 +1066,94 @@
...
+            </t>
+            <t>
+                A collected annotation MUST include the following information:
+                <list>
+                    <t>
+                        The name of the annotation, which MUST be identical to the keyword
```
awwright: How about simply "The name of the keyword that produced the annotation"

jsonschema-core.xml
```diff
@@ -1066,6 +1066,94 @@
...
+                        The schema location of the attaching keyword, as a list of URIs
+                        constructed as described below
+                    </t>
+                    <t>
+                        The attached value(s)
+                    </t>
+                </list>
+            </t>
+            <t>
+                If multiple schema locations attach values to the same instance location,
```
awwright: What's an "instance location"? Maybe just "instance"?

jsonschema-core.xml
```diff
@@ -1066,6 +1066,94 @@
...
+                Applications MAY make decisions on which of multiple annotation values
+                to use based on the schema location that contributed the value.
+                This is intended to allow flexible usage.  Collecting the schema location
+                facilites such usage.
+            </t>
+            <t>
+                For example, one application may consider a "description" annotation that is
+                in the same schema object as a "$ref" to override any "description" in the
+                reference's target schema.  A different application may prefer to concatenate
+                all "description" annotations based on whatever ordering it defines.
```
awwright: If we're going to specify sometimes annotations can override one another, there should be a deterministic algorithm defined. Consider how CSS ranks which property is highest priority.

Presenting all the generated values as equals is probably the most theoretically pure way, but I'm not sure how practical this is for schema authors. I don't think I'll have a solid opinion on this until we see implementations.

jsonschema-core.xml
```diff
@@ -1066,6 +1066,94 @@
...
+            </t>
+            <section title="Annotations and Assertions">
+                <t>
+                    If any keyword in a schema object produces a false assertion
+                    result, then all annotations from all keywords in that schema
+                    object, and any of its subschemas or referenced schemas, MUST
+                    be discarded.
```
awwright: This could be shortened for clarity, I think.
"Annotations are only generated from instances that validate against the schema."

And care is needed so we don't confuse people implementing "not". The note about "not" is only (and should only be) non-normative.

jsonschema-core.xml
handrews: Never mind, I see what you mean.  Will update.

jsonschema-core.xml
handrews: @awwright I'm going to add an example to make it more clear that we're not just punting on the question of precedence, but rather providing tools for applications to define their own in a clear way.

jsonschema-core.xml
handrews: @awwright I'm going to remove most of the "Annotations and Applicators" section to get rid of the details around `not` and similar cases, and try to clarify why I'm using the "must be discarded" language in the "Annotations and Assertions" section.

handrews:
@awwright I've made several updates and tried to incorporate most of your feedback.  I've removed a number of potentially confusing details- hopefully people will understand how they follow from the simpler descriptions.

I also added an example of how an application might make use of multiple values.  Which hopefully will make the point and usefulness more clear.  If you still really want the spec to define behavior for every keyword, let's split that out into a new issue (in which case I will remove the controversial language from this PR so that we can merge the uncontroversial parts).

The two things I did not change are:

* the line about boolean schemas.  I agree that it should be obvious, but since boolean schemas produce assertion results despite not having keywords, I feel like it needs to be explicit that they do not produce annotation results.  Even though I can't think of what they would possibly produce.

* "instance location", which is terminology that, along with "schema location", I've been using extensively across multiple PRs.  It appears in all three documents at this point.  If you think there is a serious problem with it, please file a new issue and we will work it out.  But it is not specific to this PR.

jsonschema-core.xml
```diff
@@ -1140,6 +1140,191 @@
...
+            <t>
+                Annotations are collected by keywords that explicitly define
+                annotation-collecting behavior.  Note that boolean schemas cannot
+                produce annotations as they do not make use of keywords.
+            </t>
+            <t>
+                A collected annotation MUST include the following information:
+                <list>
+                    <t>
+                        The name of the keyword that defines the annotation
```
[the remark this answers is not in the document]
handrews: Yeah, I'll change that and then commit, thanks!

vearutop:
If [validation error response](https://github.com/json-schema-org/json-schema-spec/issues/396) is  also powered by annotations, then boolean schema that produces the error should be in annotation, shouldn't it?

Issue #591, opened 2018-05-24
Rename post-split "dependencies" to "schemaDependencies"


handrews:
Previously, we split the old `dependencies` keyword in two:

* `requiredDependencies` is an assertion in the validation vocabulary, and behaves the same as the old string array form of `dependencies`
* `dependencies` is an applicator in the core spec's applicator vocabulary, and only supports the old schema form

This makes `dependencies` incompatible from draft-07 to draft-08: anything using the string form is now _incorrect_ and will not work.  This is also confusing for schema authors migrating to draft-08: there is a non-obvious change and they need to change how they think about this keyword.

Instead, let's do a true split, and make the schema case `schemaDependencies`.  This does two things:

1. It makes the behavior clear from the name, in a way that is consistent with `requiredDependencies`
2. It allows schema authors to continue using `dependencies` with its pre-draft-08 semantics as an _extension_ keyword

The second point is particularly compelling.  It also gives implementors the ability to continue to support the existing `dependencies` behavior with draft-08 (presumably as an opt-in, or even better with a custom meta-schema using the new vocabulary support allowing schema authors to declare `dependencies` as an extension- we could even offer a "draft-08-transitional" meta-schema that provides this).

I've been worried about breaking the `dependencies` keyword in this draft, and I think this is a much better solution for both clarity and migration support.

Thoughts?  Paging some folks who weighed in on other `dependencies` issues: @erayd @markchart @pipobscure

awwright:
We had `properties` and `additionalProperties` before, would the same critique apply to that, even if we didn't have `requiredProperties`?

handrews:
@awwright 

> We had `properties` and `additionalProperties` before, would the same critique apply to that, even if we didn't have `requiredProperties`?

I don't follow.  This is about a single keyword being split into two keywords, and the best way to name them while avoiding breaking existing implementations.  What does that have to do with `properties` or `additionalProperties`?  Yes, it's true that `properties` and `required` are single words that _could_ be two-word keywords matching the form of other `*Properties*`.

But the point here is not that `dependencies` is somehow bad because it's one word.  The problem is that changing the behavior of `dependencies` **breaks things** for people moving from an earlier draft to draft-08.  Replacing it with two new keywords allows for compatibility.  The symmetry of the resulting keywords is nice but only in the sense of "hey we can allow for compatibility and still make sense", not "we need to make everything a two-word keywords".

awwright:
[...]

Issue #528, opened 2018-01-07
Move "dependencies" to applicability group


epoberezkin:
As requested by @handrews in https://github.com/json-schema-org/json-schema-spec/issues/513#issuecomment-355772558

The fact you can rewrite a keyword using if/then/else/required doesn't mean that it makes the keyword itself asserting any conditions. This schema:

```json
{
  "properties": {
    "foo": {"type": "number"}
  }
}
```

can be rewritten as

```json
{
  "if": {"required":["foo"]},
  "then": {
    "properties": {
      "foo": {"type": "number"}
    }
  }
}
```

"properties" keyword still belongs to the applicability group, even though you can rewrite it using "required", same as "dependencies".

Compare keywords definitions (ignoring property syntax of "dependencies" for now, I'll come back to it later):

"properties" (http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.5.4): 

> Validation succeeds if, for each name that appears in both the instance and as a name within this keyword's value, the child instance for that name successfully validates against the corresponding schema.

"dependencies" (http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.5.7):

> If the dependency value is a subschema, and the dependency key is a property in the instance, the entire instance must validate against the dependency value.

So, for "properties", if a property name appears in the instance, you have to apply a corresponding subschema to the child instance.

For "dependencies", if a property name appears in the instance, you have to apply a corresponding subschema to the current instance.

The logic of "properties" and "dependencies" only differs in that the former applies subschemas to the child instance, and the latter to the current instance. Neither of them is making assertions on their own.

Now back to the property syntax of "dependencies". We have two possible points of view here:

1. We can consider it just a syntactic sugar that allows to use `<array of strings>` as a shortcut for `{"required": <array of strings>}`. It's not dissimilar from the boolean form of the schema, where you can use `false` as a shortcut for `{"not":{}}`. In this case, "dependencies" keyword still applies a certain schema to the current instance, without making any assertions of its own.
2. We can interpret it is as a separate thing completely that makes its own assertions. In this case the property syntax does not belong to applicability group. But the downside of this view is that it creates an overlap with "required" keyword by allowing two ways of asserting that property is required.

Regardless which way of looking at property syntax of "dependencies" we prefer, the schema syntax of "dependencies" still belongs to the applicability group, as it only uses a defined logic to apply subschemas to the current instance, without making any assertions of its own, which is consistent with how applicability is defined.

[...]

PR #614, opened 2018-06-19
Rename "dependencies" to "schemaDependencies"

handrews:
I also noticed that it was in the wrong section.  As an
in-place conditional applicator, it belongs with "if", "then",
and "else".  Not with the object child applicators.

Apparently, I had not made the meta-schema changes at all.
The "$comment" follows the precedent established from leaving
"definitions" in the meta-schema after the rename to "$defs".

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -16,7 +16,9 @@
     "properties": {
         "additionalItems": { "$ref": "#" },
         "additionalProperties": { "$ref": "#"},
+        "schemaDependencies": { "$ref": "#" },
         "dependencies": {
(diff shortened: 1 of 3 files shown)


erayd:
Per my comment on [#528](https://github.com/json-schema-org/json-schema-spec/issues/591#issuecomment-398558962), I don't think that `dependencies` should be renamed.

No objection to moving it; I agree that it belongs with `if` / `then` / `else`.

handrews:
Per discussion in #528, `dependent*` seems like it fits better than `*Dependencies`, so I'm taking this down and re-submitting (to avoid questions of which approvals are up-to-date)

Issue #350, opened 2017-08-18
How to handle multiple "rel" values per RFC 5988


handrews:
@dret has been looking at this issue with respect to JSON serialization for his linkset I-D, and is going in the direction that `"rel"` should always be an array.  We could match this (and potentially other things from the linkset work, as having JSON Hyper-Schema LDOs be compatible seems like it would be nice, see also #7 Link Attributes in JSON Hyper-Schema).

This would look like:
```JSON
{
    "links": [
        {
            "rel": ["self", "canonical"],
            "href": "foos/{id}",
            "targetSchema": {"$ref": "#"}
        }
    ]
}
```

There are two obvious alternatives for JSON Schema:

1.  Use space-separated values in a single string per RFC 5988 HTTP Link header serialization.  This discards the value of JSON as a structured data format, so I'm fairly strongly against this.  It looks like:

```JSON
{
    "links": [
        {
            "rel": "self canonical",
            "href": "foos/{id}",
            "targetSchema": {"$ref": "#"}
        }
    ]
}
```

2.  Accept issue #319 and allow `"$ref"` for LDOs:

```JSON
{
    "links": [
        {
            "rel": "self",
            "allOf": [{"$ref": "#/definitions/singleFoo"}]
        },
        {
            "rel": "canonical",
            "allOf": [{"$ref": "#/definitions/singleFoo"}]
        },
    ],
    "linkDefinitions": {
        "singleFoo": {
            "href": "foos/{id}",
            "targetSchema": {"$ref": "#"}
        }
    }
}
```

Of course, we could both accept `"linkDefinitions"` (#319) and specify that `"rel"` MUST be an array, as there are other use cases for `"linkDefinitions"`.  So we may want to satisfy those use cases and also be compatible with linkset serializations, which would require doing both of these things.

We could also allow `"rel"` to be either an array or a single value.  I dislike that sort of thing, but there is a precedent with `"type"` where it is also true that the vast majority of uses call for only a single value.  That way we could give people the option of linkset compatibility but also allow for a JSON Hyper-Schema-specific shorthand.

dret:
it definitely would be nice to have consistent patterns across JSON models for links. but it most definitely is not necessary. just look at how HAL drives everything from the relation: then you end up with a quite different view of how to represent links.

handrews:
[...]

PR #615, opened 2018-06-19
Allow multiple values for "rel"

handrews:
Addresses #350 

This follows the precedent of "type" by allowing "rel" to be
either a single string, or an array of strings.

Also paging @mokkabonna

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -456,8 +456,15 @@
                 </t>
                 <section title="rel" anchor="rel">
                     <t>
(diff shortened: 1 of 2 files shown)


philsturgeon:
What is the use-case here?

handrews:
@philsturgeon multiple rel values are well established in RFC 5988/8288 (the theoretical model and the HTTP `Link` header serialization).  They are also well established in the HTML `<link />` element, and while I reject analogies with HTML that limit Hyper-Schema, I do think that the non-presentation aspects of HTML are a good reference for a minimum baseline functionality.

PR #636, opened 2018-07-11
Updated Schema.json to force full paterns

barkermn01:
As partial pattern matching going against the point of matching a pattern for validation of an input allowing a input to pass validation even though it could end in an incompatible way to the definition.

For Example:
 `"pattern": "([1-9]+[0-9]*)(,\\s{0,1}[1-9][0-9]*)*"` would allow `1,3, fr` because the 1, 3 matches the pattern however because the pattern is not enforcing complete input validation the `, fr` does not break the validation

This pull request implements a fix in the schema to force patterns to start with `^` and end in `$` forcing the pattern to be `"pattern": "^([1-9]+[0-9]*)(,\\s{0,1}[1-9][0-9]*)*$"`

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -199,6 +199,13 @@
                     </list>
                 </t>
                 <t>
+                     Validation rules should make sure they validate the complete input.
+                    <list>
+                        <t>Validation rules should begin with "^"</t>
+                        <t>Validation rules should terminate with "$"</t>
+                    </list>
+                </t>
+                <t>
                     Finally, implementations MUST NOT take regular expressions to be
                     anchored, neither at the beginning nor at the end. This means, for instance,
                     the pattern "es" matches "expression".
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -205,6 +205,7 @@
                         <t>Validation rules should terminate with "$"</t>
                     </list>
                 </t>
+                <t>
                     Keywords that use regular expressions, or constrain the instance value
                     to be a regular expression, are subject to the interoperability
                     considerations for regular expressions in the
--- a/schema.json
+++ b/schema.json
@@ -99,7 +99,8 @@
         "minLength": { "$ref": "#/$defs/nonNegativeIntegerDefault0" },
         "pattern": {
             "type": "string",
-            "format": "regex"
+            "format": "regex",
+            "pattern": "^(\\^(.[^\\$]*)\\$)$"
         },
         "additionalItems": { "$ref": "#" },
         "items": {


handrews:
@barkermn01 Thank you for volunteering a PR!

However, this is not in line with the direction of the specification.  Patterns in JSON Schema have never been implicitly anchored, and it has never been a goal to force explicit anchoring.

Additionally, the section you put back in jsonschema-validation.xml now lives in the JSON Schema Core spec (it applies to more than just validation), and is referenced in the paragraph after your change.

Fore these reasons, I'm closing this.  I encourage you to file issues to discuss change proposals, but be aware that we will not be changing the regular expression behavior with regards to start and end anchoring.

barkermn01:
@handrews could i ask why? as per Slack 

"I can think of more cases where you would want anchored validation and not partial even with exact validation you could easily support partial validation than anything by adding `.*` to the end of your rule."

barkermn01:
I'm sorry I think you have misunderstood what this change does as, your comment "we should not be modifying expected behaviour" this change does not change expected behaviour it just enforces anchoring on pattern specifiers. there for it forces users to use anchors in patterns otherwise their schema fails validations against the schema.json.

handrews:
@barkermn01 that is a change in the expected behavior.  The current behavior, which has been the case for many years and many, possibly all, drafts (longer than I have been a part of this project) is that regular expressions do not need to be anchored.  This is an intentional feature.  It is not a bug.  Suddenly forcing everyone to anchor their expressions would be a regression in functionality.

> "I can think of more cases where you would want anchored validation and not partial even with exact validation you could easily support partial validation than anything by adding .* to the end of your rule."

This doesn't sound like me.  I disagree with it, so if I wrote that then I have no idea what I was thinking at the time.  But I doubt I wrote that.

barkermn01:
> > "I can think of more cases where you would want anchored validation and not partial even with exact validation you could easily support partial validation than anything by adding .* to the end of your rule."

> This doesn't sound like me. I disagree with it, so if I wrote that then I have no idea what I was thinking at the time. But I doubt I wrote that.

That was me in slack as I was hoping for a chat about it in slack but nothing was said sorry if that seemed like i said you had set it. can we move this conversation into slack as i really would like to know a use case that would allow anything then (pattern) then anything in a validation system beacuse i have now been thinking about this for 2 days can't think of a single one?

Relequestual:
This isn't a case of "does it have a use case?", and more a case of, "regex works like this, and we're using regex". We should not prevent a user from doing something they may ligitimatly expect to do. There's no gaurantee that a validator uses the metaschema to validate a schema, the spec doesn't require it.

Maybe someone has a shop that sells apples, and they want to make sure each description has the word "apple" in it. Doesn't matter where. Must be string, must contain the word "apple". Non anchored regex.

barkermn01:
Ok that makes is a hypothetical use case that I double would ever be a use case since use cases are supposed to be real-world uses but I will concede it's plausible. but then this has made me think maybe the reason i got this wrong is the terminology is incorrect then?

I have double checked this with a dictionary before making this request and if you agree i will open an issue about the matter. the dictionary deffinitions below:

> **format**
_noun_ the way in which something is arranged or set out.
_verb_ (especially in computing) arrange or put into a format.

Verses

> **matches**
_noun_ a person or thing that is equal to another in quality or strength
_verb_ be equal to (something) in quality or strength.

So perhapse we change format to matches as what we are doing is supplying a must match case and not a must conform to this format. we could possibly change it, add to the specification support for both a match clause &/or a format close where format uses anchoring and match does not. I think this would be clearer for understanding and better usagage of the language?

handrews:
@barkermn01 the language in the spec is very typical regular expression language.  While not all systems that work with regular expressions have identical language (e.g. Perl's `=~` is referred to as the "matching operator" and is not anchored, while in Python, unanchored usage is done with `re.search()`, while `re.match()` is anchored at *one end only*), our usage is well within the mainstream.

This is an area of the spec that is quite mature and heavily used over many years, without complaints.  We will not be changing it.

Relequestual:
We're done here.

PR #650, opened 2018-09-08
Fix entry point URI in example description

dsanders11:
Since the URI in the `GET` is `https://api.example.com`, presumably the text was supposed to be referring to that URI as well.


handrews:
Thanks!

PR #653, opened 2018-09-17
$recursiveRef part 1 of 3: Dynamic vs Lexical scope

handrews:
***NOTE:** I'm breaking this up into three parts to keep each one a bit more focused.  The first two are on the core spec file, but do not overlap in line numbers or even sections.  There are some references between them but figuring them out should be straightforward.  I'm not entirely happy with this write up (particularly the 2nd part), but I have been struggling with this for at least two months and just need to get something out there to unblock progress.   Hopefully we can edit this into something workable.*

This is some groundwork for #558 $recursiveRef/$recursiveAnchor,
formalizing lexical and dynamic notions of scope in ways that
should be quite intuitive given how the concepts are used in
a wide variety of programming languages.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -570,13 +570,70 @@
                         As with the root schema, a subschema is either an object or a boolean.
                     </t>
                 </section>
+                <section title="Lexical Scope and Dynamic Scope">
+                    <t>
+                        While most JSON Schema keywords can be evaluated on their own,
+                        or at most need to take into account the values or results of
+                        adjacent keywords in the same schema object, a few have more
+                        complex behavior.
+                    </t>
+                    <t>
+                        The lexical scope of a keyword is deterimined by the nested JSON
+                        data structure of objects and arrays.  The largest such scope
+                        is an entire schema document.  The smallest scope is a single
+                        schema object with no subschemas.
+                    </t>
+                    <t>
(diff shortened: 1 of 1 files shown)


Relequestual:
Updated referenced issue number ;)

jsonschema-core.xml
```diff
@@ -570,13 +570,70 @@
...
+                    </t>
+                    <t>
+                        Keywords MAY be defined with a partial value, such as a URI-reference,
+                        which must be resolved against another value, such as another
+                        URI-reference or a full URI, which is found through the lexical
+                        structure of the JSON document.  The "$id" core keyword and
+                        the "base" JSON Hyper-Schema keyword are examples of this sort
+                        of behavior.  Additionally, "$ref" and "$recursiveRef" from
+                        this specification resolve their values in this way, although
+                        they do not change how further values are resolved.
```
Relequestual: Thinking about this, I wonder if this is correct...
It IS correct, if we define that you must only have unique keys for each object, which I'm not sure we do, and is only a SHOULD according to the JSON specification.
If we already define such, please direct me to where.
If not, should we?
If not, I can explain a schema where this point would then be false!

handrews: We explicitly exclude (or rather, defer behavior to the JSON parser that's being used) the duplicate property scenario here: https://tools.ietf.org/html/draft-handrews-json-schema-01#section-4.2.1

PR #654, opened 2018-09-17
$recursiveRef part 2 of 3: $recursiveRef and $recursiveAnchor

handrews:
***NOTE:** I'm breaking this up into three parts to keep each one a bit more focused. The first two are on the core spec file, but do not overlap in line numbers or even sections. There are some references between them but figuring them out should be straightforward. I'm not entirely happy with this write up (particularly the 2nd part), but I have been struggling with this for at least two months and just need to get something out there to unblock progress. Hopefully we can edit this into something workable.*

Addresses #558 

These keywords allow for recursive extension of schemas,
in the way needed by meta-schemas.  This is not Object-Oriented
extension, rather it is a shortcut for the kind of "allOf"
combination that is already allowed, but without the need to
redeclare all recursive keywords in the extension.

It works by both the original target of the reference and the
dynamically chosen target of the reference opting in to the
dynamic behavior with "$recursiveAnchor", and the source of
the reference opting in by using "$recursiveRef".

This ensures that regular "$ref" usage is unchanged and
and predictable, and also that schemas not intended for
recursive extension cannot be extended without being changed.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -205,11 +205,11 @@
                         The schemas to be applied may be present as subschemas comprising all or
                         part of the keyword's value.  Alternatively, an applicator may refer to
                         a schema elsewhere in the same schema document, or in a different one.
-                        The mechanism for identifying such referred schemas is defined by the
+                        The mechanism for identifying such referenced schemas is defined by the
                         keyword.
                     </t>
                     <t>
-                        Applicator keywords also define how subschema or referred schema
+                        Applicator keywords also define how subschema or referenced schema
                         boolean <xref target="assertions">assertion</xref>
                         results are modified and/or combined to produce the boolean result
                         of the applicator.  Applicators may apply any boolean logic operation
@@ -626,20 +626,19 @@
                         and used with caution when defining additional keywords.
                     </t>
                 </section>
-                <section title="Referred and Referring Schemas" anchor="referred">
+                <section title="Referenced and Referencing Schemas" anchor="referenced">
                     <t>
                         As noted in <xref target="applicators" />, an applicator keyword may
                         refer to a schema to be applied, rather than including it as a
                         subschema in the applicator's value.  In such situations, the
-                        schema being applied is known as the referred (or referenced) schema,
-                        while the schema containing the applicator keyword is the referring
-                        (or referencing) schema.
+                        schema being applied is known as the referenced schema, while
+                        the schema containing the applicator keyword is the referencing schema.
                     </t>
                     <t>
                         While root schemas and subschemas are static concepts based on a
-                        schema's position within a schema document, referred and referring
+                        schema's position within a schema document, referenced and referencing
                         schemas are dynamic.  Different pairs of schemas may find themselves
-                        in various referred and referring arrangements during the evaluation
+                        in various referenced and referencing arrangements during the evaluation
                         of an instance against a schema.
                     </t>
                     <t>
@@ -1010,8 +1009,8 @@
                 <t>
                     Several keywords can be used to reference a schema which is to be applied to the
                     current instance location. "$ref" and "$recursiveRef" are an applicator
-                    keywords, applying the referred schema to the instance.  "$recursiveAnchor"
-                    is a helper keyword that controls how the referred schema of "$recursiveRef"
+                    keywords, applying the referenced schema to the instance.  "$recursiveAnchor"
+                    is a helper keyword that controls how the referenced schema of "$recursiveRef"
                     is determined.
                 </t>
                 <t>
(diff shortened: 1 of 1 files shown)


handrews:
@Relequestual yeah, I don't really like the basic/extension either, couldn't come up with anything better and just wanted it posted.  I'll think through some alternatives (and work on the rest of the feedback-thanks!).  I don't follow the logic of "extendedBasic" so I'll just go back to the drawing board entirely.  Maybe also talk about the relationship between what each of them validates in terms of instances more, rather than relying on the naming to convey behavior.

gregsdennis:
Maybe just use `foo`/`bar`. Those seems generally accepted. I'm not sure you necessarily need to use descriptive identifiers here.

jsonschema-core.xml
```diff
@@ -953,35 +953,234 @@
...
+            }
+        }
+    }
+}
+]]>
+                            </artwork>
+                            <postamble>
+                                Now lets consider the evaluation of the "extension" schema.
+                                Note that the "$ref": "basic" was not changed, as it works
+                                just fine as a normals static reference.  And the
```
gregsdennis: /normals/normal/

jsonschema-core.xml
```diff
@@ -953,35 +953,234 @@
...
+                                schema, the "$recursiveRef" in the basic schema actually
+                                referrs to the "extension" schema's root schema.
+                            </postamble>
+                        </figure>
+                    </section>
+                </section>
+
+                <section title="Guarding Against Inifinite Recursion">
+                    <t>
+                        A schema MUST NOT be run into an infinite loop against an instance. For
```
gregsdennis: > A schema MUST NOT be run into an infinite loop against an instance.

This doesn't really make sense.  The explanation afterward does, but I don't get this sentence.

jsonschema-core.xml
```diff
@@ -953,35 +953,234 @@
...
+                        example, if two schemas "#alice" and "#bob" both have an "allOf" property
```
gregsdennis: Can we remove the `allOf` usage from this example now that `$ref` can be used alongside other kewords?

handrews: I think it's still valid as written, so as with the previous comment if you want to figure out a better way to do this whole paragraph let's discuss in an issue or separate PR.

jsonschema-core.xml
handrews: I believe I just copy-pasted that from wherever it was before, so if we want to consider reworking it please file a new issue on it.  I'm fine with discussing it, it's just not changed here so I don't want to add to the PR.

jsonschema-core.xml
```diff
@@ -953,35 +953,234 @@
...
+                            references are statically determined.  Since the
+                            <spanx style="verb">"things"</spanx> array is in the
+                            combined schema, its referred schema is the combined
+                            schema.  But the <spanx style="verb">"recursive"</spanx>
+                            property in the basic schema still points to the root
+                            of that basic schema, and therefore will not see the
+                            description of the <spanx style="verb">"things"</spanx>
+                            property.  What we want is for it to resolve
+                            to the combined schema as well.
+                        </postamble>
```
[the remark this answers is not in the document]
handrews: Um... maybe?  The whole transcluded/included thing does not really help me.

handrews:
@Relequestual @gregsdennis I have finally gotten back to this and added a second commit based on the feedback here.  I more or less completely rewrote the description and example and I think the new version is much more straightforward.  Where the text was still the same, I applied the typo / wording corrections.

I also replaced "referred" and "referring" with "referenced" and "referencing" as that feels much more straightforward- I'd even previously changed a paragraph to use the parenthetically, so just being consistent about it seems better.

As noted, the "Guarding against..." section was only moved by this PR, not changed, so I have not changed it based on the feedback here.  Please open an issue or submit a PR if you want to pursue that.

philsturgeon:
This looks great. To be honest I did have a hard time understand what the use case was for this based on just reading the description for this PR and the contents of the changes as they are now. It all made sense once I read https://github.com/json-schema-org/json-schema-spec/issues/558 and saw that hyperschema was having to copy and paste stuff, but I am not sure how that, or anything else that made it more clear, could go into the PR. 

It's probably fine, ship it! 💃

handrews:
@Relequestual told me that it was fine to have @philsturgeon review this as he is on vacation and/or busy for a while.  So I will go ahead and merge this even though he has not lifted the "requested changes" flag.

Of course, we can keep working on the wording here, and probably will once we get to the feature-complete full spec review.

Relequestual:
Yup that's fine =]

PR #657, opened 2018-09-24
Require xml2rfc < 2.10 in travis build

dlax:
From version 2.10, option --basename changed its semantics and is now
expected to be a directory, which does not fit our current usage.

Alternatively, we could drop this option altogether (and thus avoid
version pinning) since I'm not sure what this is for.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 install:
-- pip install xml2rfc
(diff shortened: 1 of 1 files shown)


handrews:
This fixes the build, so let's go with it- thanks!  We can sort out basename later if we want.

Issue #602, opened 2018-06-06
"unevaluatedProperties" and extension keywords (keyword meta-data)


handrews:
Since `unevaluatedProperties` #556 and `unevaluatedItems` #557 depend on the results of other keywords, not just in the immediate schema object but in subschemas, we need to decide how extension keywords can or cannot impact that behavior.  There are two cases:

1.  New [object child applicators](https://github.com/json-schema-org/json-schema-spec/blob/3a517363ebe16441f4c0ba74a0737b742aa76132/jsonschema-core.xml#L1217) or [array child applicators](https://github.com/json-schema-org/json-schema-spec/blob/3a517363ebe16441f4c0ba74a0737b742aa76132/jsonschema-core.xml#L1159)

2.  New [in-place applicators](https://github.com/json-schema-org/json-schema-spec/blob/3a517363ebe16441f4c0ba74a0737b742aa76132/jsonschema-core.xml#L1008)

For child applicators, as noted in https://github.com/json-schema-org/json-schema-spec/issues/530#issuecomment-392608099, we should **not** allow them to change the behavior of `unevaluated*`.  This follows from `additionalProperties` and `additionalItems` which do not change as a result of new keywords.  They are defined just in terms of `properties`/`patternProperties` or `items`.

For new in-place applicators, which could *contain* `*properties` or `*items` keywords, the situation is more complex.

### _TL;DR:_

* we need to allow `*properties` and `*items` to affect `unevaluated*` even when they are in subschemas of an extension in-place applicator.
* we can solve this generally by declaring some keyword meta-data in vocabularies
* implementations that just want to implement standard vocabularies can hardwire stuff and not worry about handling meta-data at runtime, so it won't be a huge burden on implementations

### Example

In our brave new world of multi-vocabulary schemas, let's pretend someone decides to create an extension keyword `patternSchemaDependencies` which is a cross between `patternProperties` and `schemaDependencies` (the old schema form of `dependencies`).  So, if the instance is an object, and at least one property matches a pattern in `patternSchemaDependencies`, then that pattern's subschema is applied to the *current* instance location, making it an *in-place applicator*.

Consider this schema using that keyword (and assume `patternSchemaDependencies` is properly declared in the meta-schema referenced by `$schema`, and in whatever vocabulary stuff we come up with, and that the implementation will only process the schema if it understands the extension vocabulary, etc. see #561 for details)

```JSON
{
    "patternSchemaDependencies": {
        "^foo": {
[...]

Issue #567, opened 2018-03-13
Core vocabulary?


handrews:
In a multi-vocabulary system (#561), should it be possible to explicitly reference the core vocabulary?  Should it be required to do so?

Keep in mind that vocabularies are listed in meta-schemas, not in individual schemas, so the number of times your average schema author will need to do this is quite small.

The main use case for listing the core vocabulary is identifying which version of core is in use (e.g. draft-08), since it is not necessary for all vocabulary draft numberings to match.  In particular, extension vocabularies and meta-schema will likely not indicate a draft number, so anything that relies on matching `draft-NN` in the URI is not feasible.

Here is what a core vocabulary meta-schema would probably look like (note that per #513, the applicators are their own vocabulary even if the live in the core specification):

```JSON
{
  "$id": "http://json-schema.org/draft-08/vocabularies/core",
  "$schema": "http://json-schema.org/draft-08/schema#",
  "type": ["object", "boolean"],
  "properties": {
    "$id": {"type": "string", "format": "uri-reference"},
    "$schema": {"type": "string", "format": "uri"},
    "$recurse": {"type": "boolean", "const": true},
    "$ref": {"type": "string", "format": "uri-reference"},
    "$defs": {"type": "object", "additionalProperties": {"$recurse": true}},
    "$comment": {"type": "string"}
  }
}
```

My inclination would be to publish such a vocabulary schema, and use it in the meta-schemas.  This is the most consistent and clear approach, and allows the most flexibility for extensions.

handrews:
We decided that if `$vocabulary` is used at all, the core vocab is mandatory to list.  Merged #671.

Issue #564, opened 2018-03-09
Required vs Optional vocabularies


handrews:
This is a follow-on from #561 (Basic vocabulary support).  I would like to get this into draft-08 as well, but if it stalls, we can move forward without it.  Hence the separate issue.

As proposed in #561, `$vocabularies` will provide a list of vocabulary URIs that an implementation can use to make a decision about whether or how to process the schema.  The most basic form of this proposal would just list the URI, and the implementation has to decide what to do with unrecognized URIs.

A more useful form would allow for a meta-schema to indicate which of its constituent vocabularies **MUST** be understood in order to process a schema in accordance with the meta-schema.  This could look something like the following (_**This example is contrived**_ and not intended to be compelling.  But I wanted to use vocabularies that already exist.)

```JSON
{
    "$id": "http://json-schema.org/draft-08/hyper-schema",
    "$vocabularies": {
        "http://json-schema.org/draft-08/vocabularies/core-applicator": true,
        "http://json-schema.org/draft-08/vocabularies/validation-assertions": true,
        "http://json-schema.org/draft-08/vocabularies/hyper-schema": false
    }
}
```
or some similar syntax.  In this version `true` means:

_In order to process a schema that uses this meta-schema, you MUST understand applicators and validation.  **If** you understand hyper-schema, then you can use that as well, but if you have no idea what this means and treat the hyper-schema keywords as unknown and ignore them, the schema author allows that._

So applicators and validation are mandatory, and hyper-schema is opportunistic.  If you don't understand the first two, you're not going to be able to do anything useful.  But an application may want to run a hyper-schema through an implementation optimized to support only validation and its dependencies (core applicators), and you wouldn't want it to choke on the extra keywords.

On the other hand, if you fed this to an implementation optimized to attach simple links (no validation involved) to locations in an instance, then it *should* reject anything with this meta-schema, because the schema author is saying that understanding validation is mandatory.

If your hyperlinks are being applied conditionally based on whether a part of the instance validates or not, then you cannot ignore validation keywords and still get the correct output.

--------

More compelling use cases come when you are mixing vocabularies like an annotation vocabulary that adds code generation hints to a validation schema.  Code generation really won't work with validation (at least the way I'm thinking of it right now).  But you really don't want code generation to confuse regular validation, because maintaining separate schemas for those purposes is too much duplication.

handrews:
Merged #671.

Issue #561, opened 2018-03-08
Basic vocabulary support


handrews:
## Proposal

* Vocabularies
   * Group keyword definitions and their semantics (the proposal does not add any way to formally define semantics in JSON, they're defined in the specs just as they are now)
   * Are identified by URIs, which (like meta-schema URIs) should not automatically be retrieved
   * The URI SHOULD point to a meta-schema describing **only** that vocabulary's keyword
   * SHOULD NOT overlap with or redefine keywords from other vocabularies
       * The hyper-schema meta-schema includes validation, but the hyper-schema vocabulary is _only_ `base` and `links`, and the keywords in the LDOs
   * Vocabularies are file-scoped; keywords MUST have the same semantics throughout a single schema document

* Meta-Schemas
    * In a meta-schema, `$vocabularies` takes a list of URIs identifying the vocabularies described by the meta-schema
    * Like `$schema` in a schema, `$vocabularies` must be in the root object of the meta-schema
    * Meta-schemas SHOULD validate the combination of vocabularies they declare
    * Combining vocabularies is facilitated by `$recurse` (#558)
    * Meta-schemas MAY further constrain that vocabulary combination
    * Meta-schemas MAY describe keywords that are not in any declared vocabulary
    * Meta-schemas that do not declare a vocabulary, or that declare additional keywords, create an anonymous vocabulary (_this just fits existing meta-schemas into the vocabulary concept_)

## Examples:

_**NOTE:** "core-applicators" (stuff moved by #513) and "validation-assertions" (stuff left behind by #513) are not final names or vocabulary boundaries, I literally made them up while typing, please do not complain about whether they are "correct"._

### The applicators (per #513) as a vocabulary

This is where `$recurse` ( #558 ) would primarily be used.  This assumes that `dependencies` has been split per #528, and the applicator version is still called `dependencies`, while the string form is re-named and left in the validation vocabulary.

```JSON
{
    "$id": "http://json-schema.org/draft-08/vocabularies/core-applicators",
    "$schema": "http://json-schema.org/draft-08",
    "type": ["object", "boolean"],
    "properties": {
        "allOf": {"type": "array", "items": {"$recurse": true}},
        "anyOf": {"type": "array", "items": {"$recurse": true}},
        "oneOf": {"type": "array", "items": {"$recurse": true}},
        "not": {"$recurse": true},
        "if": {"$recurse": true},
        "then": {"$recurse": true},
        "else": {"$recurse": true},
        "items": {
[...]

PR #671, opened 2018-11-11
"$vocabulary" and basic vocabulary support.

handrews:
_It is recommended that the commits in this PR be reviewed one at a time, to avoid formatting changes making things more complicated than they are.  See also their individual commit messages._

Addresses #561, #564, #567 

I expect the wording here to have to go through some revision, but as is often the case I decided I just needed to post *something* to get the conversation started.

Key points:

* When declaring the use of a vocabulary, you can indicate whether it is required or not, but that is all
* ~For now, I'd rather not mandate any sort of vocabulary description file.  As we get feedback, we will no doubt learn what is best put there.  On the other hand, if implementation maintainers would like to hook specific keyword detection into their extension mechanism, I could reconsider and add such a thing as another PR. *[EDIT: For a [variety of reasons](https://github.com/json-schema-org/json-schema-spec/issues/602#issuecomment-437949419),~ I will probably try to do this now after all]** ***[EDIT TO EDIT: actually I probably won't, see further comments in the issue]***

* This stuff gets a bit complicated.  As with recursive references, most people will not need to write these things or understand their details, so in my view a relatively complex approach is acceptable.  Plus, I've been talking this up for nearly a year and no one has proposed anything simpler yet :-)

### DIFF
--- a/applicator.json
+++ b/applicator.json
@@ -3,6 +3,7 @@
     "$id": "https://json-schema.org/draft-08/applicator",
     "$recursiveAnchor": true,
     "$vocabulary": {
+        "https://json-schema.org/draft-08/vocabularies/core": true,
         "https://json-schema.org/draft-08/vocabularies/applicator": true
     },
 
--- a/applicator.json
+++ b/applicator.json
@@ -1,0 +1,60 @@
+{
+    "$schema": "http://json-schema.org/draft-08/schema#",
+    "$id": "https://json-schema.org/draft-08/applicator",
+    "$recursiveAnchor": true,
+    "$vocabulary": {
+        "https://json-schema.org/draft-08/vocabularies/applicator": true
+    },
+
+    "title": "Applicator vocabulary meta-schema",
+    "$defs": {
+        "schemaArray": {
+            "type": "array",
+            "minItems": 1,
+            "items": { "$recursiveRef": "#" }
+        }
+    },
+    "properties": {
+        "additionalItems": { "$recursiveRef": "#" },
+        "unevaluatedItems": { "$recursiveRef": "#" },
+        "items": {
+            "anyOf": [
+                { "$recursiveRef": "#" },
+                { "$ref": "#/$defs/schemaArray" }
+            ]
+        },
+        "contains": { "$recursiveRef": "#" },
+        "additionalProperties": { "$recursiveRef": "#" },
+        "unevaluatedProperties": {
+            "type": "object",
+            "additionalProperties": {
+                "$recursiveRef": "#"
+            }
+        },
+        "properties": {
+            "type": "object",
+            "additionalProperties": { "$recursiveRef": "#" },
+            "default": {}
+        },
+        "patternProperties": {
+            "type": "object",
+            "additionalProperties": { "$recursiveRef": "#" },
+            "propertyNames": { "format": "regex" },
+            "default": {}
+        },
+        "dependentSchemas": {
+            "type": "object",
+            "additionalProperties": {
+                "$recursiveRef": "#"
+            }
+        },
+        "propertyNames": { "$recursiveRef": "#" },
+        "if": { "$recursiveRef": "#" },
+        "then": { "$recursiveRef": "#" },
+        "else": { "$recursiveRef": "#" },
+        "allOf": { "$ref": "#/$defs/schemaArray" },
+        "anyOf": { "$ref": "#/$defs/schemaArray" },
+        "oneOf": { "$ref": "#/$defs/schemaArray" },
+        "not": { "$recursiveRef": "#" }
+    }
+}
--- a/core.json
+++ b/core.json
@@ -1,0 +1,50 @@
+{
+    "$schema": "http://json-schema.org/draft-08/schema#",
+    "$id": "https://json-schema.org/draft-08/core",
+    "$recursiveAnchor": true,
+    "$vocabulary": {
+        "https://json-schema.org/draft-08/vocabularies/core": true
+    },
+
+    "title": "Core vocabulary meta-schema",
+    "type": ["object", "boolean"],
+    "properties": {
+        "$id": {
+            "type": "string",
+            "format": "uri-reference"
+        },
+        "$schema": {
+            "type": "string",
+            "format": "uri"
+        },
+        "$ref": {
+            "type": "string",
+            "format": "uri-reference"
+        },
+        "$recursiveRef": {
+            "type": "string",
+            "format": "uri-reference"
+        },
+        "$recursiveAnchor": {
+            "type": "boolean"
+        },
+        "$vocabulary": {
+            "type": "object",
+            "propertyNames": {
+                "type": "string",
+                "format": "uri"
+            },
+            "additionalProperties": {
+                "type": "boolean"
+            }
+        },
+        "$comment": {
+            "type": "string"
+        },
+        "$defs": {
+            "type": "object",
+            "additionalProperties": { "$recursiveRef": "#" },
+            "default": {}
+        }
+    }
+}
--- a/core.json
+++ b/core.json
@@ -26,7 +26,9 @@
             "format": "uri-reference"
         },
         "$recursiveAnchor": {
-            "type": "boolean"
+            "type": "boolean",
+            "const": true,
+            "default": false
         },
         "$vocabulary": {
             "type": "object",
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -879,9 +879,10 @@
                     -->
                 </t>
                 <t>
-                    Values for this property are defined in this and other documents and by other
-                    parties.  JSON Schema implementations SHOULD implement support for current
-                    and previous published drafts of JSON Schema vocabularies as deemed reasonable.
+                    Values for this property are defined elsewhere in this and other documents,
+                    and by other parties.  JSON Schema implementations SHOULD implement support
+                    for current and previous published drafts of JSON Schema vocabularies
+                    as deemed reasonable.
                 </t>
             </section>
             <section title='The "$vocabulary" Keyword'>
@@ -1037,7 +1038,7 @@
                     <preamble>
                         This meta-schema explicitly declares both the Core and Applicator
                         vocabularies, and combines their meta-schemas with an "allOf".
-                        It additional restricts the usage of the Applicator vocabulary
+                        It additionally restricts the usage of the Applicator vocabulary
                         by forbidding the keyword prefixed with "unevaluated".  It also
                         describes a keyword, "localKeyord", that is not part of either
                         vocabulary.  Note that it is its own meta-schema,
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1094,7 +1094,7 @@
                         meta-schemas declares its corresponding vocabulary, this new
                         meta-schema must re-declare them for itself.  It would be
                         valid to leave the core vocabulary out of the "$vocabulary" keyword,
-                        but it MUST be referenced through the "allOf" keyword in order
+                        but it needs to be referenced through the "allOf" keyword in order
                         for its terms to be validated.  There is no special case for
                         validation of core keywords.
                     </postamble>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -807,49 +807,50 @@
 
         </section>
 
-        <section title='The "$schema" Keyword'>
-            <!-- TODO a custom $schema keyword might also be used to enforce minimum required functionality of a validator -->
-            <t>
-                The "$schema" keyword is both used as a JSON Schema version identifier and the
-                location of a resource which is itself a JSON Schema, which describes any schema
-                written for this particular version.
-            </t>
-            <t>
-                The value of this keyword MUST be a <xref target="RFC3986">URI</xref>
-                (containing a scheme) and this URI MUST be normalized.
-                The current schema MUST be valid against the meta-schema identified by this URI.
-            </t>
-            <t>
-                If this URI identifies a retrievable resource, that resource SHOULD be of
-                media type "application/schema+json".
-            </t>
-            <t>
-                The "$schema" keyword SHOULD be used in a root schema.
-                It MUST NOT appear in subschemas.
-            </t>
-            <t>
-                <cref>
-                    Using multiple "$schema" keywords in the same document would imply that the
-                    vocabulary and therefore behavior can change within a document.  This would
-                    necessitate resolving a number of implementation concerns that have not yet
-                    been clearly defined.  So, while the pattern of using "$schema" only in root
-                    schemas is likely to remain the best practice for schema authoring,
-                    implementation behavior is subject to be revised or liberalized in
-                    future drafts.
-                </cref>
-                <!--
-                    In particular, the process of validating an instance, including validating a
-                    schema as an instance against its meta-schema, only allows for a single set
-                    of rules across the entire instance document.  There is no equivalent of
-                    changing the meta-schema partway through the validation for non-schema
-                    instances.
-                  -->
-            </t>
-            <t>
-                Values for this property are defined in other documents and by other parties.
-                JSON Schema implementations SHOULD implement support for current and previous
-                published drafts of JSON Schema vocabularies as deemed reasonable.
-            </t>
+        <section title="Meta-Schemas and Vocabularies">
+            <section title='The "$schema" Keyword'>
+                <t>
+                    The "$schema" keyword is both used as a JSON Schema version identifier and the
+                    location of a resource which is itself a JSON Schema, which describes any schema
+                    written for this particular version.
+                </t>
+                <t>
+                    The value of this keyword MUST be a <xref target="RFC3986">URI</xref>
+                    (containing a scheme) and this URI MUST be normalized.
+                    The current schema MUST be valid against the meta-schema identified by this URI.
+                </t>
+                <t>
+                    If this URI identifies a retrievable resource, that resource SHOULD be of
+                    media type "application/schema+json".
+                </t>
+                <t>
+                    The "$schema" keyword SHOULD be used in a root schema.
+                    It MUST NOT appear in subschemas.
+                </t>
+                <t>
+                    <cref>
+                        Using multiple "$schema" keywords in the same document would imply that the
+                        vocabulary and therefore behavior can change within a document.  This would
+                        necessitate resolving a number of implementation concerns that have not yet
+                        been clearly defined.  So, while the pattern of using "$schema" only in root
+                        schemas is likely to remain the best practice for schema authoring,
+                        implementation behavior is subject to be revised or liberalized in
+                        future drafts.
+                    </cref>
+                    <!--
+                        In particular, the process of validating an instance, including validating a
+                        schema as an instance against its meta-schema, only allows for a single set
+                        of rules across the entire instance document.  There is no equivalent of
+                        changing the meta-schema partway through the validation for non-schema
+                        instances.
+                    -->
+                </t>
+                <t>
+                    Values for this property are defined in other documents and by other parties.
+                    JSON Schema implementations SHOULD implement support for current and previous
+                    published drafts of JSON Schema vocabularies as deemed reasonable.
+                </t>
+            </section>
         </section>
 
         <section title="Base URI and Dereferencing">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1057,10 +1057,10 @@
                         vocabularies, and combines their meta-schemas with an "allOf".
                         It additionally restricts the usage of the Applicator vocabulary
                         by forbidding the keyword prefixed with "unevaluated".  It also
-                        describes a keyword, "localKeyord", that is not part of either
+                        describes a keyword, "localKeyword", that is not part of either
                         vocabulary.  Note that it is its own meta-schema,
                         as it relies on both the Core vocabulary (as all schemas do)
-                        and the Applicator vocabulary (for "allOf"0.
+                        and the Applicator vocabulary (for "allOf").
                     </preamble>
                     <artwork>
 <![CDATA[
@@ -2424,7 +2424,7 @@ Link: <http://example.com/my-hyper-schema#>; rel="describedby"
                 </t>
                 <t>
                     When using the media type application/schema+json, the "schema" parameter
-                    MAY be supplied, If supplied, it SHOULD contain the same URI as identified
+                    MAY be supplied. If supplied, it SHOULD contain the same URI as identified
                     by the "$schema" keyword, and MAY contain additional URIs.  The "$schema"
                     URI MUST be considered the schema's canonical meta-schema, regardless
                     of the presence of alternative or additional meta-schemas as a media type
@@ -2564,7 +2564,7 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
                     <list>
                         <t>Type name: application</t>
                         <t>Subtype name: schema+json</t>
-                        <t>>Required parameters: N/A</t>
+                        <t>Required parameters: N/A</t>
                         <t>
                             Optional parameters:
                             <list style="hanging">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -790,7 +790,7 @@
                 <t>
                     A schema that itself describes a schema is called a meta-schema.
                     Meta-schemas are used to validate JSON Schemas and specify which vocabulary
-                    it is using.
+                    they are using.
                 </t>
                 <t>
                     Authors of extensions to JSON Schema are encouraged to write their own
@@ -819,7 +819,7 @@
             <t>
                 The role of the meta-schema is to constrain the structure of conforming schemas,
                 as well as simplify the process of composing multiple vocabularies into a usable
-                feature set.  Schema authoring is obviously expected to be a common activity, so
+                feature set.  Schema authoring is expected to be a common activity, so
                 schema authors need only understand how to reference a single meta-schema.
             </t>
             <t>
@@ -880,18 +880,16 @@
                 </t>
                 <t>
                     Values for this property are defined elsewhere in this and other documents,
-                    and by other parties.  JSON Schema implementations SHOULD implement support
-                    for current and previous published drafts of JSON Schema vocabularies
-                    as deemed reasonable.
+                    and by other parties.
                 </t>
             </section>
             <section title='The "$vocabulary" Keyword'>
                 <t>
-                    The "$vocabulary" keyword, like all keywords in meta-schemas,
-                    relates to the schema described by the meta-schema, rather than
-                    to the meta-schema itself.  Specifically, "$vocabulary" declares
+                    The "$vocabulary" keyword, which appears in a meta-schema, declares
                     what sets of keywords are expected to be used in schemas described
-                    by this meta-schema.
+                    by that meta-schema, and with what semantics.  This is conceptually
+                    analogous to how most other keywords used in meta-schemas describe
+                    the syntax of keywords used in schemas described by that meta-schema.
                 </t>
                 <t>
                     The value of this keyword MUST be an object.  The property names in the
@@ -939,6 +937,14 @@
                     must be repeated in the root of each schema document intended
                     for use as a meta-schema.  This is demonstrated in
                     <xref target="example-meta-schema">the example meta-schema</xref>.
+                </t>
+                <t>
+                    If "$vocabulary" is absent, an implementation MAY determine
+                    behavior based on the meta-schema if it is recognized from the
+                    URI value of the referring schema's "$schema" keyword.
+                    If the meta-schema, as referenced by the schema, is not recognized,
+                    then implementations MAY assume the use of all vocabularies in this
+                    specification and the companion Validation specification.
                 </t>
             </section>
             <section title="Detecting a Meta-Schema">
@@ -992,6 +998,11 @@
                     associated with the declared vocabularies describe.  This allows for
                     restricting usage to a subset of a vocabulary, and for validating
                     locally defined keywords not intended for re-use.
+                </t>
+                <t>
+                    However, meta-schemas SHOULD NOT contradict any vocabularies that
+                    they declare, such as by requiring a different JSON type than
+                    the vocabulary expects.  The resulting behavior is undefined.
                 </t>
                 <t>
                     Meta-schemas intended for local use, with no need to test for
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -941,78 +941,79 @@
                     meta-schemas.  All such declarations must be repeated in the root
                     of each schema document intended for use as a meta-schema.
                 </t>
-                <section title="Best Practices for Vocabulary and Meta-Schema Authors">
-                    <t>
-                        Meta-schema authors SHOULD NOT use "$vocabulary" to combine multiple
-                        vocabularies that define conflicting syntax or semantics for the same
-                        keyword.  As semantic conflicts are not generally detectable through
-                        schema validation, implementations are not expected to detect such
-                        conflicts.  If conflicting vocabularies are declared, the resulting
-                        behavior is undefined.
-                    </t>
-                    <t>
-                        Vocabulary authors SHOULD provide a meta-schema that validates the
-                        expected usage of the vocabulary's keywords.  Such meta-schemas
-                        SHOULD NOT forbid additional keywords, and MUST NOT forbid
-                        the "$id", "$schema", or "$vocabulary" keywords in the root schema.
-                    </t>
-                    <t>
-                        It is RECOMMENDED that meta-schema authors reference each vocabulary's
-                        meta-schema using the <xref target="allOf">"allOf"</xref> keyword,
-                        although other mechanisms for constructing the meta-schema may be
-                        appropriate for certain use cases.
-                    </t>
-                    <t>
-                        Meta-schemas MAY impose additional constraints, including describing
-                        keywords not present in any vocabulary, beyond what the meta-schemas
-                        associated with the declared vocabularies describe.  This allows for
-                        restricting usage to a subset of a vocabulary, and for validating
-                        locally defined keywords not intended for re-use.
-                    </t>
-                    <t>
-                        Meta-schemas intended for local use, with no need to test for
-                        vocabulary support in arbitrary implementations, can safely omit
-                        "$vocabulary" entirely.
-                    </t>
-                </section>
-                <section title="The JSON Schema Core Vocabulary">
-                    <t>
-                        Keywords declared in in this specification that begin with "$" make up
-                        the JSON Schema Core vocabulary.  These keywords are either required in
-                        order process any schema or meta-schema, including those split across
-                        multiple documents, or exist to reserve keywords for purposes that
-                        require guaranteed interoperability.
-                    </t>
-                    <t>
-                        The Core vocabulary MUST be considered mandatory at all times, in order
-                        to bootstrap the processing of further vocabularies.  Meta-schema
-                        authors MAY explicitly list the Core vocabulary, but do not need
-                        to do so.  If present, the Core vocabulary entry MUST have a value
-                        of true indicating that it is required.  The behavior of a value
-                        of false is undefined.
-                    </t>
-                    <t>
-                        The current URI for the Core vocabulary is:
-                        <eref target="https://json-schema.org/draft-08/vocabularies/core"/>.
-                    </t>
-                    <t>
-                        The current URI for the corresponding meta-schema is:
-                        <eref target="https://json-schema.org/draft-08/core"/>.
-                    </t>
-                </section>
-                <section title="Example Meta-Schema With Vocabulary Declarations">
-                    <figure>
-                        <preamble>
-                            This meta-schema explicitly declares both the Core and Applicator
-                            vocabularies, and combines their meta-schemas with an "allOf".
-                            It additional restricts the usage of the Applicator vocabulary
-                            by forbidding the keyword prefixed with "unevaluated".  It also
-                            describes a keyword, "localKeyord", that is not part of either
-                            vocabulary.  Note that it is its own meta-schema,
-                            as it relies on both the Core vocabulary (as all schemas do)
-                            and the Applicator vocabulary (for "allOf"0.
-                        </preamble>
-                        <artwork>
+            </section>
+            <section title="Best Practices for Vocabulary and Meta-Schema Authors">
+                <t>
+                    Meta-schema authors SHOULD NOT use "$vocabulary" to combine multiple
+                    vocabularies that define conflicting syntax or semantics for the same
+                    keyword.  As semantic conflicts are not generally detectable through
+                    schema validation, implementations are not expected to detect such
+                    conflicts.  If conflicting vocabularies are declared, the resulting
+                    behavior is undefined.
+                </t>
+                <t>
+                    Vocabulary authors SHOULD provide a meta-schema that validates the
+                    expected usage of the vocabulary's keywords.  Such meta-schemas
+                    SHOULD NOT forbid additional keywords, and MUST NOT forbid
+                    the "$id", "$schema", or "$vocabulary" keywords in the root schema.
+                </t>
+                <t>
+                    It is RECOMMENDED that meta-schema authors reference each vocabulary's
+                    meta-schema using the <xref target="allOf">"allOf"</xref> keyword,
+                    although other mechanisms for constructing the meta-schema may be
+                    appropriate for certain use cases.
+                </t>
+                <t>
+                    Meta-schemas MAY impose additional constraints, including describing
+                    keywords not present in any vocabulary, beyond what the meta-schemas
+                    associated with the declared vocabularies describe.  This allows for
+                    restricting usage to a subset of a vocabulary, and for validating
+                    locally defined keywords not intended for re-use.
+                </t>
+                <t>
+                    Meta-schemas intended for local use, with no need to test for
+                    vocabulary support in arbitrary implementations, can safely omit
+                    "$vocabulary" entirely.
+                </t>
+            </section>
+            <section title="The JSON Schema Core Vocabulary">
+                <t>
+                    Keywords declared in in this specification that begin with "$" make up
+                    the JSON Schema Core vocabulary.  These keywords are either required in
+                    order process any schema or meta-schema, including those split across
+                    multiple documents, or exist to reserve keywords for purposes that
+                    require guaranteed interoperability.
+                </t>
+                <t>
+                    The Core vocabulary MUST be considered mandatory at all times, in order
+                    to bootstrap the processing of further vocabularies.  Meta-schema
+                    authors MAY explicitly list the Core vocabulary, but do not need
+                    to do so.  If present, the Core vocabulary entry MUST have a value
+                    of true indicating that it is required.  The behavior of a value
+                    of false is undefined.
+                </t>
+                <t>
+                    The current URI for the Core vocabulary is:
+                    <eref target="https://json-schema.org/draft-08/vocabularies/core"/>.
+                </t>
+                <t>
+                    The current URI for the corresponding meta-schema is:
+                    <eref target="https://json-schema.org/draft-08/core"/>.
+                </t>
+            </section>
+            <section title="Example Meta-Schema With Vocabulary Declarations">
+                <figure>
+                    <preamble>
+                        This meta-schema explicitly declares both the Core and Applicator
+                        vocabularies, and combines their meta-schemas with an "allOf".
+                        It additional restricts the usage of the Applicator vocabulary
+                        by forbidding the keyword prefixed with "unevaluated".  It also
+                        describes a keyword, "localKeyord", that is not part of either
+                        vocabulary.  Note that it is its own meta-schema,
+                        as it relies on both the Core vocabulary (as all schemas do)
+                        and the Applicator vocabulary (for "allOf"0.
+                    </preamble>
+                    <artwork>
 <![CDATA[
 {
   "$schema": "https://json-schema.org/draft-08/core-app-example#",
@@ -1037,26 +1038,25 @@
   }
 }
 ]]>
-                        </artwork>
-                        <postamble>
-                            As shown above, even though each of the referenced standard
-                            meta-schemas declares its corresponding vocabulary, this new
-                            meta-schema must re-declare them for itself.  It would be
-                            valid to leave the core vocabulary out of the "$vocabulary" keyword,
-                            but it MUST be referenced through the "allOf" keyword in order
-                            for its terms to be validated.  There is no special case for
-                            validation of core keywords.
-                        </postamble>
-                    </figure>
-                    <t>
-                        The standard meta-schemas that combine all vocabularies defined by
-                        the Core and Validation specification, and that combine all vocabularies
-                        defined by those specifications as well as the Hyper-Schema specification,
-                        demonstrate additional complex combinations.  These URIs for these
-                        meta-schemas may be found in the Validation and Hyper-Schema specifications,
-                        respectively.
-                    </t>
-                </section>
+                    </artwork>
+                    <postamble>
+                        As shown above, even though each of the referenced standard
+                        meta-schemas declares its corresponding vocabulary, this new
+                        meta-schema must re-declare them for itself.  It would be
+                        valid to leave the core vocabulary out of the "$vocabulary" keyword,
+                        but it MUST be referenced through the "allOf" keyword in order
+                        for its terms to be validated.  There is no special case for
+                        validation of core keywords.
+                    </postamble>
+                </figure>
+                <t>
+                    The standard meta-schemas that combine all vocabularies defined by
+                    the Core and Validation specification, and that combine all vocabularies
+                    defined by those specifications as well as the Hyper-Schema specification,
+                    demonstrate additional complex combinations.  These URIs for these
+                    meta-schemas may be found in the Validation and Hyper-Schema specifications,
+                    respectively.
+                </t>
             </section>
         </section>
 
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -810,9 +810,9 @@
         <section title="Meta-Schemas and Vocabularies">
             <section title='The "$schema" Keyword'>
                 <t>
-                    The "$schema" keyword is both used as a JSON Schema version identifier and the
-                    location of a resource which is itself a JSON Schema, which describes any schema
-                    written for this particular version.
+                    The "$schema" keyword is both used as a JSON Schema feature set identifier and
+                    the location of a resource which is itself a JSON Schema, which describes any
+                    schema written for this particular feature set.
                 </t>
                 <t>
                     The value of this keyword MUST be a <xref target="RFC3986">URI</xref>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -808,6 +808,44 @@
         </section>
 
         <section title="Meta-Schemas and Vocabularies">
+            <t>
+                Two concepts, meta-schemas and vocabularies, are used to inform an implementation
+                how to interpret a schema.  A schema S declares its meta-schema M with the "$schema"
+                keyword, and meta-schemas declare vocabularies with the "$vocabulary" keyword.
+                The vocabularies declared in M are those that are expected to be used in S.
+                The meta-schema M may itself use a different set of vocabularies, which are
+                declared in its own meta-schema, M'.
+            </t>
+            <t>
+                The role of the meta-schema is to constrain the structure of conforming schemas,
+                as well as simplify the process of composing multiple vocabularies into a usable
+                feature set.  Schema authoring is obviously expected to be a common activity, so
+                schema authors need only understand how to reference a single meta-schema.
+            </t>
+            <t>
+                Meta-schema authoring is an advanced usage of JSON Schema, so the design of
+                meta-schema features emphasizes flexibility over simplicity.
+            </t>
+            <t>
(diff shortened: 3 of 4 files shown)


core.json
```diff
@@ -0,0 +1,50 @@
...
+                "format": "uri"
+            },
+            "additionalProperties": {
+                "type": "boolean"
+            }
+        },
+        "$comment": {
+            "type": "string"
+        },
+        "$defs": {
```
gregsdennis: In the PR where we added this, it was determined that we want to keep `definitions` for backward compatibility.  Would that go here or in another meta-schema component?

applicator.json
```diff
@@ -0,0 +1,60 @@
...
+        "propertyNames": { "$recursiveRef": "#" },
+        "if": { "$recursiveRef": "#" },
+        "then": { "$recursiveRef": "#" },
+        "else": { "$recursiveRef": "#" },
+        "allOf": { "$ref": "#/$defs/schemaArray" },
+        "anyOf": { "$ref": "#/$defs/schemaArray" },
+        "oneOf": { "$ref": "#/$defs/schemaArray" },
+        "not": { "$recursiveRef": "#" }
+    }
+}
```
gregsdennis: Are *assertions.json* and *annotations.json* documents to come in the future?  Has any thought been put to how these sub-meta-schemas will be combined to represent the *draft-08* (or whatever it'll be named) meta-schema as an atomic entity?

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+                vocabularies to be combined in different ways, and for meta-schema authors
+                to impose additional constraints such as forbidding certain keywords, or
+                performing unusually strict syntactical validation, as might be done
+                during a development and testing cycle.
             </t>
             <t>
-                Values for this property are defined in other documents and by other parties.
-                JSON Schema implementations SHOULD implement support for current and previous
-                published drafts of JSON Schema vocabularies as deemed reasonable.
+                A schema is considered to be processed as meta-schema if it was
```
gregsdennis: How would processing of a meta-schema differ from processing any other schema?  I ask because Manatee.Json processes all of the *draft-0x* meta-schemas like any other schema (partially to ensure self-validation).

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+                    When processing a schema that uses unrecognized vocabularies, keywords
+                    declared by those vocabularies are treated like any other unrecognized
+                    keyword, and ignored.
+                </t>
+                <t>
+                    The "$vocabulary" keyword SHOULD be used in the root schema of any schema
+                    document intended for use as a meta-schema.  It MUST NOT appear in subschemas.
+                </t>
+                <t>
+                    The "$vocabulary" keyword MUST be ignored in a schema documents that
```
gregsdennis: documents / document

applicator.json
```diff
@@ -0,0 +1,60 @@
+{
+    "$schema": "http://json-schema.org/draft-08/schema#",
+    "$id": "https://json-schema.org/draft-08/applicator",
+    "$recursiveAnchor": true,
+    "$vocabulary": {
+        "https://json-schema.org/draft-08/vocabularies/applicator": true
```
gregsdennis: I'm confused by this.  Does the URI here imply usage or definition?  If usage, why is the *applicator.json* meta-schema declaring it uses itself?  If definition, why is the value an object (what case would there be to define multiple vocabularies in the same document)?

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+                        order process any schema or meta-schema, including those split across
+                        multiple documents, or exist to reserve keywords for purposes that
+                        require guaranteed interoperability.
+                    </t>
+                    <t>
+                        The Core vocabulary MUST be considered mandatory at all times, in order
+                        to bootstrap the processing of further vocabularies.  Meta-schema
+                        authors MAY explicitly list the Core vocabulary, but do not need
+                        to do so.  If present, the Core vocabulary entry MUST have a value
+                        of true indicating that it is required.  The behavior of a value
```
gregsdennis: Why have a boolean value at all if the value must be `true` and `false` is undefined?  Why not just have an array of URIs?

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+                        The current URI for the corresponding meta-schema is:
+                        <eref target="https://json-schema.org/draft-08/core"/>.
+                    </t>
+                </section>
+                <section title="Example Meta-Schema With Vocabulary Declarations">
+                    <figure>
+                        <preamble>
+                            This meta-schema explicitly declares both the Core and Applicator
+                            vocabularies, and combines their meta-schemas with an "allOf".
+                            It additional restricts the usage of the Applicator vocabulary
```
gregsdennis: ~additional~ / additionally

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+                            by forbidding the keyword prefixed with "unevaluated".  It also
+                            describes a keyword, "localKeyord", that is not part of either
+                            vocabulary.  Note that it is its own meta-schema,
```
gregsdennis: > Note that it is its own meta-schema...

This is a nice thing to have for this example.  Can we use the same language to describe the same idea for *applicators.json* and *core.json*?  Not sure where it'd go, but it seems necessary.

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+]]>
+                        </artwork>
+                        <postamble>
+                            As shown above, even though each of the referenced standard
+                            meta-schemas declares its corresponding vocabulary, this new
+                            meta-schema must re-declare them for itself.  It would be
+                            valid to leave the core vocabulary out of the "$vocabulary" keyword,
+                            but it MUST be referenced through the "allOf" keyword in order
+                            for its terms to be validated.  There is no special case for
+                            validation of core keywords.
```
gregsdennis: If *core* can be left out of `$vocabularies` do we need *applicators*?  Wouldn't that also be inferred through the `allOf`?

core.json
handrews: @gregsdennis no, `definitions` and also `dependencies` will only appear in the draft-08 versions of the  existing "schema" and "hyper-schema" documents.  Backwards compatibility here means "backwards compatibility for users of the meta-schemas that used to contain them".

Those keywords are not part of the proper vocabularies.  There were no formal vocabularies before draft-08, so there is nothing requiring compatibility.

handrews: @gregsdennis we could also make a tiny compatibility vocabulary for the "schema" meta-schema that contained those two keywords, and then implementations that want to offer compatibility can check for that vocabulary.

I'm kind of waiting to see if this approach gets a green light at all before sorting out the details of how many vocabularies we have and what's in them.

applicator.json
handrews: @gregsdennis yes, there's just rather a lot to sorting out the exact vocabulary granularities and compositions, so I want to do that separately, preferably after getting at least some buy in / revisions on this PR.

jsonschema-core.xml
handrews: At this point, it's mostly just "if it's a meta-schema, you MUST stop processing if you don't recognize a required vocabulary in `$vocabulary`, otherwise you MUST ignore `$vocabulary`"

If you have a meta-schema M that requires vocabs A, B, and C, and it's own meta-schema M' requires vocabs A, C, and D (meaning that M itself uses A, C, and D, even though it describes A, , and C), then you need to be able to validate M without understanding B.  M describes B, but does not require it.  M' requires D, even though M only uses but does not describe it.

handrews: I say "mostly" because I remember AJV doing something else different, but I think that had to do with it getting confused about needing the meta-schema already validated before it could validate the meta-schema or some other weird circular thing that I don't think the spec requires.

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+            </section>
+            <section title='The "$vocabulary" Keyword'>
+                <t>
+                    The "$vocabulary" keyword is used within meta-schemas to declare
+                    the sets of keywords that schemas described by the meta-schema are
+                    expected to use.
+                </t>
+                <t>
+                    The value of this keyword MUST be an object.  The property names in the
+                    object MUST be URIs, (containing a scheme) and this URI MUST be normalized.
```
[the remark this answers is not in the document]
handrews: Nope.  URI start with schemEs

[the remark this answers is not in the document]
handrews: This wording is used in many places in the spec.

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
             </t>
+            <section title='The "$schema" Keyword'>
+                <t>
+                    The "$schema" keyword is both used as a JSON Schema feature set identifier and
+                    the location of a resource which is itself a JSON Schema, which describes any
+                    schema written for this particular feature set.
+                </t>
+                <t>
+                    The value of this keyword MUST be a <xref target="RFC3986">URI</xref>
+                    (containing a scheme) and this URI MUST be normalized.
```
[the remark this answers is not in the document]
handrews: Nope.

jsonschema-core.xml
handrews: Yeah, although I think I'd rather remove the "a" and go with "in schema documents"- does that make sense?  I'm never sure what's better with this sort of thing.

applicator.json
handrews: @gregsdennis great questions that get to the heart of why this was such a difficult PR to write up- I expect to have to revise this before it makes sense to people so I'm glad you're focusing on this.

As covered in the intro section with the S, M, M' example, where M is the meta-schema for S, and M' is the meta-schema for M, the vocabularies in `$vocabulary` within M are declaring the expected usage within S.  Meta-schemas describe the expected / allowed behavior of schemas that conform to them.

The whole reason for this is that meta-schemas will often declare usage of multiple vocabularies.  It's extremely rare for only one vocabulary to be sufficient.  The "schema" meta-schema will require at least four, counting core, and the "hyper-schema" meta-schema will require at least five (it depends on what we decide to do with `format` and/or `content*`, which I'll work on after we establish the basics here).

So, most people will never write meta-schemas.  If they do, it will probably be fairly rare.  Many people will write schemas.  Meta-schemas have to deal with the complexity of assembling a usable overall vocabulary from the fine-grained re-usable component vocabularies.  Schema authors do not deal with that- they reference a meta-schema that is documented to do whatever it is they are trying to do.  Most schema authors and users should not care about vocabularies directly, most of the time.

handrews: @gregsdennis there is a specific point that the core vocabulary need not be explicitly listed.  I'm of two minds on that, maybe we just say that it MUST always be present and it MUST always have a value of `true`.

jsonschema-core.xml
handrews: will fix, thanks

jsonschema-core.xml
handrews: The value of `false` has undefined behavior **for the core vocabulary only**.  The behaviors of the booleans is defined in general earlier in the document.

You can't not use the core vocabulary- you need it to bootstrap processing (assuming you don't just hardcode it, which you can do because it's mandatory and I expect most people will).  Basically, I expect implementations to ignore whether core is present in `$vocabulary` but some people will like having it there.  We could either say that `false` is an error or just say that if you do this and weird stuff happens it's your own fault.  I prefer the latter.

jsonschema-core.xml
handrews: I don't actually understand what you're asking here.  The example `$ref`s both the core and applicator meta-schemas, so it is already validating `$vocabulary` which is present in the core vocabulary's meta-schema.

I'm not sure what language you are looking to re-use from here to those other meta-schemas.  They are neither restricting vocabulary terms nor are they describing extra keywords.  Nor are either of them their own meta-schemas (they require keywords like `type`, for example).  So these four lines do not seem to apply.

jsonschema-core.xml
handrews: `$vocabulary` not `$vocabularies` – someone at some point suggested that it should be singular like `type`.  I don't have my heart set on either form, but I decided that was a reasonable analogy, and the set of individual vocabularies listed under `$vocabulary` are combined to form _the_ vocabulary described by the meta-schema, so it seemed to work.

Anyway, on to your actual question :-)

We need _applicators_ because that whole thing was a balancing act between people who believed that applicators are less replaceable and more fundamental than assertions (hence moving them to core) and people who thought they were also totally replaceable (hence making them a separate vocabulary that you must explicitly include).

I'm not re-litigating that.  It was not as traumatic as `unevaluated*` but only because it didn't take multiple years to sort out.

handrews: @gregsdennis 

> Wouldn't that also be inferred through the allOf?

I tried to work out how this would be implemented once, and the general case is challenging.  You need to determine whether there is a `$ref` to the meta-schema within any dynamic scope chain that starts in the root schema object of the entry point schema, and consists only of in-place applicators (including `$ref` and `$recursiveRef`).  This means that you have to understand which keywords are in-place applicators, including those created by 3rd parties as extensions.

If you have a meta-schema that does a lot of extending and composing, this gets hard to spot for end-users as well.

jsonschema-core.xml
gregsdennis: Another question: What would signify that a schema is a meta-schema so that an implementation would know to process it differently?

applicator.json
gregsdennis: It sounds like `$vocabularies` is kind of an equivalent of the JS `imports` statement for schema keywords.  Unless a keyword is defined within one of the "imported" meta-schemas, a validator will ignore it.

Do you think we're getting close to compiler-like behavior here?

jsonschema-core.xml
gregsdennis: Wait... is it its own meta-schema because it declares itself in `$schema`?  If that's the case, then it has to link back to the *draft-08* meta-schemas somehow, which is the purpose of `$vocabulary`.  It's like saying, "This meta-schema is self-validating, but also must validate against these other meta-schemas."

I think I'm starting to get this.

jsonschema-core.xml
gregsdennis: So the `$vocabulary` (see, I did it right) entries are used to validate *this* meta-schema, whereas the `allOf` entries are used to validate schemas that *use* this meta-schema, correct?  I'm not sure the reason why *core* is required (that schemas that use this meta-schema need it) is clear from this paragraph.

handrews: * A single URI-identified vocabulary reserves some set of keywords, with the semantics defined by whatever specification document defined the vocabulary.  Maybe that's one of our drafts, someone else's W3C recommendation, or some wiki page on your local intranet.  As it stands, the keyword semantics have to be interpreted by a human.
* Putting a URI-identified vocabulary in the `$vocabulary` set of a meta-schema M says that _schemas that declare M as their meta-schema:_ `"$schema": M` are expected to use those keywords according to the vocabulary's semantics
* That meta-schema M must **also** decide what possible schemas should be validated for this combination of vocabularies, and to what degree.
    * It could just compose the meta-schemas published by the vocabulary authors (which should be the minimal reasonable set of validation rules for the vocabulary- the _core.json_ and _applicator.json_ meta-schemas illustrate this.
    * It could be a much more strict set of rules, for example applying `"unevaluatedProperties": false` all over the place to forbid further extension and catch spelling errors.  This is particularly common for test environments, but also some applications aren't interested in evolution or compatibility, they are interested in reliability.
    * It could be a much more lax set of rules- maybe you need to conform to only one or the other vocabulary's meta-schema.  I have no use case for this, but it came up once, I forget why.
  

So, every keyword in a meta-schema **describes the schemas**.  Except `$schema`, which declares the meta-schema's meta-schema.

`$vocabulary`, however, is _not_ and exception.  Just as how `"type": "object", "properties": {...}` applies to the schema (not the meta-schema), `$vocabulary` is a usage declaration ***for the schemas*** not the meta-schema.

I should find some way to emphasize how this is not unusual behavior.  Everything in a schema document affects the document it describes.   In the case of metas-schemas, those documents are schema.  `$schema` is the only exception, identifying a meta-schema that applies to the document containing `$schema`.

jsonschema-core.xml
handrews: @gregsdennis yes!

There's some convoluted bootstrapping, and a certain amount of "yeah, probably everyone needs to understand some minimal vocabulary set" that we'll work out.

Meta-schemas are not vocabularies.  And vocabularies are not meta-schemas.  So you will never see a vocabulary URI in `$schema`, nor will you see a meta-schema URI in `$vocabulary`.

They collect vocabularies to inform an implementation what they're doing, because it's really quite complex to infer such a thing out of the meta-schema itself.  It's not exactly impossible but it's hard.

In the case of the example, it happens to work that it only needs core and applicator, and the extra constraints (restricting `unevaluated*` and describing a locally defined keyword) don't get in the way of what it needs.  So its `$id` and `$schema` are the same, and neither are directly related to the contents of `$vocabulary`.

jsonschema-core.xml
handrews: @gregsdennis annotations.  Hyper-Schema is a perfect example.  I want to have the hyper-schema meta-schema declare that schemas that `$schema`-link it make use of the core, applicator, assertion, basic annotation, and hyper-schema vocabularies.  The hyper-schema vocabulary is just `links`, `base`, and the sub-keywords inside LDOs.  If you feed a hyper-schema to a validator that has no idea what hyper-schema is, I still want the validator to work.  The hyper-schema keywords do not change standard validation at all.  So one variation on the "hyper-schema" meta-schema will have something like:

```JSON
{
    "https://json-schema.org/draft-08/vocabularies/core": true,
    "https://json-schema.org/draft-08/vocabularies/applicator": true,
    "https://json-schema.org/draft-08/vocabularies/assertions": true,
    "https://json-schema.org/draft-08/vocabularies/basic-annotations": false,
    "https://json-schema.org/draft-08/vocabularies/hyper-schema": false
}
```

Someone might want to use the same meta-schema but with a `$vocabulary` where all four of those are `true`, because they know they are only using it where hyper-schema and annotations like `readOnly` are critical.

So we can write a meta-schema that validates all of these, and then publish a meta-schema that references it with this set of vocabularies (mandatory validation, optional annotations and hyper-schema), and one that makes them all mandatory.  Different options will be preferred for different use cases.

applicator.json
handrews: It is sort of compiler-like, but let me know if you have any other ideas on how to support machine-detectible re-usuable chunks of functionality.  I've been poking at this idea for most of a year, and no one has proposed a viable alternative of which I am aware (and really I think you're the only person who has given a large amount of feedback - thanks for that!)

I will say that the OpenAPI steering group is very interested in this feature as I have sold them on it as a way to solve their persistent code generation tooling problems.

jsonschema-core.xml
handrews: As the line you attached this thread to says:

>  A schema is considered to be processed as meta-schema if it was identified as such in a schema's "$schema".  Implementations MAY also offer alternate ways to process schemas using meta-schema rules.

So if you are processing a schema (which is generally how things start), and you follow a `$schema` URI to another document, that document is a meta-schema in this context because it was the direct target of `$schema`.  So `$vocabulary` is respected in its root schema (but nowhere else).

gregsdennis: So the only way to know whether a schema is a meta-schema is to know if it's used as a meta-schema by another schema? (Say "schema" some more.)  What if I start with the meta-schema?  Is it just processed like any other schema?

applicator.json
gregsdennis: No worries.  I just brought up the compiler thing because we've historically tried to steer away from logic/behavior.

jsonschema-core.xml
gregsdennis: So if my implementation doesn't know what `basic-annotations` is, it just leaves out those annotations?  This means that implementations will need to expose a way to register new vocabularies by their spec (e.g. `https://json-schema.org/draft-08/vocabularies/basic-annotations`) that then links them to their associated meta-schema (e.g. `https://json-schema.org/draft-08/basic-annotations`).  Only then will my implementation know what to do with that.

Also, the `false` allows my implementation to continue if it doesn't know about `basic-annotations`, but it should return an error if it doesn't know about `assertions`.

Is all that right?

jsonschema-core.xml
```diff
@@ -800,49 +800,257 @@
...
+                        <artwork>
+<![CDATA[
+{
+  "$schema": "https://json-schema.org/draft-08/core-app-example#",
+  "$id": "https://json-schema.org/draft-08/core-app-example",
+  "$recursiveAnchor": true,
+  "$vocabulary": {
+    "https://json-schema.org/draft-08/vocabularies/core": true,
+    "https://json-schema.org/draft-08/vocabularies/applicator": true
+  }.
```
gregsdennis: period vs comma

jsonschema-core.xml
gregsdennis: If `$vocabulary` applies to the schema, not the meta-schema that declares it, what's the difference between that and the `allOf` in the example?  Why have both?

It seems to me that the vocabularies could be inferred from the `allOf` contents.

Conversely, the validation facilitated by the `allOf` could proceed by interpreting the vocabularies declared.

jsonschema-core.xml
handrews: As the text says:

> Implementations MAY also offer alternate ways to process schemas using meta-schema rules.

I don't mean to be a jerk here, but I don't understand what's not clear.  There are two ways that I know of that implementations process meta-schemas:

* They read `"$schema"` and either fetch the resource or load it from cache.  You obviously know it is a meta-schema because... `"$schema"`  Is there anything confusing about this part?
* There is some way to clearly hand a meta-schema to an implementation and tell it to treat it that way.  I don't know why you would do that, how, or what purpose it would serve.  The point is to allow an out for people who want to take it because, I dunno, they process meta-schemas up-front and decide whether they are supported by checking `$vocabulary`, and then don't process them as individual schemas come in.  I don't know, and the point is that I'm not preventing anyone from doing whatever they need like that.

If you start with the meta-schema as your schema, then it would not seem to fall into either of those two categories, so yes, it's being treated as a schema.

applicator.json
handrews: @gregsdennis I'm not sure how this qualifies- let me know if I'm missing something that you think we should consider.  There is a decision about whether the implementation supports the vocabularies that the meta-schema needs to have supported, but in many cases `$schema` itself is treated that way already.  Some implementations won't process schemas with unrecognized meta-schemas.

jsonschema-core.xml
handrews: @gregsdennis awesome, I will try to figure out how to clarify that point.

Thanks for all of your feedback, I knew posting it that it would be too confusing but I couldn't figure out how to make it better on my own.

applicator.json
```diff
@@ -0,0 +1,60 @@
...
+    },
+
+    "title": "Applicator vocabulary meta-schema",
+    "$defs": {
+        "schemaArray": {
+            "type": "array",
+            "minItems": 1,
```
vearutop: Should `minItems` be defined in this vocabulary or it should be included with another vocabulary?

handrews: @vearutop it is included in another vocabulary, specifically the assertions vocabulary defined in the Validation spec.  Note that this document, _applicator_, still uses _schema_ as its meta-schema.  Once I add `$vocabulary` to _schema_ it will reference the assertion vocabulary, making it available for use in _applicator_.

This is why _applicator_ is not its own meta-schema.  It uses more keywords than it defines.

handrews:
As I've thought about this more, I've decided to go ahead and try writing up a vocabulary definition file, [for reasons](https://github.com/json-schema-org/json-schema-spec/issues/602#issuecomment-437949419).

applicator.json
handrews: @gregsdennis I think I'm leaning towards your position that the core vocabulary MUST be explicitly included and set to true, and that the behavior of anything else is undefined (but it is RECOMMENDED that an error be raised).

It's slightly annoying but arguably no more annoying that anything else in this complex feature, and we can always relax it later easily enough.  Consistency is probably better for now.

handrews: I've now made it required.

jsonschema-core.xml
handrews: @gregsdennis I think I have fixed/clarified this, please let me know / mark the conversation resolved if the new text works.

handrews:
In the most recent commit, I have tried to make it clear at the start of the `$vocabulary` section that `$vocabulary`, like pretty much all keywords in meta-schemas, applies to the schemas that the meta-schema describes.  It is actually quite normal in that respect.

jsonschema-core.xml
gregsdennis: I can't mark it resolved.

jsonschema-core.xml
gregsdennis: I think this is resolved based on our offline conversation.

jsonschema-core.xml
handrews: @gregsdennis I attempted to address this in the most recent version, so I'm marking this thread as resolved and we can start a new conversation attached to the current spec lines for whatever I did not manage to address fully.

applicator.json
handrews: Marked resolved because I assume that with no further reply for two weeks that my answer was sufficient :-)

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
             </t>
             <t>
-                The "$schema" keyword SHOULD be used in a root schema.
-                It MUST NOT appear in subschemas.
+                The role of a vocabulary is to declare which keywords (including sub-keywords
+                such as those in JSON Hyper-Schema's Link Description Object) are in use,
+                and with which semantics.  The semantics are indicated by the document
+                that publicizes the vocabulary URI.  At this time, there is no machine-readable
+                description of keywords other than validation rules, which appear in the
+                meta-schema.
```
gregsdennis: > At this time, there is no machine-readable description of keywords other than validation rules, which appear in the meta-schema.

👍

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
+                    When processing a schema that uses unrecognized vocabularies, keywords
+                    declared by those vocabularies are treated like any other unrecognized
+                    keyword, and ignored.
+                </t>
+                <t>
+                    The "$vocabulary" keyword SHOULD be used in the root schema of any schema
+                    document intended for use as a meta-schema.  It MUST NOT appear in subschemas.
+                </t>
+                <t>
+                    The "$vocabulary" keyword MUST be ignored in schema documents that
```
jgonzalezdr: I like this requirement because it allows using the `$vocabulary` keyword in "leaf" schemas: at the end, "leaf" schemas are also associated in an abstract way with vocabulary specs (lexicon + grammar + semantics) that define how to process instances, e.g. the schemas associated with a bug management application (bugzilla-like) could be associated to vocabularies for executing commands, capturing, storing and displaying enriched text, etc.

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
+                <t>
+                    Values for this property are defined elsewhere in this and other documents,
+                    and by other parties.  JSON Schema implementations SHOULD implement support
+                    for current and previous published drafts of JSON Schema vocabularies
+                    as deemed reasonable.
+                </t>
+            </section>
+            <section title='The "$vocabulary" Keyword'>
+                <t>
+                    The "$vocabulary" keyword, like all keywords in meta-schemas,
```
jgonzalezdr: Are you sure about this? My impression is that "$ keywords" in a schema relate to the schema itself, and "non-$ keywords" relate to the instance. The same for meta-schema vs schema.

Therefore, `$vocabulary` would be the first "$ keyword" relating to the instance (schema) instead of the schema (meta-schema). Design smell?

Relequestual:
FYI, looks like you don't now need to review one commit at a time (Could be wrong. Let's find out.)

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
+  },
+  "properties": {
+    "$comment": "Not in vocabulary, but validated if used",
+    "localKeyword": {
+      "type": "string"
+    }
+  }
+}
+]]>
+                    </artwork>
```
Relequestual: "artwork" =]

jsonschema-core.xml
jgonzalezdr: However, `$vocabulary` has effect on the interpretation of the instance (which circumstantially for meta-schemas is an schema).

If we consider meta-schemas a different kind of animal than schemas, then according to the rule of three $-keywords have no effect on the meta-schema.

Relequestual: `$schema` also has an effect on the interpretation of the instance. I don't see how this is different...

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
+                    The behavior of a false value for this vocabulary (and only this
+                    vocabulary) is undefined, as is the behavior when "$vocabulary"
+                    is present but the Core vocabulary is not included.  However, it
+                    is RECOMMENDED that implementations detect these cases and raise
+                    an error when they occur.
+                </t>
+                <t>
+                    Meta-schemas that do not use "$vocabulary" MUST be considered to
+                    require the Core vocabulary just as if "$vocabulary" were used
+                    with only the Core vocabulary present with a value of true.
```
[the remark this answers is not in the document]
jgonzalezdr: Good point here.

Maybe to add some backward compatibility, it may be additionally stated that if `$vocabulary` is not present implementations MAY try to recognize a supported vocabulary from the value of `$id`.

jsonschema-core.xml
handrews: I'll take a look at the wording.

I don't think anyone is in a disagreement over the `$` prefix.

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
+                    to the meta-schema itself.  Specifically, "$vocabulary" declares
+                    what sets of keywords are expected to be used in schemas described
+                    by this meta-schema.
+                </t>
+                <t>
+                    The value of this keyword MUST be an object.  The property names in the
+                    object MUST be URIs, (containing a scheme) and this URI MUST be normalized.
+                    The URI MAY be a URL, but the nature of the retrievable resources is
+                    currently undefined, and reserved for future use.  Vocabulary authors
+                    SHOULD NOT serve a document at that URL.  A server MAY respond with
```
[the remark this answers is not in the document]
handrews: @Relequestual sounds good, let me know if you decide it's clear or not.

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
+                <t>
+                    It is RECOMMENDED that meta-schema authors reference each vocabulary's
+                    meta-schema using the <xref target="allOf">"allOf"</xref> keyword,
+                    although other mechanisms for constructing the meta-schema may be
+                    appropriate for certain use cases.
+                </t>
+                <t>
+                    Meta-schemas MAY impose additional constraints, including describing
+                    keywords not present in any vocabulary, beyond what the meta-schemas
+                    associated with the declared vocabularies describe.  This allows for
```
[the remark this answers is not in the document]
handrews: Yeah, that's actually a really good example.  You can describe those in a meta-schema using `patternProperties` (if you need to exclude them from a `false` `additionalProperties` or `unevaluatedProperties` which is how OAS works), but you can't declare them in a vocabulary because you don't know what keywords will actually be used in practice.  I might try to incorporate this concept in b/c it shows why this ability is needed.

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: @Relequestual it might be good to have a "MAY assume..." in there.  As written, the intent is that without `$vocabulary` everything works exactly as it does today.  In practice, most implementations assume the vocabularies from the core and validation specs (which is a larger number of vocabularies than strictly the core and validation assertion vocabulary).

@jgonzalezdr saying that they MAY try to do something based on the meta-schema's `$id` is the same as tying the behavior to the value of the schema's `$schema`, _which is exactly what already happens_.  We do not need to re-state it.  It's covered in the section on `$schema`.

jsonschema-core.xml
handrews: ?

Is this just a random amusement smiley or did I misspell artwork?  I'm staring at it and don't see a problem.

jsonschema-core.xml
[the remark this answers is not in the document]
gregsdennis: To @Relequestrial's comment:

I don't think that the validation vocabulary need be required, even if a schema is for validation.

A vocabulary is a just set of keywords. If I want to create a validation vocabulary with a completely disjoint set of keywords, I shouldn't be required to also include the standard set.  Perhaps my vocabulary *isn't* disjoint. Maybe it overlaps and redefines a number of the keywords. I'm that case, it would be explicitly wrong of me to include the standard set due to conflicts.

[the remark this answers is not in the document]
jgonzalezdr: @handrews: Do you mean that this is covered by the last paragraph of the 'The "$schema" Keyword' section?

jsonschema-core.xml
```diff
@@ -800,49 +800,291 @@
...
+                        schema as an instance against its meta-schema, only allows for a single set
+                        of rules across the entire instance document.  There is no equivalent of
+                        changing the meta-schema partway through the validation for non-schema
+                        instances.
+                    -->
+                </t>
+                <t>
+                    Values for this property are defined elsewhere in this and other documents,
+                    and by other parties.  JSON Schema implementations SHOULD implement support
+                    for current and previous published drafts of JSON Schema vocabularies
```
jgonzalezdr: Section "Meta-Schemas and Vocabularies" is really fine, and it describes very well the separation of roles between meta-schemas and vocabularies. However, I find a bit confusing indicating in this section (related to meta-schemas) something regarding implementation of vocabularies.

This paragraph is carried over almost without changes from draft-05, where meta-schemas and vocabularies were not differentiated, and I find this wording confusing in the new context.

I don't really understand if it is pretended that implementations should first try to recognize the "feature set" from the plain `$schema` value, then if it is not recognized as a fallback they should analyze the meta-schema and check the feature set from the meta-schema's `$vocabulary`; or the other way around.

handrews: Ah, good catch, thanks!  I will clarify.

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: @gregsdennis the assumptions are only allowed when `$vocabulary` is **completely absent**.  Currently, there is no `$vocabulary` and pretty much ever implementation of which I'm aware assumes that schemas are using the Core and Validation specifications.  So this is just leaving that behavior in place.

If you redefine a standard keywords in your custom vocabulary, then `$vocabulary` will be present, and no assumptions will be made.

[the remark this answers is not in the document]
handrews: @jgonzalezdr @gregsdennis I tried to address this in the latest commit, although I'm not sure the wording's really great.  It might be better to not nail this down, even, and see what people do (given that implementations have no trouble making these assumptions right now without guidance).

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: I ended up not adding this into the text because I fell down a philosophical rabbit-hole about whether the `x-` extension mechanism actually counts as being defined by the vocabulary specification (I think it does).  Let's just see what people try :-)

jsonschema-core.xml
handrews: I reworked this and I think it's stronger.  I don't know that it flows as well as it could, but I think it is more clear, so thanks, @jgonzalezdr

awwright:
I've gone through my first pass-through of this and I think I understand it, let me rephrase here:

Right now most implementations don't consider the meta-schema to be machine readable. When validating schemas maybe they can use the schema, but for actually validating instances, the $schema has intrinsic meaning, unrelated to the meta-schema definition.

$vocabulary is a proposal to allow validators to validate instances against schemas written with a $schema that the validator has no special knowledge of, except that $vocabulary allows the meta-schema to now be machine readable.

In other words, we're trying to make meta-schemas be more machine-readable. 

Sound right?

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: @Relequestual OK to mark this resolved?

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: I'm marking this resolved b/c I don't think you particularly needed me to add it, @Relequestual .  Feel free to re-open/re-comment if I missed something.

jsonschema-core.xml
handrews: @jgonzalezdr please let me know if the revised language is acceptable, and if so I will mark this thread as resolved.

handrews:
@awwright yeah, that's a good summary.

I don't think that we want to make all aspects of meta-schemas machine-readable (the validation of a schema against its meta-schema does not need to change).  But vocabularies allow us to define a machine-readable component that implementations can use to manage extensibility.

For now, we just have the URI and the boolean, so we are enabling:

* loading a plugin based on the URI to handle the appropriate keywords (I don't know what those plugins will look like, let's see what people try)
* refusing to process unrecognized required vocabularies.

Future work may add more machine-readable information, but it seems best to start small.

handrews:
@awwright thinking more on this, a sub-goal of the machine readability is to provide a more useful granularity of re-use than complete meta-schemas.  I can't remember which issue included discussions of use cases for that granularity- it was before the actual vocabulary issues were filed.

But the most compelling use cases here are not so much slicing up the keywords in the three current specs, as working in an ecosystem containing many useful extension vocabularies.

A meta-schema-only approach can only deal with the conglomeration of keywords described by the entire meta-schema.  You can't recognize that different conglomerations are made of the same constituent parts, which seriously limits the utility of extensions.

jsonschema-core.xml
Relequestual: Amusement =]

jsonschema-core.xml
```diff
@@ -873,18 +873,16 @@
...
             </section>
             <section title='The "$vocabulary" Keyword'>
                 <t>
-                    The "$vocabulary" keyword, like all keywords in meta-schemas,
-                    relates to the schema described by the meta-schema, rather than
-                    to the meta-schema itself.  Specifically, "$vocabulary" declares
+                    The "$vocabulary" keyword, which appears in a meta-schema, declares
                     what sets of keywords are expected to be used in schemas described
-                    by this meta-schema.
+                    by that meta-schema, and with what semantics.  This is conceptually
```
Relequestual: By saying "that meta-schema", it sounds like you're saying that `$vocabulary` will contain meta-schemas, which is not true, as it contains identifiers for vocabularies, right?
As such, I suggest replacing `that` with, `the associated`.

jsonschema-core.xml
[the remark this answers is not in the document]
Relequestual: I understand this now, but my question is, why `Vocabulary authors SHOULD NOT serve a document at that URL`. For what reason?

jsonschema-core.xml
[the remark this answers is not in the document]
Relequestual: @gregsdennis when I said "that is for validation", I meant "that want to use the validation vocabulary keywords".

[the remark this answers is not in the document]
Relequestual: I'm happy with the change, and I agree @handrews that we shouldn't nail it down completly, and leave a little flexibility for any ways it might need nailing down later.
As I authored this comment, I'm resolving it.

Relequestual:
@handrews Re your last comment here, Yeah, I totally see what you're getting at here. Splitting the keywords up into these vocabularies makes them easier to use in new vocabularies by vocabulary authors.

I know for a fact there is some appitite on this already.

I'm not super familiar with most implementations, but ajv for example, allows you to define your own key words. I think what we (all) want to see happen is, new vocabulary can be loaded into the process if the package is included, allowing extended functionality.

ajv could migrate it's extension keywords into a vocabulary, which can then be shared without breaking interoperability.

/cc @epoberezkin

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: @Relequestual because we are reserving the potential document for future use.

jsonschema-core.xml
jgonzalezdr: @Relequestual IMO the current wording ("that meta-schema") is pretty precise and non-ambiguous. Using "associated" would be more confusing, because "appearing in" and "being associated to" are not obviously related.

jgonzalezdr: Call me picky, but the text "declares what sets of keywords" doesn't make me associate that the property names in `$vocabularies` are identifiers for vocabularies, it may make the reader expect some list of keywords or something. In this section the word "identifier" does not appear at all (just "identifying" in a cref). In contrast, in the "$schema" section it is clearly indicated that `$schema` is a "feature set identifier".

To reinforce the fact that the names of the properties in this keyword are identifiers for vocabularies, something in the likes of "The names of the object properties are used as JSON Schema vocabulary identifiers." might be stated somewhere (maybe in next paragraph after indicating that the keyword must be an object, or on a paragraph on its own just after that).

jsonschema-core.xml
```diff
@@ -800,49 +800,302 @@
...
+                    declarations of those other meta-schemas.  All such declarations
+                    must be repeated in the root of each schema document intended
+                    for use as a meta-schema.  This is demonstrated in
+                    <xref target="example-meta-schema">the example meta-schema</xref>.
+                </t>
+                <t>
+                    If "$vocabulary" is absent, an implementation MAY determine
+                    behavior based on the meta-schema if it is recognized from the
+                    URI value of the referring schema's "$schema" keyword.
+                    If the meta-schema, as referenced by the schema, is not recognized,
```
jgonzalezdr: Doesn't this clash a bit with what is stated in line 1028?

jsonschema-core.xml
```diff
@@ -800,49 +800,302 @@
...
+                    Note that the processing restrictions on "$vocabulary" mean that
+                    meta-schemas that reference other meta-schemas using "$ref" or
+                    similar keywords do not automatically inherit the vocabulary
+                    declarations of those other meta-schemas.  All such declarations
+                    must be repeated in the root of each schema document intended
+                    for use as a meta-schema.  This is demonstrated in
+                    <xref target="example-meta-schema">the example meta-schema</xref>.
+                </t>
+                <t>
+                    If "$vocabulary" is absent, an implementation MAY determine
```
jgonzalezdr: What would happen if a schema declares `"$schema": "file://draft-07/schema"` and that meta-schema is just the plain draft-07 schema (properly identified by its `$id`)? Wouldn't it be also useful for implementations to let them determine the behavior based on the meta-schema own identifier?

jsonschema-core.xml
```diff
@@ -800,49 +800,302 @@
...
+  }
+}
+]]>
+                    </artwork>
+                    <postamble>
+                        As shown above, even though each of the referenced standard
+                        meta-schemas declares its corresponding vocabulary, this new
+                        meta-schema must re-declare them for itself.  It would be
+                        valid to leave the core vocabulary out of the "$vocabulary" keyword,
+                        but it MUST be referenced through the "allOf" keyword in order
```
jgonzalezdr:
```suggestion
                        but it has to be referenced through the "allOf" keyword in order
```
Using MUST regarding an example does not quite seem a requirement.

jsonschema-core.xml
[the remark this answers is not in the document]
Relequestual: I can't think of one, but OK then.

jsonschema-core.xml
Relequestual: OK I think I missread and got confused on this one.
@jgonzalezdr Please copy your above comment to a new review comment as I'm resolving this one (should you wish to).

jsonschema-core.xml
gregsdennis: or perhaps "must" not in caps...

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: @Relequestual there's an extensive discussion of this in #602, among other places.

jsonschema-core.xml
handrews: Hmm...  not exactly but I see why it looks that way, and there is probably a missing piece.

If you can read a meta-schema (because you have a copy of it) and there is no `$vocabulary`, line 1028 is saying that you MUST assume the use of core.  This part here on line 939-940 is saying that you MAY assume additional vocabularies to produce identical behavior to the pre-`$vocabulary` word.

But we should probably put all of this guidance together, because it's easy to read 1028 as also implying that since you MUST assume the core vocabulary, you MUST NOT assume other vocabularies.  Which is probably not quite right.

handrews:
@MikeRalphson no, although I suppose we could.  JSON Schema is not consistent about abbreviations levels, sadly.

My thought here is that relatively few people write meta-schemas, so few people are going to type this keyword at all.  And they will do that at most once per meta-schema.  So spelling it out seems reasonable.

`#/$defs/...`, on the other hand, is used multiple times per schema in any non-trivial schema.  So cutting a few letters off of it makes a really big difference.  Especially because I would type `definitoins` at least 40% of the time :-)

jsonschema-core.xml
```diff
@@ -800,49 +800,302 @@
...
+                </t>
+                <t>
+                    Values for this property are defined elsewhere in this and other documents,
+                    and by other parties.
+                </t>
+            </section>
+            <section title='The "$vocabulary" Keyword'>
+                <t>
+                    The "$vocabulary" keyword, which appears in a meta-schema, declares
+                    what sets of keywords are expected to be used in schemas described
```
jgonzalezdr: Call me picky, but the text "declares what sets of keywords" doesn't make me associate that the property names in `$vocabularies` are identifiers for vocabularies, it may make the reader expect some list of keywords or something. In this section the word "identifier" does not appear at all (just "identifying" in a cref). In contrast, in the "$schema" section it is clearly indicated that `$schema` is a "feature set identifier".

To reinforce the fact that the names of the properties in this keyword are identifiers for vocabularies, something in the likes of "The names of the object properties are used as JSON Schema vocabulary identifiers." might be stated somewhere (maybe in next paragraph after indicating that the keyword must be an object, or on a paragraph on its own just after that).

jsonschema-core.xml
jgonzalezdr: @Relequestual OK, I've moved it to a different review.

jsonschema-core.xml
jgonzalezdr: I don't mean to force implementations to do it, but rather let them optionally do so, in addition to or instead of doing it based on the schema's `$schema`. 

I was just thinking that with the shift of paradigm in this draft, moving the vocabulary declaration place from the schema to the meta-schema, some implementations might prefer to always perform the vocabulary determination during meta-schema processing, not having anything hardcoded, and therefore it could be more convenient and straightforward, when a previous draft meta-schema is encountered, to fallback to recognize the vocabulary based on a meta-schema property instead of information found on a "different processing layer".

jgonzalezdr: A possible wording for my proposal could be:
> If "$vocabulary" is absent, an implementation MAY determine behavior based on the meta-schema if it is recognized from the URI value of the referring schema's "$schema" keyword. Additionally, an implementation MAY determine behavior based on the meta-schema if it is recognized from the URI value of the meta-schema's "$id" keyword. If the meta-schema is not recognized, then implementations MAY assume the use of all vocabularies in this specification and the companion Validation specification.

jsonschema-core.xml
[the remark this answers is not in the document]
Relequestual: Great thanks. Interesting! I haven't been able to read everything =]

core.json
```diff
@@ -0,0 +1,50 @@
...
+        "$ref": {
+            "type": "string",
+            "format": "uri-reference"
+        },
+        "$recursiveRef": {
+            "type": "string",
+            "format": "uri-reference"
+        },
+        "$recursiveAnchor": {
+            "type": "boolean"
```
gregsdennis: This should be

```json
{
  "const": true,
  "default": false
}
```

gregsdennis: This is probably where you remember adding it 😄

applicator.json
```diff
@@ -0,0 +1,61 @@
...
+            "additionalProperties": { "$recursiveRef": "#" },
+            "default": {}
+        },
+        "patternProperties": {
+            "type": "object",
+            "additionalProperties": { "$recursiveRef": "#" },
+            "propertyNames": { "format": "regex" },
+            "default": {}
+        },
+        "dependentSchemas": {
```
gregsdennis: Do we also need `dependentRequired`?

jsonschema-core.xml
handrews: @jgonzalezdr We're going to stick with exactly one unambiguous mechanism for now, which matches the behavior we have always had with `$schema`.  If we get feedback from the new draft that a change is needed, we'll listen to that.

applicator.json
handrews: @gregsdennis yes, but I'll do a separate PR and fix up the meta-schemas properly, as they have a variety of issues unrelated to this actual PR.

core.json
handrews: @gregsdennis I think you fixed this elsewhere?  Regardless, I'll do a separate PR for meta-schema cleanup.  I always have to audit that and the change log at the end of the draft process.

gregsdennis: I fixed the meta-schema not broken out in PR https://github.com/json-schema-org/json-schema-spec/pull/690

applicator.json
handrews: @gregsdennis actually, no, apologies for the confusion.  `dependentSchemas` is an applicator, and belongs in this meta-schema.  `dependentRequired` is an assertion, and belongs in the assertion vocabulary meta-schema, which is not part of this PR (I'll do a vocabulary-related PR on the validation and hyper-schema specs after this one is merged).

core.json
handrews: @gregsdennis OK, let me re-base and that will be a merge conflict and I'll sort it out.

handrews: @gregsdennis brain not working well today apparently.  I eventually realized this was a duplication, merged the two descriptions (it uses `default` and `type` now) and made the "regular" meta-schema recursively extend core, which is kind of the point of having it.  That should solve it.  I hope.

handrews:
OK, I think I've addressed (either with changes or explanations) all of the feedback.  @gregsdennis and @jgonzalezdr does this seem good?  Keep in mind that we can continue to consider wording improvements through the final review period.

@awwright since it's been more than two weeks since I replied to your comment with no further response from you I'm assuming you're OK with things.  Please speak up in the next couple days if not (or if you don't see this until after it is merged, just file an issue and I'll be happy to revisit it).

schema.json
```diff
@@ -1,22 +1,23 @@
...
             "type": "integer",
             "minimum": 0
         },
         "nonNegativeIntegerDefault0": {
-            "allOf": [
-                { "$ref": "#/$defs/nonNegativeInteger" },
-                { "default": 0 }
-            ]
+            "$ref": "#/$defs/nonNegativeInteger",
+            "default": 0
```
gregsdennis: 👍

core.json
handrews: @gregsdennis it's just nicer.  Some systems for things other than validation work much better with `type` and may or may not be able to infer the type from `const`.  Yes, I know it's trivial in this case, it's just nicer to be consistent in such a visible schema.

jsonschema-core.xml
jgonzalezdr: It's fine then.

jsonschema-core.xml
jgonzalezdr: I like the new wording, much clearer! :smile:

jsonschema-core.xml
jgonzalezdr: It's very clear now.

jsonschema-core.xml
```diff
@@ -800,55 +807,313 @@
...
+            </section>
+            <section title="Example Meta-Schema With Vocabulary Declarations"
+                     anchor="example-meta-schema">
+                <figure>
+                    <preamble>
+                        This meta-schema explicitly declares both the Core and Applicator
+                        vocabularies, and combines their meta-schemas with an "allOf".
+                        It additionally restricts the usage of the Applicator vocabulary
+                        by forbidding the keyword prefixed with "unevaluated".  It also
+                        describes a keyword, "localKeyord", that is not part of either
```
jgonzalezdr:
```suggestion
                        describes a keyword, "localKeyword", that is not part of either
```

jsonschema-core.xml
```diff
@@ -800,55 +807,313 @@
...
+                        vocabulary.  Note that it is its own meta-schema,
+                        as it relies on both the Core vocabulary (as all schemas do)
+                        and the Applicator vocabulary (for "allOf"0.
```
jgonzalezdr:
```suggestion
                        and the Applicator vocabulary (for "allOf").
```

jsonschema-core.xml
```diff
@@ -2025,6 +2422,14 @@ Link: <http://example.com/my-hyper-schema#>; rel="describedby"
                     When using the media type application/schema-instance+json, the "schema"
                     parameter MUST be supplied.
                 </t>
+                <t>
+                    When using the media type application/schema+json, the "schema" parameter
+                    MAY be supplied, If supplied, it SHOULD contain the same URI as identified
```
jgonzalezdr:
```suggestion
                    MAY be supplied. If supplied, it SHOULD contain the same URI as identified
```

jsonschema-core.xml
```diff
@@ -2159,7 +2564,21 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
                     <list>
                         <t>Type name: application</t>
                         <t>Subtype name: schema+json</t>
-                        <t>Required parameters: N/A</t>
+                        <t>>Required parameters: N/A</t>
```
jgonzalezdr:
```suggestion
                        <t>Required parameters: N/A</t>
```

Relequestual:
If we had a budget or such, this would be huge cause for celebration! =D

Issue #590, opened 2018-05-24
include example of header encodings


handrews:
Apparently I did not include an example of `allow` (which for some reason I really thought I had).  There's also not a direct example of it in the referenced draft spec for encoding headers in JSON.  This has led to at least one instance of someone considering `Allow: HEAD, GET, POST` to be encoded as `"allow": [["HEAD", "GET", "POST"]]`, treating compound values within and among individual header lines differently.  That was not the intention.

handrews:
Also reference the part here about how repeated headers and multiple comma-separated values have the same semantics, which is why `Allow` does not end up with a double-array.  The headers:

> Allow: HEAD, GET
> Allow: POST

are equivalent to `Allow: HEAD, GET, POST`

https://tools.ietf.org/html/rfc7230#section-3.2.2

PR #675, opened 2018-11-12
Add Allow example for targetHints

handrews:
Addresses #590 

A similar example for `headerSchema` will be posted separately as there are different concerns there which are less clear.

### DIFF
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -1470,6 +1470,29 @@ for varname in templateData:
                     In general, headers that are likely to have different values at different
                     times SHOULD NOT be included in "targetHints".
                 </t>
+                <figure>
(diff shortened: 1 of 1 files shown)


jsonschema-hyperschema.xml
```diff
@@ -1470,6 +1470,29 @@ for varname in templateData:
...
+                <figure>
+                    <preamble>
+                        As an example, an Allow header allowing HEAD, GET, and POST
+                        would be shown as follows:
+                    </preamble>
+                    <artwork>
+<![CDATA[
+{
+    "targetHints": {
+        "allow": ["HEAD", "GET", "POST]
```
dlax:
```suggestion
        "allow": ["HEAD", "GET", "POST"]
```

Issue #559, opened 2018-03-06
Wrong example when trailing slash and no path?


mokkabonna:
Not sure if this is wrong or not, but might need to be changed to be consistent. It is although a very minor issue.

https://github.com/json-schema-org/json-schema-spec/blob/7a6c26cabf2a82cb30f38c2377063239b1823816/jsonschema-hyperschema.xml#L1593-L1606

Here the contextUri is set to https://api.example.com (no trailing slash) The base has a trailing slash. I know the slash can be omitted if the path is empty. But in reality the slash is there in those cases, just omitted. For instance when doing `curl example.com -Iv` it produces `HEAD / HTTP/1.1` Chrome does the same.

Especially when the base contains a trailing slash the example should IMO contain a trailing slash as well.  

So the suggested change: 

```js
{
        "contextUri": "https://api.example.com/",
        "contextPointer": "",
        "rel": "self",
        "targetUri": "https://api.example.com/",
        "attachmentPointer": ""
}
```

handrews:
@mokkabonna I'm not sure which thing I should fix here.  In general, I usually see the `/` left off.  For the base URI I want to show a trailing slash (because it's what you usually want for URI reference resolution as opposed to concatenation), so maybe put a path component there in which case it's not ambiguous.

But you're right that whatever it is, it should be consistent.

Relequestual:
I would prefer for the slash to be left off. Personal preference though.

handrews:
The case of an empty path component is weird and confusing, so I'm going to add a path component on the `base`.  What I want to show is how you need a trailing slash on the `base` and not have a leading slash on your relative references in order to get relative references to (effectively) append.

A number of systems, notably OpenAPI, deviate from this, require a no trailing slash on the base, and require a leading slash on the pseudo-relative URI.  And then they concatenate it.  Which is completely in violation of RFC 3986, so I want to show how it is supposed to work.

@Relequestual I don't like trailing slashes in resource URIs either, but unfortunately this is a really annoying corner of RFC 3986:  there is absolutely no way to just trim a trailing slash off of a base URI that has a path component.  But if anything, this makes it more important to show the proper resolution behavior.  It's rare to end up with just the base URI anyway.  Usually you will have at most one such URI (for the entry point resource).

[...]

PR #678, opened 2018-11-15
Clarify trailing slash ref-resolution weirdness

handrews:
Addresses #559 (paging @mokkabonna)

There is no way to do RFC 3986 reference resolution that just
removes a trailing slash without duplicating things.  Every
option is confusing and weird, but avoiding trailing slashes
is increasingly common.

Let's just face that and explain that it's weird and this is
the best that can be done.


jsonschema-hyperschema.xml
```diff
@@ -2051,7 +2057,7 @@ Link: <https://api.example.com/trees/1/nodes/456>; rev="up"
 <![CDATA[{
     "$id": "https://schema.example.com/thing-collection",
     "$schema": "http://json-schema.org/draft-08/hyper-schema#",
-    "base": "https://api.example.com/",
+    "base": "https://example.com/api/",
     "type": "object",
     "required": ["elements"],
     "properties": {
```
gregsdennis: For the other instance of a trailing slash, you added a note stating that it was included intentionally to show its confusing nature.  Is this one also intentional?

handrews: That was for the `"href": "../api"` line, not the base.  I'll try to make that more clear.

handrews: @gregsdennis reworked it with more detail, see if that makes sense now.

gregsdennis: So base URIs require trailing slashes so that relative URIs can be resolved against them.  Is there any concern about attempting to resolve one relative URI against another?

handrews: this is all RFC 3986 stuff, nothing is required.  But OpenAPI and other systems have, very annoyingly, messed up their concept of a base URI so that it is concatenation instead of reference resolution, and if you do that with reference resolution nothing will work right.  You can go look at OpenAPI if you want to dig into it.

This is why I hate writing examples.  How much of RFC 3986 do I need to explain here?  That is not a rhetorical question.  Something needs to be explained because popular formats have done it wrong, but I do not want to put a whole URI reference resolution tutorial here.  It's an incredibly complicated process if you go into all of it, but in practice it's mostly pretty straightforward except for the slash thing.

gregsdennis: I'd rather not...

👍

PR #679, opened 2018-11-18
Output Formatting

### LINKED ISSUE #643: Standardized schema output (validation result, annotations, and errors) (cont.)
This issue is to summarize the discussions and progress of #396, which sought to develop a standard output format for JSON Schema.

This issue does **not** cover standardized error message wording.  This issue is for output *formatting* only.

----------------

# Data Requirements

## Validation outcome

This will be a simple boolean value indicating whether the instance passed validation.

## Annotations/Errors

Since annotations are only collected when validation passes and errors only occur when validation fails, these two sets are mutually exclusive and will never appear in the result set together.  However, they do share a similar (if not the same) structure.

- **Keyword relative location** - The location of the keyword that produced the annotation or error.  The purpose of this data is to show the resolution path which resulted in the subschema that contains the keyword.
    - relative to the root of the principal schema; should include (inline) any `$ref` segments in the path
    - JSON pointer
- **Keyword absolute location** - The direct location to the keyword that produced the annotation or error.  This is provided as a convenience to the user so that they don't have to resolve the keyword's subschema, which may not be trivial task.  It is only provided if the relative location contains `$ref`s (otherwise, the two locations will be the same).
    - absolute URI
    - may not have any association to the principal schema
- **Instance location** - The location in the JSON instance being validated.
    - relative to the root of the JSON instance
    - JSON pointer
- **Value/Message** - This is the annotation or error itself.

## Instance data

We had discussed including the instance data in the response, but we determined that it was sufficient to provide a pointer (see **Instance location** above).

Copying the data into the output can still be an option for implementations, however.

# Format

Both proposed formats return an object containing the validation outcome and a list of annotations or errors.  The unresolved bit is how to represent annotations and errors generated from subschemas.

***NOTE** I will be using errors for these examples, but annotations could be used in their place.  As mentioned before their structure is similar.*

For these examples, we will use the following schema and (invalid) instances

```js
// schema
{
  "$id":"http://schemarepo.org/schemas/user.json",
  "$schema":"http://json-schema.org/draft-07/schema#",
  "type":"object",
  "definitions":{
    "min18":{
      "type":"integer",
      "minimum":18
    },
    "username":{
      "type":"string",
      "minLength":8
    },
    "member":{
      "type":"object",
      "properties":{
        "age":{"$ref":"#/definitions/min18"},
        "username":{"$ref":"#/definitions/username"}
      }
    },
    "membershipTypes":{"enum":["admin","user"]}
  },
  "oneOf":[
    {
      "properties":{
        "member":{"$ref":"#/definitions/member"},
        "membershipType":{"$ref":"#/definitions/membershipTypes"}
      }
    },
    {
      "properties":{
        "membershipType":{"const":"guest"},
        "firstName":{"type":"string"},
        "lastName":{"type":"string"}
      },
      "additionalProperties":false
    }
  ]
}

// instance #1
{
  "member":{
      "age":5,  // doesn't meet minimum
      "username":"aName"  // doesn't meet minLength
  },
  "membershipType":"user"
}

// instance #2
{
  "membershipType":"guest",
  "firstName":"Charles",
  "lastName":"InCharge",
  "age":27  // additional properties not allowed
}
```

## Flat

This proposal outputs all annotations and errors in a flat list.  Because it's a flat list, it's easy to both build and consume.

```js
// instance #1 results
{
  "valid":false,
  "errors":[
    {
      "keywordLocation":"#/oneOf",
      "instanceLocation":"/",
      "message":"the instance did not pass any of the subschema"
    },
    {
      "keywordLocation":"#/oneOf/0/properties/member/properties/age/$ref/minimum",
      "absoluteKeywordLocation":"http://schemarepo.org/schemas/user.json#/definitions/min18/minimum",
      "instanceLocation":"/member/age",
      "message":"value is too small"
    },
    {
      "keywordLocation":"#/oneOf/0/properties/member/properties/userName/$ref/minLength",
      "absoluteKeywordLocation":"http://schemarepo.org/schemas/user.json#/definitions/username/minLength",
      "instanceLocation":"/member/username",
      "message":"value is too short"
    },
    {
      "keywordLocation":"#/oneOf/1/membershipType",
      "instanceLocation":"/member/membershipType",
      "message":"value does not match the required value"
    },
    {
      "keywordLocation":"#/oneOf/1/additionalProperties",
      "instanceLocation":"/member/member",
      "message":"additional properties are not allowed"
    }
  ]
}

// instance #2 results
{
  "valid":false,
  "errors":[
    {
      "keywordLocation":"#/oneOf",
      "instanceLocation":"/",
      "message":"the instance did not pass any of the subschema"
    },
    {
      "keywordLocation":"#/oneOf/0/properties/membershipType/$ref/enum",
      "absoluteKeywordLocation":"http://schemarepo.org/schemas/user.json#/definitions/membershipTypes/enum",
      "instanceLocation":"/membershipType",
      "message":"value is not one of the required values"
    },
    {
      "keywordLocation":"#/oneOf/1/additionalProperties",
      "instanceLocation":"/member/age",
      "message":"additional properties are not allowed"
    }
  ]
}
```

## Hierarchical

This proposal arranges the errors in a hierarchical format based on the schema.

The primary argument against the flat structure is that it's difficult (both for machines and humans) to see any association between the errors.  For example, in instance <span>#</span>1 results above, there is no immediate indication that the first three errors pertain to the first subschema of the `oneOf` and the last two errors pertain to the second subschema.  Moreover it becomes more difficult to understand that **either** the first three **or** the last two must be resolved to pass the instance, **but not all five**.  The hierarchical format aims to make that association more apparent.

The construction of this object requires rules that would need to be included in the specification, particularly the conditions that are required for a node to be present.

1. All [applicator keywords](https://github.com/json-schema-org/json-schema-spec/issues/530) (`*Of`, `$ref`, `if`/`then`/`else`, etc.) require a node.
1. Any path branching requires a node.  (Note that a branch node may not produce an annotation or error message).

An (unoptimized) algorithm for this may be

1. Build out the validation structure, creating nodes for *all* keywords in the schema.
1. As one travels back up the structure from the leaves, analyze the parent nodes as follows:
    1. If it is an applicator keyword, it must remain.
    1. If it contains multiple children, it must remain.
    1. If it contains no children, it is removed.
    1. If it contains a single child (only one descendant keyword produced an error), it is replaced by the child.

```js
// instance #1 results
{
  "valid": false,
  "errors": [
    {
      "schemaLocation": "#/oneOf",
      "instanceLocation": "/",
      "message": "the instance did not pass any of the subschema",
      "errors": [
        {
          "schemaLocation": "#/oneOf/0/properties/member/properties",
          "instanceLocation": "/",
          "errors": [
            {
              "schemaLocation": "#/oneOf/0/properties/member/properties/age/$ref",
              "instanceLocation": "/member/age",
              "message": "referenced schema failed",
              "errors": [
                {
                  "schemaLocation": "http://schemarepo.org/schemas/user.json#/definitions/min18/minimum",
                  "instanceLocation": "/member/age",
                  "message": "value is too small"
                }
              ]
            },
            {
              "schemaLocation": "#/oneOf/0/properties/member/properties/userName/$ref",
              "instanceLocation": "/member/username",
              "message": "referenced schema failed",
              "errors": [
                {
                  "schemaLocation": "http://schemarepo.org/schemas/user.json#/definitions/username/minLength",
                  "instanceLocation": "/member/username",
                  "message": "value is too short"
                }
              ]
            }
          ]
        },
        {
          "schemaLocation": "#/oneOf/1",
          "errors": [
            {
              "schemaLocation": "#/oneOf/1/membershipType",
              "instanceLocation": "/member/membershipType",
              "message": "value does not match the required value"
            },
            {
              "schemaLocation": "#/oneOf/1/additionalProperties",
              "instanceLocation": "/member/member",
              "message": "additional properties are not allowed"
            }
          ]
        }
      ]
    }
  ]
}}

// instance #2 results
{
  "valid": false,
  "errors": [
    {
      "valid": false,
      "schemaLocation": "#/oneOf",
      "instanceLocation": "/",
      "message": "the instance did not pass any of the subschema",
      "errors": [
        {
          "valid": false,
          "errors": [
            {
              "schemaLocation": "#/oneOf/0/properties/membershipType/$ref",
              "instanceLocation": "/membershipType",
              "message": "referenced schema failed",
              "errors": [
                {
                  "schemaLocation": "http://schemarepo.org/schemas/user.json#/definitions/membershipTypes/enum",
                  "instanceLocation": "/membershipType",
                  "error": "value is not one of the required values"
                }
              ]
            }
          ]
        },
        {
          "valid": false,
          "errors": [
            {
              "schemaLocation": "#/oneOf/1/additionalProperties",
              "instanceLocation": "/member/age",
              "message": "additional properties are not allowed"
            }
          ]
        }
      ]
    }
  ]
}
```

# Other considerations

Though the crux of this issue is the above, some other proposals have been made relating to this topic.

## Implementation domain

The specification could allow for both of these formats, allowing the implementation to choose.

## Configurable output levels

The specification could define different output settings.

1. **Basic** would be a response containing simply the boolean result indicating a pass/fail status.
1. **List**
1. **Hierarchy**
1. **VerboseHierarchy** would be an uncollapsed hierarchy (potentially useful for forms and other applications that would need a fixed path.)

gregsdennis:
Here is my first go at the output formatting.  Happy to update as necessary.  It was quite difficult coming up with examples that showed the flexibility of the output while also being concise.

~The *.md* file should be removed before merging.  It's just a composition file.~

This PR resolves #643 and relates to #396.


jsonschema-core.xml
```diff
@@ -2095,6 +2101,440 @@
...
+                        <t>
+                            There are only two vertices, but three are required.
+                        </t>
+                    </list>
+                    Note that the wording in the error messages depicted in these examples is not a
+                    requirement of this specification.  Implementations SHOULD craft error messages
+                    tailored for their audience.
+                </t>
+
+                <section title="Pass/Fail">
```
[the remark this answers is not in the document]
gregsdennis: I like "flag" for now instead of "pass/fail."  I was struggling to come up with something.  If another option doesn't surface, we'll go with that.

jsonschema-core.xml
```diff
@@ -2095,6 +2101,440 @@
...
+                        The JSON key for nested results in failed validations is "errors"; for
+                        successful validations it is "annotations".
+                    </t>
+                </section>
+
+            </section>
+
+            <section title="Output Structure">
+                <t>
+                    The output MUST be an object containing two properties
```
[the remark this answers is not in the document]
Anthropic: I meant also as in along with the Pass/Fail section below. I felt the use of MUST here with "containing two properties" may confuse people when really the only required property is `valid`. Reading this section it feels like it could cause confusion. Given the description up in "Output Levels" says to look at this section for details of the output structure it felt like more information was needed here as well as in the Pass/Fail section or at least a reference to it.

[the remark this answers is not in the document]
gregsdennis: ✔️

jsonschema-core.xml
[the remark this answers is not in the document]
gregsdennis: ✔️

jsonschema-core.xml
```diff
@@ -2095,6 +2101,442 @@
...
+            <t>
+                JSON Schema is defined to be platform-independent.  As such, to increase compatibility
+                across platforms, implementations SHOULD conform to a standard validation output
+                format.  This section describes the minimum requirements that consumers will need to
+                properly interpret validation results.
+            </t>
+
+            <section title="Format">
+                <t>
+                    JSON Schema output is defined using JSON.  Implementations MAY deviate from this
```
handrews: Can we say that it is defined with the JSON Schema instance data model as described in section 4.2.1?  That would make the convertable-to-JSON requirement very clear, as anything that fits that data model can be converted.

jsonschema-core.xml
```diff
@@ -2095,6 +2101,442 @@
...
+                    A single object that contains all of these components is considered an
+                    output unit.
+                </t>
+                <t>
+                    Implementations MAY elect to provide additional information.
+                </t>
+
+                <section title="Keyword Relative Location">
+                    <t>
+                        The relative location of the validating keyword that follows the validation
```
handrews: Can we define the term "validation path" in section 4.3.4 Lexical Scope and Dynamic Scope?  The validation path essentially captures the dynamic scope, and it would be good to make that clear.

jsonschema-core.xml
```diff
@@ -2095,6 +2101,442 @@
...
+                    </figure>
+                    <t>
+                        The JSON key for this information is "keywordLocation".
+                    </t>
+                </section>
+
+                <section title="Keyword Absolute Location">
+                    <t>
+                        The absolute, dereferenced location of the validating keyword.  The value MUST
+                        be expressed as an absolute URI, and it MUST NOT include "$ref" or
```
handrews: These are more generically termed "by-reference applicators" per section 4.3.5 Referenced and Referencing Schemas.  So maybe

> MUST NOT include by-reference applicators such as "$ref" or "$recursiveRef"

to keep it general?

jsonschema-core.xml
```diff
@@ -2416,7 +2858,8 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
                 Brad Bowman,
                 Gowry Sankar,
                 Donald Pipowitch,
-                and Dave Finlay
+                Dave Finlay,
+                and Greg Dennis
```
handrews: You have an author credit, we don't really need to put you in the acknowledgements :-)

jsonschema-core.xml
```diff
@@ -2095,6 +2101,442 @@
...
+                        <t>
+                            Detailed - Provides validation information in a condensed hierarchical
+                            structure based on the structure of the schema.
+                        </t>
+                        <t>
+                            Verbose - Provides validation information in an uncondensed hierarchical
+                            structure that matches the exact structure of the schema.
+                        </t>
+                    </list>
+                    Implementations SHOULD specify which levels of validation they support.
```
handrews: Do you need to implement all levels up to and including the most complex/verbose one?  Or could an implementation support *only* the Verbose option?  We should be clear about that in this section.

jsonschema-core.xml
gregsdennis: ✔️

jsonschema-core.xml
gregsdennis: ✔️

jsonschema-core.xml
gregsdennis: ✔️

jsonschema-core.xml
gregsdennis: that's fair.  I don't see you there...  ✔️

jsonschema-core.xml
gregsdennis: I don't think that an "up-to" requirement is necessary.  Implementations should be free to support any of the formats.  Maybe "levels" is wrong?

standardized-output-composition.md
```diff
@@ -0,0 +1,291 @@
...
+      "keyword": "minItems",
+      "error": "Expected at least 3 items but found 2"
+    }
+  ]
+}
+```
+
+### Detailed
+
+The **Detailed** structure is based on the schema and can be more readable for both humans and machines.  Having the structure organized this way makes associations between the errors more apparent.  For example, the fact that the missing `y` property and the extra `z` property both stem from the same location in the instance is not immediately obvious in the **Basic** structure.  In a hierarchy, the correllation is more easily identified.
```
Relequestual: Is it possible to build the detailed structure given the basic structure? Just curious.

standardized-output-composition.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Format
+
+JSON Schema output is defined using JSON.  Implementations MAY deviate from this as supported by their specific languages and platforms, however it is RECOMMENDED that the output be convertible to the JSON format defined herein via serialization or other means.
+
+## Output Levels **<span style="color:red">(is "levels" the right word?)</span>**
+
+This specification defines four levels of output.  See the *Output Structure* section for the requirements of each level.
+
+- **Pass/Fail** - A boolean which simply indicates whether the schema was validated.
```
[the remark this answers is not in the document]
gregsdennis: Yeah, I guess it would always be validated; whether it was successful or not is a different thing.

standardized-output-composition.md
```diff
@@ -0,0 +1,291 @@
...
+
+Beyond a simplistic pass/fail, additional information is useful to aid in debugging a schema or instance.  Each sub-result SHOULD contain the information contained within this section at a minimum.
+
+A single object that contains all of these components is considered an output unit.
+
+Implementations MAY elect to provide additional information.
+
+### Keyword relative location
+
+The relative location of the validating keyword that follows the validation path.  The value MUST be expressed as a JSON Pointer, and it MUST include `$ref` or `$recursiveRef` segments for any reference jumps **<span style="color:red">(wording?)</span>**.
```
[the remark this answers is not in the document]
gregsdennis: Including this was one of @handrews's ideas (and a good one). He felt it necessary to not only include the route to the subschema that is resolved but also the route that validation took to get there.  That there may be multiple paths that resolve to the same location is not really a concern.

I suppose in that regard it would strictly be a JSON Pointer, but it's understandable within the context of processing a schema.  Additionally, when a referencing segment is included, a direct URI to the subschema is also required so that the consumer of the output doesn't have to do the processing you mentioned.

standardized-output-composition.md
```diff
@@ -0,0 +1,291 @@
...
+
+```
+#/properties/minLength/$ref/minimum
+```
+
+The JSON key for this information is `keywordLocation`.
+
+### Keyword absolute location
+
+The absolute, dereferenced location of the validating keyword.  The value MUST be expressed as an absolute URI, and it MUST NOT include `$ref` or `$recursiveRef` segments.
```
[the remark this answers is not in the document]
gregsdennis: The answer links to the answer to said other comment :)

standardized-output-composition.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Nested results
+
+For the two hierarchical structures, this property will hold nested errors and annotations.
+
+The JSON key for nested results in failed validations is `errors`; for successful validations it is `annotations`.
+
+## Output Structure
+
+The output MUST be an object containing two properties
```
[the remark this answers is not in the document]
gregsdennis: I think this was updated in the XML version

standardized-output-composition.md
```diff
@@ -0,0 +1,291 @@
...
+
+- `valid` - a boolean value indicating the overall validation success or failure
+- `errors` - the collection of errors or annotations produced by a failed validation
+- `annotations` - the collection of errors or annotations produced by a successful validation
```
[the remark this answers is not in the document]
gregsdennis: Only the verbose output would do that.  The primary cases to consider are the applicator keywords (especially `not`). For example, an `allOf` could contain 1 failed subschema, yielding an overall failure, but the verbose output would list *all* of the results. This would need to be inside `errors` since it failed.

I agree it's a little counterintuitive, but I think it works better for automated processing of all the results are together rather than having both `errors` *and* `annotations`.

standardized-output-composition.md
gregsdennis: I *think* so. It would take quite a bit more processing because you have to look at the keyword location properties to reconstruct the hierarchy, but I don't think it's impossible.

gregsdennis:
Some of this has changed in the XML version based on @handrews's comments. It's definitely worth reading that over this file. I can probably go ahead and delete it.

jsonschema-core.xml
```diff
@@ -2095,6 +2104,447 @@
...
+                        *all* results are returned.  This includes sub-schema validation results that
+                        would otherwise be removed (e.g. annotations for failed validations,
+                        successful validations inside a `not` keyword, etc.).  Because of this, it
+                        is RECOMMENDED that each node also carry a `valid` property to indicate the
+                        validation result for that node.
+                    </t>
+                    <t>
+                        Because this output structure can be quite large, a smaller example is given
+                        here for brevity.  The full output structure of the example above can be found
+                        [here](standardized-output-verbose.json)
```
gregsdennis: @handrews how do I do external links in this XML file?  Probably not with Markdown as I have here.

jsonschema-core.xml
```diff
@@ -2095,6 +2104,447 @@
...
+}
+]]>
+                        </artwork>
+                    </figure>
+                </section>
+
+                <section title="Output validation schema">
+                    <t>
+                        For convenience, a JSON Schema has been provided to validate output generated
+                        by implementations.  It can be found [here](schema-output.json).
```
gregsdennis: Another external link.

jsonschema-core.xml
handrews: @gregsdennis I think you want `<eref>`.  The full XML RFC reference is https://tools.ietf.org/html/rfc7749 (yes, that RFC is marked obsolete, but it's the right one- the xml2rfc tool still doesn't support the new RFC for xml2rfc, ironically).

standardized-output-composition.md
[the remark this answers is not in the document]
Relequestual: OK, so my follow up question here is:
If a subschema of `not` fails, obviously the `not` means the parent schema asserts valid is true. My question is then what happens to the annotations in the subschema? They only trickle up for schemas for which validation is a positive assertion, no? /ccing @handrews on this one =]

jsonschema-core.xml
```diff
@@ -2095,6 +2104,450 @@
...
+                </t>
+                <t>
+                    Implementations MAY elect to provide additional information.
+                </t>
+
+                <section title="Keyword Relative Location">
+                    <t>
+                        The relative location of the validating keyword that follows the validation
+                        path.  The value MUST be expressed as a JSON Pointer, and it MUST include
+                        any by-reference applicators such as "$ref" or "$recursiveRef".
```
Relequestual: As in my previous now outdated comment, this will technically be a valid JSON Pointer, but it won't be resolveable when including by-reference applicators. Should we note that?

standardized-output-composition.md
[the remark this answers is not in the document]
Relequestual: I'm making a new comment based on this.

standardized-output-composition.md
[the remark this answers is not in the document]
Relequestual: Happy with the change.

standardized-output-composition.md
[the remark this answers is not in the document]
handrews: Correct, you only return annotations if the assertion result is valid.  Otherwise they are dropped from the final result set (although showing what they were before they were dropped in the more complex output format is fine and useful.

jsonschema-core.xml
gregsdennis: ✔️

handrews:
Thanks @gregsdennis I will get to this in the next day or two.  Really :-P 

The meta-schema (and XML) deployment process is manual and somewhat absurd.  I don't even know what the makefile does other than generate the .html and .txt.  I think we've gotten a PR or two around it but it was always someone wanting to do something else with them instead / in addition.  Given that they get deployed roughly twice a year, by me, it's incredibly low on my priority list to change.

jsonschema-core.xml
Anthropic: [As I originally described it](https://github.com/json-schema-org/json-schema-spec/issues/396#issuecomment-406128790) an updated version would say that an implementation:
MUST provide flag
SHOULD provide basic
RECOMMENDED to provide detailed and
OPTIONALLY provide verbose

handrews: If the verbose format is likely to be easier to implement I would prefer to allow any subset of formats be implemented.  @Anthropic do you feel strongly about this?

handrews:
This is looking great!  I think all of my concerns have been addressed.

gregsdennis:
@Anthropic I've updated the language around format support requirements.  Please have a look.

jsonschema-core.xml
```diff
@@ -2142,12 +2142,9 @@
...
                         </t>
                     </list>
-                    Implementations MUST provide support for the "flag" format.  It is RECOMMENDED
-                    that implementations support at least one of the complex formats as well.  Each
-                    format is independent of the others.  For example, if an implementation decides
-                    to support the "detailed" format, it is not also required to support the "basic"
-                    or "verbose" formats, and vice versa.  Implementations SHOULD specify in their
-                    documentation which formats of validation they support.
+                    Implementations MUST provide support for the "flag" format and SHOULD provide the
+                    "basic" and "detailed" formats.  The "verbose" format is OPTIONAL.  Implementations
```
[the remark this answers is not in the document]
gregsdennis: Thanks.  I was sure there was a way to make it smoother.

Issue #659, opened 2018-09-28
Support declaring that all defined properties are required


mfulton26:
I would like a way to make all properties defined in a schema to be required without having to explicitly list them.

mfulton26:
Supporting `"required": true` on a property schema would work for me because what I am really trying to avoid is simply needing to repeat the property names.

handrews:
There are concerns over this, including:

* We have avoided keywords that depend on each other, as they are generally more difficult to work with
* It's not entirely clear how that would interact with `patternProperties`
* People then want an `optional` keyword to carve out exceptions, and it gets messier from there- past discussions have bogged down on this IIRC.

On the plus side:

* We now have a way (annotations) to formally describe keyword interactions rather than just making them all special cases, which was part of the problem in the past.  We already now define a "properties" annotation and define `additionalProperties` and `unevaluatedProperties` in terms of that annotation (see #556 and the current spec on master).  So we could do that with `"required": true`
* Saying that `patternProperties` is not affected by this would not be that weird.  It's rare to try to get `required` and `patternProperties` to interact anyway, usually `required` just goes with `properties`.

I'm at least tentatively open to this idea.  Let's see how others chime in.

jgonzalezdr:
As commented in the PR:

Maybe the solution could be creating a new `"requiredProperties"` keyword with the semantic that I just proposed above for `"required"`, which by the way would have a "naming convention" homogeneous with the rest of validation keywords for properties.

With this more meaningful keyword name the expected / implicit semantics would be a bit more evident (also explicitly indicating that the keyword applies exclusively to properties): `"requiredProperties" : true` can be easily understood as "all defined properties are required", `"requiredProperties" : false`  as "no properties are required", and `"requiredProperties" : ["a", "b"...]` as "the listed properties are required". 

The `"required"` keyword could be deprecated at some point, as it is a bit misleading. Anyway both could coexist during some time if needed.

awwright:
I mentioned elsewhere
> Unfortunately this was [previously defined](https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.7) as a boolean, but with different semantics, and this behavior is implemented in more than a few packages for reverse compatibility, e.g. [tdegrunt/jsonschema](https://github.com/tdegrunt/jsonschema/blob/ce0d4f093236639aa7a56536c10253674db7494d/lib/attribute.js#L482).
> 
[...]

PR #681, opened 2018-11-23
Added new "requiredProperties" keyword to supersede the "required" keyword

jgonzalezdr:
Added boolean values to "required" such that when set to true all the properties present in "properties" are required (related to #659).

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -462,6 +462,34 @@
                     </t>
                     <t>
                         Omitting this keyword has the same behavior as an empty array.
+                        <cref>
+                            This keyword should be deprecated.
+                        </cref>
+                    </t>
+                </section>
+
+                <section title="requiredProperties">
+                    <t>
+                        The value of this keyword MUST be either a boolean or an array.
+                        If it is an array, elements of this array, if any, MUST be strings,
+                        and MUST be unique.
+                    </t>
+                    <t>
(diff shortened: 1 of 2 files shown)


awwright:
Unfortunately this was [previously defined](https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.7) as a boolean, but with different semantics, and this behavior is implemented in more than a few packages for reverse compatibility, e.g. [tdegrunt/jsonschema](https://github.com/tdegrunt/jsonschema/blob/ce0d4f093236639aa7a56536c10253674db7494d/lib/attribute.js#L482).

I've once suggested having a "requiredProperties" keyword that functions the same as "properties" does, except that all the keywords there are also required. There was an argument against this too and I forget what it was.

jgonzalezdr:
Maybe the solution could be creating a new `"requiredProperties"` keyword with the semantic that I just proposed above, which by the way would have a "naming convention" homogeneous with the rest of validation keywords for properties. Then the `"required"` keyword could be deprecated at some point. Anyway both could coexist during some time if needed.

handrews:
@jgonzalezdr we have a good discussion going on in issue #659, so I'd like to close this until that discussion resolves.  Otherwise, people will think this is the approved solution and review it as such.

If we decide on this direction, we will re-open this PR.  But the `requiredProperties` solution also seems worth discussing.  Let's resolve that before having an active PR attracting attention.

PR #688, opened 2018-12-02
Added Appveyor CI

jgonzalezdr:
Added an Appveyor CI configuration file that generates the text and HTML outputs from the xml RFCs and packages these outputs as artifacts.

The difference from the current Travis CI is the packaging of artifacts, that can be easily accessed by everybody on the net, making therefore access to the latest draft version easier in human-readable HTML form instead of in "half-human-readable" XML form. E.g., the json-schema web can be updated as shown in this [commit](https://github.com/jgonzalezdr/json-schema-org.github.io/commit/900525d63f15a9087eb067a9213acbd987818749).

The only thing that should be done in the project side is signing in to [Appveyor](https://ci.appveyor.com) using the project's GitHub account and enable monitoring of the json-schema-spec repository (all that can be done in less than 5 minutes).

If this PR is accepted, automatic deploying of official releases in GitHub based on tags and automatic deploying of the latest release as a "latest" pre-release can also be achievable with some additional changes.

I hope you find this interesting, otherwise just close the PR.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,24 @@
+version: draft-08.build-{build}
+
+image: Visual Studio 2017
+
+environment:
+  OUTPUT_DIR: output
+  DOCS: jsonschema-core jsonschema-validation jsonschema-hyperschema
+
+cache:
+- '%LOCALAPPDATA%\pip\Cache'
+  
+install:
+- pip install xml2rfc
+
+before_build:
+- ps: if ($Env:APPVEYOR_REPO_TAG_NAME) { $Env:TAG = $Env:APPVEYOR_REPO_TAG_NAME } else { $Env:TAG = "latest" }
+
+build_script:
+- cmd: mkdir %OUTPUT_DIR%
+- cmd: FOR %%x IN (%DOCS%) DO xml2rfc %%x.xml -o %OUTPUT_DIR%\%%x-%TAG%.txt --text
+- cmd: FOR %%x IN (%DOCS%) DO xml2rfc %%x.xml -o %OUTPUT_DIR%\%%x-%TAG%.html --html
+
+artifacts:
+- path: $(OUTPUT_DIR)\*.*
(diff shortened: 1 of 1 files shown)


jgonzalezdr:
BTW, if you just want to play a bit with it and check this PR, just fork it to your personal account, sign in with it in Appveyor and force a build there.

handrews:
@jgonzalezdr I appreciate the work here, but I'm not sure this is what we actually want.  We do not want folks to casually find the work-in-progress drafts.  It is far too easy for someone to get confused about what a "real draft" vs "draft of a draft" means.

People who are motivated to work on the drafting process will also be motivated to either read the XML or install xml2rfc and build the HTML themselves.

When we get to the final review period for a draft (after all of the big PRs are in and we're just cleaning up flow and wording), we publish HTML of that, clearly marked as the work-in-progress "last call before publication" documents on json-schema.org.  That is a manual process which is fine because it happens at most twice a year.

@Relequestual @awwright @Julian @philsturgeon thoughts here?

jgonzalezdr:
@handrews I get your point. In any case, if you don't publish it in the web, it may be convenient for people working in the drafting process to have it available in a semi-obscure place, but as you said, for most  of that people the XML is enough.

johandorland:
I wouldn't mind having a place to look at the in-progress PR's without going through all the hoops of checking out a PR and generating the HTML myself.  I do see the danger of these build artifacts becoming too public so it probably wouldn't hurt to add a big fat red warning banner somewhere on the top, just in case.

handrews:
@Relequestual what do you think?  I'm personally against this and really I have no capacity to think about anything until after we get draft-08 done.  I'd really like for folks to hold off on general enhancement ideas because we're 2 months overdue as it is and reading proposals is work that I can barely fit in right now.

Julian:
Personally I'd defer to what makes your life easier at the minute @handrews, it's mostly the author that needs to have rendered versions of the thing while it's being worked on, so much as this is appreciated @jgonzalezdr, if it doesn't make your life easier (or gives you something extra to think about), suspect we should pass.

jgonzalezdr:
I'll just close this PR until after draft-08 is done, I'll reopen it later if anyone finds it interesting.

PR #690, opened 2018-12-03
Added '$recursive*' to the meta-schema

gregsdennis:
This supports https://github.com/json-schema-org/json-schema-spec/pull/653, https://github.com/json-schema-org/json-schema-spec/pull/654, and https://github.com/json-schema-org/json-schema-spec/pull/655.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -52,6 +52,14 @@
             "type": "string",
             "format": "uri-reference"
         },
+        "$recursiveRef": {
+            "type": "string",
+            "format": "uri-reference"
+        },
+        "$recursiveAnchor": {
+            "const": true,
+            "default": false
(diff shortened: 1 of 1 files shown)


handrews:
Hmm.... I'm pretty sure I added this but I think I must have messed up the merge.  I'll double check tonight- this is correct, I just want to figure out why it wasn't already there.

gregsdennis:
It looks like the appropriate `$ref`s have been migrated to `$recursiveRef`, but the keywords weren't added to the property list.

handrews:
@gregsdennis apparently I did indeed forget to add the fields themselves!  That's kind of hilarious, thanks for catching it.

PR #691, opened 2018-12-05
Fixed some minor spelling errors

jgonzalezdr:


jsonschema-core.xml
```diff
@@ -1930,7 +1930,7 @@
                             its elements is valid against the given schema.  This keyword
                             does not produce annotation results.
                             <cref>
-                                Should it produce a set of the indices for which the
```
handrews: Both spellings are correct, and a search of RFCs on ietf.org shows that both are roughly equally popular.  I would prefer to stick with indices as that's what we've always used in this project and changing it means we will almost certainly end up inconsistent.

jgonzalezdr: Ooops sorry, I'm so used to use "indexes" that I didn't confirm that "indices" was actually correct.

jgonzalezdr:
BTW, I opened this PR just to help you catch some minor spelling errors before final draft, please feel free to correct that in any PR and squash the changes there.

PR #695, opened 2018-12-17
Explain schema media type param and Accept

### LINKED ISSUE #607: What is the meaning of multiple schemas in a Content-Type header?
Section 10.2 of the latest draft (http://json-schema.org/latest/json-schema-core.html#rfc.section.10.2) gives the following example:

```
Content-Type: application/json;
          schema="http://example.com/alice http://example.com/bob"
```

What is the meaning of multiple schemas in a Content-Type header?  Does this say that the HTTP response conforms to both the "alice" and "bob" schemas?  Or to either the "alice" or "bob" schema?  In either of those cases, why wouldn't you express that through a single schema?  

How is a well-designed client supposed to process the response when it receives a Content-Type header with multiple schemas?

Thanks,

-Dan

handrews:
Resolves #607 

This explains how to use HTTP content negotiation to request
representations that conform to multiple schemas, or conform
to at least one schema out of several.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2912,7 +2912,8 @@ Content-Type: application/json;
                 </figure>
 
                 <t>
-                    Multiple schemas are whitespace separated:
+                    Multiple schemas are whitespace separated, and indicate that the
+                    instance conforms to all of the listed schemas:
                 </t>
 
                 <figure>
@@ -2923,6 +2924,33 @@ Content-Type: application/json;
 ]]>
                     </artwork>
                 </figure>
+
+                <t>
+                    Media type parameters are also used in HTTP's Accept request header:
+                </t>
+
+                <figure>
+                    <artwork>
+<![CDATA[
+Accept: application/json;
+          schema="http://example.com/qiang http://example.com/li",
+        application/json;
+          schema="http://example.com/kumar"
+]]>
+                    </artwork>
+                </figure>
+
+                <t>
+                    As with Content-Type, multiple schema parameters in the same string
+                    requests an instance that conforms to all of the listed schemas.
+                </t>
+
+                <t>
+                    The Accept header also allows repeating the same media type with
+                    different parameters, which requests that the instance conforms
+                    to one parameter value or the other.  Such an instance can conform
+                    to all separately specified parameters but is not guaranteed to do so.
+                </t>
 
                 <t>
                     <cref>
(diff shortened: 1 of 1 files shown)


awwright:
I like this, but I want to double-check my reading of media types and repeated media-type parameters first.

jsonschema-core.xml
```diff
@@ -2198,6 +2199,33 @@ Content-Type: application/json;
...
+        application/json;
+          schema="http://example.com/kumar"
+]]>
+                    </artwork>
+                </figure>
+
+                <t>
+                    As with Content-Type, multiple schema parameters in the same string
+                    requests an instance that conforms to all of the listed schemas.
+                </t>
```
awwright: Good, this appears to be consistent with how e.g. profile is used.

jsonschema-core.xml
```diff
@@ -2198,6 +2199,33 @@ Content-Type: application/json;
...
+
+                <t>
+                    The Accept header also allows repeating the same media type with
+                    different parameters, which requests that the instance conforms
+                    to one parameter value or the other.  Such an instance can conform
+                    to all separately specified parameters but is not guaranteed to do so.
+                </t>
```
awwright: I think I misread this the first time through, I thought this was saying a media type could have a property specified multiple times like `application/json;schema="http://example.com/a";schema="http://example.com/b"` but I don't know what that would even mean. I suspect it's invalid.

This is all technically correct, so I don't have an objection as such, but I'd be clearer to me _without_ this paragraph. I think the example above is sufficient to get the point across.

I can't find if there's a good way to distinguish the "type slash subtype" media type, versus the media type of a document (including its media type parameters).

handrews: @awwright thanks, I'll take a look at the wording.  I think I was trying to cover a question from the issue that was filed (this came from an outside request filed a number of months ago), but maybe it can be trimmed or reworded.

handrews: @awwright thanks for catching this, it really was confusing when I went back and read it.  I think I have clarified it by just highlighting the example.  Please let me know if this works for you.

danfoygel: @handrews thanks for making these changes - one minor suggested tweak.  Instead of "conforms to at least one of the identified schemas", it might be clearer to say "conforms to both the first and second schema or conforms to the third".

PR #704, opened 2019-01-06
Add in the missing check that patternProperties are regexen.

Julian:
This seems to already appear in Draft 7.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -115,6 +115,7 @@
         "patternProperties": {
             "type": "object",
             "additionalProperties": { "$ref": "#" },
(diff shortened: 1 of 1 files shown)


Julian:
Thanks!

Relequestual:
Given `format: 'regex'` isn't actually in draft-06, should this have been merged?

Relequestual:
I notice that the draft-07 changelog says `format: regex` was removed by mistake previously.
It was in draft-03, but not in draft-04, and not added back again till draft-07.

PR #708, opened 2019-02-07
Clarifying "const" usage.

djpadz:
It would seem to me that `const` is the same as a single-valued `enum`.  If that's correct, then that should be noted.  If it's not, then the difference should also be noted.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -243,6 +243,10 @@
                         The value of this keyword MAY be of any type, including null.
                     </t>
                     <t>
+                        Use of this keyword is functionally equivalent to an
+                        <xref target="json-schema">"enum"</xref> with a single value.
+                    </t>
+                    <t>
                         An instance validates successfully against this keyword if its value is
                         equal to the value of the keyword.
                     </t>


Relequestual:
Thanks for your contribution! You are correct. There are lots of equivolencies in JSON Schema. This info likely belongs on the Understandibg JSON Schema website, probably not in the spec document.

Is there anywhere else in the spec where we detail similar comparisons?

awwright:
We document equivalencies where theres no-effect behavior, e.g. how `minLength: 0` is the same as no keyword. It might be worth it to mention that `enum` and `const` have similar behavior if we can keep it under a few words.

djpadz:
FWIW, I think that it's important to point out equivalencies wherever they may exist, not so much for those who are writing specs, but for those who are writing code generators.  It'd save a fair amount of confusion, not to mention unintentional interpretation.  Basically, something like this would tell an implementer that `const` can be coded as:

```python
def const(what):
    return enum([what])
```

handrews:
@djpadz there are too many ways for things to be equivalent for us to address _within_ the spec, which is more than long enough already.  However, adding notes for implementors (of anything- code generation included) to the json-schema.org site would be much appreciated!

That said, I don't feel strongly either way about this particular note.  If @Relequestual and @awwright are in favor of it, I'm fine with merging it.

PR #712, opened 2019-02-23
Update year to 2019

handrews:


handrews:
OK, this is the most trivial fix ever, and the entire point was to get the build to run as cleanly as possible, and now xml2rfc is segfaulting in Travis.  I am 100% sure that changing "2018" to "2019" did not cause a ***segmentation fault****.  Also, the first run of the build, core passed and the latter two failed, and then when I re-ran it, all three failed.

Which pretty clearly means it's an environmental error.  I would like to post other PRs without the date error, so I'm merging this anyway.

PR #713, opened 2019-02-23
Note undefined behavior with unknown $ref targets

### LINKED ISSUE #687: Resolving $refs defined in unknown keywords
How should a json-schema (draft7) implementation resolve `$ref`s defined in unknown keywords?

JSON-Schema-Test-Suite defines schemas such as this, and I assume they are valid:

```json
{
    "tilda~field": {"type": "integer"},
    "slash/field": {"type": "integer"},
    "percent%field": {"type": "integer"},
    "properties": {
        "tilda": {"$ref": "#/tilda~0field"},
        "slash": {"$ref": "#/slash~1field"},
        "percent": {"$ref": "#/percent%25field"}
    }
}
```

Take this example below:

```json
{
    "$id": "http://example.com/root.json",
    "definitions": {
        "A": { "type": "integer" }
    },
    "properties": {
        "$id": {
            "type": "string"
        },
        "attributes": {
            "$ref": "#/tilda~0field/slash~1field/$id"
        }
    },
    "tilda~field": {
        "$id": "t/inner.json",
        "slash/field": {
            "$id": {
                "$id": "test/b",
                "$ref": "document.json"
            }
        }
    }
}
```

Which of the following is the `$ref` at `#/tilda~0field/slash~1field/$id/$ref` resolved to?

* http://example.com/root.json/document.json
* http://example.com/t/document.json
* http://example.com/t/test/document.json

Which of the `$id`s in `#/tilda~0field` must be considered as baseURI for the `$ref` in question and why.

PS I'm not sure if this is the right place to post this. If I need to clarify further or post this elsewhere, please let me know.

handrews:
Fixes #687 (in the sense of "it's clear what you are and are not required to implement", not in the sense of "solves the underlying applicator keyword recognition issue", which will be part of the draft after this as it is quite involved").

The most common situation here, which are basically alternate
names for "definitions" or "$defs", will generally work, but
this makes clear that there is no obligation to attempt to
guarantee the proper behavior in such situations.

We might want to narrow this behavior down in the future,
but for now this at least clearly removes any possible burdensome
and totally impractical inferencing requirements.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1552,7 +1552,19 @@
                         nested subschemas, which would be subject to the processing rules for
                         "$id".  Therefore, having a reference target in such an unrecognized
                         structure cannot be reliably implemented, and the resulting behavior
-                        is undefined.
+                        is undefined.  Similarly, a reference target under a known keyword,
+                        for which the value is known not to be a schema, results in undefined
+                        behavior in order to avoid burdening implementations with the need
+                        to detect such targets.
+                        <cref>
+                            These scenarios are analogous to fetching a schema over HTTP
+                            but receiving a response with a Content-Type other than
+                            application/schema+json.  An implementation can certainly
+                            try to interpret it as a schema, but the origin server
+                            offered no guarantee that it actually is any such thing.
+                            Therefore, interpreting it as such has security implications
+                            and may produce unpredictable results.
+                        </cref>
                     </t>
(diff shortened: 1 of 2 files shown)


handrews:
@Relequestual I'll do the change log as a batch near the end- lots of things are missing, but the issues are all tagged with the milestone so it won't be hard to go through and write it up.

Issue #697, opened 2018-12-17
Define Validation spec vocabularies


handrews:
Now that the `$vocabulary` PR is nearing approval, we need to figure out how many vocabularies are present in the Validation spec, and what to call them.

### Minimal arrangement

I'm guessing something like:

* `validation`: The validation assertions in [section 6](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6)
* `basic-format`: The `format` keyword and its standardized values in [section 7](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-7)
* `embedded-content`: The `content*` keywords in [section 8](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-8)
* `basic-metadata`: The annotations in [section 9 (formerly section 10)](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-10)

I'm not super-attached to those names, but I do think this nicely encapsulates both the different purposes of some groups of keywords, and the optional nature of `format` and `content*`.  Optional keyword groups were always particularly confusing for users and implementors.

The reason for `basic-format` is that it's extensible so I assume other people will probably have vocabularies using `format` in the name.  But I could be convinced to go with `standard-format`, `format`, or the plural of any of those (`standard-formats`, etc.).

### Multiple format vocabularies?

While `format` currently says that if you implement the keyword at all, you SHOULD implement all of the formats, this has become increasingly burdensome as the set of standard formats has expanded.  

We could break them up and provide vocabularies that each only declare the semantics for a subset of the standard formats.  An example division could be:

* `date-time-formats`
* `internet-formats` (hostnames, IP addresses, email, URIs, IRIs)
* `json-pointer-formats` (JSON Pointer and Relative JSON Pointer)
* `regex-format`

The `internet-formats` could be split up more, although I do wonder at what point it's more trouble than its worth.

@Julian @johandorland @gregsdennis as implementors what would you find useful here?

Julian:
From a practical standpoint, what I do to be honest is that formats that have an existing library supporting them get supported, and ones that don't... don't :/ until someone writes one.

E.g., IIRC there isn't a decent RFC 3339 parser for `time` that I could find, so it didn't go in.

Which isn't to say something like this wouldn't be useful -- probably it would just from categorizing purposes.

johandorland:
[...]

PR #714, opened 2019-02-23
Identify vocabularies, update meta-schemas

### LINKED ISSUE #612: Name meta schemas something other than "draft-"
We often get complaints about how confusing the draft numbering is, not just from people new to JSON Schema, but from people such as members of the IETF JSON working group.

We kept the sequential draft numbering for meta-schemas mostly because of inertia, and because that's how many of us (and others) would talk about future work even while things were stalled on draft-04 (the last meta-schema where the number appeared in at least one of the IETF draft names).

I propose dropping this for "draft-08" and calling the meta-schemas something else, ideally along with a shift to HTTPS.  Since we will have a larger set of modular vocabulary meta-schemas (in addition to the core+applicators+validation+basic-annotation, and all of that +hyper-schema meta schemas that we have now, which will be retained for convenience), it feels like a good opportunity to reconsider how we identify these.

One option might be to take a cue from modern JavaScript and use years, or years+month as we at least sometimes publish two updates a year.  Assuming a September publication of the next set:

`meta` would be for the convenience meta-schemas, which are all that most people would use:

`https://json-schema.org/09-2018/meta/schema`
`https://json-schema.org/09-2018/meta/hyper-schema`

`vocab` would be for per-vocabulary meta-schemas, which are mostly building blocks for the convenience meta-schemas (see #561)

`https://json-schema.org/09-2018/vocab/core`
`https://json-schema.org/09-2018/vocab/applicators`
`https://json-schema.org/09-2018/vocab/validation`
`https://json-schema.org/09-2018/vocab/basic-annotation`
`https://json-schema.org/09-2018/vocab/hyper-schema`

I don't like the name `basic-annotation` but that's not the point, please don't fixate on the exact list of vocabularies or their name, we will sort that out as the vocab PRs get written.

I'm not entirely sure what to do with the LDO schema.  It's not really a meta-schema, as LDOs on their own are not schemas.  Although they contain schemas.  Possibly still `meta`?  Or maybe something else like `components`?  If we go forward with a hypermedia operations object, there would likely be another schema of this sort.

`https://json-schema.org/09-2018/components/links`

Thoughts?

### LINKED ISSUE #641: Make ordering of keys consistent
It's the little details that matter in life -- we love consistency, so let's bring it to all of the schemas.

One example of many...

```
"$schema": "http://json-schema.org/draft-08/hyper-schema#",     "$id": "http://json-schema.org/draft-08/hyper-schema-output",
"$id": "http://json-schema.org/draft-08/hyper-schema#",         "$schema": "http://json-schema.org/draft-08/schema#",
"title": "JSON Hyper-Schema",                                   [Missing]
...
```

handrews:
Fixes #612 and #697 (Edit: and now fixes #641 )

Includes verbiage on the likelihood of updated meta-schemas
with later dates.

The hyper-schema output format has not yet been updated to
conform to the recent annotion/error output work.

I ended up calling the meta-data keywords URIs "meta-data" because "meta/meta" was ridiculous, but "meta/meta-data" is tolerable.


handrews:
I suspect CI will continue to fail with segfaults in xml2rfc ¯\_(ツ)_/¯

handrews:
Now also fixes #641

meta/validation.json
```diff
@@ -97,5 +70,32 @@
                 }
             ]
         }
+    },
+    "$defs": {
```
gregsdennis: I don't know... I like `$defs` earlier so that I know a thing when it's referenced.

jsonschema-core.xml
```diff
@@ -1881,17 +1888,24 @@
...
+            </t>
             <t>
                 The current URI for this vocabulary, known as the Applicator vocabulary, is:
-                <eref target="https://json-schema.org/draft-08/vocabularies/applicator"/>.
+                <eref target="https://json-schema.org/draft/2019-03/vocab/applicator"/>.
             </t>
             <t>
                 The current URI for the corresponding meta-schema is:
-                <eref target="https://json-schema.org/draft-08/applicator"/>.
+                <eref target="https://json-schema.org/draft/2019-03/meta/applicator"/>.
```
gregsdennis: Maybe not in this PR, but should we have language identifying this `vocab`/`meta` arrangement as a suggested pattern?

meta/validation.json
handrews: Having worked with a large set of schemas, often with large definitions sections, it was maddening trying to find the rest of the top-level schema for me.

I eventually got fed up and hand-reformatted around 60 schema files, IIRC.  It was that annoying.  `$defs` is a single keyword you can search for, so it's really easy to jump to.

I'm not going to fight to the death over this, just making the point that it wasn't a coin flip.

jsonschema-core.xml
handrews: Possibly- definitely not in this PR but it's an interesting idea.  Might be better in a best practices section on the site or understanding JSON Schema?  Not sure- feel free to file it as an issue on the spec repo if you want.

gregsdennis:
@handrews would you say that the sub-meta-schemas (e.g. *core.json*, *format.json*, etc.) would be a valid source for a machine-readable list of valid keywords defined by the vocabulary?  I know we were trying to come up with a distinct vocabulary format, but I think this will do well.

meta/validation.json
gregsdennis: I see it analogous to coding in C, where you have to declare your variables before you use them.  I can see your point of making it more like an appendix, though, too.

Meh.

handrews: I kind of see it like an embedded library.  I typically want to see the main structure first, and then drill down into what libraries it uses.  Sometimes I don't care about the details, so
I might have reason to trust that `"#/$defs/postalAddress"` does what I want, but I definitely want to see where it is used.

I also sometimes split schemas from `$defs` out into separate files, and it's nice (to me anyway) that the top section doesn't change beyond updating the `$ref` when that happens.

handrews:
@gregsdennis 

> would you say that the sub-meta-schemas (e.g. core.json, format.json, etc.) would be a valid source for a machine-readable list of valid keywords defined by the vocabulary? I know we were trying to come up with a distinct vocabulary format, but I think this will do well.

It's at least theoretically possible but is harder than it sounds.  In a flat specification where it's immediately clear what properties are allowed on the schema object, it's easy.  But once you start combining things, then in order to find all of the keywords you have to understand what other keywords are in-place applicators.

Which means that if there are extension vocabulary keywords being used in the meta-schema to describe your extension, then you get into a chicken and egg problem of not being able to tell which objects in the meta-schema are describing the schema objects, and which are not.

I'm pretty sure that it's about 1000 times easier to define a straightforward approach than it is to overload meta-schemas.  Meta-schemas are designed to tell you whether a schema is valid to use or not.  They are not designed to tell you how the schema works.  Let's leave them with a single responsibility rather than complicating the concept.

hyper-schema.json
```diff
@@ -1,22 +1,22 @@
 {
-    "$schema": "http://json-schema.org/draft-08/hyper-schema#",
-    "$id": "http://json-schema.org/draft-08/hyper-schema",
+    "$schema": "http://json-schema.org/draft/2019-03/hyper-schema#",
+    "$id": "http://json-schema.org/draft/2019-03/hyper-schema",
+    "$vocabulary": {
+        "https://json-schema.org/draft/2019/03/vocab/core": true,
```
dlax: should be "2019-03" instead of "2019/03", no? same for "schema.json" below.

jsonschema-core.xml
```diff
@@ -1058,11 +1058,18 @@
                 </t>
                 <t>
                     The current URI for the Core vocabulary is:
-                    <eref target="https://json-schema.org/draft-08/vocabularies/core"/>.
+                    <eref target="https://json-schema.org/draft/2019-03/vocab/core"/>.
                 </t>
                 <t>
                     The current URI for the corresponding meta-schema is:
-                    <eref target="https://json-schema.org/draft-08/core"/>.
+                    <eref target="https://json-schema.org/draft/2019/03/meta/core"/>.
```
dlax: "2019/03" -> "2019-03"

jsonschema-validation.xml
```diff
@@ -514,6 +532,20 @@
                     format attribute and instance SHOULD succeed.
                 </t>
 
+                <t>
+                    Meta-schemas that do not use "$vocabulary" SHOULD be considered to
+                    require this vocabulary as if its URI were present with a value of true,
+                    although see the Implementation Requirements below for details.
```
dlax: *also* see

(another occurrence below)

meta/validation.json
dlax: I tend to agree with having `"$defs"` after the main structure.

gregsdennis:
@handrews, I don't think it's too difficult. The meta-schemas provide the list of keywords they define in the top-level `properties` keyword.  This works for both past versions as well as this one.  There are no semantics, but the keywords are there along with the types of values they expect.  I think it's enough to have this list and let a "vocabulary implementation" worry about the meaning behind the keyword.

handrews:
@gregsdennis the current meta-schemas happen to work that way.  There is no guarantee that all meta-schemas will work that way, nor should meta-schemas be restricted to defining keywords in a single level of `properties`.  And this does not even get into managing extensible enumerations like `format`.

This is a very complex topic, and a PR for a much simpler step is not the place to debate it.  You are welcome to file an issue, but I'm not putting any more vocabulary things in this draft.  I don't think relying on the meta-schema this way is the right way to go and I want to leave the options open based on feedback.

jsonschema-validation.xml
handrews: @dlax I did actually mean "although" here, because the Implementation Requirements explains the weird optional nature of `format`, which is an exception to how the other vocabularies (except `content*`, which is the other "although") behave.

So I see this as a contrast to be called out ("although") rather than an affirmation or extension of the normal situation ("also").  Does that make sense?

hyper-schema.json
handrews: 🤦‍♂️

dlax:
@handrews, yes, it makes sense. Thanks for clarifying!

jsonschema-validation.xml
dlax: @handrews, yes, it makes sense. Thanks for clarifying!

handrews:
@gregsdennis aside from the stuff I asked be given separate issues, are you OK with this PR?

gregsdennis:
I have no issues. Just thinking out loud.

raotkind:
🤙

meta/validation.json
Relequestual: My personal preference would be for at the start, but it's a very weekly held preference. At the end is also fine.

jsonschema-validation.xml
Relequestual: @dlax OK to resolve this?

Relequestual:
All looks good.
I have one comment and one suggestion mind, if not too annoying...

Comment: Does the meta/meta-data schema not "annotations"? If so, couldn't it be "meta/annotations-schema"?

Suggestion: Can each schema include a very breif one scentence using `description`?
It took me a few mins to understand what meta/meta-data was fore. If it took me a few, could take others a lot longer, and a simple scentence might help avoid that.

Otherwise, as I said, all looks good!
Great work Mr @handrews  =]

jsonschema-validation.xml
dlax: @Relequestual sure! It was already "resolved" from my point of view :)

handrews:
@Relequestual at one point I had things named assertion and annotation but those are keyword classifications that are much broader than either vocabulary.  So I used the name from the appropriate section of the document.

I do have applicators named that way, but I couldn't come up with anything better, and applicators are more of a specialty thing in extensions while assertions and annotations are very broad categories with lots of expected new keywords.

I'm going to go ahead and merge this- we can do the descriptions separately, and if you feel really strongly about the annotation naming thing we can follow up on that.

PR #715, opened 2019-02-23
Add "uuid" format

### LINKED ISSUE #542: Add Defined Format: UUID
The UUID format is covered in [RFC 4122](https://tools.ietf.org/html/rfc4122). This standard could be a very simple addition to the default set of Defined Formats that JSON schema comes built in with.

There is a standard format that could be encoded as a pattern: https://en.wikipedia.org/wiki/Universally_unique_identifier#Format

Something like the following would work:
```
^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$
```

handrews:
Fixes #542.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -8,6 +8,7 @@
 <!ENTITY RFC3339 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3339.xml">
 <!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
 <!ENTITY RFC3987 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3987.xml">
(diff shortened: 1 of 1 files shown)


Julian:
I haven't gone and re-read RFC4122, but from experience a *lot* of places "argue" about emitting UUIDs with or without hyphens.

IIRC the RFC says to include them in the BNF, but to me it's almost common enough to specifically call out if we intend without-hyphen to be invalid (or vice versa if we intend them to be valid).

handrews:
@Julian thanks, good point, I don't know anything about that.  Maybe I will split the "duration" change out of this so we can leave this open longer for feedback since there is a reasonable question here.

handrews:
@Julian my inclination is to start with saying that this format indicates / optionally validates that the string strictly conforms with the ABNF in the RFC (which does include the hyphens).  It's easier to relax a constraint later than add one.

Alternatively, if anyone thinks this is likely to cause problems, we can punt this until the next draft which will presumably focus on `format` a lot anyway.  I don't consider this as important as adding `duration` or fixing the reference for `hostname`.

Julian:
Sounds reasonable

PR #716, opened 2019-02-23
Update hostname RFC reference

handrews:
RFC 1123 modifies the syntax to allow a leading digit, which is
present in modern practice.


handrews:
@awwright 

> RFC 1034 ... Updated by: [18 references]
>
> Good grief

Right?  And 1123 isn't even one of the 18 listed!  I did skim through them (and the 5 that update RFC 1123) and as far as I can tell none of them further change the hostname syntax rules.

PR #723, opened 2019-02-27
Update xml2rfc required version

awwright:
Let's see if this works


awwright:
Seems to work now: https://travis-ci.org/json-schema-org/json-schema-spec/builds/499101137
This has lint warnings but still passed, is that what we had before? Or maybe, what do we want?

Issue #717, opened 2019-02-24
Meta-schema for enum should allow empty and non-unique arrays


Zac-HD:
Currently, the meta-schemas for (at least) drafts 3, 4, 6, and 7 contain the following:

https://github.com/json-schema-org/json-schema-spec/blob/4e016d927839ed264f8d182abe9dc18a7e7a2c62/schema.json#L139-L144

However, the prose specification says:

> The value of this [i.e. the "enum"] keyword MUST be an array.  
> This array SHOULD have at least one element.  Elements in the array SHOULD be unique.

Validation therefore ought to allow schema authors to use empty or non-unique arrays, because the RFC 2119 definition of `SHOULD` explicitly permits ignoring such clauses after considering the circumstances.  Alternatively, the specification could replace `SHOULD` with `MUST` for the relevant constraints.

Possibly related to #618?  First reported at Julian/jsonschema#529, via [`hypothesis-jsonschema`](https://github.com/Zac-HD/hypothesis-jsonschema).

handrews:
For such short documents, the number of meta-schema bugs is astounding :-/
(I say that having been responsible for a fair number of them, sadly).

Certainly related to #618 in that it would be fantastic if someone implemented #618.

I supposed we should update them for this (and I think `writeOnly` is missing from draft-07 as well, although I can't find a bug on it).  Which will annoy some people and please others- next draft we move to date-based versioning for meta-schemas (#714) so we can just add new versions with less confusion.  Hopefully.

The good news is that the meta-schemas are not normative, so feel free to edit your local copy.

Relequestual:
@Zac-HD Correctness is good, but I have to ask, do you have a use case for this?
I can think of one, but it's not very strong.

Zac-HD:
I have `hypothesis-jsonschema`, which takes a schema and produces documents that validate against it. I would like to generate schemas from the meta-schemas.

More generally though, the sole purpose of validation libraries is correctness. If we can't implement the spec that actually exists, why would anyone choose to use jsonschema?

Relequestual:
Ah OK, yup that makes sense.

And yes, of course, I agree. I wasn't suggesting we don't fix the issue (I made a PR and it's merged already before my comment). I was more asking, should we change the spec to require this or not.

The meta-schemas are always informative as opposed to normative, so if the meta-schema doesn't line up with the spec, we have a problem, and should fix it.

Relequestual:
[...]

PR #739, opened 2019-05-21
Fix enum definition according to spec.

Relequestual:
Issue: https://github.com/json-schema-org/json-schema-spec/issues/717

enum no longer requires minimum number of items or unique items, in accordance with the spec.


Relequestual:
Given this is a non-normative change, I don't think there's a need to wait the 14 day hold period.
Will wait for someone else to approve before merging.

PR #753, opened 2019-06-16
Deprecated clarfications around arrays

philsturgeon:


jsonschema-validation.xml
```diff
@@ -1043,8 +1043,16 @@
...
                 </t>
                 <t>
-                    A root schema containing "deprecated" with a value of true indicates the entire
-                    root schema MAY be removed in the future.
+                    A root schema containing "deprecated" with a value of true indicates that
+                    the entire resource beng described MAY be removed in the future.
+                </t>
+                <t>
+                    When the "deprecated" keyword is used to describe an array, it can be used
+                    in the tuple form to deprecate positional schemas. If "deprecated" is used
```
handrews: In general this is a great improvement.  Some nitpicks over language that I realized I used with you on slack but which is not really standardized so I was probably unclear.  Actually, I'm not even 100% clear on what I meant by them now that I try to write it out:

* "tuple form" means that the value of `items` is an array, and I guess I also meant that `additionalItems` is... hmm.. `false`?  Absent?  Technically a tuple would be a fixed-length array so I guess it should be `false`?  But tbh I never thought  this part through
* I don't know what I meant by "list form", if that was a term I used.  The single schema form (`items` is a single schema, `additionalItems` is irrelevant and therefore should be absent) describes a list of uniform type

Instead of "tuple form" / "list form" let's talk about "items" with an array of schemas value (and possibly with "additionalItems" also used) or "items" with a single schema value (in which case "additionalItems" is irrelevant and need not be discussed).

So if "additionalItems" is present, you can talk about deprecating individual positional schemas and/or deprecating the schema for positions after the last positionally-defined item.

I'm unsure what it ought to mean to set `"deprecated": true` in either the single schema form of `items` or in `additionalProperties`- I'd probably set `deprecated` in the schema containing those.  I think these cases are weird enough that we can ignore them for now and see if we get feedback.  It's the list-of-schemas form of `items` that is the interesting one to me.

jsonschema-validation.xml
```diff
@@ -1043,8 +1043,16 @@
...
+                    in the list form it can mark regular "items" as deprecated, whilst allowing
+                    "additionalItems". Vice verse, setting "deprecated" to true inside
+                    "additionalItems" would mark the additional items as deprecated, but not
+                    effect list usage.
```
handrews: You want "affect" here :)

jsonschema-validation.xml
Relequestual: I agree the language should use the same form as for the keywords being reference, such as in core...

> If "items" is an array of schemas...

I'll see if I can make some suggested changes soon.

philsturgeon:
I think I fixed it?

Relequestual:
> I'm unsure what it ought to mean...

My feeling is that people will only look to use this if they know what they want to indicate.

@philsturgeon Say the first item in `items` is deprecated. Does that mean the second item then becomes the first version? I can't see it means anything else. Right?

handrews:
@Relequestual some languages allow undefined array positions.

But really it means "it won't look like this anymore soon", it does not say what it _will_ look like.  The default assumption is that it goes away, but you could also just use it to indicate an overall change.  I guess.

handrews:
@philsturgeon @Relequestual what still needs to be done here?

philsturgeon:
I am not sure, Ben thought there was further work to be done but I thought you answered his questions, so... break the deadlock with a thumb? Or, Ben, if you can let me know of a specific actionable thing? 

--
Phil Sturgeon
@philsturgeon

jsonschema-validation.xml
```diff
@@ -1043,8 +1043,14 @@
...
-                    A root schema containing "deprecated" with a value of true indicates the entire
-                    root schema MAY be removed in the future.
+                    A root schema containing "deprecated" with a value of true indicates that
+                    the entire resource beng described MAY be removed in the future.
+                </t>
+                <t>
+                    When the "deprecated" keyword is used to describe an array, it's meaning will
+                    vary depending on how "items" is used. If "items" is a single schema, it will
+                    deprecate the contents of an array, but if "items" is an array of schemas it
+                    can be used to deprecate positional schemas.
```
Relequestual: See, I understand this, but in the case where one wants to signal deprecation of a positional array element, wouldn't it make more sense to have `deprecated` set on the associated schema inside `items`?

jsonschema-validation.xml
```diff
@@ -1043,8 +1043,14 @@
                     is going to be removed in the future.
                 </t>
                 <t>
-                    A root schema containing "deprecated" with a value of true indicates the entire
-                    root schema MAY be removed in the future.
+                    A root schema containing "deprecated" with a value of true indicates that
+                    the entire resource beng described MAY be removed in the future.
+                </t>
+                <t>
+                    When the "deprecated" keyword is used to describe an array, it's meaning will
```
Relequestual: /is used to describe an array/is applicable to an array/

I figure we have the language of applicability so we should use it.

Relequestual:
@philsturgeon I guess the objective of this PR is to align with the expectations of people who see this when enclosed in an OpenAPI spec document, right?

As such, can you encourage someone from the OpenAPI spec team to take a little look and agree that this change meets said objective?

I've made a further review. One question, one suggested language change.
I think this is certinly clearer from where we started! Direction of travel is good! =]

jsonschema-validation.xml
philsturgeon: Have you seen this little fella? It's a really good feature if you've not tried poking it before.

image

jsonschema-validation.xml
```diff
@@ -1043,8 +1043,14 @@
                     is going to be removed in the future.
                 </t>
                 <t>
-                    A root schema containing "deprecated" with a value of true indicates the entire
-                    root schema MAY be removed in the future.
+                    A root schema containing "deprecated" with a value of true indicates that
+                    the entire resource being described MAY be removed in the future.
+                </t>
+                <t>
+                    When the "deprecated" keyword is used to describe an array, it's meaning will
```
Relequestual:
```suggestion
                    When the "deprecated" keyword is applicable to an array, it's meaning will
```

jsonschema-validation.xml
Relequestual: I knew it was there but I didn't know what it looked like. Ta!

philsturgeon: I was just pushing this fix haha thanks!

jsonschema-validation.xml
philsturgeon: Isn't that what this says? Can you try the suggestion feature again to improve the words? Words are hard.

handrews:
@philsturgeon @Relequestual any update on this?

philsturgeon:
I was trying to get an OAI person to review this. I'll bump it on their slack. We'll give it a few more days then just merge.

jsonschema-validation.xml
Relequestual: No, it says when it's applicable to an array, while if you want it to mean what I've said, you mean, "when it's applicable to an item in an array.

Currenly, it reads as if you define `deprecated` in the same object as `items`, rather than in a subschema defined inside `items`. See the difference?

If your intent is for it to be defined in the same object as `items` where `items` is an array of subschemas, how would you know to which subschema it applies positionally?

Relequestual:
I've added a comment on the review.

My query is regarding, if `deprecated` changes behaviour when `items` is an array of schemas, to indicate that one of the subschemas will be deprecated, how is the position known?

See my review comment.

jsonschema-validation.xml
philsturgeon: @Relequestual I took another stab at fixing this.

> but if "items" is an array of schemas it should be defined in the sub-schema targeted for deprecation.

does that make more sense? The previous sentance is saying:

´``
"items": {
   "type": "string",
   "deprecated": true
}
```

based on your understanding you think im suggesting it suddenly jumps _up_ a keyword somehow?

No it would look like this:

´``
"items": [{
   "type": "string",
}, {
   "type": "string",
   "deprecated": true
}],
```

Now the 2nd thing is going away.

Relequestual: I'll start a new review for this as this is now outdated...

jsonschema-validation.xml
```diff
@@ -1043,8 +1043,14 @@
...
                 </t>
                 <t>
-                    A root schema containing "deprecated" with a value of true indicates the entire
-                    root schema MAY be removed in the future.
+                    A root schema containing "deprecated" with a value of true indicates that
+                    the entire resource being described MAY be removed in the future.
+                </t>
+                <t>
+                    When the "deprecated" keyword is applicable to an array, it's meaning will
+                    vary depending on how "items" is used. If "items" is a single schema, it will
```
Relequestual: > The previous sentance is saying:
> 
> "items": {
>    "type": "string",
>    "deprecated": true
> }
> 
> based on your understanding you think im suggesting it jumps up a level somehow?

`When the "deprecated" keyword is applicable to an array...`
This means, when the keyword is in a schema which is applied to an array.
`maxItems` would be an example of a keyword which is applicable to an array.
The value of `items` is an array of subschemas, which are then applied to each item in the array.

Based on your previous comments, what you mean is "When the `deprecated` keyword is applied to an item in an array by means of `items`, if `items` is a single schema, the deprecation relates to the whole array, while if `items` is an array of schemas, the deprecation relates to the corrosponding item according to the subschemas position.

philsturgeon: Ok I've put in that exact wording! Thumbs? :D

PR #756, opened 2019-06-30
added some verbiage to the vocabularies example for clarity

gregsdennis:
Some of the text in the example isn't proper "spec-ese" but in the context of an example, I think that's fine.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1074,32 +1074,32 @@
             </section>
             <section title="Example Meta-Schema With Vocabulary Declarations"
                      anchor="example-meta-schema">
+                <t>
+                    This meta-schema explicitly declares both the Core and Applicator
+                    vocabularies with its own vocabulary and combines the
+                    Core and Applicator meta-schemas with an "allOf".  It does not
+                    need to include itself in the "allOf" since it is its own
+                    meta-schema.  It additionally applies three modifications to the
+                    combination of these vocabularies:
+                    <list>
+                        <t>
+                            It restricts the usage of the Applicator vocabulary
+                            by forbidding the keyword prefixed with "unevaluated".
+                        </t>
+                        <t>
+                            It describes a keyword, "pastDate", that is defined by
+                            its own vocabulary.
+                        </t>
+                        <t>
+                            It describes a keyword, "localKeyword", that is not
+                            part of any vocabulary.
+                        </t>
+                    </list>
+                    Note that it is its own meta-schema, as it relies on both the Core
+                    vocabulary (as all schemas do) and the Applicator vocabulary (for
+                    "allOf", "properties", and "patternProperties").
+                </t>
                 <figure>
-                    <preamble>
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -1121,6 +1143,31 @@
...
+                    declares this meta-schema for their "$schema".  Note that defining
+                    it in this way does not imply the new keywords are valid within an
+                    instance to be validated by such a schema.
+                </t>
+                <t>
+                    Although the meta-schema can validate the syntax of "startDate"
+                    within a schema, it is the vocabulary that defines the logic behind
+                    the semantic meaning of "startDate".  Without an understanding of
+                    the semantics (that the instance value should represent a date after that
+                    given in the schema), an implementation can only validate the syntactic
```
Relequestual: > that the instance value should represent a date after that given in the schema

Would it be valid to add this in the comment for this field, showing that information is contained in annotation which isn't avilable in the meta-schema validation mode, which is required for "proper use" of the vocabulary?

jsonschema-core.xml
```diff
@@ -1121,6 +1143,31 @@
...
+                    usage of "startDate" (that it is a date-formatted string).  This is why
+                    a schema MUST refuse to validate a schema which uses a meta-schema
```
Relequestual: If this is purely informative and not normative (as in it's already covered earlier), then lowercase that `MUST`, imho.

Relequestual:
"Request changes" is more for me =]

jsonschema-core.xml
```diff
@@ -935,8 +935,9 @@
                     such schemas.
                 </t>
                 <t>
-                    When processing a schema that uses unrecognized vocabularies, keywords
-                    declared by those vocabularies are treated like any other unrecognized
+                    When processing a schema that uses an unrecognized vocabulary
+                    (implying that the value for the vocabulary must be false), keywords
```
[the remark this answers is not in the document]
gregsdennis: Sure. Have a go.

jsonschema-core.xml
gregsdennis: Good catch. I changed the example mid-way through. I need to update this note.

gregsdennis: Possibly, sure. But I don't know that it's necessary for this example.  I put it here to contrast the syntactic usage later in the sentence.

jsonschema-core.xml
gregsdennis: ✔️

Relequestual:
Looks like I'm using an older version of xml2rfc.
We should make sure the required version is installed when running make.

gregsdennis:
This looks good to me.

gregsdennis:
Seems like @handrews went his own way.  Probably don't need this in the spec anymore.

Might still be useful for the site, though.

handrews:
@gregsdennis I had actually not intended to invalidate this and had meant to come back and look at reconciling, but perhaps it's better for me to just finish the PR I have up and we can see what's still needed.

Issue #646, opened 2018-08-01
Explain format (and content*) more clearly


handrews:
[this is a bit stream-of-consciousness, but I wanted to get it filed because I keep forgetting- we'll clean up the ideas here on the way to PRs]

`format` confuses pretty much everyone.  I have noticed people filing issues against various implementations complaining of imperfect enforcement (I believe @Julian has received complaints about "email", and @johandorland about "hostname", and I suspect they are not alone).

`format`, `contentMediaType`, and `contentEncoding` are essentially best effort validation keywords in practice.  Many if not most implementations make at least some effort to validate `format`.  I'm not sure if anything attempts that for `content*` as they are new (at least as part of the validation spec), and they would essentially require parsing the string encoding and media type which is potentially very expensive.

Complicating the matter for `format` is the fact that many of the relatively fundamental internet-related formats such as "email" and "hostname" are very old, and conformance to specifications is rather complicated.

For "hostname", [RFC 1034](https://tools.ietf.org/html/rfc1034#section-3) forbids leading digits, but this is sometimes ignored in practice, leading to ambiguous overlap with "ipv4" as a format.  In practice, most programs that accept hostnames will also accept ipv4 addresses and just recognize that no DNS resolution is required, so this is rarely a concern.

The difficulty of validating email addresses, even on the syntactical level, is well-documented (try finding a regular expression that will do it, for instance, and if you find an actual iron-clad one, let me know).

--------

Leveraging our relatively recent keyword classification work, I think it is best to classify these primarily as _annotations_ rather than treating them as some sort of hybrid annotation+assertion.  Annotations can specify any intent, including semantic validation or parsing instructions.  The specification should provide guidance on how an implementation might directly offer handlers for such intents, and how to indicate the available level of support.

Applications can, as with any annotation, then perform additional processing if the implementation either does not offer any validation, or offers only incomplete validation.  The spec already says that implementations SHOULD offer an ability to turn semantic validation off, so we can extend that guidance (probably at the MAY level) to cover situations like allowing hooks for application-defined processing in addition to or in place of implementation-supplied validation.

And of course, all of this is dependent on an implementation supporting annotations.  As with the `additionalProperties` and `additionalItems` keywords (now in the core spec and defined in terms of annotation collection), the spec should allow for the existing sort of implementations to continue to be valid and in conformance for implementations that do not implement general annotation collection support.

johandorland:
[...]

Issue #54, opened 2016-09-17
syntactic fallback validation for "format"


handrews:
### The Problem:  "format" is frequently re-implemented using "pattern" because it is unreliable

The "format" keyword is currently defined as an optional feature of JSON Schema.  This frees implementations from the relatively burdensome requirements of performing the specified semantic validations, but also intentionally makes the feature unreliable.  As a result, schema authors frequently re-define validation schemas for fields that could be completely described with the "format" keyword were its implementation consistent.

This places an undue burden on schema writers who wish to both take advantage of any full implementations and work around any minimal implementations.

Here is an example of a document (written in YAML for human-friendliness) the provides JSON Schemas for ipv4 and ipv6 addresses for use in other schemas from the same product in place of the "format" keyword:

https://support.riverbed.com/apis/sh.common/1.0/service.yml
### The Proposal

JSON Schema can provide a standard "pattern"-based schema for each format value in its meta-schema, which will provide a documented level of **purely syntactical** validation for instances.  This requires only trivial additional work from implementations as shown below under "Mechanism".

Each such schema MUST successfully validate against all possible valid instances.  They MAY also successfully validate invalid instances due to the limits of regular expressions or the decision of the JSON Schema standard that the full pattern is too complex or has too much of a performance impact to support at all.
### Mechanism

A "formats" section would be added to the "definitions" within the meta-schema:

``` JSON
{
    "definitions": {
        "formats": {
            "definitions": {
                "ipv4": {
                    "minLength": 7,
                    "maxLength": 15,
                    "pattern": "^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$"
                },
                "email": {
                    "pattern": "you-get-the-idea"
                }
            }
        }
    }
}
```

The purpose of the nested "definitions" section is to clearly differentiate between definitions used only for format validation and definitions used to build the actual meta-schema.

If an implementation does not handle "format": "ipv4" directly, then the schema:

``` JSON
{
    "$schema": "http://json-schema.org/schema#",
    "type": "string",
    "readOnly": true,
    "format": "ipv4"
}
```

should be interpreted as:

``` JSON
{
    "$schema": "http://json-schema.org/schema#",
    "allOf": [
        {
            "type": "string",
            "readOnly": true
        },
[...]

PR #764, opened 2019-07-19
Specify "format" requirements in vocabulary terms

### LINKED ISSUE #732: Format attributes
It isn't totally clear that a validator should ignore any format attribute it can't check/doesn't understand. The closest the wording comes is in 7.2:

> Implementations MAY add custom format attributes

I think this implies that the validator is likely to come across format attributes it won't understand, but it probably ought to be explicit. Maybe something like:

> A validator MUST ignore any format attribute which it cannot check.

Or under the other two points:

> they SHOULD ignore any attribute value they don't validate

### LINKED ISSUE #759: format for non-string data types (complex validation rules)
*Transferred from the JSON Schema Slack Channel...*

**Background:** I'm working on a project, which passes workflows in JSON from clients to servers. These workflows we mostly validate with JSON Schema and (aim to) do the complex validation with `format` (see also https://github.com/Open-EO/openeo-processes/issues/67). Therefore, we have quite an [extensive list of formats](https://open-eo.github.io/openeo-api/draft/processes/#openeo-specific-formats). Some examples:
* `collection-id` checks in the database whether the specified dataset exists and can be used in the workflow. It's a string and the validation with ajv works well.
* `epsg-code` check against the EPSG code database. It's an integer and doesn't work with ajv as ajv only supports format on numbers and strings.
* We also have formats for object and arrays. These usually need to check the whole object or array and not individual items. For example, one property influences the validation of the second property or I'd like to check for two-element arrays that it doesn't contain two null values.

Specifying format on non-string data types doesn't seem very uncommon. [OpenAPI also has formats for non-string types](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#data-types), for example int32/int64 for integer and float/double for number.

**Problem:** The JSON spec itself only defines formats on string values. Also, [custom formats is described in the "Understanding JSON Schema" documentation as part of the string data type](https://json-schema.org/understanding-json-schema/reference/string.html#format). This seems to lead to the problem that several libraries only implement custom formats for strings (or a subset of JSON data types).  For example, ajv currently only supports format to be defined for strings and numbers, see also https://github.com/epoberezkin/ajv/issues/1039. kin-openapi also only supports string formats (yes, it's OpenAPI, but seems to be the same problem). 

In contrast, https://json-schema.org/latest/json-schema-validation.html#format doesn't mention this restriction and explicitly mentions for individual types that they apply for strings ('These attributes apply to string instances.'), which would not be required if format is only defined for strings anyway. Therefore, I couldn't really figure out from the spec whether `format` meant to be only for strings or for all data types.

**Proposal:** Therefore, I have asked on the Slack channel for a clarification in the spec ([see the discussion](https://json-schema.slack.com/archives/C5CF75URH/p1562596499404200)). 

@handrews wrote:
> Technically you could do all sorts of things with `format`, and yes, technically implementations that only support `format` with strings have a bug.
> *BUT*, `format` is problematic because it was an early, somewhat half-assed attempt at extensibility for complex semantic concepts.  It's not consistently supported (or, as you noted, always correctly supported), and there is a tremendous amount of debate over what is really required for proper support.
> Now we have the concept of annotation keywords, which can signify arbitrarily complex concepts to applications without dealing with `format`'s problems.

So it seems there should be a clarification in the spec that format is allowed for all data types. This would clarify and help to convince implementers to support format for more data types.

**Alternative:** In the Slack discussion, an alternative was discussed. I was proposed to use custom keywords, which seems to be supported by ajv (JS) and Manatee.Json (.Net), but I'd still want to figure out how well supported these are by the libraries out there (e.g. in Python land). I'd be happy to switch over to custom keywords if they offer the same functionality and are well supported.

It was also suggested to deprecate `format` in favor of custom keywords. @handrews wrote:

> Actually, I *really* like this idea.  `format` causes nothing but pain, really.  For users, maintainers, and for us.  It would still be supported in the forthcoming draft, but deprecated to encourage usage of the vocabulary concept.  Which would be great because _then maybe people would be motivated to use vocabularies and give us feedback on them_
> We could say that `format` will not be removed unless/until vocabularies prove to be an adequate replacement.

(Although the alternative seems to be a good way forward, clarification regarding `format` still would be worth the effort, I think.)

*For the full discussion, please [see the Slack channel](https://json-schema.slack.com/archives/C5CF75URH/p1562596499404200). My quotes are only short excerpts.*

handrews:
Fixes #732, fixes #759, addresses the `format` part of #646.
The `content*` of #646 will be done in a separate PR.

Use the true vs false values of vocabulary declaration to
indicate whether a schema author requires assertion behavior
(such as because a keyword such as "oneOf" depends on such
validation functioning correctly).

Use a false value in the standard core+validation vocabulary,
reflecting the historical lack of requirement for this
keyword to be implemented.

Define annotation behavior when the vocabulary is declared with
false, to facilitate the recommended best practice of performing
semantic validation in your application.  While not the typical
false vocabulary behavior, this seems like the best way to
reframe the historically unpredictable behavior.

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -6,7 +6,7 @@
         "https://json-schema.org/draft/2019-08/vocab/applicator": true,
         "https://json-schema.org/draft/2019-08/vocab/validation": true,
         "https://json-schema.org/draft/2019-08/vocab/meta-data": true,
-        "https://json-schema.org/draft/2019-08/vocab/format": true,
+        "https://json-schema.org/draft/2019-08/vocab/format": false,
         "https://json-schema.org/draft/2019-08/vocab/content": true,
         "https://json-schema.org/draft/2019-08/vocab/hyper-schema": true
     },
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -174,14 +174,15 @@
 
         </section>
 
-        <section title="Meta-Schema">
+        <section title="Meta-Schema" anchor="meta-schema">
             <t>
                 The current URI for the JSON Schema Validation meta-schema is
                 <eref target="http://json-schema.org/draft/2019-08/schema#"/>.
                 For schema author convenience, this meta-schema describes all vocabularies
                 defined in this specification and the JSON Schema Core specification.
                 Individual vocabulary and vocabulary meta-schema URIs are given for
-                each section below.
+                each section below.  Certain vocabularies are optional to support, which
+                is explained in detail in the relevant sections.
             </t>
             <t>
                 Updated vocabulary and meta-schema URIs MAY be published between
@@ -503,28 +504,42 @@
             </section>
         </section>
 
-        <section title='A Vocabulary for Semantic Validation With "format"' anchor="format">
+        <section title='A Vocabulary for Semantic Content With "format"' anchor="format">
 
             <section title="Foreword">
                 <t>
-                    Structural validation alone may be insufficient to validate that an instance
-                    meets all the requirements of an application. The "format" keyword is defined to
-                    allow interoperable semantic validation for a fixed subset of values which are
+                    Structural validation alone may be insufficient to allow an application to correctly
+                    utilize certain values.  The "format" annotation keyword is defined to allow schema
+                    authors to convey semantic information for a fixed subset of values which are
                     accurately described by authoritative resources, be they RFCs or other external
                     specifications.
+                </t>
+
+                <t>
+                    Implementations MAY treat "format" as an assertion in addition to an annotation,
+                    and attempt to validate the values conformance to the specified semantics.
+                    See the Implementation Requirements below for details.
                 </t>
 
                 <t>
                     The value of this keyword is called a format attribute. It MUST be a string. A
                     format attribute can generally only validate a given set of instance types. If
                     the type of the instance to validate is not in this set, validation for this
-                    format attribute and instance SHOULD succeed.
+                    format attribute and instance SHOULD succeed.  All format attributes defined
+                    in this section apply to strings, but a format attribute can be specified
+                    to apply to any instance types defined in the data model defined in the
+                    <xref target="json-schema">core JSON Schema.</xref>
+                    <cref>
+                        Note that the "type" keyword in this specification defines an "integer" type
+                        which is not part of the data model.  Therefore a format attribute can be
+                        limited to numbers, but not specifically to integers.
+                    </cref>
                 </t>
 
                 <t>
                     Meta-schemas that do not use "$vocabulary" SHOULD be considered to
-                    require this vocabulary as if its URI were present with a value of true,
-                    although see the Implementation Requirements below for details.
+                    utilize this vocabulary as if its URI were present with a value of false.
+                    See the Implementation Requirements below for details.
                 </t>
                 <t>
                     The current URI for this vocabulary, known as the Format vocabulary, is:
@@ -539,26 +554,60 @@
 
             <section title="Implementation Requirements">
                 <t>
-                    The "format" keyword functions as both an annotation
-                    and as an assertion.  While no special effort is required to
-                    implement it as an annotation conveying semantic meaning, implementing
-                    validation is non-trivial.
-                </t>
-                <t>
-                    Implementations MAY support the "format" keyword as a validation assertion.
-                    Should they choose to do so:
-
+                    The "format" keyword functions as an annotation, and optionally as an assertion.
+                    Declaring the format vocabulary in "$vocabulary" with a value of true indicates
+                    that an implementation MUST treat the keyword as an assertion.
+                    This means that:
                     <list>
                         <t>they SHOULD implement validation for attributes defined below;</t>
                         <t>they SHOULD offer an option to disable validation for this keyword.</t>
                     </list>
-
                 </t>
 
                 <t>
-                    Implementations MAY add custom format attributes. Save for agreement between
-                    parties, schema authors SHALL NOT expect a peer implementation to support this
-                    keyword and/or custom format attributes.
+                    Due to the complexity involved in fully validating some format attributes
+                    defined in this specification, implementations MAY provide only limited
+                    validation support for some format attributes.  Implementations SHOULD
+                    document any such intentional limitations.
+                </t>
+
+                <t>
+                    The <xref target="meta-schema">standard core and validation meta-schema</xref>
+                    includes this vocabulary in its "$vocabulary" keyword with a value of false,
+                    since by default implementations are not required to support this keyword
+                    as an assertion.
+                </t>
+
+                <t>
+                    When the format vocabulary is declared with a value of false, an implementation
+                    SHOULD treat "format" as an annotation keyword, to facilitate applications which
+                    wish to do their own semantic validation.
+                    <cref>
+                        This is not the normal behavior for a vocabulary declared with false, which is
+                        why the requirement is a SHOULD.  Implementations MAY ignore "format" entirely
+                        as is allowed by false vocabulary declarations.  However, due to the long history
+                        of this keyword, treating it as something of a special case seems reasonable.
+                        This may be revised in future drafts based on feedback.
+                    </cref>
+                </t>
+                <t>
+                    Implementations MAY treat "format" as an assertion when the vocabulary is declared
+                    with false, as false vocabularies are optional rather than forbidden.  However,
+                    as noted above, implementations SHOULD provide a way to disable such validation.
+                </t>
+                <t>
+                    Implementations MAY support custom format attributes. Save for agreement between
+                    parties, schema authors SHALL NOT expect a peer implementation to support such
+                    custom format attributes.  An implementation MUST NOT fail
+                    validation or cease processing due to an unknown format attribute.
+                    If treating "format" as an annotation, implementations SHOULD collect both
+                    known and unknown format attribute values.
+                </t>
+                <t>
+                    Vocabularies do not support specifically declaring different value sets for keywords.
+                    Due to this limitation, and the historically uneven implementation of this keyword,
+                    it is RECOMMENDED to define additional keywords in a vocabulary rather than
+                    additional format attributes if interoperability is desired.
                 </t>
             </section>
 
@@ -1273,6 +1322,8 @@
                 <list style="hanging">
                     <t hangText="draft-handrews-json-schema-validation-02">
                         <list style="symbols">
+                            <t>Update "format" implementation requirements in terms of vocabularies</t>
+                            <t>Grouped keywords into formal vocabuarlies</t>
                             <t>Moved "definitions" to the core spec as "$defs"</t>
                             <t>Moved applicator keywords to the core spec</t>
                             <t>Renamed the array form of "dependencies" to "dependentRequired", moved the schema form to the core spec</t>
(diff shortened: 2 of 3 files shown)


hyper-schema.json
```diff
@@ -6,7 +6,7 @@
         "https://json-schema.org/draft/2019-08/vocab/applicator": true,
         "https://json-schema.org/draft/2019-08/vocab/validation": true,
         "https://json-schema.org/draft/2019-08/vocab/meta-data": true,
-        "https://json-schema.org/draft/2019-08/vocab/format": true,
+        "https://json-schema.org/draft/2019-08/vocab/format": false,
```
Relequestual: Yup. Good call.

jsonschema-validation.xml
```diff
@@ -503,28 +504,42 @@
...
+                    and attempt to validate the values conformance to the specified semantics.
+                    See the Implementation Requirements below for details.
+                </t>
+
                 <t>
                     The value of this keyword is called a format attribute. It MUST be a string. A
                     format attribute can generally only validate a given set of instance types. If
                     the type of the instance to validate is not in this set, validation for this
-                    format attribute and instance SHOULD succeed.
+                    format attribute and instance SHOULD succeed.  All format attributes defined
```
Relequestual: > If the type of the instance to validate is not in this set, validation for this format attribute and instance SHOULD succeed.

I've never noticed this before, but it sounds not what you want...
Say you have a format "smallint". If a string is provided at the applicable instance, validation will assert TRUE if the format says it is only applicable to numeric types.

I guess this makes sense, because it's similar to how other keywords work, in that keywords are often only applicable if they target an instance of the same type.

For example, `maxLength: 5` fails validation for a string of longer than 5 characters, but does not fail validation for a numeric value `123456789` because it is not the correct applicable type.

This is covered explicitly in draft-8 core (http://json-schema.org/work-in-progress/WIP-jsonschema-core.html#rfc.section.7.4.1) 

> 7.4.1. Assertions and Instance Primitive Types
> 
> Most assertions only constrain values within a certain primitive type. When the type of the instance is not of the type targeted by the keyword, the instance is considered to conform to the assertion.
> 
> For example, the "maxLength" keyword from the companion validation vocabulary will only restrict certain strings (that are too long) from being valid. If the instance is a number, boolean, null, array, or object, then it is valid against this assertion. 

But, I feel we should maybe re-reference this section here for clarity...

jsonschema-validation.xml
```diff
@@ -539,26 +554,60 @@
...
                         <t>they SHOULD implement validation for attributes defined below;</t>
                         <t>they SHOULD offer an option to disable validation for this keyword.</t>
                     </list>
+                </t>
 
+                <t>
+                    Due to the complexity involved in fully validating some format attributes
+                    defined in this specification, implementations MAY provide only limited
+                    validation support for some format attributes.  Implementations SHOULD
+                    document any such intentional limitations.
```
Relequestual: I hear you on this.
One of the reasons I was going to shunt this issue to draft-9 was I wanted to provide a MINIMUM requirement for the MUST defined above in line 559.

Otherwise, the MUST on 559 holds no actual requirement, because it's then defined that meeting said MUST is actually meeting two SHOULD requirements (and we're back to... whatever the implementation wants to do).

The point of the issues around format are providing AT LEAST SOMEthing when format is used if required.

jsonschema-validation.xml
```diff
@@ -539,26 +554,60 @@
...
+                <t>
+                    When the format vocabulary is declared with a value of false, an implementation
+                    SHOULD treat "format" as an annotation keyword, to facilitate applications which
+                    wish to do their own semantic validation.
+                    <cref>
+                        This is not the normal behavior for a vocabulary declared with false, which is
+                        why the requirement is a SHOULD.  Implementations MAY ignore "format" entirely
+                        as is allowed by false vocabulary declarations.  However, due to the long history
+                        of this keyword, treating it as something of a special case seems reasonable.
+                        This may be revised in future drafts based on feedback.
```
Relequestual: There's been some discussion around "should all unknown keywords be collected as annotations for application use?". Doing so would prevent the need to define this explicit behaviour, which as you note, is different to when other vocabs are given a false value.

jsonschema-validation.xml
handrews: I'd rather just take the "SHOULD succeed" language out entirely.  We used to have that phrasing everywhere and this is just left over from that.  As you note, this is typical behavior, so it's better not to call it out- calling things out should be done for atypical behavior.

jsonschema-validation.xml
handrews: Yeah I decided to separate that conversation from this one.  This will work fine now, and does not prevent us from making annotation collection the default behavior later.  Either before or after publishing this draft, but we really do need to wrap this up soon, and changing the overall default behavior may be pretty contentious.

Clarifying `format` is fine to do because it is such an enormous source of complaints, and none of this is incompatible with existing behavior.

jsonschema-validation.xml
handrews: This does require at least something- it just does not define what that something is, and there is NO WAY IN HELL we will get any consensus from implementors on defining a minimum for each format attribute within the next couple weeks.

What this does is ensures that the keyword produces a validation result, and gives both the implementation and schema author _some_ level of control.  Previously, the schema author had _no_ control over how things were processed.

This does not solve everything but it's a big improvement and I'm not willing to let the perfect be the enemy of the good here.  For me personally, `format` is unsalvageable as any sort of interoperable feature and should be discouraged (if not formally deprecated or removed- it's far too heavily used for that right now) in favor of new vocabularies.

handrews: @Relequestual OK now that I'm more awake and have had a chance to think about it more (I really shouldn't- SHOULD NOT? - respond to comments right when I first wake up) I think that having the vocabulary true/false option means that we can strengthen these requirements without having a huge fight over the minimal requirements.  Which can be pretty minimal with a warning that they may be strengthened in the future.

We can do this because the default usage of the vocabulary is still `false`, which is essentially the same as what it has always been.  As is noted in the text, even when the vocabulary is `false`, an implementation MAY try to validate it anyway.

So let's do this:

* If the vocabulary is `true`, then all of the format attributes MUST be supported to some minimum level along the lines of #54.  Otherwise the implementation must raise an unsupported vocabulary error.
* If the vocabulary is `false`, then it's all MAY.  Which includes the SHOULD level that is currently specified (and, as far as I can tell, very frequently ignored to some degree or another).

Again, since the default is `false`, this is an entirely backwards-compatible change.

I will revise the PR.  I think I may move a bit of this text and quite a bit more commentary into an appendix of guidance on implementing and (if necessary e.g. because OpenAPI has a huge format registry already) extending`format`.

jsonschema-validation.xml
Relequestual: Agreed.
Is there some text about "if a keyword isn't applicable to the types it may be applied to, then it is not applied and therefore is the same as a true assertion" generally?

Take `minimum`:

> The value of "minimum" MUST be a number, representing an inclusive lower limit for a numeric instance.
> 
> If the instance is a number, then this keyword validates only if the instance is greater than or exactly equal to "minimum". 

The intent is clear, but the phrasing and wording is... not very spec like.
We don't explicitly specify what happens if the instance is not a number... 
The behaviour is, in essense, undefined (unless I've missed something).

If that's the case... maybe a spec cleanup is a candidate for draft-9.

Sorry, I've got way off topic here... but it's relevant.

jsonschema-validation.xml
Relequestual: I think this is a totally reasonable and proportionate way forward.
I'm happy with a super bare minimum level of required validation in the instance where the format vocabulary is `true`: for example, email matches `^.+@.+\..+$`. Something that catches the equivilence of a glancing "huh... well that's clearly wrong" type issue.

Ultimatly, JSON Schema is (normally) first pass on validation, apart from the instances where it's baked into a database or the like.

jsonschema-validation.xml
Relequestual: Overall I think this is fine.
However, it does lead me to consider that this is placing an extra implementation SHOULD based on a vocab being false. Do you think it's worth saying something like " generally it is not recomend that behaviour is derived from a vocabulary having a false value"? or do you think that's overkill?

gregsdennis:
This looks good to me

jsonschema-validation.xml
handrews: This is covered more clearly in the new text I had written up on Sunday.  I'm just finishing up the minimum validation requirements now and will post an update by tonight.

handrews:
@Relequestual @gregsdennis @Julian @johandorland OK I have reworked this a bit more to add clarity (including clarity on where schema authors _cannot_ expect consistent or thorough validation) and ensured that there are consistent and reasonably predictable default behaviors.

There are two ways to control whether and to what degree `format` is validated:

1. The schema author can declare the format vocabulary with either `false` (the default) or `true`
2. The code that runs the validation can be configured to never perform `format` validation, always perform it, or (by default) behave based on the vocabulary value.

* `false` + default:  `format` MUST NOT be validated (this is the overall default)
* `false` + don't validate: `format` MUST NOT be validated
* `false` + validate: `format` MUST be validated to whatever extent is available, _however_ an implementation need not provide any useful validation, so it may not be possible to fail validation (a.k.a. what we have now)
* `true` + default: `format` MUST be validated, and it MUST be possible to fail validation
* `true` + don't validate: `format` MUST NOT be validated
* `true` + validated: `format` MUST be validated, and it MSUT be possible to fail validation

This is still a complicated mess but it's a more clear complicated mess and I think it is important to establish that the overall default behavior is to not validated at all, because that is the _only_ absolutely consistent option.

gregsdennis:
> **true + fmt validation configured off**

This is the weird one.  It's basically this conversation between client and implementation:

**Client:** "Oh, so, I don't want you to ever validate `format`."
**Implementation:** "But... the meta-schema says I _have_ to."
**Client:** "Did I stutter?"
**Implementation:** "..."

jsonschema-validation.xml
```diff
@@ -539,27 +557,141 @@
...
+                    </t>
+                </section>
+                <section title="Custom format attributes">
+                    <t>
+                        Implementations MAY support custom format attributes. Save for agreement between
+                        parties, schema authors SHALL NOT expect a peer implementation to support such
+                        custom format attributes.  An implementation MUST NOT fail
+                        validation or cease processing due to an unknown format attribute.
+                        When treating "format" as an annotation, implementations SHOULD collect both
+                        known and unknown format attribute values.
```
gregsdennis: "We" should build some tests around collecting format values as annotations.

gregsdennis:
I'm happy with the current state.

handrews:
@gregsdennis LOL "did I stutter?" 😆 

The point of that option is that there are always people who do not want to run the validation because it is expensive.  I agree that that is a weird case, and I'd be fine with saying that it's an invalid combination.  Let's see what other folks think on that.

jsonschema-validation.xml
handrews: Yeah.  And really around annotation collection in general.

handrews:
@Relequestual are you OK with this?  If so I'd like to merge it.  Greg OK'd it since the last update and you are the other person who made significant comments.

jsonschema-validation.xml
philsturgeon: Do we have an issue to track this?

handrews: @philsturgeon no, but feel free to file one 😄

PR #767, opened 2019-07-24
Make the "content*" keywords annotations only

handrews:
Addresses the 'content*' part of #646.

They no longer optionally funciton as validation assertions
in order to ensure consistent and interoperable behavior.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -776,18 +776,19 @@
 
             <section title="Foreword">
                 <t>
-                    Properties defined in this section indicate that an instance contains
+                    Annotations defined in this section indicate that an instance contains
                     non-JSON data encoded in a JSON string.
-                    They describe the type of content and how it is encoded.
                 </t>
                 <t>
                     These properties provide additional information required to interpret JSON data
-                    as rich multimedia documents.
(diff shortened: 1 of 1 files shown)


jsonschema-validation.xml
```diff
@@ -801,16 +802,31 @@
...
                 </t>
                 <t>
-                    Implementations MAY support the "contentMediaType" and "contentEncoding"
-                    keywords as validation assertions.
-                    Should they choose to do so, they SHOULD offer an option to disable validation
-                    for these keywords.
+                    Implementations MAY offer the ability to decode, parse, and/or validate
+                    the string contents automatically.  However, it MUST NOT perform these
+                    operations by default, and MUST provide the validation result of each
+                    string-encoded document separately from the enclosing document.
```
gregsdennis: I see the comment below, but I'm curious what this would look like if someone expects to get multiple schema validation result objects from a single `validate` call.  At least, that's what I would expect if I were consuming an implementation and I read that.

I think a note saying that it would/could require separate validation passes to validate the contents would be nice.

handrews: @gregsdennis OK see if the new wording works for you.

gregsdennis: That looks better.

PR #769, opened 2019-07-26
Clarify annotation collection with applicators

### LINKED ISSUE #768: Clarify: all applicators collect annotations, including "contains"
As noted in #766, we need to emphasize how applicators and annotation collection works, in particular that an applicator like `contains` does collect annotations, which means that when annotation collection is enabled, you cannot short-circuit the evaluation.  I think we note this with `anyOf` but not `contains`.  Those are the only two applicators that combine validation results with OR and could therefore be short-circuited when not collecting annotations.

handrews:
Fixes #768.  No release notes because the part about "contains"
not collecting annotations was never in the release notes, and
it wasn't in the previous draft either.

In particular, "contains" collects annotations.  I'm not sure
where the exception came from, as it was not present in the
previous draft (to the extent that that draft mentioned
annotations).

The CREF actually did talk about annotation collection so
apparently this was a bit of a muddled mess anyway.

This makes "contains" behave like all other applicators,
producing both assertion and annotation results.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -654,7 +654,10 @@
             <t>
                 Evaluation of a parent schema object can complete once all of its
                 subschemas have been evaluated, although in some circumstances evaluation
-                may be short-circuited due to assertion results.
+                may be short-circuited due to assertion results.  When annotations are
+                being collected, some assertion result short-circuiting is not possible
+                due to the need to collect annotations from all subschemas, including
+                those that cannot further change the assertion result.
             </t>
             <section title="Keyword Interactions">
                 <t>
@@ -1981,6 +1984,8 @@
                         <t>
                             An instance validates successfully against this keyword if it validates
                             successfully against at least one schema defined by this keyword's value.
+                            Note that when annotations are being collected, they MUST be collected
+                            from all subschemas if any subschema validates successfully.
                         </t>
                     </section>
 
@@ -2245,16 +2250,10 @@
                         </t>
                         <t>
                             An array instance is valid against "contains" if at least one of
-                            its elements is valid against the given schema.  This keyword
-                            does not produce annotation results.
-                            <cref>
-                                Should it produce a set of the indices for which the
-                                array element is valid against the subschema?  "contains"
-                                does not affect "additionalItems" or any other current
-                                or proposed keyword, but the information could be useful,
-                                and implementation that collect annotations need to
-                                apply "contains" to every element anyway.
-                            </cref>
+                            its elements is valid against the given schema.  Note that when
+                            collecting annotations, the subschema MUST be applied to every
+                            array element even after the first match has been found.  This
+                            is to ensure that all possible annotations are collected.
                         </t>
                     </section>
                 </section>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -656,7 +656,7 @@
                 subschemas have been evaluated, although in some circumstances evaluation
                 may be short-circuited due to assertion results.  When annotations are
                 being collected, some assertion result short-circuiting is not possible
-                due to the need to collect annotations from all subschemas, including
+                due to the need to examine all subschemas for annotation collection, including
                 those that cannot further change the assertion result.
             </t>
             <section title="Keyword Interactions">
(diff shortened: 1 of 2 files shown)


jsonschema-core.xml
```diff
@@ -1981,6 +1984,8 @@
                         <t>
                             An instance validates successfully against this keyword if it validates
                             successfully against at least one schema defined by this keyword's value.
+                            Note that when annotations are being collected, they MUST be collected
+                            from all subschemas if any subschema validates successfully.
```
gregsdennis: But wouldn't annotations from failed subschemas be thrown away?

```suggestion
                            from any subschemas that validate successfully.
```

jsonschema-validation.xml
```diff
@@ -1220,9 +1220,9 @@
                         their results, without asserting any conditions of their own.
```
gregsdennis: It seems to me that this isn't precisely true.  They _do_ assert conditions of their own, but indirectly through the result(s) of the subschemas rather than on the instance.  For example, `anyOf` asserts that at least one of its subschemas pass validation.  It's a fine line.

handrews: Now we're getting into the semantics of what "on their own" means.

Applicators evaluate subschemas and combine / apply logic to their results.  Combining results is not asserting conditions of their own, so I'm not changing this.

jsonschema-core.xml
handrews: @gregsdennis yes?  That doesn't mean they aren't collected.  Yes, once you know a subschema has failed you can stop collecting, but you collect until you figure that out, *THEN* drop the annotations.

I'm really frustrated by my apparent inability to get this across.  It's discussed all over the spec- is there some magic phrase you're looking for?

jsonschema-validation.xml
gregsdennis: ¯\\\_(ツ)_\/¯

jsonschema-core.xml
gregsdennis: If I conform to "from **all** subschemas if any subschema validates successfully," then when any subschema passes validation, then I get annotations from **all** subschemas, whether or not they pass.  My wording requires that annotations only from passing subschemas are kept.

> Yes, once you know a subschema has failed you can stop collecting...

I think you mean "passed" instead of "failed."  This is `anyOf` not `allOf`.  I read what you mean, though.

My focus for this is the result, not the process.  It seems that it should be understood that an implementation has to collect until validation is certain, but I don't think we have to explain that process here.  I think it's sufficient to say that we want the annotations from all passing subschemas.  We definitely *don't* want annotations from failing subschemas, which is what your wording here tells me to do.

jsonschema-validation.xml
gregsdennis: Maybe it's sufficient to simply add "on the instance" at the end of the sentence.

handrews: "on the instance"?  Are you considering `true` and `false` to be non-instance assertions?  Hmm... I tend to think of them as constants, although I can't easily articulate a difference.

Really, what I want to get at is that the validation results of applicators are strictly a logical formula applied to the results of their subschemas.  The applicators themselves have no impact on the subschema's results.  In that sense, they do not "assert" specific things about the instance (or as a constant).

I suppose you could look at it as "asserting" that the result of the formula is `true`, but that's what I'm trying to avoid.  Applicators are more like operators than assertions, and that is the mindset that I want to convey.  Otherwise it's hard to see why they were split out.

jsonschema-core.xml
handrews: @gregsdennis I see your point here.

What I'm trying to get across is that you must _perform the process of collecting annotations_ on all of the subschemas if any of them validate successfully.  Since part of that process is dropping annotations from any subschema that fails to validate, performing the process does not mean that you will get annotations from every subschema.

So we clearly need some better wording around that.  I see the point of your suggestion and it is probably better, but let me see if I can come up with something that is more clear about both aspects.

jsonschema-validation.xml
gregsdennis: Let's just leave it, then, since you're not changing this particular thing, and people generally know what it means.

jsonschema-core.xml
handrews: @gregsdennis OK I pushed a new commit with a change to this line and another one- you should probably look at that commit's diff separately to see where they both are.

I tried to make a distinction between _examining_ a subschema (which may or may not result in actual annotations depending on its assertion outcome) and actually keeping the collected annotations.

handrews:
@gregsdennis I did an update a while back for one of your comments, does it look OK now?  I'd like to merge this if so.

jsonschema-core.xml
gregsdennis: That's a lot better.

PR #772, opened 2019-08-01
Restructure vocabulary-related sections

handrews:
Attempt to make this more understandable by removing
some duplication, putting things in the right order,
and explaning certain things more clearly.  Hopefully.

_It might be easier to just download and build this to review._

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -293,7 +293,7 @@
                 <t>
                     A JSON Schema document, or simply a schema, is a JSON document used to describe
                     an instance.
-                    A schema is itself interpreted as an instance, but SHOULD always be given
+                    A schema can itself be interpreted as an instance, but SHOULD always be given
                     the media type "application/schema+json" rather than
                     "application/schema-instance+json".  The "application/schema+json" media
                     type is defined to offer a superset of the media type parameter and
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -792,47 +798,43 @@
...
-            <t>
-                The role of the meta-schema is to constrain the structure of conforming schemas,
-                as well as simplify the process of composing multiple vocabularies into a usable
-                feature set.  Schema authoring is expected to be a common activity, so
-                schema authors need only understand how to reference a single meta-schema.
-            </t>
-            <t>
-                Meta-schema authoring is an advanced usage of JSON Schema, so the design of
-                meta-schema features emphasizes flexibility over simplicity.
+                how to interpret a schema.  Every schema has a meta-schema, typically identified
```
gregsdennis: > Every schema has a meta-schema.

Later you say "If absent from the root schema, the resulting behavior is implementation-defined."  I realize they don't _technically_ conflict, but it might be nice to have something to the effect of "even if it's not explicitly defined by the `$schema` keyword" rather that the casual "typically defined by..."

jsonschema-core.xml
```diff
@@ -365,6 +365,36 @@
                         and implementations.
                     </t>
                 </section>
+                <section title="Schema Vocabularies">
+                    <t>
+                        A schema vocabulary, or simply a vocabulary, is a set of keywords,
+                        their syntax, and their semantics.  A vocabulary is generally organized
+                        around a particular purpose.  Different uses of JSON Schema will
```
philsturgeon: could we add examples here? Validation, UI generation, etc.

jsonschema-core.xml
```diff
@@ -365,6 +365,36 @@
...
+                        involve different sets of vocabularies.
+                    </t>
+                    <t>
+                        Vocabularies are the primary unit of re-use in JSON Schema, as schema
+                        authors can indicate what vocabularies are required or optional in
+                        order to process the schema.  While keywords can be supported outside
+                        of any vocabulary, there is no analogous mechanism to indicate individual
```
philsturgeon: While arbitrary extra keywords are allowed in JSON Schema, vocabularies are how we can define their meaning in a portable way, which other tooling can then start to support. 

Or something.

handrews:
@philsturgeon @gregsdennis I made some updates based on your feedback.  How is this?

handrews:
Since @philsturgeon said nothing was blocking and I need to post a PR that builds on this I'm going to go ahead and merge with @gregsdennis's approval.  Phil, let me know if I messed anything up and I'll fix it in the next PR.

Issue #779, opened 2019-08-10
Allow "$ref" to take a schema, to make inlining easier?


handrews:
***[EDIT: See better problem statement two comments below]***

Now that `$ref` can have adjacent keywords, dereferencing/inlining is more complex.  (for the purpose of this issue, we're just talking about situations where inlining is possible, e.g. no cyclic references- inlining such use cases is common, with numerous libraries dedicated to this operation).

In informal discussions, we've recommended replacing `$ref` with an `allOf` containing just the referenced schema, *OR* if there is already an adjacent `allOf`, appending the referenced schema to that `allOf`.  This is rather cumbersome.

At the same time, we use runtime JSON Pointer-ish constructs that look like `/properties/foo/$ref/properties/bar/$ref`, etc., to record the runtime path as we traverse references.

What if we allowed replacing the `$ref` URI with the target schema?  e.g. if `{"$ref": "A"}` points to `{A}`, then it can be replaced with `{"$ref": {A}}`

`$ref` here is effectively a no-op, it just allows inlining the target without having to re-arrange the context.

Pros:
* It's much easier to explain
* It matches how we report runtime paths even when not dereferenced
* We're already changing `$ref` inlining, so now is the time to sort this out

Cons:
* It's a change
* Strongly typed languages may be unhappy about the string-or-object behavior
* ???

gregsdennis:
To be clear, are you suggesting that we support schema objects as a value for `$ref`?  If so, that feels weird and counter to the reason for having references.

Or are you just talking about how references are explained without any functional changes? If so, this is probably a good explanation.

handrews:
@gregsdennis Here's a better problem statement:

* "Dereferencing" JSON Schemas is a common operation, for example JavaScript's [json-schema-ref-parser](https://www.npmjs.com/package/json-schema-ref-parser) gets over 500K downloads a week.
* By allowing other keywords adjacent to `$ref`, we broke key assumptions those libraries rely on
    * Although note that schemas that still avoid adjacent keywords will still work just fine
* We need to provide guidance for this use case, because it is very substantial
* Having a process based on a core keyword (`$ref`) depend on a non-core keyword (`allOf`) is a pretty bad design smell
    * I should be able to work with core without needing any other vocabulary
    * By definition, it can't have dependencies

A related thing that may or may not be considered a problem:

* Conceptually, dereferencing does not change behavior, but in practice it changes the output
    * This was not previously an issue due to not having an output spec in the first place
[...]

PR #780, opened 2019-08-16
Split $anchor out of $id, define canonical vs shadowed URI behavior

### LINKED ISSUE #726: "$id": Eliminate base URI shadowing
"URI shadowing" (h/t to @johandorland for the term) refers to the scenario when there are multiple `$id`s in a schema, with at least one `$id` in a subschema of another object containing `$id` (such a the root schema object).

## When did this become a thing?

Here is the schema and various resolved URI examples, copied and pasted directly from [the current spec](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.2.4):

```
{
       "$id": "http://example.com/root.json",
       "definitions": {
           "A": { "$id": "#foo" },
           "B": {
               "$id": "other.json",
               "definitions": {
                   "X": { "$id": "#bar" },
                   "Y": { "$id": "t/inner.json" }
               }
           },
           "C": {
               "$id": "urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f"
           }
       }
   }

   The schemas at the following URI-encoded JSON Pointers [RFC6901]
   (relative to the root schema) have the following base URIs, and are
   identifiable by any listed URI in accordance with Section 5 above:

   # (document root)

         http://example.com/root.json

         http://example.com/root.json#

   #/definitions/A

         http://example.com/root.json#foo

         http://example.com/root.json#/definitions/A

   #/definitions/B

         http://example.com/other.json

         http://example.com/other.json#

         http://example.com/root.json#/definitions/B

   #/definitions/B/definitions/X

         http://example.com/other.json#bar

         http://example.com/other.json#/definitions/X

         http://example.com/root.json#/definitions/B/definitions/X

   #/definitions/B/definitions/Y

         http://example.com/t/inner.json

         http://example.com/t/inner.json#

         http://example.com/other.json#/definitions/Y

         http://example.com/root.json#/definitions/B/definitions/Y

   #/definitions/C

         urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f

         urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f#

         http://example.com/root.json#/definitions/C
```

Notice that all of the locations (the list section headers) are done as JSON Pointer fragments, relative to the root of the entire example schema.

Prior to draft-handrews-json-schema-01, the analogous example section did not show _any_ examples with JSON Pointer fragments.  Using the exact same schema, it showed these resolved URIs only:

```
   # (document root)  http://example.com/root.json#

   #/definitions/A  http://example.com/root.json#foo

   #/definitions/B  http://example.com/other.json

   #/definitions/B/definitions/X  http://example.com/other.json#bar

   #/definitions/B/definitions/Y  http://example.com/t/inner.json

   #/definitions/C  urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f
```

But you can see that the locations were done the same way- as a relative JSON Pointer fragment from the overall document root.

The key point here is that we added examples of URI shadowing in handrews-*-01, which was a clarification of draft-07 (draft-07 was originally published as handrews-*00).  So we didn't even introduce these with draft-07.  There was some confusion over it (and other aspects of `$id` and `$ref`) and we added the examples when we clarified the text.

## Why did we do this?

Good question.  The change was made in PR #550, which was primarily about removing the terminology around "internal references" vs "external references".  Which had itself been a great improvement by @awwright over the draft-04 language around "inline resolution" and "canonical resolution".  None of these prior approaches directly explained how JSON Pointer fragments work across the presence of `$id`.

The issue mentioned by that PR only mentions JSON Pointer fragments once, at the end, after the PR was submitted: https://github.com/json-schema-org/json-schema-spec/issues/545#issuecomment-369551287

I know I had questions over how that should be handled when I tried implementing what were then the new draft-06 proposals.  And I know that I ended up implementing URI shadowing.  I don't remember why.

Quite a few people weighed in on and reviewed #545 and #550, so this isn't something that slipped in by accident.

## Did anyone rely on this before handrews-*-01?

No clue.

## Did anyone actually implement what's in handrews-*-01?

I think @johandorland did?  He commented on the issue/pr.  Not sure if anyone else did.  @Julian?

## Does the test suite test this?

No.

## So, really, why did we do this?

I think the rationale was that since the locations in the existing example were given as JSON Pointer fragments, that must mean that those fragments were valid, which would definitely imply that every parent `$id` introduces a base URI that must be tracked throughout all subschemas, even if another `$id` appears.

## Can we just not?

I think so.  [JSON Pointer fragment evaluation](https://tools.ietf.org/html/rfc6901#section-6) is just defined in terms of JSON document structure, without any thought given to changing base URIs or embedding one document in another.  But media types get to specify fragment syntax and semantics, so I think we can reasonably say that it stops when the base URI is reset, and from that point on, the prior base URI no longer applies at all.

That is actually how we interpret plain name fragments.  Note in this part of the example:

```
   #/definitions/B/definitions/X

         http://example.com/other.json#bar

         http://example.com/other.json#/definitions/X

         http://example.com/root.json#/definitions/B/definitions/X
```

the `#bar` plain name fragment can be used with the innermost base `http://example.com/other.json`, but ***not*** with the outer base `http://example.com/root.json`.

So restricting the use of JSON Pointer fragments to not cross `$id` boundaries would make their behavior more consistent with how we handle plain name fragments.

Similar to #724, this considers `$id` to establish a document boundary, with the new base URI it creates applying to the document within that boundary.  Unlike #724, what is proposed here does not make any other changes to `$id`'s behavior.  In particular, the effect of an `$id` with a JSON Pointer fragment remains undefined, although we can address that separately later if we want to.

## So what would this look like?

There are two options ***WITHIN THE SCOPE OF THIS ISSUE***.  If you want to talk about other options, file them yourself :-)  Comments that go off-topic here will be deleted.

Here's the example schema again so you don't have to scroll so much:

```
{
       "$id": "http://example.com/root.json",
       "definitions": {
           "A": { "$id": "#foo" },
           "B": {
               "$id": "other.json",
               "definitions": {
                   "X": { "$id": "#bar" },
                   "Y": { "$id": "t/inner.json" }
               }
           },
           "C": {
               "$id": "urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f"
           }
       }
   }
}
```

### Option 1: Exactly one valid URI involving a JSON Pointer fragment for each location:

```
   # (document root)

         http://example.com/root.json

         http://example.com/root.json#

   #/definitions/A

         http://example.com/root.json#foo

         http://example.com/root.json#/definitions/A

   #/definitions/B

         http://example.com/other.json

         http://example.com/other.json#

   #/definitions/B/definitions/X

         http://example.com/other.json#bar

         http://example.com/other.json#/definitions/X

   #/definitions/B/definitions/Y

         http://example.com/t/inner.json

         http://example.com/t/inner.json#

   #/definitions/C

         urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f

         urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f#
```

This is a lot more straightforward, and all of these addresses work whether the schemas were loaded as one document (as in this example), or if each of `root.json`, `other.json`, and `inner.json` had been loaded separately, and `$ref`-ed each other.

### Option 2: But what about the URI used to fetch the document?

If it's possible to retrieve this example from `http://example.com/alias.json` (as opposed to or in addition to its declared `$id` of `http://example.com/root.json`), then should fragments also work from that base?

I can see it either way.

One option is to consider fetching this from `alias.json` and discovering that its declared `$id` is `root.json` is to conceptualize this as a redirect, in which case the fragment is applied to the redirect target (at least in HTTP).  h/t to @jdesrosiers for this mental model.  In this sense, anywhere we have `root.json` we would also require supporting `alias.json`

On the other hand, JSON Schema repeatedly talks about URIs as identifiers rather than locators, so I think we could rationally say that the `$id` is how the schema needs to be referenced.  It's fine if it's fetched from elsewhere outside of the reference process (like being pre-loaded from a local filesystem) but does that mean that we should support those loading URIs when an `$id` is provided?  (obviously if no `$id`, then the loading URI is the only available base).

Supporting the loading URI as a shadowed base, but nothing else, might be a bit odd, or it might be more intuitive.  I really have no idea at this point 😆

### LINKED ISSUE #729: Make `$id` conform to RFC 3986 suggestion for base URI elements
Per [RFC 3986 section 4.3](https://tools.ietf.org/html/rfc3986#section-4.3):

> Some protocol elements allow only the absolute form of a URI without
   a fragment identifier.  For example, defining a base URI for later
   use by relative references calls for an absolute-URI syntax rule that
   does not allow a fragment.

Let's _please_ follow this advice and require `$id` to resolve to an absolute URI, without a fragment.  By this I mean that it _cannot contain a fragment_, but MAY be a relative URI reference (resolved in the same way that it is currently).  If it makes things easier, rather than "MUST NOT contain a fragment", we can say that "any fragment MUST be ignored."  Either option makes it absolutely clear that fragments in `$id` have no useful effect.

Currently, there are three behavioral cases of fragments in `$id`:

* `"$id": "#foo"`: Plain name fragment definitions.  We can replace with `"$anchor": "foo"`, analogous to fragment definition keywords in several other media types.  Very easy to explain, clearly separate from any concerns over base URIs or embedded documents.
* `"$id": "https://example.com/root#"`: Empty fragment in root schema object `$id`s.  This has absolutely no effect compared to the same thing without a fragment at all.  It primarily exists as an explicitly supported case because older meta-schema `id`s were written this way, and we've never bothered to change that.
* _**ALL** other cases have undefined behavior!_

Per [section 8.2.3 of the current draft](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.2.3)

>  The effect of using a fragment in "$id" that isn't blank or doesn't
   follow the plain name syntax is undefined.  [[CREF3: How should an
   "$id" URI reference containing a fragment with other components be
   interpreted?  There are two cases: when the other components match
   the current base URI and when they change the base URI.  ]]

This note explicitly makes the behavior of JSON Pointer-fragment-only `$id`s undefined, and notes that _we have no idea_ how fragment-plus-other-component `$id`s behave, meaning that their behavior is also undefined in the current spec.

---------

So there are several problems with the current behavior which this addresses:

1.  We are currently going against the best practice for base URI elements, which complicates explaining how base URIs work to those that don't already understand them.
1.  By having one form of fragment-only URI references perform a useful function, we give the impression that fragments in `$id` as a whole should work.  However, this is incorrect- _**every** other non-trivial usage_ has undefined behavior.  Although some of it looks like it should work, at least to some people.
1.  Because the two well-defined behaviors (setting the self identifier which is also the base URI vs defining a plane name fragment) are, due to the rules around base URIs and fragments, apparently completely separate from each other, schema authors and implementors find `$id` confusing.  It requires a fairly deep understanding of RFC 3986 to understand why these behaviors can possibly work in the same keyword at all.

Removing fragments from `$id` (and using `$anchor` for plain name fragment definition instead) solves these problems.

* We would be in line with the informal RFC 3986 suggestion regarding base URI elements
* Saying that fragments in `$id` MUST be ignored matches how base URIs are computed in RFC 3986
* We would (not coincidentally) now have a self/base-setting element that behaves analogous to such elements in other media types, where fragments are either forbidden or clearly have no purpose
* To the relative layperson, there would no longer appear to be two disjoint use cases for `$id`

---------

Some of the fragment-in-`$id` cases that technically have undefined behavior happen to be effective no-ops, as with the the empty fragment case.  (Recall that the empty fragment is technically a [JSON Pointer for the entire document](https://tools.ietf.org/html/rfc6901#section-6)).

So sometimes we see schemas like this generated by tools:

```JSON
{
    "definitions": {
        "foo": {
            "$id": "#/definitions/foo"
        }
    }
}
```
which is basically a no-op (note that the `$id` fragment matches the actual position).  So if we say that fragments in `$id` MUST be ignored, its's still a no-op and that's fine.  I would prefer an implementation to raise an error in this case going forward, but I don't feel strongly about that.

My point here is that the only other fragment usage I've seen in the wild doesn't actually do anything.

--------

@awwright if I'm reading https://github.com/json-schema-org/json-schema-spec/issues/719#issuecomment-468579737 you are more or less OK with this idea?  I agree that the names could be better, but keeping `$id` for the self+base form is less disruptive.

@jdesrosiers you are encouraged to comment on the merits of this if you'd like, but please do *not* discuss your approach from #724 here, as we have already determined (in #179) that mixing the two discussions confuses everyone.

--------

Everyone: fair warning, I feel very strongly about this.  Unlike #726 (Eliminate base URI shadowing), which I think we should do but if we don't that's fine, I'm incredibly fed up with the problems with `$id` and feel like many of them come from having such a wide rage of syntactically valid values that have undefined semantics.

If you want to keep the current situation of exactly one use of fragments in `$id` having defined semantics, and _all other possible uses having undefined semantics_, please make an active case for why such a thing is desirable.

Yes, changing `"$id"` to `"$anchor"` is a breaking change.  That is unfortunate.  But I feel like `$id` as it stands is so problematic that this is worthwhile.  And importantly, this just forbids one part of `$id`'s current syntax, and leaves the rest of `$id`'s well-defined behavior alone.

Although perhaps @Julian can speak to how difficult this would be in his implementation, as I think he found the prior change of `id` to `$id` particularly disruptive.

handrews:
***[EDIT:** I pushed a commit Saturday just before 1PM california time that I think simplifies things considerably- see the commit log message for details.]*

*THERE ARE THREE COMMITS (originally, and then more added)*
While this is all one coherent change, and the commits don't fully stand alone, I have organized them to make reviewing a bit easier.

* First I define a schema resource
* Then I deal with URI shadowing (#726) _somewhat obsoleted by the 4th commit_
* Then I deal with `$anchor` (#729) _and_ update the examples
* Feedback on URI shadowing and fragments in `$id`
* Update meta-schemas
* Fix more examples
* Add changelog

Fixes #729.  Fixes #726. 

This implements the recent decision to strongly discourage using fragments that cross a base URI change (e.g. an `"$id"` appearance).

This is done by describing schemas identified by absolute URIs as resources (per the literal definition of Univeral Resource Identifier), and declaring the URI resulting from the `"$id"` to be the resource's canonical URI.

While URIs for subschemas with `"$id"` and their children can be constructed using the base URI from a parent schema, these URIs are non-canonical, and their behavior is undefined.  This is on the grounds that switching between embedding and referencing schema resources should behave essentially identically.

A difficulty with how annotation collection works in the event of such as switch is noted in a CREF (related to #779)

This also adds the $anchor keyword in place of the fragment-only form of `$id`, and updates the schema identification examples for `$anchor` and canonical vs non-canonical URIs.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1654,7 +1654,7 @@
                             An implementation MAY choose not to support addressing schemas
                             by non-canonical URIs. As such, it is RECOMENDED that schema authors only 
                             use canonical URIs, as using non-canonical URIs may reduce the 
-                            transportability of a schema.
+                            schema interoperability.
                             <cref>
                                 This is to avoid requiring implementations to keep track of a whole
                                 stack of possible base URIs and JSON Pointer fragments for each,
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2,6 +2,7 @@
 <!DOCTYPE rfc SYSTEM "rfc2629.dtd" [
 <!ENTITY RFC2119 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.2119.xml">
 <!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
+<!ENTITY RFC6596 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6596.xml">
 <!ENTITY RFC6839 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6839.xml">
 <!ENTITY RFC6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
 <!ENTITY RFC7049 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.7049.xml">
@@ -402,10 +403,22 @@
                         additional keywords that are not part of a formal vocabulary.
                     </t>
                 </section>
-                <section title="Root Schema and Subschemas" anchor="root">
+                <section title="Root Schema and Subschemas and Resources" anchor="root">
+                    <t>
+                        A JSON Schema resource is a schema which is
+                        <xref target="RFC6596">canonically</xref> identified by an
+                        <xref target="RFC3986">absolute URI</xref>.
+                    </t>
+                    <t>
+                        As discussed in section
+                        <xref target="id-keyword" format="counter"></xref>, a JSON Schema document
+                        can contain multiple JSON Schema resources.
+                    </t>
                     <t>
                         The root schema is the schema that comprises the entire JSON document
-                        in question.
+                        in question.  The root schema is always a schema resource, where the
+                        URI is determined as described in section
+                        <xref target="initial-base" format="counter"></xref>.
                     </t>
                     <t>
                         Some keywords take schemas themselves, allowing JSON Schemas to be nested:
@@ -1455,7 +1468,7 @@
                     to other schemas by specifying their URI.
                 </t>
 
-                <section title="Initial Base URI">
+                <section title="Initial Base URI" anchor="initial-base">
                     <t>
                         <xref target="RFC3986">RFC3986 Section 5.1</xref> defines how to determine the
                         default base URI of a document.
@@ -1466,8 +1479,8 @@
                         situation identifiable by a URI of any known scheme.
                     </t>
                     <t>
-                        If a schema document defines no explicit base URI with "$id" (embedded in content),
-                        the base URI is that determined per
+                        If a schema document defines no explicit base URI with "$id"
+                        (embedded in content), the base URI is that determined per
                         <xref target="RFC3986">RFC 3986 section 5</xref>.
                     </t>
                     <t>
@@ -1475,6 +1488,11 @@
                         implementation-specific default URI MAY be used as described in
                         <xref target="RFC3986"> RFC 3986 Section 5.1.4</xref>.  It is RECOMMENDED
                         that implementations document any default base URI that they assume.
+                    </t>
+                    <t>
+                        Unless the "$id" keyword described in the next section is present in the
+                        root schema, this base URI SHOULD be considered the canonical URI of the
+                        schema document's root schema resource.
                     </t>
                 </section>
 
@@ -3349,6 +3367,7 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
         </references>
 
         <references title="Informative References">
+            &RFC6596;
             &RFC7049;
             &RFC7231;
             &RFC8288;
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1652,7 +1652,9 @@
                         </t>
                         <t>
                             An implementation MAY choose not to support addressing schemas
-                            by non-canonical URIs.
+                            by non-canonical URIs. As such, it is RECOMENDED that schema authors only 
+                            use canonical URIs, as using non-canonical URIs may reduce the 
+                            transportability of a schema.
                             <cref>
                                 This is to avoid requiring implementations to keep track of a whole
                                 stack of possible base URIs and JSON Pointer fragments for each,
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1653,7 +1653,7 @@
                         <t>
                             An implementation MAY choose not to support addressing schemas
                             by non-canonical URIs. As such, it is RECOMENDED that schema authors only 
-                            use canonical URIs, as using non-canonical URIs may reduce the 
+                            use canonical URIs, as using non-canonical URIs may reduce
                             schema interoperability.
                             <cref>
                                 This is to avoid requiring implementations to keep track of a whole
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -479,7 +479,7 @@
             <t>
                 Defining and referencing a plain name fragment identifier within an
                 "application/schema+json" document are specified
-                in the <xref target="id-keyword">"$id" keyword</xref> section.
+                in the <xref target="anchor">"$anchor" keyword</xref> section.
             </t>
             <t>
             </t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3579,6 +3579,9 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
                             <t>Additional guidance on initial base URIs beyond network retrieval</t>
                             <t>Allow "schema" media type parameter for "application/schema+json"</t>
                             <t>Better explanation of media type parameters and the HTTP Accept header</t>
+                            <t>Use "$id" to establish canonical and base absolute-URIs only, no fragments</t>
+                            <t>Replace plain-name-fragment-only form of $id with $anchor</t>
+                            <t>Clarified that the behavior of JSON Pointers across $id boundary is unreliable</t>
                         </list>
                     </t>
                     <t hangText="draft-handrews-json-schema-01">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1467,6 +1467,11 @@
                     identified by <xref target="RFC3986">URI</xref>, and can embed references
                     to other schemas by specifying their URI.
                 </t>
+                <t>
+                    Several keywords can accept a relative <xref target="RFC3986">URI-reference</xref>,
+                    or a value used to construct a relative URI-reference.  For these keywords,
+                    it is necessary to establish a base URI in order to resolve the reference.
+                </t>
 
                 <section title="Initial Base URI" anchor="initial-base">
                     <t>
@@ -1498,63 +1503,87 @@
 
                 <section title='The "$id" Keyword' anchor="id-keyword">
                     <t>
-                        The "$id" keyword defines a URI for the schema, and the base URI that
-                        other URI references within the schema are resolved against.
-                        A subschema's "$id" is resolved against the base URI of its parent schema.
-                        If no parent schema defines an explicit base URI with "$id", the base URI
-                        is that of the entire document, as determined per
-                        <xref target="RFC3986">RFC 3986 section 5</xref>.
+                        The "$id" keyword identifies a schema resource with its
+                        <xref target="RFC6596">canonical</xref> URI.
+                    </t>
+                    <t>
+                        Note that this URI is an identifier and not necessarily a network locator.
+                        In the case of a network-addressable URL, a schema need not be downloadable
+                        from its canonical URI.
                     </t>
                     <t>
                         If present, the value for this keyword MUST be a string, and MUST represent a
-                        valid <xref target="RFC3986">URI-reference</xref>.
-                        This value SHOULD be normalized, and SHOULD NOT be an empty fragment &lt;#&gt;
-                        or an empty string &lt;&gt;.
+                        valid <xref target="RFC3986">URI-reference</xref>.  This URI-reference
+                        SHOULD be normalized, and MUST resolve to an
+                        <xref target="RFC3986">absolute-URI</xref> (without a fragment).
+                    </t>
+                    <t>
+                        This URI also serves as the base URI for relative URI-references in keywords
+                        within the schema resource, in accordance with
+                        <xref target="RFC3986">RFC 3986 section 5.1.1</xref> regarding base URIs
+                        embedded in content.
+                    </t>
+                    <t>
+                        The presence of "$id" in a subschema indicates that the subschema constitutes
+                        a distinct schema resource within a single schema document.  Furthermore,
+                        in accordance with <xref target="RFC3986">RFC 3986 section 5.1.2</xref>
+                        regarding encapsulating entities, if an "$id" in a subschema is a relative
+                        URI-reference, the base URI for resolving that reference is the URI of
+                        the parent schema resource.
+                    </t>
+                    <t>
+                        If no parent schema object explicitly identifies itself as a resource
+                        with "$id", the base URI is that of the entire document, as established
+                        by the steps given in the <xref target="initial-base">previous section.</xref>
                     </t>
                     <section title="Identifying the root schema">
                         <t>
-                            The root schema of a JSON Schema document SHOULD contain an "$id" keyword with
-                            an <xref target="RFC3986">absolute-URI</xref> (containing a scheme, but no fragment),
-                            or this absolute URI but with an empty fragment.
-                            <!-- All of the standard meta-schemas use an empty fragment in their id/$id values. -->
+                            The root schema of a JSON Schema document SHOULD contain an "$id" keyword
+                            with an <xref target="RFC3986">absolute-URI</xref> (containing a scheme,
+                            but no fragment).
                         </t>
                     </section>
-                    <section title="Changing the base URI within a schema file">
+                    <section title="JSON Pointer fragments and embedded schema resources">
                         <t>
-                            When an "$id" sets the base URI, the object containing that "$id" and all of
-                            its subschemas can be identified by using a JSON Pointer fragment starting
-                            from that location.  This is true even of subschemas that further change the
-                            base URI.  Therefore, a single subschema may be accessible by multiple URIs,
-                            each consisting of base URI declared in the subschema or a parent, along with
-                            a JSON Pointer fragment identifying the path from the schema object that
-                            declares the base to the subschema being identified.  Examples of this are
-                            shown in section <xref target="idExamples" format="counter"></xref>.
-                        </t>
-                    </section>
-                    <section title="Location-independent identifiers">
-                        <t>
-                            Using JSON Pointer fragments requires knowledge of the structure of the schema.
-                            When writing schema documents with the intention to provide re-usable
-                            schemas, it may be preferable to use a plain name fragment that is not tied to
-                            any particular structural location.  This allows a subschema to be relocated
-                            without requiring JSON Pointer references to be updated.
+                            Since JSON Pointer URI fragments are constructed based on the structure
+                            of the schema document, an embedded schema resource and its subschemas
+                            can be identified by JSON Pointer fragments relative to either its own
+                            canonical URI, or relative to the containing resource's URI.
                         </t>
                         <t>
-                            To specify such a subschema identifier,
-                            the "$id" keyword is set to a URI reference with a plain name fragment (not a JSON Pointer fragment).
-                            This value MUST begin with the number sign that specifies a fragment ("#"),
-                            then a letter ([A-Za-z]),
-                            followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"),
-                            colons (":"), or periods (".").
-                        </t>
-                        <t>
-                            The effect of using a fragment in "$id" that isn't blank or doesn't follow the
-                            plain name syntax is undefined.
+                            Conceptually, a set of linked schemas should behave identically whether
+                            each schema is a separate document connected with
+                            <xref target="references">schema references</xref>, or is structured as
+                            a single document with one or more schema resources embedded as
+                            subschemas.
                             <cref>
-                                How should an "$id" URI reference containing a fragment with other components
-                                be interpreted?  There are two cases:  when the other components match
-                                the current base URI and when they change the base URI.
+                                Note that when using schema references, the reference keyword
+                                appears in the runtime path in the standard output format for
+                                errors and annotations.  This means that while the validation
+                                outcome is unchanged when switching between an embedded schema
+                                resource and a referenced one, the runtime paths of annotations
+                                do change.  A future draft may allow directly replacing the value
+                                of the reference keyword with its target while leaving the keyword
+                                itself in place in order to make embedding vs referencing transparent
+                                to annotation collection.  This would allow replacing
+                                {"$ref": "/foo"} with {"$ref": {"type": "string"}}, assuming
+                                the schema at "verb">"/foo" consists of just a "type" keyword with
+                                value "string".  Feedback on this topic is highly encouraged.
                             </cref>
+                        </t>
+                        <t>
+                            Since URIs involving JSON Pointer fragments relative to the parent
+                            schema resource's URI cease to be valid when the embedded schema
+                            is moved to a separate document and referenced, applications and schemas
+                            SHOULD NOT use such URIs to identify embedded schema resources or
+                            locations within them.  The effect of using such URIs is undefined.
+                            Implementations MAY produce an error requiring that the canonical
+                            URI for the embedded resource be used.
+                        </t>
+                        <t>
+                            Examples of such URIs with undefined behavior, as well as the appropriate
+                            canonical URIs to use instead, are provided in section
+                            <xref target="idExamples" format="counter"></xref>.
                         </t>
                     </section>
                     <section title="Schema identification examples" anchor="idExamples">
@@ -1639,7 +1668,7 @@
                     </section>
                 </section>
 
-                <section title="Schema References">
+                <section title="Schema References" anchor="references">
                     <t>
                         Several keywords can be used to reference a schema which is to be applied to the
                         current instance location. "$ref" and "$recursiveRef" are applicator
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -410,11 +410,6 @@
                         <xref target="RFC3986">absolute URI</xref>.
                     </t>
                     <t>
-                        As discussed in section
-                        <xref target="id-keyword" format="counter"></xref>, a JSON Schema document
-                        can contain multiple JSON Schema resources.
-                    </t>
-                    <t>
                         The root schema is the schema that comprises the entire JSON document
                         in question.  The root schema is always a schema resource, where the
                         URI is determined as described in section
@@ -441,6 +436,15 @@
                     </t>
                     <t>
                         As with the root schema, a subschema is either an object or a boolean.
+                    </t>
+                    <t>
+                        As discussed in section
+                        <xref target="id-keyword" format="counter"></xref>, a JSON Schema document
+                        can contain multiple JSON Schema resources.  When used without qualification,
+                        the term "root schema" refers to the document's root schema.  In some
+                        cases, resource root schemas are discussed.  A resource's root schema
+                        is its top-level schema object, which would also be a document root schema
+                        if the resource were to be extracted to a standalone JSON Schema document.
                     </t>
                 </section>
             </section>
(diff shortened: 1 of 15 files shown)


jsonschema-core.xml
```diff
@@ -1483,63 +1566,97 @@
...
+                                    </t>
+                                </list>
+                            </t>
+                            <t hangText="#/$defs/B">
+                                <list style="hanging">
+                                    <t hangText="base URI">https://example.com/other.json</t>
+                                    <t hangText="canonical URI with pointer fragment">
+                                        https://example.com/other.json#
+                                    </t>
+                                    <t hangText="Fragment relative to root.json (undefined behavior)">
```
jdesrosiers: This one's not undefined

jsonschema-core.xml
```diff
@@ -1483,63 +1566,97 @@
...
+                            <t hangText="#/$defs/B/$defs/Y">
+                                <list style="hanging">
+                                    <t hangText="base URI">https://example.com/t/inner.json</t>
+                                    <t hangText="canonical URI with plain fragment">
+                                        https://example.com/t/inner.json#bar
+                                    </t>
+                                    <t hangText="canonical URI with pointer fragment">
+                                        https://example.com/t/inner.json#
+                                    </t>
+                                    <t hangText="Fragment relative to other.json (undefined behavior)">
```
jdesrosiers: This one's not undefined

jsonschema-core.xml
```diff
@@ -1483,63 +1566,97 @@
...
+                            </t>
+                            <t hangText="#/$defs/C">
+                                <list style="hanging">
+                                    <t hangText="base URI">
+                                        urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f
+                                    </t>
+                                    <t hangText="canonical URI with pointer fragment">
+                                        urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f#
+                                    </t>
+                                    <t hangText="Fragment relative to root.json (undefined behavior)">
```
jdesrosiers: This one's not undefined

gregsdennis:
Does the `$anchor` functionality differ from that of `$id` when `$id`'s value is "anchor-like"?  (i.e Is this existing functionality just split it to another keyword?)  If so, does `$id` no longer support what `$anchor` now does?

jsonschema-core.xml
handrews: Did you mean to comment on this one?  It says "canonical URI with pointer fragment", as it has the `#` tacked on to the `urn:...`.  The next one relative to `root.json` is undefined.

handrews: @jdesrosiers oh whoops, the sidescroll bar obscure the line for me :-/

`https://example.com/root.json#/$defs/C` is indeed undefined.  The `urn:...` is an absolute URI and the canonical URI of that schema.  URIs with fragments are only canonical if they use the canonical base.

`https://example.com/root.json#/$defs/C` is effectively the _retrieval URI_ of that subschema.  It's how you get there, but if you are writing an application or another schema that `$ref`s this, you SHOULD use `urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f` or `urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f#` (which are equivalent due to JSON Schema's fragment rules).

handrews: The more practical reason that they are undefined is that if you changed the example so it looks like this:

```JSON
{
    "$id": "https://example.com/root.json",
    "$defs": {
        "A": { "$anchor": "foo" },
        "B": {
            "$ref": "other.json",
        },
        "C": {
            "$ref": "urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f"
        }
    }
}
```

Then `https://example.com/root.json#/$defs/C` no longer points to the same thing as `urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f`.  It points the `$ref` object.

So if you switch from embedded to reference, that `#/$defs/C` pointer does ***not*** become _invalid_, but it does point to a different thing.  That ultimately behaves the same, but is still not identical.

So I decided that was the criteria for "undefined behavior".  You can evaluate the pointer and it's still valid, but you might be surprised at what you get.

I am open to suggestion on how to present that better, I agree it is puzzling because "undefined" kind of implies that evaluating the JSON Pointer might not be defined.  That's not really what I meant, I mean that it might not always give you the same thing for two schema arrangements that are logically the same.

handrews:
@gregsdennis `"$anchor": "foo"` has the same effect that `"$id": "#foo"` used to have.

You can no longer write `"$id": "#foo"`, as the spec says that the value of `$id` _MUST resolve to an absolute-URI (without a fragment)._

But I agree that that's not entirely obvious from the current PR.  I will add something to clarify the change.  It might be a CREF rather than main text, and it will definitely be in the changelog... which I forgot to write.  Lemme add that...

jsonschema-core.xml
handrews: @jdesrosiers OK I ended up reworking the wording a lot in one of the commits I just added.  See what you think of that.  I'm not thrilled with it, but I think it's at least in the right direction?  Further suggestions welcome.

jdesrosiers: > Then `https://example.com/root.json#/$defs/C` no longer points to the same thing as `urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f`. It points the `$ref` object.

This is contradictory to the way I understood it. To me, the following two are equivalent.

`{ "$ref": "urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f" }`
`{ "$id": "urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f", "type": "string" }`

Pointing to the first results in the second. I thought that's what was meant by an `$id` being equivalent to an embedded `$ref`. At least that's what I meant by it when I introduced the concept.

But, I realize that with the changes to `$ref`, that's not even possible anymore. A reference is no longer `{ "$ref": "..." }`, it's the value of the value of a keyword called `$ref`. That changes the dynamics significantly. Now I understand why you proposed that the following should be equivalent.

`{ "$ref": "urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f" }`
`{ "$ref": { "$id": "urn:uuid:ee564b8a-7a87-4125-8c96-e9f123d6766f", "type": "string" } }`

Given the new definition of `$ref`, I think this is the only way the embedding concept makes sense.

handrews:
@jdesrosiers thanks- you spotted a legitimately confusing case, and that is very helpful!  I'm honestly not sure what the right thing to do is with the fragment URI that expresses the location of the embedded schema resource in the parent schema resource.  It does not seem right to say that it is undefined, so let's see how this wording holds up and whether I need to tweak it further.

jsonschema-core.xml
handrews: @jdesrosiers yeah, allowing adjacent keywords changes things.

It also opens up a substantial use case of annotations that can layer semantics (such as OO inheritance or composition) on top of references, which couldn't be done in the old system, and also couldn't be done if we eliminated `$ref` and just used URI strings in place of schema objects.  These use cases specifically rely on being adjacent to something that is recognizably a reference.

jsonschema-core.xml
```diff
@@ -1394,87 +1412,179 @@
...
-                            an <xref target="RFC3986">absolute-URI</xref> (containing a scheme, but no fragment),
-                            or this absolute URI but with an empty fragment.
-                            <!-- All of the standard meta-schemas use an empty fragment in their id/$id values. -->
+                            The root schema of a JSON Schema document SHOULD contain an "$id" keyword
+                            with an <xref target="RFC3986">absolute-URI</xref> (containing a scheme,
+                            but no fragment).
                         </t>
                     </section>
-                    <section title="Changing the base URI within a schema file">
+                    <section title="JSON Pointer fragments and embedded schema resources"
```
Relequestual: Reviewed this far...
Excited by this change so far for sure =D
Death to URI shadowing!

jsonschema-core.xml
```diff
@@ -1394,87 +1412,179 @@
...
-                            Using JSON Pointer fragments requires knowledge of the structure of the schema.
-                            When writing schema documents with the intention to provide re-usable
-                            schemas, it may be preferable to use a plain name fragment that is not tied to
-                            any particular structural location.  This allows a subschema to be relocated
-                            without requiring JSON Pointer references to be updated.
+                            Conceptually, a set of linked schemas should behave identically whether
+                            each schema is a separate document connected with
+                            <xref target="references">schema references</xref>, or is structured as
+                            a single document with one or more schema resources embedded as
+                            subschemas.
```
Relequestual: Is it worth mentioning "as per section (https://github.com/json-schema-org/json-schema-spec/blob/9ac32f405ccadc48fa0ed732839f29d49832aec1/jsonschema-core.xml#L1458-L1465)" (whatever section that is)?

I say this, because when reading this section, I'm left wondering when I (as a schema author) should use `$id` vs `$anchor`

jsonschema-core.xml
```diff
@@ -1394,87 +1412,179 @@
...
+                        When writing schema documents with the intention to provide re-usable
+                        schemas, it may be preferable to use a plain name fragment that is not tied to
+                        any particular structural location.  This allows a subschema to be relocated
+                        without requiring JSON Pointer references to be updated.
+                    </t>
+                    <t>
+                        The "$anchor" keyword is used to specify such a fragment.
+                    </t>
+                    <t>
+                        If present, the value of this keyword MUST be a string, which MUST start with
```
Relequestual: /this keyword/$anchor/
I think this is more in keeping with most of the document.

(My preference would be to change all instances of "this keyword" to "[keyword]", making the document more easy to scan and search for specific keyword requirements as a reference document.

jsonschema-core.xml
handrews: @Relequestual I don't understand what this paragraph has to do with `$anchor` at all?

Keep in mind that when it comes to sorting out people's confusion migrating from the old `$id`, that goes on the web site.  The spec is primarily to be read on its own.  I can't see any reason why someone reading this for the first time would think it has anything to do with `$anchor`.

handrews: `$anchor` has only previously been mentioned once, in the section on fragments, with a forward reference.  It's not properly introduced until several sections after this.

jsonschema-core.xml
handrews: @Relequestual I'll do some of this, but at this stage I do not want to spend to much time fine-tuning wording.  If it's understandable, let's get this done.

handrews: @Relequestual actually I went and looked and I'm really consistent about using "this keyword".  I don't want to chase it all over- you are welcome to do a stylistic revamp when you take on the next draft 😃

handrews:
The update I just pushed is simply a rebase onto master to fix the conflicts from merging the other PRs.  The only conflict was the `$schema` and `$id` values of the vocab example (they were totally changed in the example PR, and I just had the trailing `#` in this PR).

There shouldn't be any significant change to the diff, as I made a point to keep the PRs as non-overlapping as possible.

handrews:
@gregsdennis I answered the one question you have- are you OK with this change now or have you not had time to finish reviewing?

@johandorland @Julian @kayess are any of you planning to weigh in?  If so, can you give an estimated date of when you might get to it?  If not I'll just wait for at least @Relequestual and @awwright (and @philsturgeon if he ever gets back to a laptop from his bike tour 🚴 ).

This is the last big mandatory thing, and is a prerequisite for the last optional proposal I'm still working on, and we'd really like to get to publishing, so... reviews are needed!  I know it's summer and folks are busy and on vacation and stuff, so letting me know when you might get to it would be very helpful.

Julian:
Thumbs up from me! Thanks as usual @handrews (both for a good change and for managing to catch my attention :/)

gregsdennis:
My question is answered

handrews:
I have now posted #788, which sits on top of this change.  In my opinion the combination of this change and that one is far superior to this one on its own. As this stands, there are quite a few weird cases and inconsistencies. Those are fixed by #788, although it requires that `$ref` accept a schema as well as a URI-reference.  Hopefully that will make more sense in the context of an actual PR with examples and everything.

Relequestual:
I'll plan to review changes this week.

jsonschema-core.xml
Relequestual: Fair enough.

jsonschema-core.xml
```diff
@@ -1476,87 +1508,221 @@
...
                     <t>
                         If present, the value for this keyword MUST be a string, and MUST represent a
-                        valid <xref target="RFC3986">URI-reference</xref>.
-                        This value SHOULD be normalized, and SHOULD NOT be an empty fragment &lt;#&gt;
-                        or an empty string &lt;&gt;.
+                        valid <xref target="RFC3986">URI-reference</xref>.  This URI-reference
+                        SHOULD be normalized, and MUST resolve to an
+                        <xref target="RFC3986">absolute-URI</xref> (without a fragment).  Therefore,
+                        "$id" MUST NOT contain a non-empty fragment, and SHOULD NOT contain an
+                        empty fragment.
```
Relequestual: I've read and re-read this one paragraph a few times.
MUST NOT contain a non-empty fragment...
SHOULD NOT contain an empty fragment...

Aren't these inverse situations and so should have inverse requirement keywords?
I may just be not properly understanding what a "non-empty" fragment is. Not sure. Tired.

philsturgeon: Double negative. “MUST NOT contain a fragment other than #”

Relequestual: Oooooooh I get it now. Morning rather than sleep time.
A non empty fragment would be `#something` and an empty fragment is `#`.
Allow empty fragment becaus legacy, but prefer no fragment at all.

Relequestual: Looks like this was just me... =]

Relequestual:
My suggestion of explicitly suggesting schema authors avoid non-canonical URIs still stands.

>                             An implementation MAY choose not to support addressing schemas
>                             by non-canonical URIs.

... "As such, it is RECOMENDED that schema authors only use canonical URIs, as using non-canonical URIs may reduce the transportability of a schema."

https://github.com/json-schema-org/json-schema-spec/pull/780/files#diff-f8351ad4b29f76d24d34d5b0f50ea16bR1655

jsonschema-core.xml
```diff
@@ -1476,87 +1508,221 @@
...
+                            functioning similarly to a retrieval URI for a resource.  While valid,
+                            examining the resolved value and either using the "$id" (if the value
+                            is a subschema), or resolving the reference and using the "$id" of the
+                            reference target, is preferable.
                         </t>
                         <t>
-                            The effect of using a fragment in "$id" that isn't blank or doesn't follow the
-                            plain name syntax is undefined.
+                            An implementation MAY choose not to support addressing schemas
+                            by non-canonical URIs.
```
Relequestual:
```suggestion
                            by non-canonical URIs. 
                            As such, it is RECOMENDED that schema authors only 
                            use canonical URIs, as using non-canonical URIs may reduce the 
                            transportability of a schema.
```

jsonschema-core.xml
handrews: OK I'm leaving this as-is.  "non-empty fragment" is phrasing that we have used in various places.

jsonschema-core.xml
```diff
@@ -1476,87 +1508,223 @@
...
+                            examining the resolved value and either using the "$id" (if the value
+                            is a subschema), or resolving the reference and using the "$id" of the
+                            reference target, is preferable.
                         </t>
                         <t>
-                            The effect of using a fragment in "$id" that isn't blank or doesn't follow the
-                            plain name syntax is undefined.
+                            An implementation MAY choose not to support addressing schemas
+                            by non-canonical URIs. As such, it is RECOMENDED that schema authors only 
+                            use canonical URIs, as using non-canonical URIs may reduce the
```
Relequestual:
```suggestion
                            use canonical URIs, as using non-canonical URIs may reduce
```

Just a gramatical fix on this one. Otherwise, golden.
You can just accept this suggested change =]

awwright:
Our naming of "$anchor" here is a little bit unfortunate, in the rest of the Web it's backwards: "id" is a plain label (like a CNAME) and "anchor" is a URI.

awwright:
By and large this seems like it'll be more accessible to Web developers than it was before; but the biggest hurdle I have is figuring out the significance of a "canonical URI".

Why not define a rel=self link, instead of a rel=canonical relationship?

handrews:
@awwright regarding our `$anchor` vs HTML's `id` (and I guess XML's?  I've never looked):  Yeah, that's unfortunate, and I noticed that.  I'm not sure what to do about it.  This part of our `$id` was never as widely used so it feels safer to change.

Also, if we tried to change `"$id": "#foo"` to `"$id": "foo"` (note the change from URI-reference to plain label) while changing `"$id": "https://example.com/bar"` to `"$somethingElse": "https://example.com/bar"` that would seem to be even more confusing, as *no* current use of `$id` would be left unchanged.

I'm open to suggestion, including leaving `$id` the way this PR leaves it but using a different name for `$anchor`, but I couldn't come up with anything better than this.  At least `$anchor` is reasonably accurate in the sense that in older versions of HTML the idiom `<a name="bar">` was referred to as an anchor.  But I did not want to use `$name` as that seemed bad for some other reason that escapes me at the moment.

Would `$name` be better?  And if so, should `$recursiveAnchor` be changed as well?

handrews:
@awwright 

> By and large this seems like it'll be more accessible to Web developers than it was before; but the biggest hurdle I have is figuring out the significance of a "canonical URI".
>
> Why not define a rel=self link, instead of a rel=canonical relationship?

Can you elaborate on what that would look like?  I found the "canonical" language helpful as the specific issue is understanding that, out of multiple possibly usable URIs, this is the one that should be used and the only one that can really be relied on.

Do you think that discussing this as a "self" relation instead is more accurate?  And (at least as important), do you think that more people will find discussion in terms of "self" intuitive?

awwright:
I'll sit down later and compare the two link relations, but off the top of my head, rel=self is more accurate.

My understanding of rel=canonical is more like "these are technically different resources, but the preferred way to access it is via that one over there". Mostly for search engine optimization, so you can tell that two different pages actually contain the same blog article.

handrews:
@awwright thanks, I'll also take a look at the wording and see if using "self" makes sense to me.  I do follow your reasoning here.

handrews:
To follow up on a discussion with @awwright on slack, this was resolved in favor of keeping "canonical" through the following reasoning:

After a discussion of how "self" and "canonical" might differ (a resource accessed through URI X might give X as its "self" but Y as its "canonical", which seems to be a thing that happens with web pages at least based on SEO guidance), @awwright asked:

> Do we want to convey a “most correct URI”, if any of the URIs can suffice?

To which my response was yes, that was ultimate the decision from issue #726: _`"$id"`: Eliminate base URI shadowing_

@awwright I'm taking your "I suppose that makes sense" from yesterday as approval since you didn't follow up with any further question or response.  If that is an error, please let me know and I can make a new PR to address any remaining concerns.

handrews:
Also, while I agree that our naming of `$id` and `$anchor` is somewhat unfortunate, that's come up before and no one has had a better idea then or now.  I wish we could figure something out, but we do want to preserve the more common uses of `$id`.

PR #782, opened 2019-08-17
Improve vocab example per gregsdennis suggestions

handrews:
This is heavily based on @gregsdennis 's work in PR #756.

Hopefully it is now more clear why I went off and updated a bunch
of other things before coming back to this.  What I found while
reviewing that PR is that the example wasn't great to begin with,
and in a few places was outright wrong.  This, understandably,
led to some of the enhancements to the example not being
quite right either.

Most notably, the old example should not have been set up
as its own meta-schema.  I can barely even figure out why
I must have done that, it makes so little sense 😕 

The example now follows the best practices from the previous
section by showing separate general purpose and single-vocabulary
meta-schemas, and discusses how they are used together.

I also switched "startDate" to "minDate" so that the parallel in behavior
with "minLength", "minItems", etc. was more intuitive.  And used `pattern`
in addition to `format` because of the changes to how `format` is handled
by default (another thing that wasn't done yet when @gregsdennis 
originally wrote his PR).

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1208,6 +1208,15 @@
                             must be repeated in the root of each schema document intended
                             for use as a meta-schema.  This is demonstrated in
                             <xref target="example-meta-schema">the example meta-schema</xref>.
+                            <cref>
+                                This requirement allows implementations to find all vocabulary
+                                requirement information in a single place for each meta-schema.
+                                As schema extensibility means that there are endless potential
+                                ways to combine more fine-grained meta-schemas by reference,
+                                requiring implementations to anticipate all possibilities and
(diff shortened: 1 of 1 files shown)


handrews:
@Relequestual I changed the example name to something similar to what you suggested (and noticed an inconsistency which got fixed along the way so that's a bonus).

@gregsdennis adding a schema (instance for the meta-schema) might be a good idea but I'm getting bogged down trying to come up with one- there are few things I'm worse at than coming up with examples.

Since you both approved it, I'm going to go ahead and merge (the update for the example name was trivial).  @gregsdennis if you want to add such a valid schema instance, I would be in favor of that.  I also think we can get away without it and, as you note, put it on the web site.

Issue #512, opened 2017-11-28
Move "definitions" to core (as "$defs"?)


handrews:
In #505 I commented that `definitions` fits better into core than validation (and @Anthropic agreed).

The point of `definitions` is simply to reserve a keyword under which you can safely collect named schemas, without worrying that they will get accidentally interpreted by an extended vocabulary.

`definitions` also allows packaging multiple schemas in a file for re-use as `$ref` targets without having to explicitly assign `$id`s.  In that sense, it is most closely associated with `$ref` and `$id`, both of which are part of core.

Furthermore, having a designated re-use location has nothing to do with validation or any other particular JSON Schema use case.  Reading through the validation spec it seems quite out of place.  For the last several drafts, there has even been a specific note in its description that it does *not* affect validation.

When you have to explain how the keyword has nothing to do with the purpose of the specification, that's probably a sign that the keyword is in the wrong place.

Anyone seriously object to moving this?

erayd:
Moving this into core sounds like a good idea. Core makes much more sense as a logical home for the definitions keyword, as the keyword itself doesn't actually imply any action - it's simply a well-known location where schema definitions may be stored.

In order to maintain consistency with other core keywords, it seems sensible to rename it to `$definitions`. Doing so will improve obviousness going forward, and seems unlikely to adversely impact users.

handrews:
@erayd I suppose there wouldn't be any harm in officially reserving `$definitions` and perhaps deprecating `definitions` but noting that extension vocabularies MUST NOT give it different behavior.  It's not like references to `#/definitions/...` would stop working :-P

erayd:
@handrews That sounds like a good idea.

erayd:
On a related note, would it also be possible to specify that arbitrary schema fragments SHOULD NOT appear outside of `[$]definitions` (other than locations already defined by the spec obviously, e.g. under `properties`), and that implementations MAY treat any schema that is defined in a nonstandard location as equivalent to `{}`?

The particular use-case I'm thinking of is a validation implementation which compiles a schema definition into procedural logic for direct execution against a document. Any circumstance which results in an inbound remote reference to a nonstandard location would necessarily imply lazy compilation, which adds significant implementation complexity and may be highly undesirable - in such cases, this would allow the implementation to ignore inbound references to a nonstandard location.

handrews:
[...]

PR #783, opened 2019-08-18
Clarify the nature of schema placeholder keywords

### LINKED ISSUE #778: Review draft-08/core/9.4 - $defs: Clarify its a recommendation
In requests for reviewing draft-08 I would like to comment on

**Core-Spec, Section 9.4. Schema Re-Use With "$defs"**

> The "$defs" keyword provides a standardized location for schema authors to inline re-usable JSON Schemas into a more general schema. [...]

In the original issue #512 @handrews [stated](https://github.com/json-schema-org/json-schema-spec/issues/512#issuecomment-353403603) that `$defs` is actually more of a recommendation while `definitions` and basically any other arbitrary keys (see https://github.com/json-schema-org/json-schema-spec/issues/512#issuecomment-352612038) should continue to work, too.

Reading previous versions of the specification as well as draft-08, though, it is easy to read it as if `$defs` (and formerly `definitions`) were *the only possible spec-conformant place* where to put subschemas.  It is not very clear that it has been a recommendation. I guess the fine distinction to catch is the use of the phrase *a standardized location* rather than *the standardized location* and an absence of MUST. However, *standardized* sounds more reliable than it obviously is and doesn't indicate a *recommendation* at all.

I think the spec would profit from using the wording *the recommended location* or from becoming even more clear with something like

> It MUST NOT be assumed that `$defs` is *the only permitted* location for subschemas. This is, because schemas and subschemas are identified by an URI which itself is agnostic to the actual keyword used to locate subschemas (see 8.3.2 Dereferencing). Schema authors SHOULD consider using `$defs` over any alternative for conventional purposes, though.

handrews:
Fixes #778 (paging @about-code).

Acknowledge that "$defs" and placeholders like it also indicate
that their values are schemas, which is relevant to whether
they can be detected as a proper "$ref" target or place where
"$id" may appear.

This ensures that similar extension keywords are understood
to have the same concerns as extension applicator keywords.

I did not add anything about `$defs` being recommended, because
now that it is covered in the same way as the applicator vocab,
it is clear that you can add more extension keywords like it.

And I think it is inherently clear that keywords in Core are recommended.

[See later comments from me about a few further changes]

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1871,7 +1871,7 @@
                     <section title="References to Possible Non-Schemas" anchor="non-schemas">
                         <t>
                             Subschema objects (or booleans) are recognized by their use with known
-                            applicator keywords or with placeholder keywords such as
+                            applicator keywords or with location-reserving keywords such as
                             <xref target="defs">"$defs"</xref> that take one or more subschemas
                             as a value.  These keywords may be "$defs" and the standard applicators
                             from this document, or extension keywords from a known vocabulary, or
@@ -2032,7 +2032,7 @@
 
             <section title='Comments With "$comment"'>
                 <t>
-                    This keyword is reserved a location for comments from schema authors
+                    This keyword reserves a location for comments from schema authors
                     to readers or maintainers of the schema.
 
                     The value of this keyword MUST be a string. Implementations MUST NOT present this
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1851,7 +1851,9 @@
                     <section title="References to Possible Non-Schemas">
                         <t>
                             Subschema objects (or booleans) are recognized by their use with known
-                            applicator keywords.  These keywords may be the standard applicators
+                            applicator keywords, or with placeholder keywords such as
+                            <xref target="defs">"$defs"</xref> that take one or more subschemas
+                            as a value.  These keywords may be "$defs" and the standard applicators
                             from this document, or extension keywords from a known vocabulary, or
                             implementation-specific custom keywords.
                         </t>
@@ -1973,7 +1975,7 @@
                     </section>
                 </section>
 
-                <section title='Schema Re-Use With "$defs"'>
+                <section title='Schema Re-Use With "$defs"' anchor="defs">
                     <t>
                         The "$defs" keyword provides a standardized location for schema
                         authors to inline re-usable JSON Schemas into a more general schema.
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -307,7 +307,7 @@
                     <t>
                         Object properties that are applied to the instance are called keywords,
                         or schema keywords.  Broadly speaking, keywords fall into one
-                        of three categories:
+                        of four categories:
                         <list style="hanging">
                             <t hangText="assertions:">
                                 produce a boolean result when applied to an instance
@@ -318,6 +318,10 @@
                             <t hangText="applicators:">
                                 apply one or more subschemas to a particular location
                                 in the instance, and combine or modify their results
+                            </t>
+                            <t hangText="reserved locations:">
+                                do not directly affect results, but reserve a place
+                                for a specific purpose to ensure interoperability
                             </t>
                         </list>
                     </t>
@@ -1003,6 +1007,22 @@
                         </t>
                     </section>
                 </section>
+            </section>
+            <section title="Reserved Locations">
+                <t>
+                    A fourth category of keywords simply reserve a location to hold re-usable
+                    components or data of interest to schema authors that is not suitable
+                    for re-use.  These keywords do not affect validation or annotation results.
+                    Their purpose in the core vocabulary is to ensure that locations are
+                    available for certain purposes and will not be redefined by extension
+                    keywords.
+                </t>
+                <t>
+                    While these keywords do not directly affect results, as explained in section
+                    <xref target="non-schemas" format="counter"></xref> unrecognized
+                    extension keywords that reserve locations for re-usable schemas may have
+                    undesirable interactions with references in certain circumstances.
+                </t>
             </section>
         </section>
         <section title="The JSON Schema Core Vocabulary">
@@ -1848,10 +1868,10 @@
                         </t>
                     </section>
 
-                    <section title="References to Possible Non-Schemas">
+                    <section title="References to Possible Non-Schemas" anchor="non-schemas">
                         <t>
                             Subschema objects (or booleans) are recognized by their use with known
-                            applicator keywords, or with placeholder keywords such as
+                            applicator keywords or with placeholder keywords such as
                             <xref target="defs">"$defs"</xref> that take one or more subschemas
                             as a value.  These keywords may be "$defs" and the standard applicators
                             from this document, or extension keywords from a known vocabulary, or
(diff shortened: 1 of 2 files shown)


jsonschema-core.xml
```diff
@@ -1789,7 +1789,9 @@
                     <section title="References to Possible Non-Schemas">
                         <t>
                             Subschema objects (or booleans) are recognized by their use with known
-                            applicator keywords.  These keywords may be the standard applicators
+                            applicator keywords, or with placeholder keywords such as
```
about-code: Sorry for my late feedback. To be honest I am not sure if this would clarify things considerably compared to before. I weren't able to draw the line in the opposite direction (as a reader of section *Schema Re-Use With "$defs"* right to this section and the essence of its meaning). But that might be caused by my personal situation of "not being in the specs" in such depth, so isn't meant to disqualify your answer. Though, the newly introduced term *placeholder keyword*  would require its own definition.

I would accept this PR as closing my issue but would also accept the issue being closed without any further action if majority doesn't see any necessity for clarification.

Thank you.

Just a little note: I can't comment on things in a timely manner ATM so don't make your progress depend on me.

handrews: Thanks, @about-code!  That's helpful feedback, I'll tweak it a bit more (we don't want people to have to already be experts when reading the standard- it's not a tutorial but it should be accessible).  If you have time to look at the update, great, but thanks for letting me know that you might not be available.

Relequestual:
> In the original issue #512 @handrews stated that $defs is actually more of a recommendation while definitions and basically any other arbitrary keys (see #512 (comment)) should continue to work, too.
- @about-code in the related issue

`$defs` is the recognised keyword, changed form `definitions` in previous draft-7.

Essentially, your question is, should `definitions` continue to work with draft-8?

The answer is, maybe, but you shouldn't rely on such, because the behaviour is now undefined.

Some implementations may choose to support referencing using arbitrary keys, but it's not required. There are lots of "super edge" cases that we would need to consider to define the behaviour, so rather than forbid it, leaving it undefined leaves implementations to document specific support for users who don't need their schemas to be portable. "behaviour is undefined" essentially translates to, "Probably not a good idea, as it's complex, but we won't stop you doing it if you really want, and understand it will likely mean your implementation isn't portable".

I hope that gives some additional clarification.

jsonschema-core.xml
Relequestual: /applicator keywords, or with placeholder keywords such as/applicator keywords or with placeholder keywords, such as

handrews:
@Relequestual @about-code OK I actually did some serious-ish reworking.

I had come up with the idea of `$defs` and `$comment` as "reserved location" keywords, and forgot that I'd never actually put that in.  So I added a fourth keyword category, which I think makes it obviously you can add more keywords like that.  I'm not going to outright say "you can make a keyword exactly like `$defs` because we don't really want people to do that.  But it should be clear that it's valid now.

@Relequestual `definitions` is actually sort of still reserved by the default meta-schema.  I clarified that in the validation spec.  The idea of calling that particular meta-schema the "default meta-schema" is part of #780. 

I updated both `$defs` and `$comment` to use wording about reserving locations, which also avoids the question of how forceful "standard" is.  Reserving a location does not imply that all other locations are unusable.  It just implies that this location is _reliably_ usable.

I think this is a good improvement overall, so thanks for bringing it to our attention, @about-code !

gregsdennis:
> @Relequestual Some implementations may choose to support referencing using arbitrary keys, but it's not required.

For example, mine does this. It allows `$ref` to point anywhere, but if what it finds is not a schema, it throws an exception.

handrews:
@gregsdennis also, presumably if you have a structure like:

```JSON
{
  "$id": "https://this.example.com/schemas/foo",
  "$ref": "#/a/b/c",
  "a": {
    "b": {
      "$id": "https://that.example.com/schemas/bar",
      "c": {
        "$ref": "somewhere#/$defs/whatever"
      }
    }
  }
}
```

where `a`, `b`, and `c` are all unrecognized keywords and all of them take schemas as values, you will end up evaluating the `$ref` in `c` incorrectly because you did not take into account the `$id` in `b`.

Or if you're implementation does look at that `$id`, and if `b` in fact instead does _not_ take a schema, but allows arbitrary keys which take schemas (like `properties`), then that would also produce the wrong `$ref` resolution behavior.

@gregsdennis @Relequestual , the point here isn't about implementations choosing to "support" arbitrary keys.  The point is that _it is absolutely not possible to guarantee the correct behavior of nested arbitrary unknown keys._  There is no way to do it.  You have to decide whether you look for `$id` in intervening levels, but since there are two possibilities and you can't implement both at once, you are _guaranteed_ to produce the wrong behavior for one use case or the other.

This is a strange corner case, but it is critically important not to over-promise spec behavior.

You *cannot* do this correctly, and shouldn't claim to do so.

However, you can get away with (for lack of a better term) schema duck typing for a single level of unknown keyword, because it's not possible to insert an unexpected `$id` anywhere.  But that's it.  Claiming to match whatever assumptions people make beyond that is in violation of the spec.  You can document a consistent behavior, but understand that there are cases where it will be wrong.

gregsdennis:
Yeah, hopefully people don't do that because it's really weird.

handrews:
@gregsdennis 

> Yeah, hopefully people don't do that because it's really weird.

LOL have you met people? 😆 

Are you OK with these changes as I've updated things?

jsonschema-core.xml
```diff
@@ -1004,6 +1007,22 @@
                     </section>
                 </section>
             </section>
+            <section title="Reserved Locations">
+                <t>
+                    A fourth category of keywords simply reserve a location to hold re-usable
+                    components or data of interest to schema authors that is not suitable
+                    for re-use.  These keywords do not affect validation or annotation results.
```
about-code: Here I'd suggest to use the term *placeholder keyword* at least once, maybe even emphasized. For example: 

> These *placeholder keywords*  do not affect validation[...] 

Then when the term *placeholder keyword* is used further down below such as in section *References to Possible Non-Schemas* it is terminologically clear that *placeholder keywords*  is the phrase used to refer to keywords falling into the category of *reserved locations*.

Overall the recent additions make things much more consistent and comprehensible, I think. :+1:

jsonschema-core.xml
```diff
@@ -1948,8 +1969,8 @@
 
             <section title='Comments With "$comment"'>
                 <t>
-                    This keyword is reserved for comments from schema authors to readers or
-                    maintainers of the schema.
+                    This keyword is reserved a location for comments from schema authors
```
about-code: > is reserved a location for

I guess it should be 

> *is a* reserved location

jsonschema-core.xml
handrews: Oh, I'd actually meant to get rid of "placeholder" and use the "location" terminology.  I'll sort that out and fix the out-of-order wording, too- thanks!

handrews:
OK, @about-code I made both of those places consistent by using "location-reserving keywords" and "reserves a location", respectively.

Since those changes are pretty trivial, and @Relequestual 's only specific change request was a comma that I fixed yesterday, I'm going to go ahead and merge.  @Relequestual seemed to approve the changes aside from that grammar thing.  And Greg approved yesterday.

PR #785, opened 2019-08-19
Fix verbose example to use `$defs`

garethsb-sony:
Follows on from 5b088711e78d472905ed37481d5ca055f370ba05.

### DIFF
--- a/output/verbose-example.json
+++ b/output/verbose-example.json
@@ -5,7 +5,7 @@
   "errors": [
     {
       "valid": true,
-      "keywordLocation": "#/definitions",
+      "keywordLocation": "#/$defs",
       "instanceLocation": "#"
     },
     {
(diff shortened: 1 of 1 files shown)


garethsb-sony:
This _Verbose_ example is also missing the `"error"` messages given in the matching _Basic_ and _Detailed_ examples, e.g. https://github.com/json-schema-org/json-schema-spec/blob/c5a60784b7794c5c236ca9fa815175a8580ee1eb/jsonschema-core.xml#L2750-L2785. Is that an oversight? Potentially a little confusing for a Verbose example to not be as verbose as it could be?

handrews:
This looks great to me- I'm just leaving it open to make sure @gregsdennis sees the question about `"error"`

Although for something like `"error": "Additional property 'z' found but was invalid." ` we recently discussed on slack that for this sort of low-level output, it would be more accurate for the error to be about the field matching `additionalProperties` but failing the `false` schema.

The idea being that people will probably write more human-friendly tools for analyzing verbose output rather than reading raw JSON, so they could render that error however they like.  But noting the error against the `false` schema is more consistent with how `additionalProperties` is actually supposed to function as an applicator keyword.

@gregsdennis what are your feelings on this?  We can open a separate issue if needed and merge this for now.

gregsdennis:
Probably an oversight.  I have no strong feels about fixing it here or in another PR.  Additionally, I don't think that I required having an `error` property for applicators.  Maybe I did; I don't recall, and I can't be bothered to go look it up right now.

> it would be more accurate for the error to be about the field matching additionalProperties but failing the false schema

:+1:

PR #787, opened 2019-08-23
Improve guidance on annotations and output

### LINKED ISSUE #786: Standardize handling of overlapping defaults
We've been working with handling defaults in react-jsonschema-form and the logic seems more complicated than one might initially expect. https://github.com/mozilla-services/react-jsonschema-form/pull/1376

Here are some examples of the scenarios we've come across:
- Imagine your schema is:
```
{
          type: "object",
          default: {
            level1: { level2: ["root-default-1", "root-default-2"] },
          },
          properties: {
            level1: {
              type: "object",
              properties: {
                level2: {
                  type: "array",
                  items: [
                    {
                      type: "string",
                      default: "child-default-1",
                    },
                    {
                      type: "string",
                    },
                  ],
                },
              },
            },
          },
        }
```

Then the default value should (probably, intuitively) be `{level1: {level2: { ["child-default-1", "root-default-2"] } }` instead of `{level1: {level2: { ["root-default-1", "root-default-2"] } }`.

- Another schema:
```
{
          type: "object",
          properties: {
            level1: {
              type: "array",
              default: [
                {
                  item: "property-default-1",
                },
                {},
              ],
              additionalItems: {
                type: "object",
                properties: {
                  item: {
                    type: "string",
                    default: "additional-default",
                  },
                },
              },
              items: [
                {
                  type: "object",
                  properties: {
                    item: {
                      type: "string",
                    },
                  },
                },
              ],
            },
          },
        }
```
Here, because the length of `items` is 1, the default should probably be `          level1: [
            { item: "property-default-1" },
            { item: "additional-default" },
          ]` rather than `          level1: [
            { item: "property-default-1" },
            {}
          ]`

You can see more examples at https://github.com/mozilla-services/react-jsonschema-form/pull/1376/files#diff-0de170772e45d2602c103dd916607787.

I was wondering if there might be some value in standardizing this behavior, perhaps in the JSON Schema spec.

The [spec](http://json-schema.org/latest/json-schema-validation.html#rfc.section.10.2) says:
> There are no restrictions placed on the value of this keyword. When multiple occurrences of this keyword are applicable to a single sub-instance, implementations SHOULD remove duplicates.

Perhaps it would be good to more clearly define what "removing duplicates" is, and define an algorithm for doing so. It's true that `default` has no effect on validation, though, but it might still be good to somehow standardize how to handle overlapping defaults.

handrews:
Fixes #786.

Remove wording that is no longer meaningful now that we have
recommended output formats.

Also do not use meta-schema as example for relative and absolute
paths because my brain broke trying to figure out what it was doing.

Finally, a few tweaks in the output description where I felt it was 
a bit confusing- @gregsdennis let me know if these are wrong or if
you feel they're a net negative.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -813,21 +813,25 @@
                     multiple applicable occurrences of the keyword with different values.
                 </t>
                 <t>
-                    The default behavior is simply to collect all values in a list in
-                    indeterminate order.  Given the extensibility of keywords, including
-                    applicators, it is not possible to define a universally predictable
-                    order of processing.
-                </t>
-                <t>
                     Unlike assertion results, annotation data can take a wide variety of forms,
                     which are provided to applications to use as they see fit.  JSON Schema
                     implementations are not expected to make use of the collected information
                     on behalf of applications.
                 </t>
                 <t>
+                    Unless otherwise specified, the value of an annotation keyword's
+                    annotation is the keyword's value.  However, other behaviors are possible.
+                    For example, <xref target="json-hyper-schema">JSON Hyper-Schema's</xref>
+                    "links" keyword is a complex annotation that produces a value based
+                    in part on the instance data.
+                </t>
+                <t>
                     While "short-circuit" evaluation is possible for assertions, collecting
                     annotations requires examining all schemas that apply to an instance
                     location, even if they cannot change the overall assertion result.
+                    The only exception is that subschemas of a schema object that has
+                    failed validation MAY be skipped, as annotations are not retained
+                    for failing schemas.
                 </t>
 
                 <section title="Collecting Annotations">
@@ -863,7 +867,9 @@
                         If the same keyword attaches values from multiple schema locations
                         to the same instance location, and the annotation defines a process
                         for combining such values, then the combined value MUST also be associated
-                        with the instance location.
+                        with the instance location.  The <xref target="output">output formats</xref>
+                        described in this specification that include annotation information
+                        meet this requirement.
                     </t>
                     <section title="Distinguishing Among Multiple Values">
                         <t>
@@ -2649,13 +2655,13 @@
                     <figure>
                         <artwork>
 <![CDATA[
-#/properties/minLength/$ref/minimum
+#/properties/width/$ref/minimum
 ]]>
                         </artwork>
                     </figure>
                     <t>
-                    Note that this pointer may not be resolvable due to the inclusion of these
-                    applicator keywords.
+                    Note that this pointer may not be resolvable by the normal JSON Pointer process
+                    due to the inclusion of these by-reference applicator keywords.
                     </t>
                     <t>
                         The JSON key for this information is "keywordLocation".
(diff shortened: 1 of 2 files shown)


gregsdennis:
This looks fine to me.

handrews:
@epicfaace if you have a moment to take a look I'd welcome you're feedback on this PR.

jsonschema-core.xml
```diff
@@ -812,22 +812,26 @@
                     location, annotation keywords need to specify any unusual handling of
                     multiple applicable occurrences of the keyword with different values.
                 </t>
-                <t>
-                    The default behavior is simply to collect all values in a list in
-                    indeterminate order.  Given the extensibility of keywords, including
-                    applicators, it is not possible to define a universally predictable
-                    order of processing.
```
epicfaace: Seems like the spec is moving towards defining rules for handling annotations in overlapping subschemas (as this paragraph is removed, and the wording in `readOnly` and `writeOnly` show), which is good.

epicfaace:
Thanks for pointing me to this. The same logic you mentioned about defaults in #786 (and which is discussed in [7.6.1.1](http://json-schema.org/work-in-progress/WIP-jsonschema-core.html#rfc.section.7.6.1.1)) could be used to argue against, however, the guidance added in this PR about how to handle overlapping annotations, such as:

>             <section title='"readOnly" and "writeOnly"'>
>                 <t>
>                     The value of these keywords MUST be a boolean.  When multiple occurrences
>                     of these keywords are applicable to a single sub-instance, the resulting
>                     behavior SHOULD be as for a true value if any occurrence specifies a true value,
>                     and SHOULD be as for a false value otherwise.
>                 </t>

(For example, what if an implementor would prefer that the value of `readOnly` would only apply to the most specific subschema? or least specific?)

I'd prefer that the schema be explicit in these cases, as the PR does. My question is this: If it's able to be explicit about handling attributes such as `readOnly` / `writeOnly`, then why not `defaults` as well?

handrews:
@epicfaace for `readOnly` and `writeOnly` there is a pretty strong consensus on appropriate behavior.  We recommend this with the RFC 2119 term `SHOULD` to indicate that this behavior is expected to be common, and those who might take advantage of common behavior for interoperability should use this behavior.

However, `SHOULD` is not `MUST`.  Application wishing to define alternate behavior for these keywords can still do so without contradicting the specification at all.  `SHOULD` is a strong recommendation, but not a requirement.

Over the course of numerous years on this project, nothing remotely resembling a consensus around `default` has **ever** emerged.  We have been unable to even get an agreement on whether the value of `default` ought to be valid against the schema in some way (this is more complex than it sounds and I'm not going to get into it- you can search issues, particularly closed issues, if you really want to slog through it).  `examples` is similar, although for controversy no annotation comes close to the arguments and attitudes around `default`.

`title` and `description` are not really controversial, but at the same time, various people have requested diametrically opposed behaviors, all for reasonable reasons.  So we do not take a position.

The goal here is to provide the appropriate amount of guidance for a given keyword, not to provide a one-size-fits-all approach to keyword definitions.  Some keywords are helped by a recommendation (but you'll note that I weakened them from `MUST` to `SHOULD` specifically to ensure that all annotations have that freedom to be used differently).  But some definitely are not.

PR #788, opened 2019-08-24
More consistent concept of schema resources

### LINKED ISSUE #779: Allow "$ref" to take a schema, to make inlining easier?

handrews:
***[EDIT: I am going to do a revision of this attempting to keep the nice "schema resource" concept and facilitate the bundling use case without needing the `$ref` schema value.  There are many things about the approach in this PR (788) that I like, but having sorted out more details I may have one more option for comparison.]***

Fixes #779.

_To me, this is essentially part of #780.  Where we end up with
here is much more sensible and consistent, with fewer weird caveats.
It acknowledges a use case that I have seen and used often, while
making clear that many "dereferencing" cases, while common and
potentially workable depending on what you're doing, are in fact
likely to break things._

This further clarifies the schema resource concept and
how it is used.  In particular, schema resources are
crucial to determining when it is safe to embed a
reference target into its referring schema.

This also requires allowing "$ref" to take a schema value.

The bundling use case should be more clear given this
guidance.  While many reference target embedding transformations
are likely to break things and are therefor to be discouraged,
when framed in terms of schema resources a clear use case emerges.

This also relaxes a restriction on "$schema" that has no functional
impact but avoids requiring implementations to detect and handle
a rather complex case (embedding schema resources with different
"$schema" values).  And also means that embedding can work without
having to change the embedded value by removing the "$schema"
keyword.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -100,7 +100,7 @@
             <t>
                 This specification defines JSON Schema core terminology and mechanisms, including
                 pointing to another JSON Schema by reference,
-                dereferencing a JSON Schema reference,
+                dereferencing a JSON Schema reference or embedding its target,
                 specifying the vocabulary being used,
                 and defining the expected output.
             </t>
@@ -410,11 +410,6 @@
                         <xref target="RFC3986">absolute URI</xref>.
                     </t>
                     <t>
-                        As discussed in section
-                        <xref target="id-keyword" format="counter"></xref>, a JSON Schema document
-                        can contain multiple JSON Schema resources.
-                    </t>
-                    <t>
                         The root schema is the schema that comprises the entire JSON document
                         in question.  The root schema is always a schema resource, where the
                         URI is determined as described in section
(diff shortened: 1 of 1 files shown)


handrews:
I'm going to take this down in favor of another approach.  May revive if that doesn't work.

PR #789, opened 2019-08-24
Best practices for generative use cases adding semantics to references

handrews:
_Note that this was at one point going to be part of #788, 
but it turned out they could be done separately._

This is going to be a powerful technique for generative use
cases, which have been observed either requiring references
(as opposed to subschemas) in certain places, or implementing
various sorts of inferencing around references on their own
or in conjunction with other keywords.

For example, one OpenAPI documentation tool assumes that
in an "allOf" of several references, the first is a base
class of some sort.  While this is not correct behavior
by either JSON Schema or OpenAPI, it demonstrates a feature
built by making implicit assumptions about certain references.

This section provides a recommendation for making such
behavior explicit, and also guides writers of such tools
to the concept of extension vocabularies and annotations.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2139,6 +2139,67 @@
                             </cref>
                         </t>
                     </section>
+                    <section title="Assigning semantic meaning to references.">
+                        <t>
+                            While the presence of references is expected to be transparent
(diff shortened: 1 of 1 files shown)


handrews:
I'm taking this down to re-post it with a better-organized best practices section.

PR #790, opened 2019-08-24
Trivial: Consolidate core examples/best practices.

handrews:
_There are **no** content changes, just moved sections_

This improves the flow and will work better as we add a couple
more best practices sections before publishing.

The examples within core tend to rely on understanding more
of the core keywords than have necessarily been introduced.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1302,177 +1302,6 @@
                         interoperable across implementations.
                     </t>
                 </section>
-                <section title="Best Practices for Vocabulary and Meta-Schema Authors">
-                    <t>
-                        Vocabulary authors SHOULD
-                        take care to avoid keyword name collisions if the vocabulary is intended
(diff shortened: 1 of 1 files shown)


handrews:
Or should this be an appendix (or a couple of appendices)?  I don't think there are any normative requirements in these sections at all, and moving it down there would substantially reduce the epic size of the normative (or mostly normative) portion of the document.

Similar to some of the Hyper-Schema recommendations that we moved down http://json-schema.org/draft-07/json-schema-hypermedia.html#rfc.appendix.A

@awwright do you have thoughts on this?  I'm actually leaning towards an appendix and might update this to reflect that (we can always go back, or reject this PR entirely).

handrews:
OK I like this a lot better as appendixes, will repost.

awwright:
An appendix does seem like the appropriate place to put this, let's try it out and see how it works

PR #791, opened 2019-08-25
 Move core examples/best practices to appendixes

handrews:
***NOTE**: The first commit is just a move, the second changes the RFC 2119 terms.  They should be examined separately.*

These sections are not normative, and moving them to an
appendix reduces the length of the sizable normative
portion of the spec.  And also improves flow, I think.

The examples within core tend to rely on understanding more
of the core keywords than have necessarily been introduced,
so they work better after the entire vocabulary has been
introduced.

Vocabulary authorship is not part of the JSON Schema specification,
so this guidance should not use RFC 2119 terms.

PR #792, opened 2019-08-25
Add guidance on bundling and references

handrews:
This should make clear what use cases are considered valid
and safe, vs when someone is on their own in terms of figuring
that out.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3597,7 +3597,7 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
             </t>
             <t>
                 To make such use case-specific semantics explicit, the best practice
-                is to create an annotation keyword can be created for use in the same
+                is to create an annotation keyword for use in the same
                 schema object alongside of a reference keyword such as "$ref".
             </t>
             <figure>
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -3414,6 +3414,57 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
...
+                    Attempting to remove all references and produce a single schema document does not,
+                    in all cases, produce a schema with identical behavior to the original form.
+                </t>
+                <t>
+                    Since "$ref" is now treated like any other keyword, with other keywords allowed
+                    in the same schema objects, fully supporting non-recursive "$ref" removal in
+                    all cases can require relatively complex schema manipulations.  It is beyond
+                    the scope of this specification to determine or provide a set of safe "$ref"
+                    removal transformations, as they depend not only on the schema structure
+                    but also on the intende usage.
```
philsturgeon:
```suggestion
                    but also on the intended usage.
```

handrews:
I've moved these commits over to #791 along with the commits for #793 in order to tidy up how all of these stacked PRs need to merge.  At this point, everything in #791 (which includes #792 and #793 now) was approved in its original PR by @Relequestual and/or @philsturgeon .

PR #793, opened 2019-08-25
Generative use case guidance

handrews:
This is going to be a powerful technique for generative use
cases, which have been observed either requiring references
(as opposed to subschemas) in certain places, or implementing
various sorts of inferencing around references on their own
or in conjunction with other keywords.

For example, one OpenAPI documentation tool assumes that
in an "allOf" of several references, the first is a base
class of some sort. While this is not correct behavior
by either JSON Schema or OpenAPI, it demonstrates a feature
built by making implicit assumptions about certain references.

This section provides a recommendation for making such
behavior explicit, and also guides writers of such tools
to the concept of extension vocabularies and annotations.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3589,6 +3589,68 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
             </section>
         </section>
 
+        <section title="References and generative use cases">
+            <t>
+                While the presence of references is expected to be transparent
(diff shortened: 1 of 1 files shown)


Relequestual:
Given this section you've added has no requirement keywords, and is for illustration, does it fit better in an appendix as opposed to the spec body?

handrews:
@Relequestual It is in an appendix.  Any `<section>` tag after the `</middle>` closing tag is an appendix.  There's no `<appendix>` tag :-)

PR #795, opened 2019-08-28
Clarify "$recursiveAnchor" behavior in terms of base URIs

handrews:
So.. I know I said I was done, but... I went to add a bit about `$id` and `$anchor` being identification keywords (alongside assertions, annotations, applicators, and reserved locations), because it seemed like it might fix a conceptual gap and help folks understand what we're doing with the split, and then I realized that `$recursiveAnchor` fit into this category if we think about it as dynamically setting the base URI for resolving `$recursiveRef`.  So that's the first commit.

This led to a much more straightforward way of explaining how these keywords work, so I reworked that whole section.  I feel like with the new presentation, the elaborate justification is not required.  That's the second commit (including removing the old example).

The third commit is the new example, explained in terms of the new presentation of how the keywords work.  It's down in the Appendixes with the other big example sections.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -310,6 +310,10 @@
                         or schema keywords.  Broadly speaking, keywords fall into one
                         of four categories:
                         <list style="hanging">
+                            <t hangText="identifiers:">
+                                control schema identification through setting the schema's
+                                canonical URI and/or changing how the base URI is determined
+                            </t>
                             <t hangText="assertions:">
                                 produce a boolean result when applied to an instance
                             </t>
@@ -702,6 +706,25 @@
                     "<xref target="additionalItems" format="title"/>" and
                     "<xref target="additionalProperties" format="title"/>" keywords in this
                     document.
+                </t>
+            </section>
+            <section title="Identifiers" anchor="identifiers">
+                <t>
+                    Identifiers set the canonical URI of a schema, or affect how such URIs are
+                    resolved in <xref target="references">references</xref>, or both.
+                    The Core vocabulary defined in this document defines several
+                    identifying keywords, most notably "$id".
+                </t>
+                <t>
+                    Canonical schema URIs MUST NOT change while processing an instance, but
+                    keywords that affect URI-reference resolution MAY have behavior that
+                    is only fully determined at runtime.
+                </t>
+                <t>
+                    While custom identifier keywords are possible, vocabulary designers should
+                    take care not to disrupt the functioning of core keywords. For example,
+                    the "$recursiveAnchor" keyword in this specification limits its URI resolution
+                    effects to the matching "$recursiveRef" keyword, leaving "$ref" undisturbed.
                 </t>
             </section>
             <section title="Applicators" anchor="applicators">
@@ -1517,7 +1540,8 @@
                         without requiring JSON Pointer references to be updated.
                     </t>
                     <t>
-                        The "$anchor" keyword is used to specify such a fragment.
+                        The "$anchor" keyword is used to specify such a fragment.  It is an
+                        identifier keyword that can only be used to create plain name fragments.
                     </t>
                     <t>
                         If present, the value of this keyword MUST be a string, which MUST start with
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -1563,206 +1587,84 @@
...
-                            complicated the original schema, the more redefinitions are necessary
-                            in the extension.  This leads to a verbose and error-prone extension,
-                            which must be kept synchronized with the original schema if the
-                            original changes its recursive fields.
-                            This approach can be seen in the meta-schema for JSON Hyper-Schema
-                            in all prior drafts.
+                            For an example using these keyword, see appendix
+                            <xref target="recursive-example" format="counter" />.
+                            <cref>
+                                The difference between the hyper-schema meta-schema in previous
```
philsturgeon: Should we be talking about hyper-schema in core?

handrews: @philsturgeon it's in a CREF.  It's not really part of the spec, it's just commentary and it's by far the clearest example of why this feature is needed.  The old hyper-schema meta-schema is unmaintainable.  I broke it a lot trying to keep it in sync.

Also, this hyper-schema reference is not new, so if you want to take it out, that's a separate issue.  I just moved it and maybe reworded it.

jsonschema-core.xml
```diff
@@ -1587,206 +1587,75 @@
...
-                                schema document, and MUST NOT appear in any subschema.
+                                Finally, the value of "$recursiveRef" is resolved against the
+                                new base URI determined according to "$recursiveAnchor" producing
+                                the final resolved reference URI.
                             </t>
                             <t>
-                                The value of "$recursiveAnchor" MUST be of type boolean, and
-                                MUST be true.  The value false is reserved for possible future use.
+                                Note that in the absence of "$recursiveAnchor" (and in some cases
+                                when it is present", "$recursiveRef"'s behavior is identical to
```
Relequestual: Assuming there should be a closing bracket here?

jsonschema-core.xml
```diff
@@ -3307,6 +3316,81 @@ User-Agent: product-name/5.4.1 so-cool-json-schema/1.0.2 curl/7.43.0
...
+            <t>
+                The outermost is "#", which is the root schema of the "strict-tree"
+                schema, so we use its base URI of "https://example.com/strict-tree",
+                which produces a final resolved URI of
+                "https://example.com/strict-tree#" for the "$recursiveRef".
+            </t>
+            <t>
+                This way, the recursion in the "tree" schema recurses to the root
+                of "strict-tree", instead of only applying "strict-tree" to the
+                instance root, but applying "tree" to instance children.
```
Relequestual: This is great and pretty clear. Easy reading.

jsonschema-core.xml
```diff
@@ -1587,206 +1587,75 @@
...
-                                to that of "$ref".
+                                "$recursiveAnchor" is used to dynamically identify a base URI
+                                at runtime for "$recursiveRef" by marking where such a calculation
+                                can start, and where it stops.  This keyword MUST NOT affect the
+                                base URI of other keywords, unless they are explicitly defined
+                                to rely on it.
+                            </t>
+                            <t>
+                                If set to true, then when the containing schema object is used
+                                as a dynamic reference target, a new base URI is determined
```
Relequestual: "as a dynamic reference target (by use of `$ref`)"
Is this what you mean by that phrase?

Relequestual:
One quick fix. One query.

Looking at the example makes the use clear, but I SHOULD be able to mostly understand it without the example. I'm not 100% confident that's the case. I'll re-read a few more times.

Relequestual:
I can understand why `$recursiveAnchor` and `$recursiveRef` are in individual sections, but I think in doing so you've made it really hard to understand before reading the example. I can't quite match up the spec words to the way in the process works and is explained in the example.

The example is great BTW!

Maybe the spec could detail a step by step algorithm for resolving `$recursiveRef`?
This could be a new section, or as part of a merged those two keywords section (if you decide to try that).

I'll say it again, the example makes it really clear. I've not looked at this for a while, but looking at the example now, I immediatly understand how it works. Such quick understanding is really encouraging and makes me (and presumidly others) feel smart. 💡

jsonschema-core.xml
handrews: Heh, you don't like infinite asides?

Relequestual:
It's a HUGE improvement! Bravo!

jsonschema-core.xml
handrews: Not quite, but I changed the wording so it should be more clear now.

jsonschema-core.xml
handrews: Fixed.

handrews:
@Relequestual I did not add a step-by-step listing (I tried that originally but it got rather hard to do precisely because of too many conditionals).  However, I added a paragraph in the intro section before the link to the example that informally summarizes how it works.  Does that help?

PR #802, opened 2019-09-13
fix invalid JSON in 7.7.1.1. Distinguishing Among Multiple Values

notEthan:
fix nesting and structure of JSON example. the existing JSON is indented incorrectly, contains a closing `}` with no corresponding `{` and doesn't parse.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -891,29 +891,28 @@
 {
     "title": "Feature list",
     "type": "array",
-        "items": [
-            {
-                "title": "Feature A",
-                "properties": {
-                    "enabled": {
-                        "$ref": "#/$defs/enabledToggle",
-                        "default": true
-                    }
-                }
-            },
-            {
-                "title": "Feature B",
-                "properties": {
-                    "enabled": {
-                        "description": "If set to null, Feature B
-                                        inherits the enabled
-                                        value from Feature A",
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -891,35 +891,32 @@
...
-                                        value from Feature A",
-                        "$ref": "#/$defs/enabledToggle"
-                    }
+            }
+        },
+        {
+            "title": "Feature B",
+            "properties": {
+                "enabled": {
+                    "description": "If set to null, Feature B inherits the
```
handrews: Please re-wrap this the way it was before.  The way this gets formatted in the plain text output this line ends up too wide.

jsonschema-core.xml
```diff
@@ -891,35 +891,32 @@
...
-    },
+        }
+    ],
     "$defs": {
         "enabledToggle": {
             "title": "Enabled",
-            "description": "Whether the feature is enabled (true),
-                            disabled (false), or under
-                            automatic control (null)",
+            "description": "Whether the feature is enabled (true), disabled
```
handrews: Please re-wrap this the way it was before.  The way this gets formatted in the plain text output this line ends up too wide.

notEthan:
I suspected messing with that might be a mistake. restored the original wrapping.

Issue #824, opened 2019-12-09
Vertices is not defined in the example in 10.4 Output Sturucture


bwicklund:
The following schema is defined 
```
{
  "$id": "https://example.com/polygon",
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "$defs": {
    "point": {
      "type": "object",
      "properties": {
        "x": { "type": "number" },
        "y": { "type": "number" }
      },
      "additionalProperties": false,
      "required": [ "x", "y" ]
    }
  },
  "type": "array",
  "items": { "$ref": "#/$defs/point" },
  "minItems": 3
}
```

With the following instance given:

```
[
  {
    "x": 2.5,
    "y": 1.3,
  },
  {
    "x": 1,
    "z": 6.7
  }
]
```

You call out the errors this instance will produce as the following: 

```
The second element in the "vertices" property is missing a "y" property.
The second element in the "vertices" property has a disallowed "z" property.
There are only two vertices, but three are required.
```

**vertices** is not defined anywhere; only the object name "point" is mentioned in the schema, It would be a lot of any interpreter to infer these individual points to be "vertices."

I would suggest this edit:
```
The second "point" object is missing a "y" property.
The second "point" object has a disallowed "z" property.
There are only two "point"', but three are required.
```

bwicklund:
Pull Request: https://github.com/json-schema-org/json-schema-spec/pull/825

gregsdennis:
Discussion continues is the referenced PR.

Relequestual:
Given the PR was merged, I'm closing this issue.
Please feel free to comment or re-open if need be.

PR #825, opened 2019-12-09
Update to 10.4 "Output Structure"

bwicklund:
Per the following issue: https://github.com/json-schema-org/json-schema-spec/issues/824

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2636,13 +2636,13 @@ https://example.com/schemas/common#/$defs/count/minimum
                     Specifically, the errors it will produce are:
                     <list>
                         <t>
-                            The second "point" object is missing a "y" property.
+                            The second object is missing a "y" property.
                         </t>
                         <t>
-                            The second "point" object has a disallowed "z" property.
+                            The second object has a disallowed "z" property.
                         </t>
                         <t>
-                            There are only two "point", but three are required.
+                            There are only two objects, but three are required.
                         </t>
                     </list>
                     Note that the error message wording as depicted in these examples is not a
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2636,13 +2636,13 @@ https://example.com/schemas/common#/$defs/count/minimum
                     Specifically, the errors it will produce are:
                     <list>
                         <t>
-                            The second element in the "vertices" property is missing a "y" property.
+                            The second "point" object is missing a "y" property.
                         </t>
                         <t>
-                            The second element in the "vertices" property has a disallowed "z" property.
+                            The second "point" object has a disallowed "z" property.
                         </t>
                         <t>
-                            There are only two vertices, but three are required.
+                            There are only two "point", but three are required.
                         </t>
                     </list>
                     Note that the error message wording as depicted in these examples is not a


gregsdennis:
Why is this an improvement? It's not fixing anything. There's not actually a typo in the existing text.

jsonschema-core.xml
```diff
@@ -2636,13 +2636,13 @@ https://example.com/schemas/common#/$defs/count/minimum
                     Specifically, the errors it will produce are:
                     <list>
                         <t>
-                            The second element in the "vertices" property is missing a "y" property.
+                            The second "point" object is missing a "y" property.
```
gregsdennis: I would suggest just saying `The second item` and leaving out the type.

jsonschema-core.xml
```diff
@@ -2636,13 +2636,13 @@ https://example.com/schemas/common#/$defs/count/minimum
...
                         </t>
                         <t>
-                            The second element in the "vertices" property has a disallowed "z" property.
+                            The second "point" object has a disallowed "z" property.
```
gregsdennis: Ditto

jsonschema-core.xml
```diff
@@ -2636,13 +2636,13 @@ https://example.com/schemas/common#/$defs/count/minimum
...
                         </t>
                         <t>
-                            There are only two vertices, but three are required.
+                            There are only two "point", but three are required.
```
gregsdennis: `There are only two items`.

jsonschema-core.xml
bwicklund: How about `The second object`? A parser would know it is an object.

gregsdennis: That's fine.  Like I mentioned, this portion of the text is prose, not machine output.  The reason I suggested the edit was the edit below: `There are only two "point"` sounds weird, and I don't know how to fix it, so just reword it.

bwicklund: I totally understand it is prose, but it also feels like and an example of potential error output so I think it should read as real as possible.

Relequestual:
Interesting case here.
What would happen if the `$ref` was inlined?
You would then only know what "it" was based on any provided annotations, for example `title`.
Alterntativly, use `properties` key, OR, say "at the document root".
My feeling is the correct "fix" for this would be to include the annotation.
Additionally, we could suggest a workflow for determining the subject of the error.

Later in 10.4.2, we give the BASIC output with the error "Expected at least 3 items but found 2".

Maybe the mentioned error message was simply not sanity checked.
Either way, the ISSUE is correct, but I'm not sure on the intent or the best resolution right now.

Relequestual:
In terms of "what's the process", let's open a new issue and get it defined properly =]

handrews:
@gregsdennis I tried really hard to get all of the example lines within the limits of the IETF text rendering format, or at least down to only giving specification build warnings instead of errors.  I'm fairly certain I changed it to "points" to save a few letters and must have missed "vertices" here.

handrews:
@Relequestual @gregsdennis the process is to tag these things with the appropriate milestone, which would be draft-08-patch1.  Which I've been doing as appropriate since 2019-09 was published.  See also the big table at https://json-schema.org/specification-links.html which records milestones along with every other type of version thingy we use.

Relequestual:
I think two is the magic number here...

PR #839, opened 2020-01-28
Make sure xref is wrapped in `t` tag to fix #814

### LINKED ISSUE #814: strange formatting in Relative JSON Pointers section 4. Evaluation
referencing the source https://github.com/json-schema-org/json-schema-spec/blob/draft-handrews-json-schema-02/relative-json-pointer.xml#L121-L140

and the web site:
https://json-schema.org/draft/2019-09/relative-json-pointer.html#rfc.section.4

the following text (starting with an `<xref>` tag) appears at entirely the wrong place in the output on the site:

> RFC 6901, Section 4 ("Evaluation"), with the modification that the initial reference being used is the reference currently being held (which may not be root of the document).

it should follow "then evaluation proceeds as per" but instead appears before the `<list>` preceding it.

weird.

Relequestual:
Looks like if you have a `t` tag with a `list` tag, you're not expected to have other tags at the same level.

I couldn't find the documentation to confirm this, but this change does seem to fix the issue.

### DIFF
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -133,9 +133,12 @@
                         the new referenced value is that object.
                     </t>
                 </list>
+            </t>
+            <t>
                 If the remainder of the Relative JSON Pointer is a JSON Pointer, then
-                evaluation proceeds as per <xref target="RFC6901">RFC 6901, Section 4</xref>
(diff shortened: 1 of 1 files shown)


Relequestual:
Good call. Noted

PR #840, opened 2020-01-28
Rephrase potentially confusing paragraph

### LINKED ISSUE #817: Fix awkward paragraph in 8.2.4.5 "Loading a Reference Schema"
In [section 8.2.4.5](https://tools.ietf.org/html/draft-handrews-json-schema-02#section-8.2.4.5), the 2nd paragraph reads:

> When schemas are downloaded, for example by a generic user-agent that doesn't know until runtime which schemas to download, see Usage for Hypermedia ([Section 11](https://tools.ietf.org/html/draft-handrews-json-schema-02#section-11)).

which I had trouble parsing just now while referring someone to that section.  I _think_ what's going on here is that:

* We want to acknowledge that while implementations are not required to automatically fetch referenced schemas, sometimes that will be a thing that happens.
* In particular, acknowledging systems where the exact schemas needed are only known at runtime.
* Section 11 is specifically about ways to fetch schemas that allow meta-data that can work with RFC 8288 links and/or media types and their parameters.  This is relevant, but not the only option- loading from files is also common, for example.

We should figure out what we're trying to say here and say it more clearly, without assuming HTTP or similar protocols.

Relequestual:
Resolves #817

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1509,7 +1509,7 @@
                         </t>
                         <t>
                             An implementation MAY choose not to support addressing schemas
-                            by non-canonical URIs. As such, it is RECOMENDED that schema authors only 
+                            by non-canonical URIs. As such, it is RECOMENDED that schema authors only
                             use canonical URIs, as using non-canonical URIs may reduce
                             schema interoperability.
                             <cref>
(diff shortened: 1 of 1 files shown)


handrews:

PR #843, opened 2020-01-29
Fix/832

### LINKED ISSUE #832: JSON Schema Core - 11.2/13.1, 13.2 - inconsistent treatment of multiple schema items
JSON Schema Core, 11.2, about the media type parameter "schema":
"Multiple schemas are whitespace separated, and indicate that the instance conforms to all of the listed schemas:"
This statement is followed by an example using content type "_application/json_".

In 13.1 and 13.2, however, dealing with media types "_application/schema+json_" and "_application/schema-instance+json_", the role of the media type parameter "schema" is defined slightly differently:
"The instance SHOULD successfully validate against at least one of these meta-schemas." (13.1)
"The instance SHOULD successfully validate against at least one of these schemas." (13.2)

**PROPOSAL**
- If the different treatment of multiple "schema" values in media type "application/json" versus media types "application/schema+json" and "application/schema-instance-json" is intended, make the (surprising) difference explicit, e.g. adding a note to 11.2
- Otherwise, correct 11.2 (presumably)

Relequestual:
Fixes #832

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2950,7 +2950,7 @@ Link: <https://example.com/my-hyper-schema#>; rel="describedby"
                 <figure>
                     <artwork>
 <![CDATA[
-Content-Type: application/json;
+Content-Type: application/schema-instance+json;
           schema="https://example.com/my-hyper-schema#"
 ]]>
                     </artwork>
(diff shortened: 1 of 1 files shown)


handrews:
@Relequestual oof now I remember what's going on here.  There were two ideas and I'm not 100% sure how to resolve them.  I'm going to elaborate in the issue instead of in the PR :-)

handrews:

PR #844, opened 2020-02-03
Don't refer to JSON values as "productions".

### LINKED ISSUE #735: "What's a 'production'?"
The phrase "A JSON null production" is alien to me.
It reportidly comes from ABNF.
The wikipedia page for EBNF uses the term, but not the RFC related to ABNF that we link to.

Given it doesn't add to the understanding (for me, and others), I discussed with @awwright and he agrees we should remove it.

Looks like this simply means "rule" or the result of following a production rule. We can rephrase this to make it less unintittive for those unfamiliar with computer science terms like this (like myself).

Reference: https://cs.stackexchange.com/q/67417

Relequestual:
Fixes #735

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -211,12 +211,12 @@
                         depending on the type:
 
                         <list style="hanging">
-                            <t hangText="null:">A JSON "null" production</t>
-                            <t hangText="boolean:">A "true" or "false" value, from the JSON "true" or "false" productions</t>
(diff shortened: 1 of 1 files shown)


awwright:
This is fine by me; but is it possible that "rule" is a better term to use?

Relequestual:
@awwright I can't see how it would fit. Could you give me an example please in case I'm missing something? =]

awwright:
I'm looking at the ABNF RFC and its preferred term seems to be "rule"
I think "value" is clear enough too, though.

PR #845, opened 2020-02-03
$ref across schema versions clarification

### LINKED ISSUE #808: $ref (and kin) across schema versions
The spec doesn't define (and also doesn't mention that the behavior is undefined) how the reference keywords (e.g. `$ref`) work when referencing into schemas declaring other spec versions.

### Example 1

I define schema `http://my.com/schema` as a draft 7 schema, but I need to reference schema `http://that.one.guy.com/schema` for one of my properties.  Happily, that guy's schema is also defined at draft 7.

One day in the not-to-distant future, that guy upgrades their schema to draft 2019-09.  Now my draft 7 schema (without being altered) references a draft 2019-09 schema.

### Example 2

I define schema `http://my.com/schema` as a draft 7 schema, but I need to reference schema `http://that.one.guy.com/schema` for one of my properties.  Happily, that guy's schema is also defined at draft 7.

One day in the not-to-distant future, I upgrade my schema to draft 2019-09, but it still references a draft 7 schema, which I don't control.

---

How is the validator supposed to handle this?

Ideally, I'd think the validator would take each schema document separately, and (if it supports the draft declared by the document, otherwise error) it would evaluate the reference under the draft it declares.  However, I can see where a validator would get this wrong and evaluate the entire reference tree based on the draft declared by the root schema.

This should be either defined or explicitly stated that it's undefined.

Relequestual:
Note that referenced schemas which identify as needing to be processed differently MAY be supported, or an error SHOULD be thrown

Fixes #808

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1767,6 +1767,10 @@
                             URI to identify more than one schema. When multiple schemas try to identify
                             as the same URI, validators SHOULD raise an error condition.
                         </t>
+                        <t>
+                            As with all schemas, a referenced schema SHOULD be processed according to its define "$schema"
+                            and "$vocabulary" values, if defined.
+                        </t>
                     </section>
                     <section title="Dereferencing">
                         <t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1768,7 +1768,7 @@
                             as the same URI, validators SHOULD raise an error condition.
                         </t>
                         <t>
-                            As with all schemas, a referenced schema SHOULD be processed according to its define "$schema"
+                            As with all schemas, a referenced schema SHOULD be processed according to its "$schema"
                             and "$vocabulary" values, if defined.
                         </t>
                     </section>


jsonschema-core.xml
```diff
@@ -1767,6 +1767,10 @@
                             URI to identify more than one schema. When multiple schemas try to identify
                             as the same URI, validators SHOULD raise an error condition.
                         </t>
+                        <t>
+                            As with all schemas, a referenced schema SHOULD be processed according to its define "$schema"
```
handrews: I feel like this ought (OUGHT?) to be MUST, except that I don't think we ever actually say that `$schema` MUST be respected at all?  But we do say that `$vocabulary` MUST be respected at least to the point of the implementation failing in the face of required but unsupported vocabularies.  So that sort-of implies a MUST for respecting `$schema`.  Maybe?

-----

Looking at this now, it occurs to me that how a schema is loaded and processed is a very fundamental thing that deserves more emphasis than we are giving it.  Loading a _referenced_ schema is an important use case, but generally loading schemas is the same whether you found the schema directly, through `$ref`, or through hypermedia.  As @awwright noted in #808, `$ref` is not really a special case for this.

I think we should have a section about loading schemas that covers the general load-and-process model, and notes any specifics needed for `$ref` and for hypermedia.  My initial thought was to put this early in the spec (general considerations?) but you really need to understand the core vocabulary to do all of the steps.

So I say we should add a new top-level section between section 8 (Core Vocabulary) and section 9 (Applicator Vocabulary) on schema loading.  It should include most or all of:

8.2.4.5 Loading a referenced schema _(but just Loading a schema)_
8.2.4.6 Dereferencing _(maybe tweak to cover other ways you get a schema other than just `$ref`)_
8.2.4.3 Guarding against infinite recursion
8.2.4.4 References to possible non-schemas
8.1.4 Detecting a meta-schema
11 Usage in hypermedia _(demote to subsection?)_

While this sounds like a lot, it's mostly just moving stuff around.

I've gotten several questions that basically boil down to "How do I load schemas?"  Consolidating this information would make an essential process much more clear and emphasize its importance.  It would also trim the extremely long Core Vocabulary section and avoid some of the really deep nesting that causes key sections to not show up in the Table of Contents right now.

Thoughts?

handrews: (maybe I should have put all that in the issue, can move if we'd prefer)

gregsdennis: Not to diminish @handrews' excellent comments, but maybe remove the initial "define" as it reads somewhat redundant with the later "if defined."

Relequestual: Sounds reasonable. As such, want to create a new issue for this?

handrews: OK- yeah I'm happy with you merging this after taking into account @gregsdennis 's comment, and I'll file the above so that we can consider it separately.

Relequestual: I'd probably go as far as to say "Implementations MAY throw an error if a $schema is not provided"... but hey =]

Relequestual: I'll hold this as unresolved till I see a new issue.

Relequestual: Yeah makes sense

handrews: @Relequestual no, it's not fair- there is no requirement to provide `$schema`.  A linter can raise an issue, but an implementation?  Definitely not.

handrews: JSON Schema implementations should generally be forgiving.  If the spec specifically allows something (and saying that `$schema` SHOULD be present is more or less equivalent to saying that it MAY be absent) then all implementations should always accept that thing even if we are clearly indicating that doing so is a bad idea.

Things that are bad ideas but neither outright forbidden (requiring an error) nor explicitly have undefined or implementation-defined behavior (allowing an error as that is one possible behavior) need to be allowed.  Not warnings, not errors, allowed.  A linter can do "warnings" for those things as a separate piece.

Relequestual: Apologies, it wasn't clear enough that suggesting implementers throw an error on lack of `$schema` was not a serious suggestion.

handrews: @Relequestual ah, I see.  I usually read these right after waking up in hopes of catching you and Phil at the end of your days, so subtleties are lost on me.

handrews:

Issue #848, opened 2020-02-05
Enumerate key architectural principles


handrews:
There are some underlying principles that the JSON Schema project has been based on, but we've never really written them out.  A concise list would be appropriate in the Core spec, as they might head off a few questions.

They also provide some general guidance on how to design keywords for extension vocabularies.

* JSON Schema works with all possible instance resources that can be parsed into its data mode, including all syntactically legal JSON documents with well-defined behavior per RFC 8259†.  No aspect of the instance data's structure or contents is constrained.
* It is possible (and RECOMMENDED?) to resolve all URI references (e.g. `$id`, `$ref`, and `$recursiveRef`) and locate all defined fragments (`$anchor`)  used in JSON Schema functionality just from the schemas, without an instance. 
* The result of applying a schema object to an instance is a function only of the schema object and its dynamic-scope subschemas.  Assuming URI references were already resolved, it is never necessary to know the context (parent schema) of the schema object.
* Keywords SHOULD be independent, but MAY be defined in terms of the assertion and/or annotation results of other keywords within the same schema object.
* Keywords behavior fits into one or more classifications, but MUST have the same classification(s) regardless of its value.††

†Note that a JSON object with duplicate property names is syntactically legal but does **not** have well-defined behavior per RFC 8259.  There are also limits on floating point interoperability.

††this is why we split `dependencies` which changed classification depending on whether it was an object of schemas or an object of string arrays.

_Are there any other such principles?_  I don't think I want to include general principles like "least power", just the ones that couldn't be found anywhere else but within JSON Schema.

gregsdennis:
I think something about it being a constraint-based system, and as a result, concepts like inheritance do not fit well.

Relequestual:
I feel like we should say something about Meta-Data Annotations and the semantic meaning of such keywords, but I'm not sure what... I know that's not directly helpful but it might spark some other thoughts.

handrews:
Everything in JSON Schema is addressed using URI-references (including full URIs) ***or*** Relative JSON Pointer.  Those are the two things that we use today (URIs everywhere, Relative JSON pointer in Hyper-Schema).

This principle would have flushed out problems with OpenAPI's `discriminator` keyword, which addresses schemas by their name in the `definitions` (OAS2) or `components` (OAS3) structure.

handrews:
[...]

PR #856, opened 2020-02-11
First pass a list of principles.

handrews:
These inform how JSON Schema is designed as a system.

The section is placed after Definitions as otherwise the
principles don't have clear meaning.

This is a first pass at #848.

I definitely want real feedback here.  This is a long-ish list.  What should be cut?  Can the wording be condensed?  Does this feel like it serves a real purpose?

(Using the new "draft pull request" feature to ensure that it doesn't get merged too quickly :-)

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -455,6 +455,68 @@
 
         </section>
 
+        <section title="Principles">
+            <t>
+                JSON Schema is built on several principles, which both explain various design
+                choices, and provide guidance on how to extend the specification effectively.
+            </t>
+            <!-- Links will be added to relevant spec parts after the sections are reorganized
+                 to create the new schema loading section -->
+            <t>
+                <list>
+                    <t>
+                        All syntactically legal JSON documents with well-defined behavior are fully
+                        supported as instance data.  Notably, JSON objects with duplicate property
+                        names are syntactically legal but not well-defined.
+                    </t>
+                    <t>
+                        JSON Schema is defined over a data model, not JSON text.  Therefore
+                        JSON Schema can be used with any data format that can be reliably parsed
+                        into the data model.
+                    </t>
+                    <t>
+                        Locations in schemas and instances are always identified by URIs,
+                        JSON Pointers, or Relative JSON Pointers.  JSON Schema treats URIs as
+                        identifiers only.  Locating and retrieving URI-identified resources
+                        is outside of the scope of this specification.
+                    </t>
+                    <t>
+                        Schema URIs referenced or defined in JSON Schema's Core vocabulary MUST be
+                        resolvable without instance data.
+                    </t>
+                    <t>
+                        Once schema URIs are resolved, the result of applying a schema object
+                        to instance data is data is a function of that schema object and its
+                        dynamic-scope subschemas.  Parent schemas have no effect on results.
+                    </t>
+                    <t>
+                        Each keyword's behavior falls into one or more behavioral classifications.
+                        A keyword's classification MUST be the same for all values of the keyword.
+                    </t>
+                    <t>
+                        Keywords produce a boolean assertion result, and an annotation result that
+                        can be of any type.  Depending on the classification, the assertion result
+                        might always be true, or the annotation result might always be empty.
+                    </t>
+                    <t>
+                        Keyword results MAY be defined in terms of the results of other keywords
+                        in the same schema object, as long as no cyclic dependencies are produced.
+                    </t>
+                    <t>
+                        JSON Schema assertions form a constraint system.  The empty schema,
+                        <spanx style="verb">{}</spanx>, allows everything and constrains nothing.
+                        Each assertion keyword adds constraints; keywords MUST NOT remove constraints.
+                    </t>
+                    <t>
+                        JSON Schema annotations provide a flexible way to convey information for
+                        an application to use.  The information can be used with instance data,
+                        or can be used directly from schemas to generate systems for use with
+                        conforming instances.
+                    </t>
+                </list>
+            </t>
+        </section>
+
         <section title="Fragment Identifiers" anchor="fragments">
             <t>
                 In accordance with section 3.1 of <xref target="RFC6839">RFC 6839</xref>,
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -486,7 +486,7 @@
                     </t>
                     <t>
                         Once schema URIs are resolved, the result of applying a schema object
-                        to instance data is data is a function of that schema object and its
+                        to instance data is a function of that schema object and its
                         dynamic-scope subschemas.  Parent schemas have no effect on results.
                     </t>
                     <t>


jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                        identifiers only.  Locating and retrieving URI-identified resources
+                        is outside of the scope of this specification.
+                    </t>
+                    <t>
+                        Schema URIs referenced or defined in JSON Schema's Core vocabulary MUST be
+                        resolvable without instance data.
+                    </t>
+                    <t>
+                        Once schema URIs are resolved, the result of applying a schema object
+                        to instance data is data is a function of that schema object and its
```
Relequestual: "data is data is"

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                        dynamic-scope subschemas.  Parent schemas have no effect on results.
```
Relequestual: Does "dynamic-scope schemas" cover the recursive anchor feature?

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                    </t>
+                    <t>
+                        Each keyword's behavior falls into one or more behavioral classifications.
+                        A keyword's classification MUST be the same for all values of the keyword.
+                    </t>
+                    <t>
+                        Keywords produce a boolean assertion result, and an annotation result that
+                        can be of any type.  Depending on the classification, the assertion result
+                        might always be true, or the annotation result might always be empty.
```
Relequestual: Annotations are keywords and they do not produce any assertion result, right?

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                <list>
+                    <t>
+                        All syntactically legal JSON documents with well-defined behavior are fully
+                        supported as instance data.  Notably, JSON objects with duplicate property
+                        names are syntactically legal but not well-defined.
+                    </t>
+                    <t>
+                        JSON Schema is defined over a data model, not JSON text.  Therefore
+                        JSON Schema can be used with any data format that can be reliably parsed
+                        into the data model.
```
gregsdennis: I'm guessing that this opens up the possibility for JSON Schema to be represented in YAML?

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                    </t>
+                    <t>
+                        Locations in schemas and instances are always identified by URIs,
+                        JSON Pointers, or Relative JSON Pointers.  JSON Schema treats URIs as
+                        identifiers only.  Locating and retrieving URI-identified resources
+                        is outside of the scope of this specification.
```
gregsdennis: Locating and retrieving is out of scope, but loading is in scope?  This is a fine line and maybe we should define "locating", "retrieving", and "loading".

jsonschema-core.xml
gregsdennis: Is it still a subschema if it's external but referenced?

jsonschema-core.xml
gregsdennis: This defines that *all* keywords produce both, but "annotation" keywords always produce a `true` assertion, making the assertion result moot.

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                        A keyword's classification MUST be the same for all values of the keyword.
+                    </t>
+                    <t>
+                        Keywords produce a boolean assertion result, and an annotation result that
+                        can be of any type.  Depending on the classification, the assertion result
+                        might always be true, or the annotation result might always be empty.
+                    </t>
+                    <t>
+                        Keyword results MAY be defined in terms of the results of other keywords
+                        in the same schema object, as long as no cyclic dependencies are produced.
```
gregsdennis: "in the same schema object or its subschemas"?  Maybe explicitly call out that subschema results can be acted on?

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                    </t>
+                    <t>
+                        JSON Schema assertions form a constraint system.  The empty schema,
+                        <spanx style="verb">{}</spanx>, allows everything and constrains nothing.
+                        Each assertion keyword adds constraints; keywords MUST NOT remove constraints.
```
gregsdennis: "MUST NOT remove *or alter* contraints."

Also, I still think that we should mention that this does not conform to inheritance models such as polymorphism.

jsonschema-core.xml
handrews: yes, and yes.  "Dynamic Scope" is explained in the definitions section above the principles section, so people will have seen that term if they're reading top to bottom.  Can also xref it- I figure this section should have a number of xrefs but I'll put them in when we have the basic contents agreed to.

jsonschema-core.xml
handrews: @gregsdennis that is correct.  Might need to also clarify what "empty" annotation results mean.

I put this in because in terms of generic extensions, just always returning the same sorts of things makes sense instead of "wait, what kind of keyword was this?  does it return this kind of thing? yes? no?"  If not otherwise specified, keywords return a `true` validation result.

This is open to debate, it just made the most sense as I was typing it.

(this is exactly why laying out principles is a good idea, though! I don't think we've ever really discussed this)

jsonschema-core.xml
handrews: It's always been a possibility.  The "data  model not txt" language is already present in the spec- it has its own section, even.   I think alternate representations are even explicitly called out (although maybe with CBOR rather than YAML, but I think a non-JSON case gets mentioned somewhere).  Although maybe only for instances and not schemas, but schemas are also instance so... 🤷‍♂

jsonschema-core.xml
handrews: Good point.  Will add to definitions section.

jsonschema-core.xml
handrews: The results of a keyword include the results of any subschemas.  I'd rather keep these things orthogonal.  If we need to have a principle about subschema result roll-up we can?  Not sure if that's needed?

jsonschema-core.xml
handrews: What sort of "alter" do you have in mind?  99% sure I agree, just want to be clear if something has come up in practice.

And yeah, probably worth a comment.

gregsdennis: I think that's me drifting off into an inheritance mindset where you can override members.

The case I can think of is redefining something like `maximum` alongside a `$ref` to something that has `maximum` defined within.  But in that case, the lower one would win out because the instance would just have to satisfy both `maximum`s.  Similar for property definitions: the instance would have to validate against both definitions.

This is fine, but it doesn't do what people thinking about inheritance would think it should do.

jsonschema-core.xml
gregsdennis: I don't think it's a design principle, more of a consequence of processing.

jsonschema-core.xml
handrews: Hmm.... yeah, that's an important point, I'll see if I can come up with concise wording.

jsonschema-core.xml
Relequestual: Heck, I sometimes write my schemas in YAML and pre-build.

jsonschema-core.xml
Relequestual: The trouble is, we DO have keywords which alter the behaviour of other keywords, so I don't think you can say that.

handrews: @Relequestual There's a wording that will work in here somewhere, but that's part of why I questioned "alter" as a word choice.  Just to make sure we're all on the same page about the issue, I think the simplest version of this is:

```JSON
{
    "type": "object",
    "additionalProperties": {"type": "integer"}
}
```
vs
```JSON
{
    "type": "object",
    "properties": {"foo": true},
    "additionalProperties": {"type": "integer"}
}
```
where the `properties` keyword technically adds a constraint that is specific to `foo`, but because that carves `foo` out of `additionalProperties`, from the point of view of that instance property the constraints are _changed_, and given that the schema used for `foo` is `true`, the effect is to remove the `"type": "integer"` constraint.

handrews: There might be something here about applicators vs assertions.  You can use applicators to change which subschema(s) _apply_ to a location in the instance.  But adding assertions to a given schema object always further constrains.  The only way to effectively lift a constraint is to use applicators to change which schema object is used.

gregsdennis: > but because that carves `foo` out of `additionalProperties`, from the point of view of that instance property the constraints are *changed*...

I don't consider this as `additionalProperties` changing its behavior.  `additionalProperties` is *defined* to depend on the existence of `properties`, so the behavior of seeing `foo` only when it's not handled by `properties` is consistent with that definition.

Maybe "override" would be a better word here?

handrews: I intentionally didn't say (or at least didn't mean to say) that `additionalProperties` changed its behavior.  Rather, from the black box validation point of view, a constraint appeared to be lifted when a keyword was added.  In a proper constraint layering system, the set of valid instances after adding any keyword would be a subset of the set of valid instances before adding the keyword.

However, this instance: `{"foo": "hello", "bar": 2}` fails validation against my first example schema, because `"hello"` is not an integer.  But it _passes_ validation against my second example schema, because _adding_ `properties` caused a different subschema (`true`) to apply to `"hello"`.

I'm 99% certain that if we just look at assertions, adding an assertion always reduces the set of valid instances (ignoring complex situations like `"not"` where adding to its subschema reduces that subschema's valid set, which expands the inverted set accepted by `"not"`.  Hmm... that's an applicator behavior, too.  Not the "which subschema applies to what" part but the "perform boolean logic on the subschema results to produce the applicator result."

I think it is always true that adding a pure **assertion** keyword (that is not also an applicator and/or annotation) to a schema object always reduces the set of valid instances according to that schema object (including its subschemas, because subschemas are never affected by parent schemas once you resolve URI references- that's one of the other principles).

Applicators allow much more complex behavior, but it's still just changing how the subschemas **results** are combined.  This is why there is never any override or merge or replacement.  The operations are performed on subschema results, never on individual keywords.  Aside from information returned in annotation results, subschemas are black boxes; the parent schema has no idea what keywords they contain.

jsonschema-core.xml
Relequestual: Cool. Fine. Just checking.

jsonschema-core.xml
Relequestual: OK. Fair.

Relequestual: Looking at the document, I can't see this defined explicitly. Are you saying it's implicit with the above wording? If so, could we ALSO make it explicit with requirement keywords? Tests will then have a requirements keyword to point to in the spec then.

jsonschema-core.xml
Relequestual: Could the issue here be it's "assertions and annotations must not remove constraints" but applicator keywords will alter how constraints are applied?

`properties` and `additionalProperties` are both applicators.
An argument could be made that when using `if/then/else`, the instance data actually modifies how the constraints are applied also. BUT, they are not removed, just applied differently.

Maybe it's worth noting that while constraints cannot be removed, the application of the schema they belong to may be altered as a side effect of the annotation results of other keywords?

jsonschema-core.xml
handrews: @Relequestual I think we should make it explicit somewhere else.  I specifically don't want any requirement keywords in the Principles section:  They're principles, not requirements.  They should drive the requirements.  I will make a separate PR for clarifying the nature of results.

handrews: (I'm taking the MAY out of one of the other principles and changing it to "can")

jsonschema-core.xml
handrews: I don't think there's anything to do here?  I'm going to mark this as resolved.

jsonschema-core.xml
handrews: @Relequestual yeah what you're saying here is getting very close to it, I think.

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                    </t>
+                    <t>
+                        Locations in schemas and instances are always identified by URIs,
+                        JSON Pointers, or Relative JSON Pointers.  JSON Schema treats URIs as
+                        identifiers only.  Locating and retrieving URI-identified resources
+                        is outside of the scope of this specification.
+                    </t>
+                    <t>
+                        Schema URIs referenced or defined in JSON Schema's Core vocabulary MUST be
+                        resolvable without instance data.
```
karenetheridge, line 485: What does "resolvable without instance data" mean?

handrews: Pretty much what it literally says- don't overthink it ;-)

A URI reference in `$ref` or `$id` is resolved against the base URI, which is the nearest parent `$id`.  If there isn't one, the schema resource's retrieval URI is used (per RFC 3986 generic rules for establishing base URIs).

What this point was trying to establish is that you can safely pre-process a schema so that all of those URI references are converted to full URIs (starting with a scheme) before you're given any instance data.  This makes applying the schema to any instance faster and simpler, because you don't have to keep track of base URIs and resolve references when you encounter them.

--------

However, this point doesn't quite work in 2019-09: the URI in `$recursiveRef` can't, based on the implementation mechanism described, be resolved like this, because what `$recursiveAnchor` does is replace the original base URI (from `$id` or the retrieval URI) with the base URI of the dynamic-scope outermost schema object containing `"$recursiveAnchor": true`, if any such schema object exists in the dynamic scope.

We could actually change the internal mechanism for this to be simply "replace the entire resolved reference from `$recursiveRef` with the URI of the dynamic-scope outermost etc.", because right now the only legal value of `$recursiveRef` is `"#"`, and both of these mechanisms produce the same result with that URI reference.

If we were to lift that restriction and allow any URI reference as a `$recursiveRef` value, then the two mechanisms produce different results, and require keeping track of base URIs at runtime.

See #868 for more details and hopefully eventually a decision on whether to require static (without instance) resolution for all keywords and a change to `$recursive*`, or whether to change this architectural principle.

jsonschema-core.xml
```diff
@@ -455,6 +455,68 @@
...
+                        names are syntactically legal but not well-defined.
+                    </t>
+                    <t>
+                        JSON Schema is defined over a data model, not JSON text.  Therefore
+                        JSON Schema can be used with any data format that can be reliably parsed
+                        into the data model.
+                    </t>
+                    <t>
+                        Locations in schemas and instances are always identified by URIs,
+                        JSON Pointers, or Relative JSON Pointers.  JSON Schema treats URIs as
```
karenetheridge, line 479: Sorry, I only just noticed this now. Are relative json pointers now fair game to be used inside $refs? I think this is the first time that's been mentioned explicitly in the spec.  (The metaschema itself just uses 'format: uri-reference' for the $ref syntax, and there are no examples of relative JSON pointers used.)  It would be good to clarify this explicitly (probably in a separate patch).

handrews: @karenetheridge ignore this PR, it's way outdated.  I should really just take it down, I may or may not revisit it, or this info may go on the web site, or something id.

Issue #849, opened 2020-02-05
Consolidate information on loading schemas


handrews:
It occurs to me that how a schema is loaded and processed is a very fundamental thing that deserves more emphasis than we are giving it.

Loading a referenced schema is an important use case, but generally loading schemas is the same whether you found the schema directly, through `$ref`, or through hypermedia. As @awwright noted in #808, `$ref` is not really a special case for this.  Additionally, encountering a separate schema resource as a subschema is really another case of loading schemas.  We should focus around the resource concept more than the document concept.

Questions around this have been common.  The question of indicating a schema from within an instance somehow is one.  The question of setting a base URI for a schema read from a filesystem or passed in with no context is another.  And then there are questions around `$ref`s among different schema versions (#808) which is related to the question of how to treat embedded schema resources in a single document.

I think we should have a section about loading schemas that covers the general load-and-process model, and notes any specifics needed for `$ref`, embedded resources, hypermedia, or other cases. My initial thought was to put this early in the spec (general considerations?) but you really need to understand the core vocabulary to do all of the steps.

So I say we should add a new top-level section between section 8 (Core Vocabulary) and section 9 (Applicator Vocabulary) on schema loading. It should include most or all of:

8.2.4.5 Loading a referenced schema (but just Loading a schema)
8.2.4.6 Dereferencing (maybe tweak to cover other ways you get a schema other than just `$ref`)
8.2.4.3 Guarding against infinite recursion
8.2.4.4 References to possible non-schemas
8.1.4 Detecting a meta-schema
(new section) Connecting Instances to Schemas
11 Usage in hypermedia (as subsection)
(new subsection) Usage outside of hypermedia (needs better title)

While this sounds like a lot, it's mostly just moving stuff around.

I've gotten several questions that basically boil down to "How do I load schemas?" Consolidating this information would make an essential process much more clear and emphasize its importance. It would also trim the extremely long Core Vocabulary section and avoid some of the really deep nesting that causes key sections to not show up in the Table of Contents right now.

Relequestual:
This all sounds reasonable.
What sort of information do you want to convey in section "Usage outside of hypermedia"?

handrews:
@Relequestual here's the sort of question that I hope people won't need to ask as much: https://github.com/RicoSuter/NJsonSchema/issues/1090

[...]

PR #857, opened 2020-02-11
Consolidate schema loading sections

handrews:
We are adding a section dealing with loading, processing,
and referencing schemas.

These sections have been moved without any changes to text or
formatting beyond adjusting the indentation to fit the new
location.

This is the first step for #849.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1306,30 +1306,6 @@
                         as those listed here.
                     </t>
                 </section>
-                <section title="Detecting a Meta-Schema">
-                    <t>
-                        Implementations MUST recognize a schema as a meta-schema if it
-                        is being examined because it was identified as such by another
-                        schema's "$schema" keyword.  This means that a single schema
-                        document might sometimes be considered a regular schema, and
-                        other times be considered a meta-schema.
-                    </t>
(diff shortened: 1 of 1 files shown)


handrews:
@gregsdennis @johandorland @Julian tagging y'all in as implementors to see if this direction makes sense.  See #849 for context- this PR is just moving blocks around unchanged so if possible I'd like to get it merged quickly so I can put up actual real changes for longer review.

Julian:
The outline in #849 certainly seems to make sense to me if I understand it (which there's always a chance I don't when dereferencing is involved). So +1 on this one I think?

gregsdennis:
These seem like the appropriate sections.  Lift-and-shift doesn't concern me so much as the changes that'll come later to make it read a bit more fluently (to suggest that these parts are _supposed_ to be together).

Relequestual:
We shouldn't worry about xml2rfc complaining about figure length as the resuling text document does the right thing when wrapping.

handrews:
@Relequestual yeah the length warnings are not new.

@Julian I ended up including the part on dereferencing in the new section because the question of how to resolve references is closely related to the question of what it means to "load" a schema so that you have it available to reference.

PR #879, opened 2020-03-18
Fixing Wording in Annotation Section

vanou:
This commit fixes wording in Annotation section in json-schema-core.

**(Before)**
Unless otherwise specified, **the value of an annotation keyword's annotation** is the keyword's value. However, other behaviors are possible.

**(After)**
Unless otherwise specified, *the value of an annotation keyword* is the keyword's value. However, other behaviors are possible.


Relequestual:
Thanks!

Issue #815, opened 2019-10-20
This url - https://json-schema.org/draft/2019-09/vocab/core doesn't work


opethe1st:
This url in the latest draft as defined here https://json-schema.org/draft/2019-09/schema. You can go to the url and check that this is one of the urls under "$vocabulary"

opethe1st:
I could work on a PR if that works too.

notEthan:
this URL is just an identifier which does not correspond to a document. it sort of corresponds to the meta document at https://json-schema.org/draft/2019-09/meta/core (referenced in the allOf property of the metaschema), though I can't quite explain the relationship.

I was confused by this for a while too. I think it needs improved clarification in the spec.

opethe1st:
I know from the spec it doesn't have to correspond to a retrievable document but this particular case looked like an oversight to me. 
Plus it is actually better develop UX to be able to click in an editor on the url and get the document being referenced. 

RIght now, I would have to jump through hoops to find the validation schema.

handrews:
> I know from the spec it doesn't have to correspond to a retrievable document but this particular case looked like an oversight to me.

It doesn't correspond to a retrievable document because we don't know what to put in the document until people try to implement the vocabulary feature and tell us what should go there.

Some of the inconclusive discussions involved can be found at #602 and at https://github.com/json-schema-org/json-schema-spec/issues/563#issuecomment-438869464

PR #889, opened 2020-04-08
use tag uris for vocabularies

notEthan:
this follows #874 (which is included in the diff). 

instead of using a transport protocol for vocabularies, which have no document to transport, why not use a uri that is simply an identifier? the tag URI seems like it was built for this. (I think most of the json schema members are familiar, but a tag URI is described at [taguri.org](https://www.taguri.org/), [wikipedia](https://en.wikipedia.org/wiki/Tag_URI), [its RFC](https://tools.ietf.org/html/rfc4151).)

this would help alleviate confusion on the part of people looking for retrievable resources at vocabulary URIs (#815).

### DIFF
--- a/hyper-schema.json
+++ b/hyper-schema.json
@@ -2,13 +2,13 @@
     "$schema": "https://json-schema.org/draft/2019-09/hyper-schema",
     "$id": "https://json-schema.org/draft/2019-09/hyper-schema",
     "$vocabulary": {
-        "https://json-schema.org/draft/2019-09/vocab/core": true,
-        "https://json-schema.org/draft/2019-09/vocab/applicator": true,
-        "https://json-schema.org/draft/2019-09/vocab/validation": true,
-        "https://json-schema.org/draft/2019-09/vocab/meta-data": true,
-        "https://json-schema.org/draft/2019-09/vocab/format": false,
-        "https://json-schema.org/draft/2019-09/vocab/content": true,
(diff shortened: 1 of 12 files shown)


Relequestual:
YIKES, this totally a non trivial change.
Closing this down till we / others can discuss in an issue you might want to create to see if implementers are for this idea.
It's certinly not something we're going to do in the short term future.

notEthan:
uh ... what's wrong with this issue? do I need to open one without a diff attached for some reason? I didn't expect this to just be merged, I opened it for discussion. obviously it's not trivial.

Relequestual:
This isn't an issue, it's a pull request. A pull request DOES signify "I want to merge this". It can even have a linked issue.

notEthan:
github treats pull request as an issue with commits attached (that's probably oversimplifying, but I do know you can transform an issue into a pull request by attaching a branch to it). but that's not an argument I'm here to have. I will open an issue without the attached commit.

karenetheridge:
Github has *just* announced that it's now possible to convert a pull request to 'draft' status, which prevents it from being merged, and is useful for posting code for review/discussion: https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/

ssilverman:
It's often the case that "http" and "https" schemes are used for URIs, even though not retrievable. Yes, they look like URLs, but it's a common practice.

PR #916, opened 2020-05-09
Fix $anchor plain name format to match XML's NCName production

ssilverman:
This uses the US-ASCII part of the production.
See: https://www.w3.org/TR/2006/REC-xml-names11-20060816/#NT-NCName

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1391,9 +1391,11 @@
                         identifier keyword that can only be used to create plain name fragments.
                     </t>
                     <t>
-                        If present, the value of this keyword MUST be a string, which MUST start with
-                        a letter ([A-Za-z]), followed by any number of letters, digits ([0-9]),
-                        hyphens ("-"), underscores ("_"), colons (":"), or periods (".").
+                        If present, the value of this keyword MUST be a string and MUST start with
+                        a letter ([A-Za-z]) or underscore ("_"), followed by any number of letters,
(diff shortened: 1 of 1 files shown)


handrews:
@awwright I think you're the only person who would remember why the existing rules were done the way they were.  If you are OK with this, I'm OK with this.

In section 5: Fragments, we reference the [W3C best practices](https://www.w3.org/TR/2012/WD-fragid-best-practices-20121025/) document, which at one point talks about the [NCName production](https://www.w3.org/TR/xml-names11/#NT-NCName) in the context of plain name fragments.

Whatever we do, section 5 and the section for `$anchor` need to agree.

ssilverman:
Just to be certain, the formatting and use of `eref` is okay? I'm thinking of instead wrapping that around some text. Do you guys have a style preference?

handrews:
@ssilverman I'd make an `<xref...>NCName production</xref>` for better readability.  I can't remember if that's already declared as an external document at the top or and bottom, but if not there are other W3C specs referenced- you'll figure it out :-) 

You can't include an anchor in that reference, but "NCName production" will be enough for anyone to locate it within the spec.

ssilverman:
Updated to xref.

handrews:
@ssilverman you're getting a build error through Travis.  I think you need to add a corresponding `<!ENTITY` declaration at the top.  xrefs to external specifications require both that and an entry under `references`.

ssilverman:
I added a missing "date" element; that seems to be what the DTD wanted.

PR #917, opened 2020-05-10
contentSchema: type MUST be a schema

ssilverman:


handrews:
@ssilverman this should actually just say at the top of the `contentSchema` section that its value MUST be a schema, instead of saying that it contains a schema if blah blah blah.

The type of the keyword value is unaffected by the presence or absence of `contentMediaType`.  The reason `contentSchema` is ignored without `contentMediaType` is that there's no way to know how to apply it if you can't figure out the media type.  But that's independent of the value type.

Good catch noticing that I left the MUST out ,though.

ssilverman:
I'm still thinking about it...

ssilverman:
The reason that I chose this language is because the first paragraph implies that the property is a schema _if_ the instance is a string and "contentMediaType" is present. This seems to imply that the value may or may not be a schema if those two conditions aren't satisfied.

This also means that if "contentMediaType" is not present, then we don't really know what "contentSchema" can be. Maybe, then, all three paragraphs need to change.

ssilverman:
Ok, I tried a re-word. Let me know what you think.

handrews:
@ssilverman 

> the first paragraph implies that the property is a schema if the instance is a string and "contentMediaType" is present. This seems to imply that the value may or may not be a schema if those two conditions aren't satisfied.

While I understand why you're thinking about that, I think you're overthinking it 😄 

No keywords change their data type requirements based on the instance.  They may be ignored, but their syntax absolutely cannot be determined by the instance, or implementations couldn't validate and process schemas up front.

jsonschema-validation.xml
```diff
@@ -1015,15 +1015,17 @@
 
             <section title="contentSchema">
                 <t>
-                    If the instance is a string, and if "contentMediaType" is present, this
-                    property contains a schema which describes the structure of the string.
+                    If the instance is a string containing valid JSON, and if "contentMediaType" is
```
handrews, line 1018: There is no need for the instance to be JSON.  `contentMediaType` specifically either forbids or recommends against (I forget) attempting to parse the instance according to the media type.

jsonschema-validation.xml
```diff
@@ -1015,15 +1015,17 @@
...
+                    present, this property is used to validate the string. If the string is not
+                    valid JSON, then the validation fails.
                 </t>
                 <t>
                     This keyword MAY be used with any media type that can be mapped into
                     JSON Schema's data model.
                 </t>
                 <t>
-                    The value of this property SHOULD be ignored if "contentMediaType" is not present.
+                    The value of this property MUST be a schema. It is ignored if "contentMediaType"
```
handrews, line 1027: You should (SHOULD 😛 ) keep the "SHOULD be ignored" wording.  

Another minor consistency nitpick:  Throughout the core spec, the phrasing is "MUST be a valid JSON Schema" (which shows you how far off I was when I wrote this section!)

jsonschema-validation.xml
ssilverman: This is a point that's always confused me. If the instance isn't JSON, how can the contents of "contentSchema", a JSON schema, validate it?

jsonschema-validation.xml
ssilverman: Done both.

jsonschema-validation.xml
handrews: @ssilverman JSON Schema is not defined over JSON, it is defined over a data model that derives from (but is not precisely identical to) the JSON data model.  The Core spec has a whole section on the data model, and notes that JSON Schema can be used with any media type that can be mapped into that data model.  Defining such mappings is outside the scope of this specification.  A lot of them are trivial at least in typical usage: YAML, for example, as long as you don't use the advanced features that most people don't know exist.

So the instance media type may not have been JSON (or at least not plain `application/json`) in the first place, and the content in the string might be that same other media type.

JSON Schema is very careful not to put restrictions on the `content*` keywords regarding media types.  This is also why implementations must not, by default, try to parse the embedded content and apply the schema.  That's an open-ended burden, and parsing and loading an arbitrary media type is a major security concern.

karenetheridge: It sounds like the change at this position should be reverted entirely? is there anything else to discuss there?

handrews: @karenetheridge I think

> this property contains a schema

is the language that @ssilverman originally found confusing.  Because it almost sounds like a syntactic requirement but is not.  Since he fixed the syntactic requirement further down, maybe there's a better wording that doesn't cause that confusion?  Or we can just revert it and assume that the last paragraph covers the formalities.

karenetheridge: How about:

> This keyword only applies to instance data of the "string" type. If "contentMediaType" is present, this property contains a schema which describes the structure of the string.

ssilverman: We shouldn't revert this entirely, because it still needs to add the "MUST". I've force pushed that one change.

ssilverman:
Just force pushed changes with this message: "contentSchema: type MUST be a schema and instance MUST follow the data model".

jsonschema-validation.xml
```diff
@@ -1017,13 +1017,16 @@
                 <t>
                     If the instance is a string, and if "contentMediaType" is present, this
                     property contains a schema which describes the structure of the string.
+                    The string MUST be in a format compatible with the data model defined in the
+                    <xref target="json-schema">core JSON Schema.</xref>
```
karenetheridge, line 1021: I stared at this for a minute wondering what this was trying to say. Perhaps this addition would help? `"; that is, a valid JSON string."`

jsonschema-validation.xml
```diff
@@ -1176,9 +1179,9 @@
                     the entire resource being described MAY be removed in the future.
                 </t>
                 <t>
-                    When the "deprecated" keyword is applied to an item in an array by means of
```
karenetheridge, line 1179: I think handrews is taking care of whitespace changes already in other cleanup commits.

jsonschema-validation.xml
ssilverman: That’s what I originally thought at first too. That’s incorrect, however. I refer you to the previous resolved comments in this PR.

ssilverman: See: https://github.com/json-schema-org/json-schema-spec/pull/917#discussion_r428091385
(You even replied to it.)

jsonschema-validation.xml
ssilverman: Fixed.

handrews: Yeah I usually do a formatting-only commit somewhere for this sort of thing.  I'm fine with tacking things you find onto other PRs although they should be done as separate commits.

We should probably write guidance like this down somewhere, but "don't mix formatting changes and content changes in the same commit" is a good rule of thumb.

jsonschema-validation.xml
karenetheridge: ok right.

jsonschema-validation.xml
```diff
@@ -1017,13 +1017,16 @@
                 <t>
                     If the instance is a string, and if "contentMediaType" is present, this
                     property contains a schema which describes the structure of the string.
+                    The string MUST be in a format compatible with the data model defined in the
```
handrews, line 1020: There's not really a MUST here, because these are annotations, not assertions.  They are telling the application that it can use the data in some way.  That does not guarantee that the data is suitable.

Just because a piece of JSON data is annotated with `title` and `description` does not mean that the given title and description are accurate, relevant, or even safe to display to end-users.  It is up to the application to decide whether additional validation is needed, and what that would look like.

`contentEncoding`, `contentMediaType`, and `contentSchema` should be treated the same as if received a MIME attachment with MIME header `Content-Transfer-Encoding` matching `contentEncoding` and MIME header `Content-Type` matching `contentMediaType`.  And separately, you are given a JSON Schema, and told that it is relevant to that attached resource.

There is no automatic application of the schema.  You can decode and parse the attachment or not.  If you do, you can apply the schema or not.  You have to decide how to parse the attachment, based on its media type, into the JSON Schema content model.

It doesn't make sense to say "a format compatible with the data model" because we do not define mappings from media types into the data model.  Applications are required to come up with that on their own.  Some (like the `application/jwt` example in the spec) are fairly obvious.  Others area not, and may potentially work in several different ways that could be suitable for different use cases.

JSON Schema doesn't enforce any of that.  The application first of all has to decide whether to trust the instance author (if you have `"contentMediaType": "application/javascript"`, there is a lot of security risk to `eval()`ing it, for example).

These fields are also like HTTP headers- `Content-Type` is not guaranteed to be accurate by HTTP.  The sender may or may not have set it correctly, and some user-agents will do a bit of parsing to try to figure out the "real" media type.

But all of this means that these are not MUST assertions.  A MUST is something that an implementation has to take action on, and there is none of that with these keywords.  As noted in the introductory material for this section, `contentSchema` cannot affect the outcome of the schema that contains it.  If run, it MUST be done as a separate step (more accurately, the instance string MUST NOT be ***automatically*** decoded, parsed, and validated.

karenetheridge: > A MUST is something that an implementation has to take action on

I wouldn't interpret a MUST here as "the implementation must verify that this is a valid schema"\*, but rather "you can assume that this is a valid schema, and if anything takes action on this and it *isn't* a valid schema, well, we already told you it must be one, so too bad, you get to keep both halves".

\* After all, there are other places in the spec that assert a "MUST" that the implementation doesn't necessarily check.  e.g. in an if/then/else assertion, if the "if" subschema evaluates true, we may never bother to actually verify that the subschema underneath the "else" path actually is a spec-compliant schema, because we never go down that path.  To require the implementation to verify every MUST assertion in the spec incurs an unnecessary efficiency burden.

ssilverman: @handrews my understanding from your comment (https://github.com/json-schema-org/json-schema-spec/pull/917#discussion_r428091385) is that there's a schema to be applied to some data model. I get your point on the MUST. Maybe changing to a SHOULD would suffice? Pushing now...

ssilverman: @handrews My logic: On one hand, you've said that the whole point of there being a schema (https://github.com/json-schema-org/json-schema-spec/pull/917#discussion_r428091385) is to apply it to some data model that the instance uses, but on the other you're saying it can't be a MUST, which makes sense because these aren't assertions. I've changed to SHOULD because why else would there be a schema in the first place? Resolving this conversation because I just force pushed the change.

handrews:
@ssilverman It still doesn't make sense to say SHOULD follow the data model.  What does that even mean?  It's completely media-type specific and application-specific.  How that gets sorted out is not the responsibility of the specification.

There's no clear definition for what "follow the data model" means.  There is a data model.  And you have a document of some media type that you parsed out of a string based on `contentMediaType` and possibly `contentEncoding`.  Does it follow the data model?  I don't know, can the application figure out a way to map it into the data model?  Then yes.  Is that going to be the same for all applications?  Possibly not.  Mapping XML to a JSON-ish structure, for example, can be done in several ways.  Does that mean XML follows the data model?  How would you even know without knowing what mapping the application is using?

What about some media type we've never heard of?  Does it follow the data model?  That depends, can whoever is designing that application figure something out that works well enough?  It's outside the scope of the spec.

The `content*` keywords are just information about content in the string.  One of those pieces of information happens to be a schema.  These pieces of information may or may not be accurate.  JSON Schema can't enforce that.  These keywords never fail validation, not even `contentSchema`.

`contentSchema` is just a schema that you can try to apply to the content, assuming you can parse it into something that allows that.  We don't want to say anything that restricts possible creative ways people decide to do that.  And we ***definitely*** do not want a situation where people keep dropping by here asking if media type X "follows the data model" and if so how that works.  That is outside of our scope.

karenetheridge:
> And we definitely do not want a situation where people keep dropping by here asking if media type X "follows the data model" and if so how that works. That is outside of our scope.

The only possible answer to that, IMO, is "go write an extension for your implementation of choice and come back and let us know".

handrews:
As a relevant example, I'm also working on a PR advising OpenAPI how to use these keywords in OAS 3.1, which will be the first version to support them.  They have a few media types they want to explicitly map (most notably `multipart/form-data` but also some others), so OAS will explain how to do the mapping.

There are several OAS-specific complications there, and it is better for JSON Schema to leave a lot of room in this area than to try to nail anything down.

ssilverman:
I'm still confused what you want here, for these reasons:
1. In your comment (https://github.com/json-schema-org/json-schema-spec/pull/917#discussion_r428091385), you say that JSON Schemas apply to something that has the JSON data model.
2. What on earth is "contentSchema" _for_ if not to apply to a JSON data model? By not even saying that the instance "SHOULD follow the JSON data model", that makes "contentSchema" meaningless as a JSON schema.
3. Why, then, is "contentSchema" even defined to "MUST be a JSON schema"?

I've just removed the `The string SHOULD be in a format compatible with the data model defined in the <xref target="json-schema">core JSON Schema</xref>.` part, left in the "contentSchema MUST be a JSON schema part", and then force pushed.

I'm still very unclear on why "contentSchema" "MUST" be a JSON schema if it's not even going to be applied to something that doesn't follow the JSON data model.

ssilverman:
I also remove the `Also changed one other period to be outside its <xref> tag, for consistency.` part.

handrews:
@ssilverman I am sorry for the confusion.  This is why we usually file issues first for anything beyond at typo, although long-time contributors sometimes go straight to PR for small things.

When you opened this PR I assumed you were just trying to make this change (as I just merged it), and there didn't seem to be a reason to go make you open an issue.  We're not really sticklers for process here.

But as you poked at it you found more things to question, which is completely fine and in general a good thing.  But these things came in piecemeal and I was trying to respond to them in between bits and pieces here and there, which is not great for giving the most coherent response.  I probably should have asked for an issue laying out your overall concerns once things got at all more complicated.  So I doubt my responses were ideal even though I did try to go back over things.

Unfortunately, since you edit and force-push the same commit every time, some previous parts of the conversation were not accessible (not that I could figure out, anyway).  So I may have contradicted myself because I couldn't find the older comments (I don't always save email notifications).  It's usually a good idea to add commits if there are unresolved comments attached to existing commits.  Again, we're not sticklers about that, but it does make it easier to see how a long-running PR evolved.  We really are not bothered by "extra" commits in the history.

To answer your question (I hope):

* The value of the keyword `contentSchema` is part of our specification, and like all keywords, has MUST-level requirements on the form (even if for a keyword like `const` it allows everything so there are no MUST _restrictions_.
* The instance data is instance data, and these keywords are not assertions.  Therefore the instance data is not something on which we impose requirements that a JSON Schema implementation enforces.

This is why `contentMediaType` is written as:

> If the instance is a string, this property indicates the media type of the contents of the string.

and there _is no requirement_ that the string contents actually be a valid document of that media type.  In that case, the wording is completely intentional.

As for

> This keyword MAY be used with any media type that can be mapped into JSON Schema's data model.

it's just noting the possibility that it doesn't have to be `application/json`.  That is mostly there to prevent people from enforcing a JSON media type when the schema is used.  It doesn't do anything more than that, nor do we want it to.

I realize my previous response probably sounded odd since this does mention the data model, but again, it's hard to drop in and out as the scope of the PR shifts and sometimes I just miss things that are really obvious because I've stared at this spec for way too long that parts of it are just invisible to me at this point.

ssilverman:
@handrews Just for posterity, I'll respond and clarify. Your points were all good ones, thank you. I thought this would be a simple PR but it appeared to sparked more discussion. I'll make an issue next time first.

With regards to force pushing, I find that I tend to take on the development practices I've used most recently. I use force pushing for branches that will eventually get into the main branch so that there's a minimal number of changes. Since the reviewers in this org tend to merge the PR and not the submitter, keeping the commits as they should be, for less clutter, is necessary.

It _is_ possible to look at prior comments and discussion from pre-force-pushed commits in these ways:
1. Click on any of the "Show resolved" lines in the discussion here. It may be the case that that discussion is marked as "Outdated", but you can still expand them.
2. Links into an "Outdated" discussion still work. Try this: Copy the link and paste into the address bar of a new browser tab or window. You'll find that it goes to the correct, even if outdated, comment. It's possible this works less well on mobile, which is why you may not see this. Try either a desktop view or a browser on a desktop.

It's my new understanding that `contentSchema` is really just a suggestion, with no proscriptions on the instance contents, even though you state that it's meant to be applied to a "JSON Schema data model". (For practice going to an "Outdated" discussion: https://github.com/json-schema-org/json-schema-spec/pull/917#discussion_r428091385 :) )

karenetheridge:
> Since the reviewers in this org tend to merge the PR and not the submitter, keeping the commits as they should be, for less clutter, is necessary.

FWIW, the reviewer can rebase (and squash) before merging. There is a GUI option to squash all commits, and it's not hard to rebase and squash selected commits together from the command line as well.

As a reviewer it's much easier to follow what's going on if changes aren't applied as squashed amendments while the conversation is going on. 

One can put the PR into a work-in-progress state which prevents merges until the flag is lifted.  "let me know before merging so I have a chance to clean up the commits". Additionally, it's ugly when a pull request is merged without rebasing the commits that have gone into master in the meantime, so the author can rebase as well.

ssilverman:
I’m not sure how your comments are relevant. It’s advantageous to reduce the number of commits in a PR. This has nothing to do with rebasing. And anyone _can_ go back and see context even with force pushes.

handrews:
@ssilverman @karenetheridge all of this is a matter of style.  We often do multi-commit PRs to make reviews easier in various ways.  They are rarely squashed unless someone wants to do so.

No one's going to force you to split commits unless something is incredibly difficult to review otherwise (this is usually major reworkings with multiple steps).  But please do not try to enforce style from other projects here.

ssilverman:
@handrews For the record, I'm not trying to enforce style. I was feeling questioned and so was explaining my reasoning. I can assure you I'm very aware that there's more than one way to do things here. Please don't suggest I'm unaware of this or trying to enforce anything.

handrews:
@ssilverman It was a request in response to the overall conversation.  It was not an accusation, nor was it specifically directed at you.  I don't even think it would be a bad thing to bring in style, so I'm not sure what I'm supposed to have accused you of.

karenetheridge:
My comments were purely in response to the fact that force-pushing before the review phase has completed makes it harder to see the history (because previous commits are lost, even if the comments attached to them are not), so I was suggesting that one not force-push until ready to merge. I was not presuming the git ability level of anyone because people come from all levels of knowledge and experience and that's okay.  That's all.

PR #923, opened 2020-05-15
Fix patterns and paths in output schema

ssilverman:
Fix a raw "$" in two patterns and add a missing "$defs" in three paths

### DIFF
--- a/output/schema.json
+++ b/output/schema.json
@@ -50,16 +50,16 @@
               {
                 "properties": {
                   "keywordLocation": {
-                    "pattern": ".*/$ref/.*"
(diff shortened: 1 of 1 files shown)


output/schema.json
```diff
@@ -50,16 +50,16 @@
               {
                 "properties": {
                   "keywordLocation": {
-                    "pattern": ".*/$ref/.*"
+                    "pattern": ".*/\\$ref/.*"
```
karenetheridge, line 53: Shouldn't this be one backslash, not two?

karenetheridge: nm, it is good.

Issue #906, opened 2020-05-01
Drop the idea of automatically combining annotations


handrews:
We have this idea that some annotations get combined (e.g. if multiple `readOnly` keywords apply to the same location, the overall annotation is `true` if any of them are `true`).

However, our output recommendations don't support that.  They just return all of the information needed to implement it (all of the annotation values, where they come from, and to where they are applied).

Since annotations are intended to be processed by the application anyway, we can get the necessary effect for keywords like `readOnly` by directing applications on how they process the results.  For `readOnly` and `writeOnly`, our language actually supports that anyway:

> The value of these keywords MUST be a boolean. When multiple occurrences of these keywords are applicable to a single sub-instance, the resulting behavior SHOULD be as for a true value if any occurrence specifies a true value, and SHOULD be as for a false value otherwise.

Note that this doesn't say anything about implementations combining things, just about the resulting behavior.

------
This would involve some changes to the Core spec:

In section 7.7.1 Collecting Annotations, we would change:

> If the same keyword attaches values from multiple schema locations to the same instance location, and the annotation defines a process for combining such values, then the combined value MUST also be associated with the instance location. The output formats described in this specification that include annotation information meet this requirement.

to something like:

> When the same annotation keyword attaches values from multiple schema locations to the same instance location, the keyword MAY specify how applications SHOULD or MUST interpret the different values.

In section 7.5 Applicators, we would drop:

> Annotation results are combined according to the rules specified by each annotation keyword.

There are probably other spots, but this gets the idea across.  The whole "combine annotations" thing was never really thought through all the way in the context of output formats and other things that were added after the initial idea.

ssilverman:
This seems like a good place to ask this. If we drop "combining annotations", that leaves the rest of the sentence, and I'd like to inquire as to its intent.

The text:
```
Annotation results for "properties" keywords from multiple schemas applied to the same instance location
```

Can a short example be provided that shows where you can actually apply multiple schemas to the same instance location? I can think of, for example, "allOf", but since "additionalProperties" (I'm neglecting "unevaluatedProperties" for now) can't peer inside an "allOf", how can multiple schemas even apply?

I should be asking, actually: "What is the intent behind 'multiple schemas'" here?

ssilverman:
I'll answer my own question. The phrase **same schema object** is key in this paragraph from "annotationProperties":
[...]

PR #925, opened 2020-05-15
Split "prefixItems" from "items", drop "additionalItems", make "unevaluatedItems" respect "contains"

### LINKED ISSUE #810: "unevaluatedItems" should consider "contains"
The interaction of these seems underspecified.  @Relequestual what was the problem you ran into here exactly?  I've already forgotten 😬

### LINKED ISSUE #864: Split `items` into `items` and `tupleItems`
The `items` keyword has an object form and an array form. The object form is used for constraining the items in an array while the array form is for constraining an array that is used like a tuple. This is something that trips up people new to JSON Schema.

It also causes confusion with the `additionalItems` keyword. `additionalItems` is only defined for use with the array form of `items`, but I often see people throwing in `"additionalItems": false` when they are using the object form. I'm not sure what they think that means, but it's not uncommon to see.

# Proposal
* Redefine the `items` keyword to only support the current object form behaviour
* Introduce the `tupleItems` keyword that supports the current array form of `items`
* Rename `additionalItems` to `additionalTupleItems` to make it clear that it works with the `tupleItems` keyword not the `items` keyword.
* Rename `unevaluatedItems` to `unevaluatedTupleItems` to make it clear that it works with the `tupleItems` keyword not the `items` keyword.

## Pros
* Less confusing for JSON Schema noobs and veterans alike
* Better errors. If you try to use an array for `items` when you meant the object form (and vis versa) meta-schema validation will fail making it more clear where your problem is.

## Cons
* Redefining an existing keyword is a backwards incompatible change
* OAS is skeptical of additional changes at this stage

handrews:
Closes #810 (`unevaluatedItems` should consider `contains`)
Closes #864 (split `items`)

This reworks the array applicators in several ways:

* "prefixItems" takes on the former role of the array form of "items"
* "items" keeps its single-schema role, and takes on the role of
  "additionalItems" when "prefixItems" is present
* "contains" now produces an annotation indicating what it evaluated
* "unevaluatedItems" now respects "contains", and the language around
  interpreting the relevant annotation is (hopefully) less convoluted

Note that this does not address the change to put unevaluatedItems
into a separate vocabulary with unevaluatedProperties, which will
be done as a separate PR.  Doing that will also take care of the current ordering
problem- `contains` should appear before `unevaluatedItems`, but deferring
that keeps this diff more readable.

It is also possible that questions raised on slack today regarding `contains`/`minContains`/`maxContains` might further impact this, but any such impacts will be done as a separate, follow-on PR so as not to delay getting this section of work checked in.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3786,9 +3786,9 @@ https://example.com/schemas/common#/$defs/count/minimum
                     <t hangText="draft-handrews-json-schema-03">
                         <list style="symbols">
                             <t></t>
-                            <t></t>
-                            <t></t>
-                            <t></t>
+                            <t>Array-value "items" functionality is now "prefixItems"</t>
+                            <t>"items" subsumes the old function of "additionalItems"</t>
+                            <t>"contains" and "unevaluatedItems" interactions now specified</t>
                             <t></t>
                             <t></t>
                             <t></t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2445,9 +2445,9 @@
                         </t>
                         <t>
                             This keyword produces an annotation value which is an array of
-                            the indexes to which this keyword successfully applied its subschema,
-                            in ascending order.  The value MAY be a boolean true if the subschema
-                            was successfully applied to every index of the instance.
+                            the indexes to which this keyword validates successfully when applying 
+                            its subschema, in ascending order. The value MAY be a boolean true if the 
+                            subschema validated successfully when applied to every index of the instance.
                         </t>
                     </section>
                 </section>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2332,13 +2332,17 @@
                         </t>
                         <t>
                             Validation succeeds if each element of the instance validates
-                            against the schema at the same position, if any.
+                            against the schema at the same position, if any.  This keyword
+                            does not constrain the length of the array.  If the array is longer
+                            than this keyword's value, this keyword validates only the
+                            prefix of matching length.
                         </t>
                         <t>
                             This keyword produces an annotation value which is the largest
                             index to which this keyword applied a subschema.  The value
                             MAY be a boolean true if a subschema was applied to every
                             index of the instance, such as is produced by the "items" keyword.
+                            This annotation affects the behavior of "items" and "unevaluatedItems".
                         </t>
                         <t>
                             Omitting this keyword has the same assertion behavior as
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -703,7 +703,7 @@
                     an alternate implementation producing the same behavior is available.
                     Keywords of this sort SHOULD describe reasonable alternate approaches
                     when appropriate.  This approach is demonstrated by the
-                    "<xref target="additionalItems" format="title"/>" and
+                    "<xref target="items" format="title"/>" and
                     "<xref target="additionalProperties" format="title"/>" keywords in this
                     document.
                 </t>
@@ -931,7 +931,7 @@
 {
     "title": "Feature list",
     "type": "array",
-    "items": [
+    "prefixItems": [
         {
             "title": "Feature A",
             "properties": {
@@ -2143,11 +2143,11 @@
                             "additionalProperties" and itself
                         </t>
                         <t>
-                            "additionalItems", whose behavior is defined in terms of "items"
+                            "items", whose behavior is defined in terms of "prefixItems"
                         </t>
                         <t>
                             "unevaluatedItems", whose behavior is defined in terms of annotations
-                            from "items", "additionalItems" and itself
+                            from "prefixItems", "items", "contains", and itself
                         </t>
                     </list>
                 </t>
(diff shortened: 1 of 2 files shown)


handrews:
Note also that some of the text regarding how to combine multiple annotation values was intentionally dropped per #906 (Drop the idea of automatically combining annotations).  The remaining `unevaluatedProperties` text explains how multiple values are to be used for that keyword, and other keywords would provide their own clear conventions as needed.

I will do another PR in the future to take care of the rest of #906 but there was no point in preserving those paragraphs while reworking these sections.

jsonschema-core.xml
```diff
@@ -2324,61 +2324,49 @@
...
                             index to which this keyword applied a subschema.  The value
                             MAY be a boolean true if a subschema was applied to every
-                            index of the instance, such as when "items" is a schema.
-                        </t>
-                        <t>
-                            Annotation results for "items" keywords from multiple
-                            schemas applied to the same instance location are combined
-                            by setting the combined result to true if any of the values
-                            are true, and otherwise retaining the largest numerical value.
+                            index of the instance, such as is produced by the "items" keyword.
```
notEthan, line 2339: it seems simpler to just have `prefixItems` produce a max index annotation, and `items` produce true.

notEthan: also the language in `items` doesn't seem to account for a `true` annotation result from `prefixItems`, only numeric

notEthan:
I don't know if this is the PR for it, but it would make more sense to me if items, prefixItems, and contains all produced arrays of array indices as their annotations. I'm not fond of the mishmash of `true`, max applied index, and array of applied indices which are all now options.

this would be more consistent with the arrays of property names produced by the *properties keywords.

handrews:
@notEthan The point of the annotations is to allow the most performant creation and utilization of the information.  If an instance contains a 500-element array, there is no reason for `items` to emit a 500-element array of indices.  All you need to know is that all elements have been evaluated.  That could be done with a check against the size of the array, but there's no reason to do that check when there is no cost to just set it to `true` for `items` or `unevaluatedItems` `prefixItems` might require a check of the prefix length vs total length which is why it's a MAY and not a MUST to use a boolean- I could see dropping the boolean option for `prefixItems` although I wouldn't support it.  You can file that as an issue if you would like, I do not want to address it here.

jsonschema-core.xml
handrews: @notEthan if `prefixItems` is true then `items` doesn't do anything.  There's no need for a length check.

notEthan:
all right, conceded that an arbitrarily large array shouldn't emit an equally-large annotation array, I'll drop that idea.

I'm still going to advocate for one annotation type per keyword - items emits true, prefixItems emits a number, contains emits an array. this split seems natural to me as a part of the `items` split.

the alleged benefit of prefixItems emitting true is based on the idea that a check for a true value is more efficient than a length check against an integer. I think the performance difference between those is completely negligible, and the added complexity of accommodating multiple type options for prefixItems annotations outweighs any few cycles' difference between comparing booleans and integers.

> I could see dropping the boolean option for prefixItems although I wouldn't support it. You can file that as an issue if you would like, I do not want to address it here.

sorry I didn't file a separate issue.  but this PR is the one splitting `items`, so annotations changed by that split seem very much in scope to me. however I will not pursue it further after this comment if you remain against it.

handrews:
> sorry I didn't file a separate issue. but this PR is the one splitting items, so annotations changed by that split seem very much in scope to me. however I will not pursue it further after this comment if you remain against it.

@notEthan I'm just not doing it right now.  I very strongly oppose scope creep on issues.  This issue just splits and carries forward the previously existing annotation scheme (adding annotations for `contains`).  We spent an epic amount of time sorting this out in the issue and that's done and approved.

If you want to change the pre-existing annotation approach, you need to file an issue on that and not try to expand the scope of a major change that went through a lot of negotiation already, just because it touches the same area.

For me, I need to get PRs completed.  Someone else can decide if your suggestion warrants a change in this draft, I'm trying to stay out of that because I have more than enough work to do with the issues already assigned to the draft.

karenetheridge:
Is the bikeshedding on the "prefixItems" name totally done (vs. alternatives like tupleItems, sequentialItems etc).  The spec should make it very clear what "prefix" is referring to -- for example "all items that come at the beginning of the list, and processed before additionalItems considers what items it should evaluate against".

handrews:
@karenetheridge yes, the naming is 100000000% final.  Please see the issue for that whole saga, but it will not be reopened.  Or rather, if you want to re-open it, ask @Relequestual and y'all go off and figure it out and tell me what happens 😛 All of my arguments for `prefixItems` are in the issue and don't need further elaboration.  I was generally fine with `tupleItems` but someone had objections there, I think even beyond my concern that the relationship between `tupleItems` and `items` was not clear from the names.

The dependency of `items` (not `additionalItems` which has been removed) on `prefixItems` is all that needs to be said in terms of _order_ of processing.  Technically, since you can implement `items` without annotations, you don't actually need to process them fully in that order anyway.  It's an implementation detail and does not belong in the spec.

I will add a line about its _conceptual_ relationship with `items`, though, that's a good catch.  It needs to be clear that you can have a "prefix" without anything following it.

handrews:
@karenetheridge I agree that using `prefixItems` on its own is a little awkward, but so is using `additionalItems` on its own, which is an important use case that people often don't even notice.  Naming is hard.

notEthan:
@handrews I think maybe you meant "`additionalProperties` on its own" ? additionalItems on its own had no effect (and is now gone). (not meaning to nitpick, just to clarify)

handrews:
@notEthan yeah thanks, fixed

handrews:
@karenetheridge updated the text, see if this is better.  I am trying to keep this fairly minimal as keywords specs are not tutorials, but I agree that a bit more explanation was called for.

jsonschema-core.xml
```diff
@@ -2141,11 +2141,11 @@
                             "additionalProperties" and itself
                         </t>
                         <t>
-                            "additionalItems", whose behavior is defined in terms of "items"
+                            "items", whose behavior is defined in terms of "prefixItems"
                         </t>
                         <t>
                             "unevaluatedItems", whose behavior is defined in terms of annotations
-                            from "items", "additionalItems" and itself
+                            from "prefixItems", "items", "contains", and itself
```
gregsdennis, line 2148: I still firmly disagree with `contain` influencing `unevaluatedItems`.  Can we separate the `items` split and the `contains` issue into two PRs, please?

jsonschema-core.xml
```diff
@@ -2399,7 +2391,7 @@
                         <t>
                             The behavior of this keyword depends on the annotation results of
                             adjacent keywords that apply to the instance location being validated.
-                            Specifically, the annotations from "items" and  "additionalItems",
+                            Specifically, the annotations from "prefixItems", "items", and "contains",
```
gregsdennis, line 2394: Not `contains`

jsonschema-core.xml
```diff
@@ -2408,33 +2400,25 @@
...
-                            "unevaluatedItems" MUST be ignored.  However, if none of these
-                            annotations are present, "unevaluatedItems" MUST be applied to
-                            all locations in the array.
-                        </t>
-                        <t>
-                            This means that "items", "additionalItems", and all in-place applicators
-                            MUST be evaluated before this keyword can be evaluated.  Authors of
-                            extension keywords MUST NOT define an in-place applicator that would need
-                            to be evaluated before this keyword.
+                            This means that "prefixItems", "items", "contains", and all in-place
```
gregsdennis, line 2412: not `contains`

handrews:
@gregsdennis take the `contains` thing up with @Relequestual in the issue or in slack, and let me know the outcome.  I'm not going to bother to disentangle this until there's a decision.  The issue was marked block pending the spit issue, and that issue got resolved, so I went ahead.

I feel strongly that these should interact, and I don't have anything further to say on it.  If you sell @Relequestual on overruling me that's cool, I'll change it then.

jsonschema-core.xml
gregsdennis: I've been externally convinced, though I still don't like the names.  I'll open a new issue for the naming of things.

handrews: @gregsdennis we spent an epic number of comments debating the naming, with participation from the OpenAPI steering committee as well.  In the end, @Relequestual made a call and that's what I implemented.

I do apologize for not realizing the `contains` things wasn't signed off on, but `prefixItems` was [signed off in this comment](https://github.com/json-schema-org/json-schema-spec/issues/864#issuecomment-598313804) and then after more comments, was [reaffirmed in this comment](https://github.com/json-schema-org/json-schema-spec/issues/864#issuecomment-601147496).

jsonschema-core.xml
```diff
@@ -2459,6 +2443,12 @@
                             array element even after the first match has been found.  This
                             is to ensure that all possible annotations are collected.
                         </t>
+                        <t>
+                            This keyword produces an annotation value which is an array of
+                            the indexes to which this keyword successfully applied its subschema,
+                            in ascending order.  The value MAY be a boolean true if the subschema
+                            was successfully applied to every index of the instance.
```
Relequestual, line 2450: I'm not happy with the phrasing `successfully applied`. I feel it's ambigious and it's the first time we've used it in this document. This is the only two instances of it.

I'd be happy with `validates successfully when applied` or similar.

```suggestion
                            This keyword produces an annotation value which is an array of
                            the indexes to which this keyword validates successfully when applying 
                            its subschema, in ascending order. The value MAY be a boolean true if the 
                            subschema validated successfully when applied to every index of the instance.
```

PR #926, opened 2020-05-15
Rel JSON Ptr forward/backward index manipulation

### LINKED ISSUE #115: Relative JSON-pointer extension
The proposal is to extend relative JSON pointer (https://tools.ietf.org/html/draft-luff-relative-json-pointer-00) to be able to refer to the items in array with some relative position.

Originally filed as: https://github.com/json-schema/json-schema/issues/223

The syntax extension can be really simple, from the current N/JSONpointer to N+M/JSONPointer (to traverse N levels up and M items forward) or N-M/JSONPointer (M items backward), where N == 0 can possibly be dropped.

That would allow, for example, to implement restrictions on the array such as ascending/descending order (using `$data` [v5 proposal](https://github.com/json-schema/json-schema/wiki/$data-%28v5-proposal%29)).

Ascending order without duplicates:

``` JSON
{
  "type": "array",
  "items": {
    "type": "integer",
    "minimum": { "$data": "-1/" },
    "exclusiveMinimum": true
  }
}
```

Objects ordered in ascending order by unique `id` property:

``` JSON
{
  "type": "array",
  "items": {
    "properties": {
      "id": {
        "type": "integer",
        "minimum": { "$data": "1-1/id" },
        "exclusiveMinimum": true
      }
    }
  }
}
```

handrews:
Closes #115 (to the extent that it is being adopted- several proposals came and went in that discussion).

This allows using a relative pointer to look ahead or behind
in arrays when calculating the initial point at which to apply
the JSON Pointer that is the suffix of the Relative JSON Pointer.

### DIFF
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -33,7 +33,7 @@
             </address>
         </author>
 
-        <date year="2019"/>
+        <date year="2020"/>
         <workgroup>Internet Engineering Task Force</workgroup>
         <keyword>JSON</keyword>
         <keyword>JavaScript</keyword>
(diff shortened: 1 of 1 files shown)


Relequestual:
This has been here forever so I'm merging it.

Relequestual:
Do we need to update references in other spec files as a result?

handrews:
@Relequestual Before we publish we need to update all of the cross-references.  I think they might not build until the target exists?  I don't remember.  I might submit them in dependency order and update the later ones before submitting?  Meh.  I really don't recall.

PR #928, opened 2020-05-20
Fix $anchor pattern in core.json

ssilverman:
This was missed in the $anchor specification update.


handrews:
Good catch.  We always have to go over the meta-schema right before publishing because we always forget something.  Actually I think we _still_ always have deployed with at least one error :/

PR #929, opened 2020-05-20
Ether/minor wording

karenetheridge:
very minor fixes to grammar and an xref link

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -457,7 +457,7 @@
 
         <section title="Fragment Identifiers" anchor="fragments">
             <t>
-                In accordance with section 3.1 of <xref target="RFC6839"></xref>,
+                In accordance with section 3.1 of <xref target="RFC6839">RFC 6839</xref>,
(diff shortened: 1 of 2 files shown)


handrews:
RFC xrefs actually work fine without text (although I should have used the `<xref target="..." />` form to make the intent more clear.  But we are inconsistent about it (the brackets happen when you don't provide text for the xref):

Screen Shot 2020-05-20 at 12 14 32 PM

I'm fine with changing it this way (and nice catch on the missing "a"s) but it should probably go on some consistency checklist somewhere for later review across the spec.

karenetheridge:
> RFC xrefs actually work fine without text

Yup, I noticed when reading the marked up document that this link rendered differently (not badly, just differently) than all the others, and then saw why when I looked at the source. Inconsistencies bug me :)

handrews:
@karenetheridge I'm all in favor of resolving the inconsistencies.  I suspect there are more like this, but that's no reason not to merge this one.  I don't think anyone else will object so I'll go ahead and do that.

Issue #809, opened 2019-10-08
Some headings are too long for IETF format


handrews:
I've never figured out what tool generates this format so I never checked, but see https://tools.ietf.org/html/draft-handrews-json-schema-02#section-8.2.4.2 for one that wraps and looks quite bad.

Also, https://tools.ietf.org/html/draft-handrews-json-schema-02#section-8.2.3 and https://tools.ietf.org/html/draft-handrews-json-schema-02#section-9.2.1 look fine but are too long in the table of contents so it refuses to link them (as far as I can tell that's what's going on).

There might be other sections with this problem, and note that many sections such as 8.2.4.2 are not in the TOC but may still look bad in the formatting.

handrews:
Most of this will get handled by the PR for #909 so I'll pick up the last bit, too.

PR #930, opened 2020-05-20
Rename $recursive* to $dynamic*, make it work with normal anchors / plain name fragments instead of base URI switching

### LINKED ISSUE #868: Handling the base URI while evaluating against an instance
This issue is about the architectural principles that:
* You can resolve all uses of the base URI as a pre-processing step
* Once you do that, evaluating a schema object is the same regardless of its parent

Technically, draft 2019-09 violates that, although in practice we can wiggle around it.  But we should decide whether these principles hold, and make sure that our spec meets our own principles!  Options include:

* We change how we describe the behavior of a few keywords, and keep both principles _(I'm leaning towards this as explained at the end)_
* We decide that you can resolve `$id` and `$ref` as a pre-processing step, but in general you might need to keep track of base URIs during evaluation with an instance
* We decide not to promote the idea of a pre-processing step at all (although you do need one to at least discover all of the static URIs that can be reference targets, so you never entirely get rid of it)

------

`$id` and `$ref` (the only draft-07 keywords to rely on the base URI) to full URIs during schema _loading_, can be pre-processed by simply setting their values to the full URIs at the same time that you find the various schemas and cache them in some sort of URI-lookup thingy.

However, in the latest draft, `$anchor`, `$recursiveAnchor`, and `$recursiveRef` rely on the base URI as well.

For `$anchor`, since it only adds a URI to associate with that schema in your cache, there's no further processing with the base URI to do there.  So it's not a problem in practice.

But `$recursiveAnchor` and `$recursiveRef` are problems.  At least in theory.  In practice, because we restrict `$recursiveAnchor` to resource root schemas and `$recursiveRef` to only have a value of `"#"`, you can handle these without needing to know the base URI.  That is, in fact, why those restrictions exist.

So we can kind-of get away with ignoring this problem for now, and we could change how we talk about these keywords to remove the base URI stuff.

---------

In general, though, these keywords as we currently describe them work by dynamically re-calculating the base URI of the URI-reference in `$recursiveRef` depending on `$recursiveAnchor`.  This was done so that we could lift those restrictions on the value of `$recursiveRef`.  Or replace these keywords with a more general `$dynamicAnchor` and `$dynamicRef` since the name "recursive" wouldn't be entirely accurate anymore.

This works as follows:

* Resolve the URI-reference in `$recursiveRef` just as you would for `$ref` to get the _initial target_
* If the initial target contains `"$recursiveAnchor": true`, walk back up the dynamic scope to find the outermost scope that also has `"$recursiveAnchor": true`, to get the _intermediate target_
* ***Re-resolve*** the URI-reference in `$recursiveRef` against the base URI from the immediate target, to produce the _final target_

The nice thing about this is that it works with any URI-reference in `$recursiveRef` (although some sorts of URI-references don't make much sense- the other reason we restricted it).  And in practice, the `"#"` restriction means that the final target is **always** the same as the intermediate target, so you never actually need to re-resolve the URI reference.  once you find the intermediate target, you're done.

But in the general case, where the intermediate and final targets could be different, you need to know, *at runtime*, the base URI for both the initial and intermediate targets.  You can't even resolve the `$recursiveRef` to a full URI because you need to know which part was the original reference in order to re-resolve it against the intermediate target's base URI.

-----------

If we want to keep the ability to preprocess the base URI to the point where we never need to worry about parent schema objects, the best way to do that would be to reserve a keyword (`$base`? `$_base`?) where an implementation could safely store the base URI during preprocessing if there's a keyword in that object that would need it.  Then, when `$recursiveRef` or `$recursiveAnchor` is encountered, and implementation could just look at that reserved location.

There are subtleties there like what to do if someone actually does try to use it as a keyword, etc.  But that's what comes to mind for me.

Thoughts?  At the moment, I'm leaning towards changing how we talk about `$recursiveAnchor` and `$recursiveRef` and saying something like "these keywords adjust the reference target within the dynamic scope" instead of "these keywords change the base URI of the reference."  That gets a bit messy with another architectural principle of "always identify things with URIs", but I feel like that is more easily finessed.  Besides, we at least _start_ with a URI, and we can figure out the URI of the final target if we wanted to.

### LINKED ISSUE #909: Improve/simplify "$recursiveAnchor" and "$recursiveRef"
These keywords depend on the concept of the [dynamic scope](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.7.1), the runtime schema traversal structure including which references were followed how often to reach the current schema+instance locations.

Note that @awwright has also filed #907 regarding this same problems space.  While I don't currently see how that proposal solves the entire use case, I am very much open to being sold on other proposals for this.

---------
Currently, `$recursiveAnchor` takes a boolean which MUST be `true`, and `$recursiveRef` takes a URI-reference which MUST either be an empty fragment or an absolute-URI.  (_Note:  The text says it MUST be `"#"`, but we actually use absolute-URIs as well so the text is in error._).

The mechanism for how these works is described a complex process of evaluating the `$recursiveRef` twice with different base URIs, depending on the presence of `$recursiveAnchor`.

This is why there's a (incorrectly specified) restriction on what kind of URI-references are legal for `$recursiveRef`.  Path references (e.g. `foo/bar`) when used with this process could produce very confusing results without any clear use case.

It's not actually that important to understand exactly how it works currently.  It's probably better to think of this as a new proposal.

We at minimum need to fix the 'MUST be `"#"`' problem, _and_ it would be nice to make this less complex.

-----------

I propose changing `$recursiveAnchor` so that it takes the same sort of value as `$anchor`, and produces a plain name fragment identifier, just like `$anchor`.

The difference between the two is that when a `$recursiveAnchor` is targeted by a `$recursiveRef`, the final resolved target will not necessarily be the locally declared fragment, but the fragment of the same name in the outermost dynamic scope that _also_ declares that fragment name with `$recursiveAnchor`.  That's a lot of word salad, so we'll go through an example.  But first:

* `$recursiveRef`s that do not point to a `$recursiveAnchor` act just like `$ref`
* `$ref`s that point to a `$recursiveAnchor` work just a if it were `$anchor`
* Only when `$recursiveRef` targets a `$recursiveAnchor` do they behave differently.

This ensures that all schemas involved (the schema object making the reference, the local subschema with the plain name fragment, and the final resolved subschema with the plain name fragment) all intentionally opt-in to this behavior.  We do not want magical behavior involving `$ref` or `$anchor`, ever.  Most people will use `$ref` and `$anchor` (or at least `$ref`), so they need to be simple and predictable.  `$recursiveRef` and `$recursiveAnchor` are advanced keywords suitable for a very specific use case.

To see how this works, let's look at:

* the standard dialect schema (https://json-schema.org/draft/2019-09/schema, which describes all vocabs included in the core and validation spec)
* an _abbreviated_ form applicator vocabulary schema (https://json-schema.org/draft/2019-09/meta/applicator),  which includes only the applicator vocabulary keywords.

In all cases, I'm omitting `$vocabulary`, `title`, and other inessentials to save space, and of course I'm using the new `$recursiveAnchor` syntax.

### Applicator Vocabulary

I'm only including enough applicators here to get the point across

```JSON
{
    "$schema": "https://json-schema.org/draft/2019-09/schema",
    "$id": "https://json-schema.org/draft/2019-09/meta/applicator",

    "$recursiveAnchor": "meta",

    "type": ["object", "boolean"],
    "properties": {
        "additionalProperties": { "$recursiveRef": "#meta" },
        "unevaluatedProperties": { "$recursiveRef": "#meta" },
        "properties": {
            "type": "object",
            "additionalProperties": { "$recursiveRef": "#meta" },
            "default": {}
        },  
        "allOf": { "$ref": "#/$defs/schemaArray" },
        "anyOf": { "$ref": "#/$defs/schemaArray" },
        "oneOf": { "$ref": "#/$defs/schemaArray" }
    },
    "$defs": {
        "schemaArray": {
            "type": "array",
            "minItems": 1,
            "items": { "$recursiveRef": "#meta" }
        }
    }
}
```
### Standard Dialect

```JSON
{
    "$schema": "https://json-schema.org/draft/2019-09/schema",
    "$id": "https://json-schema.org/draft/2019-09/schema",

    "$recursiveAnchor": "meta",

    "allOf": [
        {"$ref": "meta/core"},
        {"$ref": "meta/applicator"},
        {"$ref": "meta/validation"},
        {"$ref": "meta/meta-data"},
        {"$ref": "meta/format"},
        {"$ref": "meta/content"}
    ],  
    "type": ["object", "boolean"],
}
```

When we evaluate just the applicator meta-schema against a schema, there is only one meta-schema resource involved.  In this situation, `$recursiveAnchor` and `$recursiveRef` behave _exactly_ like `$anchor` and `$ref`.

When we evaluate the standard dialect meta-schema, the `$recursive*` keywords behave differently, as follows:

1. We begin evaluating from the root of the standard dialect meta-schema.  _This is the outermost dynamic scope_
1. Next we follow each `allOf` branch in turn, most notably the one to `"meta/applicator"`
1. Within the applicator meta-schema, we look at specific property schemas and eventually see a `"$recursiveRef": "#meta"`
1. We look for the `"#meta"` fragment declaration in the current schema resource, and find it- it is defined by `"$recursiveAnchor"`, so we have more steps to take.
1. We look at each resource in the dynamic scope, looking for the outermost resource that also declares a `#meta` fragment _with `"$recursiveAnchor"`_
1. The dialect meta-schema also has `"$recursiveAnchor": "meta"`, and is the outermost scope that does (it's also the only other scope in this example).
1. The resolved runtime target of the `"$recursiveRef": "#meta"` in the applicator vocabulary is the `"#meta"` fragment in the _dialect meta-schema_.

The effect here is that since we started from the dialect schema, which declares a `"$recursiveAnchor": "meta"`, and the applicator vocab schema also declares a `"$recursiveAnchor": "meta"`, the `"$recursiveRef": "#meta"` in the applicator vocab schema resolves to the anchor in the _dialect schema_.  Same plain name fragment, different resource.

handrews:
Closes #909 (improve/simplify/rename `$recursive*`)
Closes #868 (Let's not require runtime base URI tracking)
Also addresses 2/3 or #809 (overly long section names) but the other 1/3 is not in this part of the spec so I'll close that separately.

Instead of using a boolean (which never really fit with how the similarly named "$anchor" worked, and meant only one dynamic resolution was possible at a time) and messing with base URIs and re-resolving URI-references (a topic that creates plenty of confusion even on its own), use a fragment name the same as "$anchor" and make "$dynamicRef" work by just looking for the outermost resource with a matching "$dynamicAnchor" fragment name.

This stil requires that the initial "$dynamicRef" target URI be a valid URI, with a fragment created by "$dynamicAnchor", in order to produce the dynamic behavior.  This ensures that it is always clear which schemas are dynamic extension points and which are not.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1386,7 +1386,7 @@
                     </t>
                     <t>
                         The "$anchor" and "$dynamicAnchor" keywords are used to specify such
-                        a fragments.  They are identifier keywords that can only be used to create
+                        fragments.  They are identifier keywords that can only be used to create
                         plain name fragments, rather than absolute URIs as seen with "$id".
                         The behavior of the created fragment is identical for both keywords.
                     </t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -624,7 +624,7 @@
                     which must be resolved against another value, such as another
                     URI-reference or a full URI, which is found through the lexical
                     structure of the JSON document.  The "$id", "$ref", and
-                    "$dyanmicRef" core keywords, and the "base" JSON Hyper-Schema
+                    "$dynamicRef" core keywords, and the "base" JSON Hyper-Schema
                     keyword, are examples of this sort of behavior.
                 </t>
                 <t>
@@ -638,12 +638,12 @@
                     with an instance document.  The outermost dynamic scope is the
                     root schema of the schema document in which processing begins.
                     The path from this root schema to any particular keyword (that
-                    includes any "$ref" and "$dyanmicRef" keywords that may have
+                    includes any "$ref" and "$dynamicRef" keywords that may have
                     been resolved) is considered the keyword's "validation path."
                     <cref>
                         Or should this be the schema object at which processing
                         begins, even if it is not a root?  This has some implications
-                        for the case where "$dyanmicAnchor" is only allowed in the
+                        for the case where "$dynamicAnchor" is only allowed in the
                         root schema but processing begins in a subschema.
                     </cref>
                 </t>
@@ -658,8 +658,8 @@
                     dynamic parent, rather than examining the local lexically enclosing parent.
                 </t>
                 <t>
-                    The concept of dynamic scope is primarily used with "$dyanmicRef" and
-                    "$dyanmicAnchor", and should be considered an advanced feature
+                    The concept of dynamic scope is primarily used with "$dynamicRef" and
+                    "$dynamicAnchor", and should be considered an advanced feature
                     and used with caution when defining additional keywords.  It also appears
                     when reporting errors and collected annotations, as it may be possible
                     to revisit the same lexical scope repeatedly with different dynamic
@@ -721,8 +721,8 @@
                 <t>
                     While custom identifier keywords are possible, vocabulary designers should
                     take care not to disrupt the functioning of core keywords. For example,
-                    the "$dyanmicAnchor" keyword in this specification limits its URI resolution
-                    effects to the matching "$dyanmicRef" keyword, leaving the behavior
+                    the "$dynamicAnchor" keyword in this specification limits its URI resolution
+                    effects to the matching "$dynamicRef" keyword, leaving the behavior
                     of "$ref" undisturbed.
                 </t>
             </section>
@@ -774,7 +774,7 @@
                         For some by-reference applicators, such as
                         <xref target="ref">"$ref"</xref>, the referenced schema can be determined
                         by static analysis of the schema document's lexical scope.  Others,
-                        such as "$dyanmicRef" (with "$dyanmicAnchor"),  may make use of dynamic
+                        such as "$dynamicRef" (with "$dynamicAnchor"),  may make use of dynamic
                         scoping, and therefore only be resolvable in the process of evaluating
                         the schema with an instance.
                     </t>
@@ -1435,7 +1435,7 @@
                 <section title="Schema References" anchor="references">
                     <t>
                         Several keywords can be used to reference a schema which is to be applied to the
-                        current instance location. "$ref" and "$dyanmicRef" are applicator
+                        current instance location. "$ref" and "$dynamicRef" are applicator
                         keywords, applying the referenced schema to the instance.
                     </t>
                     <t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1509,7 +1509,7 @@
                             resolution is needed.
                         </t>
                         <t>
-                            For a full example using theses keyword, see appendix
+                            For a full example using these keyword, see appendix
                             <xref target="recursive-example" format="counter" />.
                             <cref>
                                 The difference between the hyper-schema meta-schema in pre-2019
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1407,7 +1407,7 @@
                         for details.
                     </t>
                     <t>
-                        In most cases, the normal fragment behavior both sufficies and
+                        In most cases, the normal fragment behavior both suffices and
                         is more intuitive.  Therefore it is RECOMMENDED that "$anchor"
                         be used to create plain name fragments unless there is a clear
                         need for "$dynamicAnchor".
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -623,11 +623,9 @@
                     Keywords MAY be defined with a partial value, such as a URI-reference,
                     which must be resolved against another value, such as another
                     URI-reference or a full URI, which is found through the lexical
-                    structure of the JSON document.  The "$id" core keyword and
-                    the "base" JSON Hyper-Schema keyword are examples of this sort
-                    of behavior.  Additionally, "$ref" and "$dyanmicRef" from
-                    this specification resolve their values in this way, although
-                    they do not change how further values are resolved.
+                    structure of the JSON document.  The "$id", "$ref", and
+                    "$dyanmicRef" core keywords, and the "base" JSON Hyper-Schema
+                    keyword, are examples of this sort of behavior.
                 </t>
                 <t>
                     Note that some keywords, such as "$schema", apply to the lexical
@@ -724,7 +722,8 @@
                     While custom identifier keywords are possible, vocabulary designers should
                     take care not to disrupt the functioning of core keywords. For example,
                     the "$dyanmicAnchor" keyword in this specification limits its URI resolution
-                    effects to the matching "$dyanmicRef" keyword, leaving "$ref" undisturbed.
+                    effects to the matching "$dyanmicRef" keyword, leaving the behavior
+                    of "$ref" undisturbed.
                 </t>
             </section>
             <section title="Applicators" anchor="applicators">
@@ -775,7 +774,7 @@
                         For some by-reference applicators, such as
                         <xref target="ref">"$ref"</xref>, the referenced schema can be determined
                         by static analysis of the schema document's lexical scope.  Others,
-                        such as "$dyanmicRef" and "$dyanmicAnchor",  may make use of dynamic
+                        such as "$dyanmicRef" (with "$dyanmicAnchor"),  may make use of dynamic
                         scoping, and therefore only be resolvable in the process of evaluating
                         the schema with an instance.
                     </t>
@@ -1377,8 +1376,7 @@
                         </t>
                     </section>
                 </section>
-                <section title='Defining location-independent identifiers with "$anchor"'
-                         anchor="anchor">
+                <section title="Defining location-independent identifiers" anchor="anchor">
                     <t>
                         Using JSON Pointer fragments requires knowledge of the structure of the schema.
                         When writing schema documents with the intention to provide re-usable
@@ -1387,8 +1385,32 @@
                         without requiring JSON Pointer references to be updated.
                     </t>
                     <t>
-                        The "$anchor" keyword is used to specify such a fragment.  It is an
-                        identifier keyword that can only be used to create plain name fragments.
+                        The "$anchor" and "$dynamicAnchor" keywords are used to specify such
+                        a fragments.  They are identifier keywords that can only be used to create
+                        plain name fragments, rather than absolute URIs as seen with "$id".
+                        The behavior of the created fragment is identical for both keywords.
+                    </t>
+                    <t>
+                        The base URI to which the resulting fragment is appended is the canonical
+                        URI of the schema resource containing the "$anchor" or "$dynamicAnchor"
+                        in question.  As discussed in the previous section, this is either the
+                        nearest "$id" in the same or parent schema object, or the base URI
+                        for the document as determined according to RFC 3986.
+                    </t>
+                    <t>
+                        Separately from the usual usage of URIs, "$dynamicAnchor"
+                        indicates that the fragment is an extension point when used with
+                        the "$dynamicRef" keyword.  This low-level, advanced feature
+                        makes it easier to extend recursive schemas such as the meta-schemas,
+                        without imposing any particular semantics on that extension.
+                        See the section on <xref target="dynamic-ref">"$dynamicRef"</xref>
+                        for details.
+                    </t>
+                    <t>
+                        In most cases, the normal fragment behavior both sufficies and
+                        is more intuitive.  Therefore it is RECOMMENDED that "$anchor"
+                        be used to create plain name fragments unless there is a clear
+                        need for "$dynamicAnchor".
                     </t>
                     <t>
                         If present, the value of this keyword MUST be a string and MUST start with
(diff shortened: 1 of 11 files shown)


handrews:
Note for those who saw the note that I removed from the main description: I decided not to try to do overly clever things with the hyper-schema meta-schemas, and added a commit that just updates them (and the other meta-schemas) more or less as they are.  I might come back to the idea (which involved both the hyper-schema meta-schema and the links schema being dynamically extended) but TBH it's pretty complicated and of questionable value.

Let's focus on the essentials for now.

jsonschema-core.xml
```diff
@@ -1447,103 +1464,59 @@
...
-                            In that case, the dynamic scope is examined to determine a new base URI,
-                            and the URI-reference in "$dynamicRef" is re-evaluated against that
-                            base URI.  Unlike base URI changes with "$id", changes with
-                            "$dynamicAnchor" are calculated each time a "$dynamicRef" is
-                            resolved, and do not impact any other keywords.
+                            If the initially resolved starting point URI includes a fragment that
+                            was created by the "$dynamicAnchor" keyword, the initial URI MUST be
+                            replaced by the URI (including the fragment) for the outermost schema
+                            resource in the <xref target="scopes">dynamic scope</xref> that defines
+                            an identically named fragment with "$dynamicAnchor".
```
Relequestual, line 1497: I'm going to want an example (for me) of what this means. It may be it's simplt too late in the day for me to parse this paragraph, but actually that might not be a bad measure.

handrews: @Relequestual does the example in the appendix not help?  I would like to work on that example until it is clear rather than try to come up with a different one.

handrews:
Also, I'll leave this open for the full 14-day period for major changes, but once approved I'll work on #849 and #857 on the assumption that this or something very much like it will get merged.  That work (about schema loading) needed to account for whether the old runtime base URI switching was present, which made the loading+runtime process substantially more complex.

jsonschema-core.xml
Relequestual: After taking another read, it makes sense. You just have to read it in the context of the previous paragraph. It may well have been I just failed to do that with my first reading.

The example is fine.

jsonschema-core.xml
```diff
@@ -1387,8 +1385,32 @@
                         without requiring JSON Pointer references to be updated.
                     </t>
                     <t>
-                        The "$anchor" keyword is used to specify such a fragment.  It is an
-                        identifier keyword that can only be used to create plain name fragments.
+                        The "$anchor" and "$dynamicAnchor" keywords are used to specify such
+                        a fragments.  They are identifier keywords that can only be used to create
```
jdesrosiers, line 1389: typo: "such a fragments"

jsonschema-core.xml
```diff
@@ -640,12 +638,12 @@
                     with an instance document.  The outermost dynamic scope is the
                     root schema of the schema document in which processing begins.
                     The path from this root schema to any particular keyword (that
-                    includes any "$ref" and "$recursiveRef" keywords that may have
+                    includes any "$ref" and "$dyanmicRef" keywords that may have
```
karenetheridge, line 641: typo: "dyanmicRef" -- this occurs several times throughout the document (also "dyanmicAnchor")

handrews: @karenetheridge thanks- I must have messed up one of the merge conflict resolutions

jsonschema-core.xml
```diff
@@ -1447,103 +1464,59 @@
...
-                            and the URI-reference in "$recursiveRef" is re-evaluated against that
-                            base URI.  Unlike base URI changes with "$id", changes with
-                            "$recursiveAnchor" are calculated each time a "$recursiveRef" is
-                            resolved, and do not impact any other keywords.
+                            Otherwise, its behavior is identical to "$ref", and no runtime
+                            resolution is needed.
                         </t>
                         <t>
-                            For an example using these keyword, see appendix
+                            For a full example using theses keyword, see appendix
```
karenetheridge, line 1512: "theses"

jsonschema-core.xml
Relequestual: Fixed

Relequestual:
I may have fix / fixed typo... If only I had been consistent in that... =]

Relequestual:
@karenetheridge can you review this again please? =]

jsonschema-core.xml
```diff
@@ -1387,8 +1385,32 @@
...
+                        Separately from the usual usage of URIs, "$dynamicAnchor"
+                        indicates that the fragment is an extension point when used with
+                        the "$dynamicRef" keyword.  This low-level, advanced feature
+                        makes it easier to extend recursive schemas such as the meta-schemas,
+                        without imposing any particular semantics on that extension.
+                        See the section on <xref target="dynamic-ref">"$dynamicRef"</xref>
+                        for details.
+                    </t>
+                    <t>
+                        In most cases, the normal fragment behavior both sufficies and
```
karenetheridge, line 1410: sufficies -> suffices

jsonschema-core.xml
```diff
@@ -1387,8 +1385,32 @@
...
+                        is more intuitive.  Therefore it is RECOMMENDED that "$anchor"
```
karenetheridge, line 1411: "more readable"? "less confusing"? I'm not sure that $recursiveAnchor is less *intuitive*, as much as more confusing and causes more cognitive (and parsing) overhead.

jsonschema-core.xml
```diff
@@ -3793,9 +3801,9 @@ https://example.com/schemas/common#/$defs/count/minimum
                             <t>Array-value "items" functionality is now "prefixItems"</t>
                             <t>"items" subsumes the old function of "additionalItems"</t>
                             <t>"contains" and "unevaluatedItems" interactions now specified</t>
-                            <t></t>
-                            <t></t>
-                            <t></t>
+                            <t>Rename $recursive* to $dynamic*</t>
+                            <t>$dynamicAnchor defines a fragment like $anchor</t>
+                            <t>$dynamic* no longer use runtime base URI determination</t>
```
karenetheridge, line 3806: $dynamic* never existed before, so it cannot "no longer" do anything.  Perhaps "$dynamic* does not use runtime base URI determination like $recursive* did"?

links.json
```diff
@@ -2,6 +2,7 @@
     "$schema": "https://json-schema.org/draft/2019-09/hyper-schema",
     "$id": "https://json-schema.org/draft/2019-09/links",
     "title": "Link Description Object",
+
```
karenetheridge, line 5: intentional?

meta/applicator.json
```diff
@@ -4,47 +4,47 @@
     "$vocabulary": {
         "https://json-schema.org/draft/2019-09/vocab/applicator": true
     },
-    "$recursiveAnchor": true,
+    "$dynamicAnchor": "meta",
```
karenetheridge, line 7: I'm not sure if "meta" is the best name for this, but I am flailing at coming up with a better alternative.  "subschema" maybe?

links.json
Relequestual: Yeah, it's the same with other schema files.
@handrews why do we do this?

meta/applicator.json
Relequestual: I think this is named such because it's in the "meta" applicator schema.
OK to close this?

handrews: It's called "meta" because it's the anchor for metaschemas.  It must be the same name in all of them.  It needs to not collide with any other casually likely name that a schema author might use elsewhere.  What else would you call it?

jsonschema-core.xml
Relequestual: I've made a change. If you consider that reasonable, please resolve the comment =]

links.json
Relequestual: Readability. This is fine and intended.

meta/applicator.json
Relequestual: ah right, of course. I'm sure I read this and appreciated that reason for naming at some point in the past... =]

Relequestual: I'm going to resolve this. I don't think there's any action here.

karenetheridge: ok. "metaschema" might be more clear than just "meta", but I do get that it's just a name, and naming things is subjective. :)

links.json
karenetheridge: I was inquiring about the extra whitespace.

jsonschema-core.xml
karenetheridge: I don't have an option to resolve this comment - perhaps because I am not a member of this repo? however: LGTM.

jsonschema-core.xml
Relequestual: I think this is fine. We can bikeshed phrasing later if required.

jsonschema-core.xml
Relequestual: I believe this is resolved

Relequestual:
@handrews If you approve of the changes made since your initial PR, we're basically ready.

I note that after seeing #940, the output schema needs to be updated to reflect, but that can be a separate issue.

handrews:
@Relequestual looks great.  I had to stare at `$dyanmic` for a **long** time before I could even figure out what changed 😂 
I'll go ahead and merge.

PR #940, opened 2020-05-27
output schema fixes

karenetheridge:
Fixed some errors in output/schema.json

### DIFF
--- a/output/schema.json
+++ b/output/schema.json
@@ -50,14 +50,14 @@
               {
                 "properties": {
                   "keywordLocation": {
-                    "pattern": ".*/\\$ref/.*"
+                    "pattern": "/\\$ref/"
                   }
                 }
               },
               {
                 "properties": {
                   "keywordLocation": {
-                    "pattern": ".*/\\$recursiveRef/.*"
+                    "pattern": "/\\$recursiveRef/"
                   }
                 }
               }
--- a/output/schema.json
+++ b/output/schema.json
@@ -46,7 +46,7 @@
         },
         {
           "if": {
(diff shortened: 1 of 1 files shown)


gregsdennis:
@karenetheridge Could you link to the portions of the spec that this file references, please?  Just to make sure that this change is valid.

karenetheridge:
re absoluteKeywordLocation: https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.10.3.2

re patterns: https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.6.4

gregsdennis:
@karenetheridge thanks, but I mean the output portions.  Where are the schema and instance that generated the output (they're in the spec, but I'm too busy to find them right now)?

output/schema.json
```diff
@@ -46,18 +46,18 @@
         },
         {
           "if": {
-            "oneOf": [
+            "anyOf": [
```
gregsdennis, line 49: Why this change?  Are you expecting `$ref` and `$recursiveRef` to both appear in the keyword location?

... I suppose they could....

okay.

karenetheridge: sure, if there is a $recursiveRef in the subschema at the other end of a $ref, it could certainly happen.

I have a test case with `.../$ref/$ref/$ref/$ref/..` :) -- https://github.com/karenetheridge/JSON-Schema-Draft201909/blob/master/t/errors.t#L788

Relequestual:
@gregsdennis If you'll approve, let's merge this?

Issue #941, opened 2020-05-28
Allow "contentMediaType" without "type"


handrews:
Its useful to be able to use "contentMediaType" in contexts where binary media types are present.  It does not make sense to use `"type": "string"` with binary data, because such data is not a JSON string no matter how much you try to map it into the data model.

A schema without a type can effectively be a hint not to worry about a detailed mapping.  It is a way to use JSON Schema across media types without excluding binary content.

See https://github.com/OAI/OpenAPI-Specification/pull/2200 for use cases.

handrews:
Fixed by #970.

PR #942, opened 2020-05-31
Clarify data model regarding binary data

handrews:
It is advantageous in hypermedia environments to apply untyped schemas to binary data, despite the lack of a truly suitable mapping into the data model.  This removes some JSON-specific language and provides additional guidance on media types.

See https://github.com/OAI/OpenAPI-Specification/pull/2200 for the detailed use case.  Prior to that PR, OAS was modeling unencoded binary data as strings with a custom `format`.

The expansion of contentMediaType is motivated by the need to indicate media types for binary resources in hypermedia environments.

Existing usage (e.g. OpenAPI 3.0) considered unencoded binary data to be strings, but such data violates the expectations of JSON strings. A better approach is to purely indicate the media type and avoid constraining the instance by JSON type as no JSON type is suitable.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -198,13 +198,17 @@
 
             <section title="Instance">
                 <t>
-                    A JSON document to which a schema is applied is known as an "instance".
+                    A document to which a schema is applied is known as an "instance".
                 </t>
 
                 <section title="Instance Data Model">
                     <t>
-                        JSON Schema interprets documents according to a data model. A JSON value
-                        interpreted according to this data model is called an "instance".
+                        JSON Schema interprets documents according to a data model. A value
+                        interpreted according to this data model is called an "instance".  JSON
+                        documents map trivially into the data model, with a few exceptions as
+                        noted below.  Documents of other media types MAY be treated as instances
+                        if a suitable application-defined mapping of the media type into the
+                        data model can be determined.
                     </t>
                     <t>
                         An instance has one of six primitive types, and a range of possible values
@@ -218,6 +222,12 @@
                             <t hangText="number:">An arbitrary-precision, base-10 decimal number value, from the JSON "number" value</t>
                             <t hangText="string:">A string of Unicode code points, from the JSON "string" value</t>
                         </list>
+                    </t>
+                    <t>
+                        Binary data MAY be treated as an instance, however no data type in the data model
+                        is suitable.  Therefore, only schemas such as the empty schema that do not
+                        constrain the type can be considered to pass.  Rationales for and behavior of
+                        binary data as an instance SHOULD be defined by the consuming application.
                     </t>
                     <t>
                         Whitespace and formatting concerns, including different lexical
@@ -3796,7 +3806,7 @@ https://example.com/schemas/common#/$defs/count/minimum
                             <t></t>
                             <t></t>
                             <t></t>
-                            <t></t>
+                            <t>Clarify applicability to non-JSON media types</t>
                             <t></t>
                             <t></t>
                             <t></t>
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -913,7 +913,9 @@
             <section title="Foreword">
                 <t>
                     Annotations defined in this section indicate that an instance contains
-                    non-JSON data encoded in a JSON string.
+                    non-JSON data encoded in a JSON string.  Additionally, they can be used in
+                    the context of resources of various media types to indicate binary resources
+                    not otherwise describable by JSON Schema.
                 </t>
                 <t>
                     These properties provide additional information required to interpret JSON data
@@ -945,14 +947,21 @@
                     consumer than that which processed the containing document.
                 </t>
                 <t>
-                    All keywords in this section apply only to strings, and have no
-                    effect on other data types.
+                    All keywords in this section generally apply to strings, and have no
+                    effect on other JSON data types.  Additionally, they MAY be used without
+                    type information when describing resources of other media types, subject
+                    to certain restrictions.
                 </t>
                 <t>
                     Implementations MAY offer the ability to decode, parse, and/or validate
-                    the string contents automatically.  However, it MUST NOT perform these
+                    the string contents automatically.  However, they MUST NOT perform these
                     operations by default, and MUST provide the validation result of each
-                    string-encoded document separately from the enclosing document.  This
+                    string-encoded document separately from the enclosing document.  In particular,
+                    these keywords, including "contentSchema", MUST NOT cause the containing schema
+                    to fail validation.
+                </t>
+                <t>
+                    The optional automatic decoding, parsing, and validating
                     process SHOULD be equivalent to fully evaluating the instance against
                     the original schema, followed by using the annotations to decode, parse,
                     and/or validate each string-encoded document.
@@ -1005,7 +1014,14 @@
                 <t>
                     If the instance is a string, this property indicates the media type
                     of the contents of the string.  If "contentEncoding" is present,
-                    this property describes the decoded string.
+                    this property describes the decoded string.  If the "type" keyword is
+                    absent, this keyword MAY be interpreted as describing an unencoded binary
+                    resource.  The exact meaning and behavior of this untyped usage is
+                    application-defined.
+                    <cref>
+                        For an example of application-defined untyped usage,
+                        see the forthcoming OpenAPI Specification v3.1
+                    </cref>
                 </t>
                 <t>
                     The value of this property MUST be a string, which MUST be a media type,
@@ -1015,8 +1031,9 @@
 
             <section title="contentSchema">
                 <t>
-                    If the instance is a string, and if "contentMediaType" is present, this
-                    property contains a schema which describes the structure of the string.
+                    If the instance is a string or an untyped binary resource,
+                    and if "contentMediaType" is present, this property contains a schema
+                    which describes the structure of the string.
                 </t>
                 <t>
                     This keyword MAY be used with any media type that can be mapped into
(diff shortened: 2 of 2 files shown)


awwright:
I'm trying to understand what this gets us.

Trying to use JSON Schema with non-JSON media types isn't prohibited, just undefined. Other people may attempt to define behavior in these cases, though. I don't think it needs any special treatment.

There's also this paragraph that specifies how JSON Schema applies to media types merely compatible with JSON:

> JSON Schema is only defined over JSON documents. However, any document or memory structure that can be parsed into or processed according to the JSON Schema data model can be interpreted against a JSON Schema, including media types like CBOR.

handrews:
@awwright did you look at the OAS PR that I linked?

jsonschema-core.xml
```diff
@@ -219,6 +223,12 @@
                             <t hangText="string:">A string of Unicode code points, from the JSON "string" value</t>
                         </list>
                     </t>
+                    <t>
+                        Binary data MAY be treated as an instance, however no data type in the data model
+                        is suitable.  Therefore, only schemas such as the empty schema that do not
+                        constrain the type can be considered to pass.  Rationales for and behavior of
```
ssilverman, line 229: I _think_ commas may be required: "Rationales for, and behavior of,"

jsonschema-core.xml
```diff
@@ -219,6 +223,12 @@
                             <t hangText="string:">A string of Unicode code points, from the JSON "string" value</t>
                         </list>
                     </t>
+                    <t>
+                        Binary data MAY be treated as an instance, however no data type in the data model
+                        is suitable.  Therefore, only schemas such as the empty schema that do not
```
ssilverman, line 228: The added paragraph above this one says that it's possible there's some application-defined mapping from binary data to the JSON data model. The wording of this paragraph seems to waffle between "MAY be treated as an instance" and "can't do anything unless it's an empty schema". Maybe add some language that suggests that a schema can be applied if that data model is applied. For example (changes in bold), "Binary data MAY be treated as an instance, however no data type in the data model is **directly** suitable. Therefore, only schemas such as the empty schema that do not constrain the type can be considered to pass **outright**."

I'm sure there's some better language, but this paragraph feels like it's fighting with itself if there's no acknowledgement that there exists that "application-defined mapping" in the previous paragraph.

jsonschema-validation.xml
```diff
@@ -945,14 +947,21 @@
                     consumer than that which processed the containing document.
                 </t>
                 <t>
-                    All keywords in this section apply only to strings, and have no
-                    effect on other data types.
+                    All keywords in this section generally apply to strings, and have no
+                    effect on other JSON data types.  Additionally, they MAY be used without
+                    type information when describing resources of other media types, subject
+                    to certain restrictions.
```
ssilverman, line 953: Could one or two examples of restrictions be given here?

jsonschema-validation.xml
```diff
@@ -945,14 +947,21 @@
...
                     operations by default, and MUST provide the validation result of each
-                    string-encoded document separately from the enclosing document.  This
+                    string-encoded document separately from the enclosing document.  In particular,
+                    these keywords, including "contentSchema", MUST NOT cause the containing schema
+                    to fail validation.
+                </t>
+                <t>
+                    The optional automatic decoding, parsing, and validating
                     process SHOULD be equivalent to fully evaluating the instance against
                     the original schema, followed by using the annotations to decode, parse,
```
ssilverman, line 966: Should this say instead, "The optional automatic decoding, parsing, and validating process SHOULD be equivalent to fully evaluating the instance against process SHOULD be equivalent to fully evaluating **an** instance against **a** schema, followed by using the annotations to decode, parse the original schema". The thing is, this section is about applying schemas or rules to encoded "content". if we use "evaluating **the** instance against **the** original schema", then it feels like there's confusion between the schema containing all this stuff vs. the stuff inside "content". Does this make sense?

jsonschema-validation.xml
```diff
@@ -1005,7 +1014,14 @@
...
                     If the instance is a string, this property indicates the media type
                     of the contents of the string.  If "contentEncoding" is present,
-                    this property describes the decoded string.
+                    this property describes the decoded string.  If the "type" keyword is
+                    absent, this keyword MAY be interpreted as describing an unencoded binary
+                    resource.  The exact meaning and behavior of this untyped usage is
+                    application-defined.
+                    <cref>
+                        For an example of application-defined untyped usage,
+                        see the forthcoming OpenAPI Specification v3.1
```
ssilverman, line 1023: Should a link to this be added?

jsonschema-validation.xml
```diff
@@ -1015,8 +1031,9 @@
 
             <section title="contentSchema">
                 <t>
-                    If the instance is a string, and if "contentMediaType" is present, this
-                    property contains a schema which describes the structure of the string.
+                    If the instance is a string or an untyped binary resource,
```
ssilverman, line 1034: What about: "If the instance is a string or an untyped binary resource having an application-defined mapping to the data model"

jsonschema-validation.xml
Relequestual: There isn't a link to provide yet. A cref kinda doesn't matter too much.

Relequestual:
I've read this PR, it's content, and the associated OAS PR, but I still can't understand what this is about.
I can explain which parts I don't understand, but I don't know if doing so is a good use of your time.

That being said, if you feel you'd like me to understand and approve this seemingly simple PR, I'm happy to detail what I'm missing.

jsonschema-core.xml
Relequestual: Not in this instance.

handrews:
Closes #941 (which I forgot to link)

jsonschema-validation.xml
```diff
@@ -1005,7 +1014,14 @@
                 <t>
                     If the instance is a string, this property indicates the media type
                     of the contents of the string.  If "contentEncoding" is present,
-                    this property describes the decoded string.
+                    this property describes the decoded string.  If the "type" keyword is
+                    absent, this keyword MAY be interpreted as describing an unencoded binary
+                    resource.  The exact meaning and behavior of this untyped usage is
+                    application-defined.
```
awwright, line 1020: This is effectively permitting values outside the data model, which sort of defeats the point of having a data model, doesn't it?

That's not to say this should be illegal, but maybe we can describe this inside the data model section instead? Something like "JSON Schema can be applied to data that is a superset of this data model; in this case, the instance may be outside any of the six primitive types."

handrews: @awwright I like your idea of simply extending the data model.  I'd probably recommend that we just add `binary` rather than throwing it entirely open, but I could be persuaded on that point.

What's going on here is that I'm recognizing a thing that a significant number of people (OpenAPI users) were already doing in the wild, in a way that was _more_ problematic.  OAS 3.0 treats binary data as `"type": "string"` plus a custom `format`, which is definitely wrong.  Unencoded binary data cannot be considered a JSON string in any reasonable universe.

Since OAS 3.1 is picking up the `content*` keywords, they were willing to drop the custom `format` and use of `"type": "string"` as long as there was something that they could recommend for this use case.  This is a scenario where nothing's going to be validated, but a JSON Schema is used for descriptive purposes.  Which is fine with `content*` because they don't do validation anyway.  They are annotations, so like all annotations, their behavior is application-defined.  This is particularly unusual behavior so it is called out.  Also, we want it called out so people don't come up with worse "solutions" again.

Relequestual: I would argue, let's just call binary for now, and if other use cases present, then open it up later.
I'd like to avoid speculation and stay grounded on use cases right infront of our faces first.
=]

handrews: @Relequestual @awwright just to clarify, we're now proposing to add `binary` as a type?

I feel like this should get an issue for visibility so that others can comment on it.  It's a major change from what I worked out with OAS.  I'm reasonably OK with making the change, but it's kind of a big deal.

There are also alternatives such as a `"binary": true` approach.  This would avoid breaking compatibility with anything that hardcodes the current set of valid `type` values.

I will file an issue for this.

awwright:
(re <https://github.com/json-schema-org/json-schema-spec/pull/942#discussion_r456938350> from review)

@handrews I don't think we should add things to JSON Schema that are not supported by `application/json`. And JSON does not support binary data, of course.

However whatever we write, the potential applications should be clear. If we just say "binary data" it's not clear to me how you get binary data into an instance. So a brief example may be in order.

Or we just say "Schemas may be applied to supersets of the data model defined here; we won't go into all the details but all of the rules still work as written: type-specific rules such as `"minimum"` don't apply, but generic keywords such as `"allOf"` and `"format"` do."

handrews:
@awwright I'm trying to figure out how that's different than what I'm doing.  Is the main goal here to change this:

> Documents of other media types MAY be treated as instances
                        if a suitable application-defined mapping of the media type into the
                        data model can be determined.

to say "superset of the data model"?

awwright:
@handrews That paragraph, as it is, doesn't suggest to me that JSON Schema supports a superset of the data model/domain.

So, to rephrase, I have one of two suggestions:

1. Remove the reference to "binary data" and just say "superset", and let other documents like OAI get to decide what that means, or

2. Include some example of a document that supports "binary data" because it's not obvious what that means in a document that's only supposed to be talking about `application/json` (or other strictly equivalent serializations).

Relequestual:
@awwright given @handrews tumbedup your suggestion, care to make a change suggestion to this PR?
When making a review on code, clicking the +- button allows you to make direct change suggestions, which can then be easily added to the PR.

handrews:
@Relequestual @awwright yes, please feel free to do this.  I'll get to it at some point but TBH personal things keep coming up 😐

awwright:
Mission accepted

awwright:
I haven't forgotten about this... I've re-read Core a couple times and I don't think there's anything that prohibits using annotations for non-JSON documents. So would anyone be adverse to me modifying the the Data Model and some other sections to offer better advice for these situations?

I'd issue this as a new PR because that'll be easier for me, than trying to figure out what parts of this PR to amend.

handrews:
@awwright sounds good to me, I'll close this out.

Issue #900, opened 2020-04-29
Why doesn't the applicator meta-schema define "type"?


ssilverman:
It seems like `"type": ["object", "boolean"]` is missing.

Julian:
Which one? Looks present [here](https://github.com/json-schema-org/json-schema-spec/blob/master/meta/validation.json#L10)

ssilverman:
Oops. I don't know why I put "validator". I meant "applicator".

handrews:
@ssilverman it was either an accident, or I decided that I wanted the vocabulary schemas to constrain the keywords but not the overall type (deferring that to the dialect meta-schema, meaning the schemas that are not under `/meta/` but are instead alongside it).  I don't remember, and either way it's not consistent.

Relequestual:
Given the other schemas in `/meta/` have it defined, it makes sense to add it ot the validation vocabulary meta-schema also.
@ssilverman Feel free to PR this.

Relequestual:
This update needs to be also made for the 2019-09 schemas, and then released to the website.

karenetheridge:
PR created in #968.

ssilverman:
@Relequestual I’m confused. This is already a 2019-09 schema? What else needs to change?

ssilverman:
Ok, I've sorted this out. It's a different branch.

Relequestual:
This is now resolved.
This change along with a few others have been cherrypicked into a PR against the 2019-09 branch.
Then the 2019-09 submodule for the website has been updated to the latest branch, so everyone can get the new meta-schemas.

ssilverman:
I believe this issue can be closed.

PR #945, opened 2020-06-11
[954] Fix $recursiveAnchor in schema to match spec

### LINKED ISSUE #954: Spec disagrees with meta-schema for $recursiveAnchor
The spec implies that $recursiveAnchor can be `false`, however the meta-schema limits its presence to having the value `true`.

ssilverman:
The spec allows values of false. This fixes the schema to match the spec.

Closes #954

### DIFF
--- a/meta/core.json
+++ b/meta/core.json
@@ -33,7 +33,6 @@
         },
         "$recursiveAnchor": {
             "type": "boolean",
-            "const": true,
             "default": false
         },
         "$vocabulary": {


ssilverman:
See https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2.4.2.2:
> If this keyword is set to false, the base URI is unchanged.

Also, the spec doesn't require the value to be true.

Relequestual:
Thanks!
Could you also update the 2019-09 meta schema please?
Once done, I'll merge.

karenetheridge:
>  Could you also update the 2019-09 meta schema please?

what update is needed? $recursiveAnchor isn't defined in the main meta-schema document.

Relequestual:
What appears on the website for fetching, the schemas for previous versions are included via git submodule.
This can be eased if the branch on here gets updated with the fix too.
This also applies for https://github.com/json-schema-org/json-schema-spec/issues/900 but I just didn't remember.

Relequestual:
I guess what we actually need is a second PR, because this one can be merged.

karenetheridge:
PR created in #968.

PR #950, opened 2020-06-19
Issue-949: Fixing inconsistencies in output schema

stan-sb:
1. Changing "oneOf" to "anyOf", because basic, detailed and verbose have the same definition, oneOf would fail validation.
2. Adding "error" keyword and making either "error" or "errors" required in case "valid" is false.

PR #967, opened 2020-08-13
Fix output schema impossible validation

zo:
basic, detailed and verbose output can't be valid against exactly one subschema because they share the same definition


zo:
Sorry, duplicate of #950

Issue #902, opened 2020-04-30
Refer to collections of vocabularies as described by meta-schemas as dialects


handrews:
A _dialect_ is a collection of vocabularies plus a set of validation rules for that collection.

A meta-schema that uses `$vocabulary` describes a dialect, and is considered a "dialect meta-schema."  `https://json-schema.org/draft/2019-09/schema` and `https://json-`schema.org/draft/2019-09/meta-schema` are dialect meta-schemas.

A meta-schema describing a dialect consisting of a single vocabulary and its standard syntax is considered to be a "vocabulary meta-schema" as well as being a "dialect meta-schema." `https://json-schema.org/draft/2019-09/meta/validation` and `https://json-schema.org/draft/2019-09/meta/hyper-schema` are vocabulary meta-schemas.

We should make this clear in the spec as the "dialect" terminology has proven helpful in discussions elsewhere.

Relequestual:
Relevent: 

I note that while we talk about how `$schema` represents an identifier for a "feature set", we don't define what that phrase actually means. https://github.com/json-schema-org/json-schema-spec/blob/b615488327ede58b52d21360865714b50e141969/jsonschema-core.xml#L1156-L1158

Relequestual:
Also relevant https://github.com/json-schema-org/json-schema-spec/blob/b615488327ede58b52d21360865714b50e141969/jsonschema-core.xml#L2096-L2097

handrews:
@Relequestual 

> I note that while we talk about how $schema represents an identifier for a "feature set", we don't define what that phrase actually means.

Yup, that's exactly what this is about!  We should say "dialect" rather than "feature set", although the phrase "feature set" may be useful in explaining what a dialect is.  But obviously it is not sufficient on its own to explain it.

I have a spreadsheet of dialect vs vocabulary and I have no idea why I didn't link it when I filed this.

https://docs.google.com/spreadsheets/d/18SIXnzyjXTJZgqeo5W-qIEwq-bNKXb5M76Pq_47r2Is/edit?usp=sharing

Relequestual:
I expect this will be one of the later issues to get closed out when most others for the milestone are done.
Part of a final pass, if you like.

Relequestual:
I can't quite decide if the PR was enough.
I guess all we can ask is we try to confirm this with a final pass.

handrews:
[...]

PR #970, opened 2020-08-17
Clarify how JSON Schema works with a superset of the defined data model

awwright:
Successor to #942

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -200,6 +200,15 @@
                 <t>
                     A JSON document to which a schema is applied is known as an "instance".
                 </t>
+                <t>
+                    JSON Schema is defined over "application/json" or compatible documents,
+                    including media types with the "+json" structured syntax suffix.
+                </t>
+                <t>
+                    Among these, this specification defines the "application/schema-instance+json"
+                    media type which defines handling for fragments in the URI,
+                    and the "schema" media type parameter.
+                </t>
 
                 <section title="Instance Data Model">
                     <t>
@@ -243,25 +252,6 @@
                     </t>
                 </section>
 
-                <section title="Instance Media Types">
-                    <t>
-                        JSON Schema is designed to fully work with "application/json" documents,
-                        as well as media types using the "+json" structured syntax suffix.
-                    </t>
-                    <t>
-                        Some functionality that is useful for working with schemas is
-                        defined by each media type, namely media type parameters and
-                        URI fragment identifier syntax and semantics.  These features are
-                        useful in content negotiation and in calculating URIs for specific
-                        locations within an instance, respectively.
-                    </t>
-                    <t>
-                        This specification defines the "application/schema-instance+json"
-                        media type in order to allow instance authors to take full advantage
-                        of parameters and fragment identifiers for these purposes.
-                    </t>
-                </section>
-
                 <section title="Instance Equality">
                     <t>
                         Two JSON instances are said to be equal if and only if they are of the same type
@@ -286,6 +276,22 @@
                         there is no way to define multiple properties with the same key,
                         and mere formatting differences (indentation, placement of commas, trailing
                         zeros) are insignificant.
+                    </t>
+                </section>
+
+                <section title="Non-JSON Instances">
+                    <t>
+                        It is possible to use JSON Schema with a superset of the JSON Schema data model,
+                        where an instance may be outside any of the six JSON data types.
+                    </t>
+                    <t>
+                        In this case, annotations still apply; but validation keywords will not be useful,
+                        as they will always pass or always fail.
+                    </t>
+                    <t>
+                        A custom vocabulary may define support for a superset of the core data model.
+                        The schema itself may only be expressible in this superset;
+                        for example, to make use of the "const" keyword.
                     </t>
                 </section>
             </section>


jsonschema-core.xml
```diff
@@ -288,6 +278,36 @@
...
+                        A custom vocabulary may define support for a superset of the core data model.
+                        The schema itself may only be expressible in this superset;
+                        for example, to make use of the "const" keyword.
+                    </t>
+                    <t>
+                        New data model types may be defined as either supersets of an existing type
+                        (similar to how "number" is a superset of "integer"),
+                        or a type disjoint from the core data model altogether (like "binary").
+                        Existing keywords may be updated to define behavior for the new range of values,
+                        but vocabularies that intend to be compatible with JSON Schema Validation should not
```
Relequestual, line 301: /should not/MUST NOT/ ?

jsonschema-core.xml
```diff
@@ -200,6 +200,15 @@
...
                     A JSON document to which a schema is applied is known as an "instance".
                 </t>
+                <t>
+                    JSON Schema is defined over "application/json" or compatible documents,
+                    including media types with the "+json" structured syntax suffix.
+                </t>
+                <t>
+                    Among these, this specification defines the "application/schema-instance+json"
+                    media type which defines handling for fragments in the URI,
+                    and the "schema" media type parameter.
```
Relequestual, line 210: I know this was already in the spec... but what does this actually mean? =/

jsonschema-core.xml
awwright: I'll have to review the spec again, but I'd guess vocabularies are allowed to not use the validation vocabulary.

I wonder if it would save implementation effort to not allow this?

jsonschema-core.xml
```diff
@@ -288,6 +278,36 @@
...
+                    <t>
+                        It is possible to use JSON Schema with a superset of the JSON Schema data model,
+                        where an instance may be outside any of the six JSON data types.
+                    </t>
+                    <t>
+                        In this case, annotations still apply; but validation keywords will not be useful,
+                        as they will always pass or always fail.
+                    </t>
+                    <t>
+                        A custom vocabulary may define support for a superset of the core data model.
```
karenetheridge, line 292: - does this paragraph belong in the section "Non-JSON Instances"?

- does this paragraph mean that a custom vocabulary may be expressed in terms of another custom vocabulary? Or that this is *not* permitted and the core vocabulary only may be used? (I strongly prefer the latter, as implementations could otherwise be stuck in a recursive chain of "load vocabulary X first to understand vocabulary Y". I'd at least want someone to show an example of this being done (e.g. in an appendix) to demonstrate that it allowed and provide a potential test case for implementations.

jsonschema-core.xml
```diff
@@ -288,6 +278,36 @@
...
+                        The schema itself may only be expressible in this superset;
+                        for example, to make use of the "const" keyword.
+                    </t>
+                    <t>
+                        New data model types may be defined as either supersets of an existing type
```
karenetheridge, line 297: Please clarify if that means that existing keywords can be extended, or if new keywords must be created to extend functionality. This paragraph almost implies that a vocabulary might extend the existing 'type' keyword to accept more values than number, integer, boolean, string, null, object and array.

jsonschema-core.xml
karenetheridge: It is "application/schema+json" that defines how URI fragments are handled. "application/schema-instance+json" makes no further modifications.

jsonschema-core.xml
awwright: >  does this paragraph belong in the section "Non-JSON Instances"?

If an implementations adds support for a superset of the data model, that means it's supporting data that cannot be represented by `application/json`. The way for a schema to declare using this superset would be with a custom `$schema` keyword.

> does this paragraph mean that a custom vocabulary may be expressed in terms of another custom vocabulary

I haven't really thought that far out, but that seems like a concern for the vocabularies section.

(I hope I'm understanding the question correctly.)

jsonschema-core.xml
awwright: > This paragraph almost implies that a vocabulary might extend the existing 'type' keyword to accept more values than number, integer, boolean, string, null, object and array.

This is exactly the intent.

karenetheridge: FWIW I think that's terrible. It means that the implementation of every core keyword now needs to check if another vocabulary has extended that keyword. For example: when the validator vocabulary evaluates "type": "my_new_type"`, it can no longer generate an error "unrecognized type 'my_new_type'" but instead.. what?  silently ignore it? call the other vocabulary's evaluation code to handle it? Is it going to know what vocabulary that is?

awwright: Vocabulary behavior aside, it makes sense to me that if I'm going to produce an implementation that's a superset of JSON Schema, that implementation will allow schemas like `{ type: "binary" }`.

I'll take another look and see if I can address this though.

I feel like this has come up once or twice before, or is this just déjà vu?

karenetheridge: > it makes sense to me that if I'm going to produce an implementation that's a superset of JSON Schema...

I agree.  Perhaps 'type' is the only keyword where it might make sense to expand the syntax of an existing keyword. If so, we can handle this thusly:
* update the documentation for 'type' to indicate it should explicitly ignore any unrecognized type names, to allow for new vocabularies to add their own 'type' keyword (and indicate that those vocabularies should also ignore all the core types and any other unrecognized values) - which would allow vocabularies to share the 'type' keyword between them
* explicitly document that vocabularies cannot use the same keyword name as a core vocabulary (with the single exception of 'type')
* leave as undefined behaviour what happens when two custom vocabularies choose to use the same keyword (some vocabularies may choose to error on a collision; others may wish to have overlapping behaviour as with 'type').

jdesrosiers: > I feel like this has come up once or twice before, or is this just déjà vu?

I've brought this up wrt the OAS 3.1 vocabulary definition. I know the spec says vocabularies are allowed to modify keywords, but this is incredibly problematic to implement in a system that allows users to provide their own vocabulary implementations. My implementation doesn't support adding/modifying keyword semantics other than what can be constrained by using an alternate meta-schema (such as disallowing the array form of the `type` keyword).

However, you can have a vocabulary that defines a keyword called `type` with the new semantics. That means that if your vocabulary provides a different `type` keyword, and you include the validation vocabulary, you now have two keywords with the name `type`. Therefore, you also have to provide a custom validation vocabulary definition that copies the standard one except leaves out `type`.

Adding a custom `type` keyword is something that can be supported, but you can't also use the standard validation vocabulary. If the new type requires modifications to other keywords, those would also conflict with the standard vocabularies they appear in.

jsonschema-core.xml
handrews: @Relequestual you (technically) cannot use any media type parameters with `application/json`, nor can you use JSON Pointer fragments (expressly forbidden in RFC 6901 JSON Pointer).  So `application/schema-instance+json` means "just like `application/json` except you can use JSON Pointer fragments and the schema media type parameter."

In practice, everyone ignores these restrictions, but we should provide a formal way to work with them.

@karenetheridge `application/schema+json` and `application/schema-instance+json` are independently defined.  There is an RFC for `+json` but otherwise media types are not related even if they look like it.  We, of course, define them identically for fragment and media type parameter purposes, but they are separate defintions.

jsonschema-core.xml
handrews: @awwright in the previous PR you wrote:

> So the specification defining "type": "binary" I think is out.

Which I took to mean that we would ***NOT*** be making `type` extensible.  I realize now that that is not quite what that sentence says, but I agree wih @karenetheridge and @jdesrosiers that we really shouldn't do that.  There is no way to express it in the vocabulary construction.

@jdesrosiers's efforts looking into similarly problematic attempts to square differing definitions of `exclusiveMinimum` and `exclusiveMaximum` across drafts led me to abandon that attempt, and helped me make the case to OAS to drop compatibility so we didn't need it.

> Adding a custom `type` keyword is something that can be supported, but you can't also use the standard validation vocabulary. If the new type requires modifications to other keywords, those would also conflict with the standard vocabularies they appear in.

> I know the spec says vocabularies are allowed to modify keywords

We should look at that.  What it should say is that compatible overlapping definitions are allowed, or something like that.  If a keyword has a fixed `enum` set of values in one vocabulary, and other values in another, I would argue that that is _not_ compatible (for the same reason that we abandoned the `exclusiveMinimum` stuff).

------------

I think it is fine to say that extra keywords can be added.  For example, some sort of `isBinary` assertion (since `contentMediaType` is never an assertion).

To the extent that this interacts with `type`, it just removes any implicit assumption that the absence of `type` is the same as `"type": ["null", "boolean", "number", "array", "object"]`.  If `type` is absent and you have a parsed instance, no further type checking is done as an assertion.

-------------

I think there's an open question about disambiguating keywords that could make sense for binary values but don't automatically, like `minLength`.  But the more I think about it the more I lean towards saying that you should add your own keywords, `minBinaryLength`, for example.

In practice, a modular implementation will dispatch the instance to keyword handlers based on what the vocabulary describes.  Each module should be able to make a decision on handling without coordinating with other modules.  What that really means is that a keyword can't fail an assertion in one module and pass in another, _and expect the result to be a pass_.  If any handler fails, the result is a failure.  Just like adding keywords to a schema object, constraints can only be added, never removed.

Does this make sense?

karenetheridge: This all sounds good.  I also found this in the spec (https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.2.3): "Vocabularies do not support specifically declaring different value sets for keywords." which at the moment explicitly rules out adding additional valid values to the 'type' keyword, but we could always carve out an exception for it as we do for 'format' (see one paragraph up in the spec.  Adding new keywords e.g. isBinaryType would certainly be less painful though.

handrews: @karenetheridge we didn't really carve out an exception for `format`.  It's always been defined as having an open-ended set of values and we are forced to continue supporting that for the time being.  No one should ever create a keyword like `format` again.

karenetheridge: @handrews I well understand the difficult nature of `format` :)  I am saying we should be explicit in the spec (given the questions and different understandings shown earlier in this thread) that it is not generally permitted for a vocabulary to extend the syntax, range of permitted values or other functionality of an existing core keyword.  Especially in the context of extending the defined data model set (the topic of the PR), the spec should give direction that new data types will need vocabularies to define *new* keywords to work with them, rather than extend existing keywords.

jsonschema-core.xml
Relequestual: Yes this is true. My feeling is that the basic use of type should be pretty universal, and SHOULD NOT makes it clear you better have a good reason to change the behavior of existing types.

jsonschema-core.xml
Relequestual: I think I had just forgotten what a "media type parameter" was, and looked like, and it is in the spec... so yeah I'm done with this thread. Thanks =]

jsonschema-core.xml
karenetheridge: I thought @handrews had shown elsewhere in this thread that we would not be permitting updating existing keywords, so MUST NOT is correct.

jsonschema-core.xml
Relequestual: My read on the above comments are, this PR is off, and the key is...

> To the extent that this interacts with type, it just removes any implicit assumption that the absence of type is the same as "type": ["null", "boolean", "number", "array", "object"]. If type is absent and you have a parsed instance, no further type checking is done as an assertion. - @handrews 

In addition, +1 to what @handrews further said...

> What that really means is that a keyword can't fail an assertion in one module and pass in another, and expect the result to be a pass. If any handler fails, the result is a failure. Just like adding keywords to a schema object, constraints can only be added, never removed.

@awwright do you feel you've misread the issue here?
Do you feel a new PR would make the most sense or do you want to continue to modify this one?

jsonschema-core.xml
Relequestual: Correct. I read this comment, made a comment, before reading the rest of the discussion related to this issue.

handrews: @awwright it is definitely true that a _dialect_ can omit the validation _vocabulary._ (I'm trying to promoted the "dialect" term, see #902).

We need to be careful here of the distinction between `type` (the keyword in the validation vocabulary) and the more generic notion of JSON Schema data model.  I'm a little confused by:

> New data model types may be defined as either supersets of an existing type (similar to how "number" is a superset of "integer")

as "number" is both a data model type and a value for `type`, while "integer" is ***not*** a data model type, just a value for `type`.  "integer" is a refinement/shortcut `type` assertion for the data model type "number" plus `"multipleOf": 1` (more or less, I don't recall the exact language for "integer" offhand).

A binary value would be outside of the JSON Schema data model, and therefore be an extension (not refinement or shortcut) to that data model.  Because the `type` keyword only has keywords relevant to the JSON data model, you would not be able to use `type` with an extension to the data model and have the assertion pass.

------

An important consideration here is that when we say "this keyword applies to numbers" or "this keyword applies to strings" we are talking about the JSON data model.  We can't say "this keyword applies to integers" because the only way to determine that something is in integer is to use `type: integer`.  The data model is not aware of integers.

My _slight_ preference after reading all this is to basically add one additional thing to the JSON Schema data model: none of the above.  Let's not get into allowing people to make specific extensions to the data model because the implications of that are potentially complex.  But we can explicitly allow people to use JSON Schema with data that doesn't fit the data model, assuming their implementation supports it somehow.  

With this approach, you could add a keyword that works for "values outside the primary JSON Schema data model" or whatever we want to call it.  But you can't add keywords for various different sorts of things that might show up that way.

In practice with OpenAPI, these "instances" are being documented and handed off to something else to interpret based on the JSON Schema `contentMediaType` keyword.  It's a way to say, _from within JSON Schema_, that the thing is not actually JSON and shouldn't be interpreted as such, but also shouldn't be an error.

(BTW, much thanks to @awwright for picking this up and to everyone for comments- this is going to be a much more solid change than my original idea when it's all done).

Relequestual: Follow up thoughts, I often a use case for wanting to validate a JSON structure in javascript which sometimes contains a function... You can't convert them to JSON, but, in js, it could be part of the data you want to validate, where you don't care about what the function is, just that it's a function.

awwright:
I took the paragraphs out and the behavior of validation keywords remains undefined.

Relequestual (approved):

handrews:
Fixes #941 (linking issue).  Can we merge this now?

awwright:
@handrews Go ahead

PR #977, opened 2020-08-29
Define compound schema documents (#936)

### LINKED ISSUE #936: Specify process for validating a compound schema document against meta-schema(s)
PR #914 is blocked on the question of how to validate schema document containing embedded schema resources with differing meta-schemas.

In my view, when loading a schema document, an implementation should be able to recognize embedded schema resources and treat them separately, as if they were `$ref`'d.

There are two concerns:
1. Do we support embedding draft-04 schemas?  This requires recognizing `id`. 

 I think we should say no, if you want to use draft-04 with draft 2020-NN, you have to `$ref` it.  Since draft-04 did not have vocabularies, there is no way to tell the difference between a custom draft-04 meta-schema relying on `id` and a custom meta-schema of another draft, that happens to have `id` next to it either b/c of a typo or someone did a really weird extension.

2. How does the meta-schema look?

@Relequestual says:

> When processing a schema document with any embedded schema resources, for the purposes of schema validation against meta-schemas (confirming the JSON Schema document is likely to be processable), embedded schema resources SHOULD be validated within their own JSON Schema feature set (using the appropriate meta-schema). For enclosing schema resources (which is likely the document root schema), an embedded resource SHOULD be considered as a valid schema document, with the value of `true`, for the purposes of validating the enclosing schema resource as a valid JSON Schema.

Which I almost follow but it's late and I think with the other points this deserves to be visible in an issue.

Relequestual:
Define Compound Schema Document and associated concerns

Resolves #936

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1857,8 +1857,9 @@
                 </t>
                 <section title="Bundling">
                     <t>
-                        A Compound Schema Document is created by replacing references (such as "$ref") with an
-                        embedded version of the referenced Schema Resource.
+                        A Compound Schema Document is created by taking references (such as "$ref")
+                        to an external Schema Resource and embedding the referenced Schema Resources
+                        within the referreing document.
                     </t>
                     <t>
                         Each embedded JSON Schema Resource (and the document's root) MUST identify itself
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1858,7 +1858,7 @@
                     </t>
                     <t>
                         When the Schema Resource referenced by a by-reference applicator is bundled, the Schema Resource
-                        MUST be located as a value of a "$defs" object at the containing schemas root.
+                        MUST be located as a value of a "$defs" object at the containing schema's root.
                         The key of the "$defs" for the now embedded Schema Resource MAY be the "$id" of the bundled schema
                         or some other form of application defined unique identifer (such as a UUID). This key is not
                         intended to be referenced in JSON Schema, but may be used by an application to aid the
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1880,9 +1880,7 @@
                 </section>
                 <section title="Differing and Default Dialects">
                     <t>
-                        In addition to the rules defining the use of
-                        <xref target="keyword-schema">the "$schema" keyword</xref>,
-                        if multiple schema resources are present in a single document, then
+                        If multiple schema resources are present in a single document, then
                         schema resources which do not have a "$schema" keyword in their root
                         schema object MUST be processed as if "$schema" were present with the
                         same value as for the immediately enclosing resource.
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1847,9 +1847,9 @@
                 </t>
                 <section title="Bundling">
                     <t>
-                        A Compound Schema Document is created by taking references (such as "$ref")
-                        to an external Schema Resource and embedding the referenced Schema Resources
-                        within the referreing document.
+                        The bundling process for creating a Compound Schema Document is done by taking
+                        references (such as "$ref") to an external Schema Resource and embedding the referenced
+                        Schema Resources within the referreing document.
                     </t>
                     <t>
                         Each embedded JSON Schema Resource (and the document's root) MUST identify itself
@@ -1868,6 +1868,11 @@
                     <t>
                         References in the containing schema document to the previously external Schema Resources
                         MUST NOT be changed, and now resolve to a schema using the "$id" of an embedded Schema Resource.
+                    </t>
+                    <t>
+                        While the bundling process will often be the main method for creating a Compound Schema Document,
+                        it is also possible and expected that some will be created by hand, potentially without individual
+                        Schema Resources existing on their own previously.
                     </t>
                 </section>
                 <section title="Differing and Default Dialects">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3924,7 +3924,7 @@ https://example.com/schemas/common#/$defs/count/minimum
                             <t>Rename $recursive* to $dynamic*</t>
                             <t>$dynamicAnchor defines a fragment like $anchor</t>
                             <t>$dynamic* (previously $recursive) no longer use runtime base URI determination</t>
-                            <t></t>
+                            <t>Define Compound Schema Documents (bundle) and processing</t>
                             <t></t>
                             <t></t>
                             <t></t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1878,7 +1878,7 @@
                         dialects, it is NOT RECOMMENDED that these documents are validated by applying a meta-schema
                         to the Compound Schema Document as an instance. It is RECOMMENDED that an alternate
                         validation process be provided in order to validate Compound Schema Documents
-                        (such as a differnet function).
+                        (such as a different function).
                     </t>
                     <t>
                         A Compound Schema Document in which all embedded resources idenfity as using the same
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1847,9 +1847,9 @@
 
             <section title="Compound Documents">
                 <t>
-                    A Compound Schema Document is defined as JSON document (sometimes called a "dereferenced" or
-                    "bundled" schema) which has multiple embedded JSON Schema Resources bundled into the same
-                    document to ease transportation.
+                    A Compound Schema Document is defined as JSON document (sometimes called a "bundled" schema)
+                    which has multiple embedded JSON Schema Resources bundled into the same document to
+                    ease transportation.
                 </t>
                 <t>
                     Each embedded Schema Resource MUST be treated as an individual Schema Resource, following standard
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1849,7 +1849,9 @@
                     <t>
                         The bundling process for creating a Compound Schema Document is defined as taking
                         references (such as "$ref") to an external Schema Resource and embedding the referenced
-                        Schema Resources within the referring document.
+                        Schema Resources within the referring document. Bundling is done in such a way that
+                        all URIs (used for referencing) in the base document and any referenced/embedded
+                        documents do not require altering.
                     </t>
                     <t>
                         Each embedded JSON Schema Resource MUST identify itself with an absolute URI using the "$id" keyword,
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1888,13 +1888,13 @@
                         same value as for the immediately enclosing resource.
                     </t>
                     <t>
-                        Embedded schema resources MAY specify different "$schema" values from their
-                        enclosing resource, as any schema that can be referenced can also be embedded.
+                        Since any schema that can be referenced can also be embedded, embedded schema resources MAY
+                        specify different "$schema" values from their enclosing resource.
                     </t>
                 </section>
                 <section title="Validating">
                     <t>
-                        Given a Compound Schema Document may have embedded resources which idenfiy as using different
+                        Given that a Compound Schema Document may have embedded resources which idenfiy as using different
                         dialects, it is NOT RECOMMENDED that these documents are validated by applying a meta-schema
                         to the Compound Schema Document as an instance. It is RECOMMENDED that an alternate
                         validation process be provided in order to validate Schema Documents.
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1181,16 +1181,6 @@
                         the document root schema, the resulting behavior is implementation-defined.
                     </t>
                     <t>
-                        If multiple schema resources are present in a single document, then
-                        schema resources which do not have a "$schema" keyword in their root
-                        schema object MUST be processed as if "$schema" were present with the
-                        same value as for the immediately enclosing resource.
-                    </t>
-                    <t>
-                        Embedded schema resources MAY specify different "$schema" values from their
-                        enclosing resource, as any schema that can be referenced can also be embedded.
-                    </t>
-                    <t>
                         Values for this property are defined elsewhere in this and other documents,
                         and by other parties.
                     </t>
@@ -1871,6 +1861,19 @@
                         MUST NOT be embedded by replacing the schema object from which it was referenced, or by wrapping
                         in other applicator keywords. Instead, a bundled Schema Resource MUST be located in `$defs`,
                         where the key is the `$id` of the Schema Resource, which MUST then be referenced using `$ref`.
+                    </t>
+                </section>
+                <section title="Differing and Default Dialects">
+                    <t>
+                        In addition to the rules defining the use of the "$schema" keyword <target='The "$schema" Keyword' format="counter"></xref>,
+                        if multiple schema resources are present in a single document, then
+                        schema resources which do not have a "$schema" keyword in their root
+                        schema object MUST be processed as if "$schema" were present with the
+                        same value as for the immediately enclosing resource.
+                    </t>
+                    <t>
+                        Embedded schema resources MAY specify different "$schema" values from their
+                        enclosing resource, as any schema that can be referenced can also be embedded.
                     </t>
                 </section>
                 <section title="Validating">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1843,13 +1843,13 @@
                 </t>
                 <t>
                     Each embedded Schema Resource MUST be treated as an individual Schema Resource, following standard
-                    schema loading and processing requirements, including determining voacbulary support.
+                    schema loading and processing requirements, including determining vocabulary support.
                 </t>
                 <section title="Bundling">
                     <t>
                         The bundling process for creating a Compound Schema Document is defined as taking
                         references (such as "$ref") to an external Schema Resource and embedding the referenced
-                        Schema Resources within the referreing document.
+                        Schema Resources within the referring document.
                     </t>
                     <t>
                         Each embedded JSON Schema Resource MUST identify itself with an absolute URI using the "$id" keyword,
@@ -1892,7 +1892,7 @@
                 </section>
                 <section title="Validating">
                     <t>
-                        Given that a Compound Schema Document may have embedded resources which idenfiy as using different
+                        Given that a Compound Schema Document may have embedded resources which identify as using different
                         dialects, these documents SHOULD NOT be validated by applying a meta-schema
                         to the Compound Schema Document as an instance. It is RECOMMENDED that an alternate
                         validation process be provided in order to validate Schema Documents. Each Schema Resource
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1159,7 +1159,7 @@
                     Meta-schema authoring is an advanced usage of JSON Schema, so the design of
                     meta-schema features emphasizes flexibility over simplicity.
                 </t>
-                <section title='The "$schema" Keyword'>
+                <section title='The "$schema" Keyword' anchor="keyword-schema">
                     <t>
                         The "$schema" keyword is both used as a JSON Schema feature set identifier and
                         as the identifier of a resource which is itself a JSON Schema, which describes the
@@ -1865,8 +1865,8 @@
                 </section>
                 <section title="Differing and Default Dialects">
                     <t>
-                        In addition to the rules defining the use of the "$schema" keyword
-                        <xref target='The "$schema" Keyword' format="counter"></xref>,
+                        In addition to the rules defining the use of
+                        <xref target="keyword-schema">the "$schema" keyword</xref>,
                         if multiple schema resources are present in a single document, then
                         schema resources which do not have a "$schema" keyword in their root
                         schema object MUST be processed as if "$schema" were present with the
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1852,9 +1852,9 @@
                         Schema Resources within the referreing document.
                     </t>
                     <t>
-                        Each embedded JSON Schema Resource (and the document's root) MUST identify itself
-                        with an absolute URI using the "$id" keyword, and SHOULD make use of the "$schema"
-                        keyword to identify the dialect it is using, in the root of the schema resource.
+                        Each embedded JSON Schema Resource MUST identify itself with an absolute URI using the "$id" keyword,
+                        and SHOULD make use of the "$schema" keyword to identify the dialect it is using, in the root of the
+                        schema resource.
                     </t>
                     <t>
                         When a by-reference applicator is dereferenced as part of the bundling process, the Schema Resource
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1868,7 +1868,7 @@
                     <t>
                         When a reference applicator is dereferenced as part of the bundling process, the Schema Resource
                         MUST NOT be embedded by replacing the schema object from which it was referenced, or by wrapping
-                        in other applicator keywords. In stead, a bundled Schema Resource MUST be located in `$defs`,
+                        in other applicator keywords. Instead, a bundled Schema Resource MUST be located in `$defs`,
                         where the key is the `$id` of the Schema Resource, which MUST then be referenced using `$ref`.
                     </t>
                 </section>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1859,8 +1859,15 @@
                     <t>
                         When a by-reference applicator is dereferenced as part of the bundling process, the Schema Resource
                         MUST NOT be embedded by replacing the schema object from which it was referenced, or by wrapping
-                        in other applicator keywords. Instead, a bundled Schema Resource MUST be located in `$defs`,
-                        where the key is the `$id` of the Schema Resource, which MUST then be referenced using `$ref`.
+                        in other applicator keywords. Instead, a bundled Schema Resource MUST be located as a value of
+                        a "$defs" object at the containing schemas root. The key of the "$defs" for the now embedded
+                        Schema Resource MAY be the "$id" of the bundled schema or some other form of application defined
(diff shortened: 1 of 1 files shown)


karenetheridge:
> A Compound Schema Document is created by replacing references (such as "$ref") with an embedded version of the referenced Schema Resource.

This is not how I bundle schemas together that have `$ref`s to other documents. Instead, the $referenced document is placed in a top-level `$def`inition, and the `$ref` URI is either changed to something new and unique (with the `$id` in the `$def`inition altered to match), or both the $ref and $id are left unchanged (which should work just fine too).

Replacing `$ref`erences with an embedded version doesn't work well when there are multiple `$ref`erences to the same URI (there is a lot of repetition), and it doesn't work at all with recursive references.

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                        A Compound Schema Document is created by replacing references (such as "$ref") with an
+                        embedded version of the referenced Schema Resource.
+                    </t>
+                    <t>
+                        Each embedded JSON Schema Resource (and the document's root) MUST identify itself
+                        with an absolute URI using the "$id" keyword, and SHOULD make use of the "$schema"
+                        keyword to identify the dialect it is using, in the root of the schema resource.
+                    </t>
+                    <t>
+                        When a reference applicator is dereferenced as part of the bundling process, the Schema Resource
```
karenetheridge, line 1854: "reference applicator" is a new phrase. Does it mean the `$ref` keyword?

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                        MUST NOT be embedded by replacing the schema object from which it was referenced, or by wrapping
+                        in other applicator keywords. In stead, a bundled Schema Resource MUST be located in `$defs`,
```
karenetheridge, line 1856:
```suggestion
                        in other applicator keywords. Instead, a bundled Schema Resource MUST be located in `$defs`,
```

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                        where the key is the `$id` of the Schema Resource, which MUST then be referenced using `$ref`.
+                    </t>
+                </section>
+                <section title="Validating">
+                    <t>
+                        Given a Compound Schema Document may have embedded resources which idenfiy as using different
+                        dialects, it is NOT RECOMMENDED that these documents are validated by applying a meta-schema
+                        to the Compound Schema Document as an instance. It is RECOMMENDED that an alternate
+                        validation process be provided in order to validate Compound Schema Documents
+                        (such as a differnet function).
```
karenetheridge, line 1866:
```suggestion
                        (such as a different function).
```

jsonschema-core.xml
handrews: Yes.  Or `$dynamicRef`.  It is defined somewhere in the spec (or was- I definitely used it in a number of places).  Maybe as "by-reference applicator".

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
                 </section>
             </section>
 
+            <section title="Compound Documents">
+                <t>
+                    A Compound Schema Document is defined as JSON document (sometimes called a "dereferenced" or
+                    "bundled" schema) which has multiple embedded JSON Schema Resources bundled into the same
+                    document to ease transportation.
```
handrews, line 1837: I feel like "bundled" is fine here but "dereferenced schema" has other meanings (like the result of dereferencing a single schema) so I'm a little wobbly on that term in this context.

jsonschema-core.xml
Relequestual: Yes. It's also called "by-reference applicator". I need to do a terminology consistency run.

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                </t>
+                <t>
+                    Each embedded Schema Resource MUST be treated as an individual Schema Resource, following standard
+                    schema loading and processing requirements, including determining voacbulary support.
+                </t>
+                <section title="Bundling">
+                    <t>
+                        A Compound Schema Document is created by replacing references (such as "$ref") with an
+                        embedded version of the referenced Schema Resource.
```
handrews, line 1846: I feel like we might need a distinction between references to other documents vs references within the same document.  Because really bundling means getting all of the resources into one document.  It may or may not be possible (or desirable) to dereference things.

"Bundling" could simply mean "stuff each referenced resource that is in another document under this document's `$defs`.  This doesn't even require changing any of the references since they are targeting by `$id` and that doesn't change through this basic form of bundling.

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                    </t>
+                    <t>
+                        Each embedded JSON Schema Resource (and the document's root) MUST identify itself
+                        with an absolute URI using the "$id" keyword, and SHOULD make use of the "$schema"
+                        keyword to identify the dialect it is using, in the root of the schema resource.
```
handrews, line 1851: Do we have wording around how an embedded resource defaults to the `$schema` of the enclosing resource if not specified?  I vaguely recall adding that somewhere but maybe it belongs here now?

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                    </t>
+                    <t>
+                        When a reference applicator is dereferenced as part of the bundling process, the Schema Resource
+                        MUST NOT be embedded by replacing the schema object from which it was referenced, or by wrapping
```
handrews, line 1855: This is where clarity on "dereference" vs "bundling" would be helpful.  If "bundling" is _defined_ as embedding under `$defs` (transitively until all references can be satisfied within the document), then this is not a concern for bundling.  It's just not part of that process.

**Separately**, we can discuss dereferencing which might be applied after bundling, or in various other contexts.  I think that might add clarity.  Bundling should be really, really easy to implement.  Dereferencing is substantially more complex.

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                        in other applicator keywords. In stead, a bundled Schema Resource MUST be located in `$defs`,
+                        where the key is the `$id` of the Schema Resource, which MUST then be referenced using `$ref`.
```
handrews, line 1857: This is easier if you just say that the original reference MUST be left unchanged (but could later be changed through dereferencing, which is defined separately).

jsonschema-core.xml
```diff
@@ -1830,6 +1830,48 @@
...
+                    </t>
+                </section>
+                <section title="Validating">
+                    <t>
+                        Given a Compound Schema Document may have embedded resources which idenfiy as using different
+                        dialects, it is NOT RECOMMENDED that these documents are validated by applying a meta-schema
+                        to the Compound Schema Document as an instance. It is RECOMMENDED that an alternate
+                        validation process be provided in order to validate Compound Schema Documents
```
handrews, line 1865: Does this need to be an alternate for _compound_ schema documents, or just an alternate for schema documents (schemas against meta-schemas) in general?  You should always be able to tell whether a given schema is compound, so if you know your instance is a schema that is enough.

We don't want to encourage two different meta-schema functions, because if someone validating a non-bundled schema with the "regular" meta-schema validation function, but then that schema got changed to a bundled schema, it would now require a code change to be handled correctly.  We _never_ want to require a code change to keep validation working.

karenetheridge:
Why do the referenced resources have to have absolute URI identifiers in order to be bundled? In my (admittedly simple) implementation of bundling in a json-schema-using application, which uses `$ref`s of the form `"$ref": "common_defs.json#/definitions/Foo"`, I just move the definition subschema over into the main document and only rewrite the definition name if it conflicts with another name -- so there are no `$id`s in use anywhere.

jdesrosiers:
@karenetheridge What if `common_defs.json#/definitions/Foo` has a local reference to another definition in `common_defs.json`? The reference would now point to `definitions` in the wrong document. Although simple transculsion works in many cases, the only way it always work is if the schema being referenced has an `$id`.

karenetheridge:
@jdesrosiers All the definitions being moved are also scanned for embedded `$ref`s and those schemas moved across as well, and so on recursively until all definitions are moved (with `$ref`s pointing to them rewritten as needed). It's not very hard at all as long as there are no definition names that collide, and in that case they just get a `_1` or somesuch appended to them.

jsonschema-core.xml
Relequestual: Maybe. If so it does belong here.

jsonschema-core.xml
Relequestual: The consensus in the associated issue was that a compound document would be transport only, in that you can't validate it by applying a meta schema to it as an instance.
Check the issue for the reasoning and justification.

jdesrosiers:
@karenetheridge 

Sorry, what I wrote doesn't make sense. Let me try again. You're right, you can bundle schemas without `$id`s, but it means you need to recursively rewrite all the references.

Ideally, whether a schema is bundled or not, you should get the same output. When you rewrite references, your `keywordLocation` and `absoluteKeywordLocation` values will be different. A bundle should just be a convenient delivery mechanism, it shouldn't create a new schema. I, for one, want to be able to maintain traceability to where the referenced schema originated.

Also, when you need to embed a schema of a different draft, you need to include `$schema`, which means you need to have an `$id`. There's no way around that one.

jsonschema-core.xml
jdesrosiers: @Relequestual I think @handrews was just concerned that this could be misunderstood to mean that you need a separate process to validate a schema than you do to validate a compound schema. The process for validating schema documents *is* the process for validating compound schema documents.

I think the suggestion would be to do something like ...

> It is RECOMMENDED that an alternate validation process be provided in order to validate ~Compound~ Schema Documents

The "Compound" part is implied because all schemas are potentially compound.

handrews: @Relequestual yes, what @jdesrosiers said.

Relequestual: Ok, we are on the same page now. I don't disagree. 
But, I'm now not sure about a few things. I'm saying the following having not reviewed the issue again (because this is a quick comment in a few mins spare time).

I thought in the issue we said it wouldn't be possible to identify a compound schema document vs a normal schema document, although that may have been in context of being an instance.

I'm going to HAVE to review the issue discussion again. It's fine =]

jdesrosiers: > I thought in the issue we said it wouldn't be possible to identify a compound schema document vs a normal schema document, although that may have been in context of being an instance.

Yes, that was in the context of being an instance. It's not possible to identify an instance as a schema (compound or otherwise).

Relequestual:
I boradly agree with @jdesrosiers regarding tracability. It should be EASY for people to work out which of the schemas they have bundled, back to the original files. The FHIR bundled schema is comprised of a LOT of schemas. A LOT.

> Also, when you need to embed a schema of a different draft, you need to include $schema, which means you need to have an $id. There's no way around that one.

I'm not sure I follow. `$schema` is required, but `$id` isn't specifically required in `2019-09` or indeed in current main branch iirc. How did you arrive at this conclusion?

jdesrosiers:
@Relequestual I didn't mean anything controversial with that statement. I just mean that without `$id` you can't embed a schema that is a different version as the referring schema if they are not using the same draft. Quick example ...

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "foo": { "$ref": "./schema/foo#/definitions/aaa" }
  }
}
```

```json
{
  "$id": "http://example.com/schema/foo",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "aaa": { "const": "example" }
  }
}
```

Using the method @karenetheridge described for bundling, you would get this ...

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "foo": { "$ref": "#/definitions/aaa" }
  },
  "definitions": {
    "aaa": { "const": "example" }
  }
}
```

This schema will not work properly because `const` is not a draft-04 keyword in this context. You would have to include `$schema` to not lose what draft the embedded schema is using.

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "foo": { "$ref": "#/definitions/aaa" }
  },
  "definitions": {
    "aaa": {
      "$schema": "http://json-schem.org/draft-07/schema#",
      "const": "example"
    }
  }
}
```

But, you can only use `$schema` at the root of the schema, so you need an `$id` to make `#/definitions/aaa` a root schema.

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "foo": { "$ref": "#/definitions/aaa" }
  },
  "definitions": {
    "aaa": {
      "$id": "???",
      "$schema": "http://json-schem.org/draft-07/schema#",
      "const": "example"
    }
  }
}
```

But, the referenced schema has no `$id` for us to put there. Therefore, when embedding schemas with differing drafts, the referenced schema must be addressable with an `$id`.

Relequestual:
Thanks for clarifying @jdesrosiers, crystal clear now.
I'm going to make some changes based on this, other feedback, and discussion from the associated issue!

jsonschema-core.xml
Relequestual: My upcoming update to the PR suggests...
```
                        A Compound Schema Document is created by taking references (such as "$ref")
                        to an external Schema Resource and embedding the referenced Schema Resources 
                        within the referring document.
```

Does that sound enough?

Relequestual:
I've made a few changes based on suggestions.
I still feel like something needs to be said about how $ids of embedded resources can be referenced AS IS.

Relequestual:
Ah, it's failing because I need to add an anchor first, then xref target that anchor. I'll fix tomorrow.

jsonschema-core.xml
```diff
@@ -1830,6 +1820,74 @@
...
+                    <t>
+                        Embedded schema resources MAY specify different "$schema" values from their
+                        enclosing resource, as any schema that can be referenced can also be embedded.
+                    </t>
+                </section>
+                <section title="Validating">
+                    <t>
+                        Given a Compound Schema Document may have embedded resources which idenfiy as using different
+                        dialects, it is NOT RECOMMENDED that these documents are validated by applying a meta-schema
+                        to the Compound Schema Document as an instance. It is RECOMMENDED that an alternate
```
handrews, line 1876: Is this actually a MUST/MUST NOT?  Treating the compound schema document as an instance is just flat-out incorrect and won't work in the general case.

I think the simplest way to state the requirement here is that each Schema Resource must be separately validated against its meta-schema, _unless_ all schema resources in the Compound Schema Document have the same value for `$schema` (or omit it).  I'm sure the wording could be better, but I think starting from separate per-resource validation and then (as you so a paragraph down) noting that there is a specific case where you can validate an entire document is the way to go.

jsonschema-core.xml
handrews: I still think there's still confusion around "dereferencing" and "bundling" (and reference removal) here.

* _dereferencing_, elsewhere in this document, is about following references during evaluation
* _bundling_ is simply inspecting references and to figure out which resources need to be placed in a single document- there's no dereferencing going on at all
* _reference removal_ is when you actually take the target of a reference and inline it; this is not always possible, and I don't think is something we really want to discuss.

handrews (commented):
This is good progress.

I think in general we want to be clear about the use case (bundling) vs the mechanism (compound documents) vs the notion of "dereferencing" which is actually something that happens during evaluation, at least based on how we use it elsewhere.

The question of "how do I detect and handle a compound schema document" is somewhat separate from "why would I make a compound schema document."  The "why" part is important to motivate this, but once you have a compound document it doesn't matter why it was created.

jsonschema-core.xml
```diff
@@ -1830,6 +1820,74 @@
...
+                    which has multiple embedded JSON Schema Resources bundled into the same document to
+                    ease transportation.
+                </t>
+                <t>
+                    Each embedded Schema Resource MUST be treated as an individual Schema Resource, following standard
+                    schema loading and processing requirements, including determining voacbulary support.
+                </t>
+                <section title="Bundling">
+                    <t>
+                        A Compound Schema Document is created by taking references (such as "$ref")
```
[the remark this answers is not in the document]
Relequestual: I'm making a change.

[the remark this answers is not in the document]
Relequestual: This section is supposed to explain how to bundle multiple schemas to form a compound schema document. Therefore I think this is OK, but I'll re-phrase to make it clearer.

jsonschema-core.xml
Relequestual: It doesn't seem we do. I'll mark this is resolved.

Relequestual: It doesn't seem we do. I'll mark this is resolved.

jsonschema-core.xml
Relequestual: Yeah, you're right, I've got my termonology muddled, and it might be that I wrote that in part confusion about how it would work anyway.
Lemme re-work!

Relequestual: Done. You'll see some changes here. No longer used "dereferencing".

jsonschema-core.xml
```diff
@@ -1830,6 +1820,74 @@
...
+                    <t>
+                        Each embedded JSON Schema Resource (and the document's root) MUST identify itself
+                        with an absolute URI using the "$id" keyword, and SHOULD make use of the "$schema"
+                        keyword to identify the dialect it is using, in the root of the schema resource.
+                    </t>
+                    <t>
+                        When a by-reference applicator is dereferenced as part of the bundling process, the Schema Resource
+                        MUST NOT be embedded by replacing the schema object from which it was referenced, or by wrapping
+                        in other applicator keywords. Instead, a bundled Schema Resource MUST be located as a value of
+                        a "$defs" object at the containing schemas root. The key of the "$defs" for the now embedded
```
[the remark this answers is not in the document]
Relequestual: 😅

jsonschema-core.xml
```diff
@@ -1830,6 +1820,74 @@
...
+                        Schema Resource MAY be the "$id" of the bundled schema or some other form of application defined
```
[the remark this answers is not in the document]
Relequestual: Yes.

jsonschema-core.xml
```diff
@@ -1830,6 +1820,74 @@
...
+                        unique identifer (such as a UUID). This key is not intended to be referenced in JSON Schema, but may
+                        be used by an application to aid the bundling process.
+                    </t>
+                    <t>
+                        References in the containing schema document to the previously external Schema Resources
+                        MUST NOT be changed, and now resolve to a schema using the "$id" of an embedded Schema Resource.
```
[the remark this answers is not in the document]
Relequestual: MUST NOT because you want to preserve the original reference to the Schema Resource (for the potential resulting output), as opposed to using a new one making use of `$defs` in the documents root.

jsonschema-core.xml
```diff
@@ -1830,6 +1820,74 @@
...
+                    </t>
+                </section>
+                <section title="Differing and Default Dialects">
+                    <t>
+                        In addition to the rules defining the use of
+                        <xref target="keyword-schema">the "$schema" keyword</xref>,
```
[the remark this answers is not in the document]
Relequestual: Is that overly verbose? I felt that since additional handling of $schema is being defined, it wouldn't hurt to remind people that some rules exist already.

jsonschema-core.xml
```diff
@@ -1830,6 +1820,74 @@
...
+                        if multiple schema resources are present in a single document, then
+                        schema resources which do not have a "$schema" keyword in their root
+                        schema object MUST be processed as if "$schema" were present with the
+                        same value as for the immediately enclosing resource.
+                    </t>
+                    <t>
+                        Embedded schema resources MAY specify different "$schema" values from their
+                        enclosing resource, as any schema that can be referenced can also be embedded.
```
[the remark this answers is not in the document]
Relequestual: Making it so.

jsonschema-core.xml
```diff
@@ -2405,8 +2463,8 @@
                         </t>
                         <t>
                             This keyword produces an annotation value which is an array of
-                            the indexes to which this keyword validates successfully when applying 
-                            its subschema, in ascending order. The value MAY be a boolean true if the 
+                            the indexes to which this keyword validates successfully when applying
+                            its subschema, in ascending order. The value MAY be a boolean true if the
```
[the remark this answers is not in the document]
Relequestual: Noted. I'll make it happen.

jsonschema-core.xml
Relequestual: Yes, this is what I was trying to say without actually saying it, but I guess there's no reason not to specify this.
RECOMMENDED is like SHOULD, in that there may be legitimate situations where you don't want to do that (as you've identified).

I feel it's bad form to say "blah blah MUST do x unless something else and in this situation" because it feels a little bit missleading.

I'm still going to have a stab at rephrasing this to make the intent a lot clearer, as you've articulated the intent better.

Relequestual: Actually, I've already said this is the case...

                        A Compound Schema Document in which all embedded resources idenfity as using the same
                        dialect MAY be validated by applying the appropriate meta-schema.

Just I didn't specify `$schema`

Relequestual: Ah, I see, I think you typoed. /so/do/ makes sense.

Still, update pending.

Relequestual:
I've forced pushed but something has done gone wrong or I've done something wrong with the naming of the branches and I'm not seeing my changes =/

Relequestual:
Accidentally deleted the wrong branch in github UI...

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: In that case I think the conditional "In order to produce identical output..." or something like that needs to be attached to the MUST NOT.  Or be made very clear in the framing of this section (which maybe it is and I've just lost track of that?)

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: I just don't understand how it connects to the rest of the sentence.  To me it reads like to kinda-tangentially-related things stuck together.  The rest of the sentence is clear on its own, but this one implies that I'm probably forgetting something super-relevant in some other section and I have no idea what it is.  Why do other parts of the `$schema` rules have particular relevance here?

jsonschema-core.xml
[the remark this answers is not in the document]
Relequestual: Additionally, there was some previous talk (I forget who from specifically) about doing this in a way where the process would change all the references, done in such a way which could potentially cause failures in dereferencing later.
So it's not JUST for preserving output.

jsonschema-core.xml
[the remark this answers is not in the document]
Relequestual: Actually, you're right. There's no need for this. I've removed the phrasing. Pushed.

jsonschema-core.xml
```diff
@@ -1845,6 +1835,82 @@
...
 
+            <section title="Compound Documents">
+                <t>
+                    A Compound Schema Document is defined as JSON document (sometimes called a "bundled" schema)
+                    which has multiple embedded JSON Schema Resources bundled into the same document to
+                    ease transportation.
+                </t>
+                <t>
+                    Each embedded Schema Resource MUST be treated as an individual Schema Resource, following standard
+                    schema loading and processing requirements, including determining voacbulary support.
```
karenetheridge, line 1846:
```suggestion
                    schema loading and processing requirements, including determining vocabulary support.
```

jsonschema-core.xml
```diff
@@ -1845,6 +1835,82 @@
...
+                </t>
+                <section title="Bundling">
+                    <t>
+                        The bundling process for creating a Compound Schema Document is defined as taking
+                        references (such as "$ref") to an external Schema Resource and embedding the referenced
+                        Schema Resources within the referreing document.
+                    </t>
+                    <t>
+                        Each embedded JSON Schema Resource MUST identify itself with an absolute URI using the "$id" keyword,
```
karenetheridge, line 1855: Is this referring to embedded resources in the resulting schema (after bundling has been performed), or the original source resources, or both?  My earlier questions about this being required vs. recommended still apply -- it need not be a requirement that referenced resources use absolute URIs if it is acceptable to have rewritten $refs in the result (that is, a new schema be produced that has different canonical uris at various subschemas than in the original, and the user-application understands that the locations in resulting errors or annotations will be different than had the original non-bundled schemas been used).

jsonschema-core.xml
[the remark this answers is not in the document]
karenetheridge: If you use the "output" wording, please make it clear that this includes not just evaluation success/failure, but also all the uris and paths used in the resulting annotations or errors.

jsonschema-core.xml
```diff
@@ -1845,6 +1835,82 @@
...
+                        same value as for the immediately enclosing resource.
+                    </t>
+                    <t>
+                        Since any schema that can be referenced can also be embedded, embedded schema resources MAY
+                        specify different "$schema" values from their enclosing resource.
+                    </t>
+                </section>
+                <section title="Validating">
+                    <t>
+                        Given that a Compound Schema Document may have embedded resources which idenfiy as using different
```
karenetheridge, line 1895:
```suggestion
                        Given that a Compound Schema Document may have embedded resources which identify as using different
```

jsonschema-core.xml
handrews: @Relequestual yes that sounds great!

handrews:
I think this is looking good- Any of my remaining comments are more personal style sort of things so however y'all want to resolve them (or leave as is) should be fine.

jsonschema-core.xml
Relequestual: I don't think this is exclusive to canonical URIs. I'm going with a slightly altered version of your suggestion. Thanks!

jsonschema-core.xml
Relequestual: I believe this has been resolved.

Relequestual:
@karenetheridge After your further review, I think this is good to merge.

karenetheridge (approved):

PR #981, opened 2020-09-06
Split "unevaluatedItems" and "unevaluatedProperties" into their own vocabulary.

### LINKED ISSUE #831: JSON Schema Core - 9.3.1.3. unevaluatedItems - clarification suggested
The text of _9.3.1.3. unevaluatedItems_ uses the term "adjacent keywords", whereas the text _9.3.1.2. additionalItems_ uses the term "within the same schema object". I suppose that "adjacent" and "within the same schema object" are equivalent, but I am not completely sure.

A second point to clarify. The text says: "Those two annotations, as well as 'unevaluatedItems', can also result from any and all adjacent in-place keywords". I suppose that "resulting from adjacent in-place keywords" means _directly or indirectly_ resulting annotations - e.g. includes the annotations produced by a subschema which is not the value of an adjacent in-place keyword, but the value of a nested in-place keyword contained by the value of the adjacent in-place keyword.

If my understanding is correct, I suggest to change the text slightly, 
from 
   "resulting from adjacent in-place keywords" 
to 
   "resulting from adjacent in-place keywords, either directly or indirectly via nested in-place keywords."

### LINKED ISSUE #853: Split "unevaluated*" into separate vocab for separate support indications
While responding at https://github.com/epoberezkin/ajv/issues/1082#issuecomment-583766036 I was reminded that we technically allow validators to make an independent decision about supporting `unevaluatedProperties` and `unevaluatedItems` due to their dependence on annotation collection, which is not required for basic spec conformance.

We did that before the `$vocabulary` process had been developed.  Which means that whether or not these keywords are considered errors is totally outside of that process, which is otherwise how we give schema authors control over ensuring support for the features on which they depend.

We could split those keywords out of the "applicator" vocabulary (perhaps in to an "annotated-applicator" vocabulary, or just "unevaluated", or something, the name isn't that important).

Then we could provide (or just tell people they can write, like we do with the different "format" vocabulary options) a meta-schema that is just like the standard one, but with the "unevaluated" (or whatever) vocabulary declared with `false`, or omitted entirely.

This would require meta-schema updates, but given that 2019-09 has barely been implemented at all and still does not even have a full test suite, rolling everything forward to a 2020-03 or 2020-04 that is aligned with OpenAPI 3.1 would be pretty reasonable.  I will file a separate issue for tracking the question of whether the draft-08-patch1 release should be one with a new meta-schema or not.

handrews:
Fixes #853.  Fixes #831.

Quite some time ago we agreed to split these out.  Now support (or lack of support) is all handled through `$vocabulary`, without any "the vocabulary is supported except these keywords might error out if the implementation just didn't feel like it" behavior.

If the implementation doesn't support annotations, it won't be able to process anything that requires the unevaluated vocabulary.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -304,9 +304,6 @@
                 <t>
                     A JSON Schema MUST be an object or a boolean.
                 </t>
-                <t>
-                    Keywords within the same schema object are referred to as adjacent keywords.
-                </t>
                 <section title="JSON Schema Objects and Keywords">
                     <t>
                         Object properties that are applied to the instance are called keywords,
@@ -338,6 +335,9 @@
                         SHOULD only produce assertion results based on their subschemas'
                         results.  They should not define additional constraints independent
                         of their subschemas.
+                    </t>
+                    <t>
+                        Keywords within the same schema object are referred to as adjacent keywords.
                     </t>
                     <t>
                         Extension keywords, meaning those defined outside of this document
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -304,6 +304,9 @@
                 <t>
                     A JSON Schema MUST be an object or a boolean.
                 </t>
+                <t>
+                    Keywords within the same schema object are referred to as adjacent keywords.
+                </t>
                 <section title="JSON Schema Objects and Keywords">
                     <t>
                         Object properties that are applied to the instance are called keywords,
@@ -674,7 +677,8 @@
             <section title="Keyword Interactions">
                 <t>
                     Keyword behavior MAY be defined in terms of the annotation results
-                    of <xref target="root">subschemas</xref> and/or adjacent keywords.
+                    of <xref target="root">subschemas</xref> and/or adjacent keywords
+                    (keywords within the same schema object) and their subschemas.
                     Such keywords MUST NOT result in a circular dependency.
                     Keywords MAY modify their behavior based on the presence or absence
                     of another keyword in the same
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2479,6 +2479,28 @@
 
         <section title="A Vocabulary for Unevaluated Locations">
             <t>
+                The purpose of these keywords is to enable schema authors to apply
+                subschemas to array items or object properties that have not been
+                successfully evaluated against any dynamic-scope subschema of any
+                adjacent keywords.
+            </t>
+            <t>
+                These instance items or properties may have been unsuccessfully evaluated
+                against one or more adjacent keyword subschemas, such as when an assertion
+                in a branch of an "anyOf" fails.  Such failed evaluations are not considered
+                to contribute to whether or not the item or property has been evaluated.
+                Only successful evaluations are considered.
+            </t>
+            <t>
+                Recall that adjacent keywords are keywords within the same schema object,
+                and that the dynamic-scope subschemas include reference targets as well as
+                lexical subschemas.
+            </t>
+            <t>
+                The behavior of these keywords depend on the annotation results of
+                adjacent keywords that apply to the instance location being validated.
+            </t>
+            <t>
                 Meta-schemas that do not use "$vocabulary" SHOULD be considered to
                 require this vocabulary as if its URI were present with a value of true.
             </t>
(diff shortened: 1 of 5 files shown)


jdesrosiers (approved):

handrews:
Also fixes #831

jsonschema-core.xml
```diff
@@ -2570,6 +2474,124 @@
...
 
+        <section title="A Vocabulary for Unevaluated Locations">
+            <t>
+                Meta-schemas that do not use "$vocabulary" SHOULD be considered to
+                require this vocabulary as if its URI were present with a value of true.
+            </t>
+            <t>
+                The current URI for this vocabulary, known as the Unevaluated Applicator
+                vocabulary, is:
+                &lt;https://json-schema.org/draft/2019-09/vocab/unevaluated&gt;.
```
karenetheridge, line 2510: shouldn't this be draft/2020-XX/..., or will that be changed later?

handrews: We'll fix it all at once later.  People (possibly me, I don't recall) pushed for it earlier last time and I had to keep changing the XX and it was a giant waste of time.

karenetheridge:
Now that these keywords depend on annotation results from *another* vocabulary, the evaluation order of the vocabularies is now significant, so it should be explicitly stated what that ordering should be (e.g. this vocabulary always runs last after all others defined by the core specification).

Relequestual:
@karenetheridge Want to file a new issue for that? This PR is about to be merged.

handrews:
> Want to file a new issue for that? This PR is about to be merged.

@Relequestual @karenetheridge #995 

(which for some reason still has no comments despite addressing something that has been brought up repeatedly in the last few weeks... is this thing on?)

jsonschema-core.xml
```diff
@@ -2566,6 +2477,138 @@
...
+                subschemas to array items or object properties that have not been
+                successfully evaluated against any dynamic-scope subschema of any
+                adjacent keywords.
+            </t>
+            <t>
+                These instance items or properties may have been unsuccessfully evaluated
+                against one or more adjacent keyword subschemas, such as when an assertion
+                in a branch of an "anyOf" fails.  Such failed evaluations are not considered
+                to contribute to whether or not the item or property has been evaluated.
+                Only successful evaluations are considered.
```
Relequestual, line 2492: I've been thinking about this section and its phrasing.
It's the first time in the whole document we mention "evaluation", and we don't really define what it means.

People (including myself previously) incorrectly assume that if a location has been evaluated (had something applied to it), then it counts as having been evaluated.

I'm going to riff some ideas...

We say here "successful evaluations", we mean if an application doesn't result in validation errors, and therefore the applied schema or keyword is correctly applicable to the instance location.

For example, if you had a schema...
`anyOf [ { title> unicycle, props > wheels > const > 1 }, { title: bicycle, props > wheels > const > 2 } ]`

If your instance was a unicycle, the first subschema item in the `anyOf` would be applied and _**successfully evaluated**_. Because it only has one wheel, we know it's a unicycle. It is not a bicycle, so that whole subschema is not successfully evaluated.

The instance is successfully evaluated as (or determined to be) a unicycle.

We can further justify this terminology by the fact that evaluation can be short-circuited, and therefore there's no expectation for a schema to be fully evaluated (although that would be a weak argument, and probably confuse things).

---

We need to clearly define what successful evaluation means, and what unsuccessful evaluation means, with justification (IMHO). I'm not currently convinced I can do that.

I'm happy for this to be a new issue in order to get this PR merged, but it would be essential to 2020-NN.

Relequestual:
> 
> 
> > Want to file a new issue for that? This PR is about to be merged.
> 
> @Relequestual @karenetheridge #995
> 
> (which for some reason still has no comments despite addressing something that has been brought up repeatedly in the last few weeks... is this thing on?)

yokay, I'm still reading the rest of the issue. If I see no problems, it's merging time.

jsonschema-core.xml
karenetheridge: This is a good point, and it should be made clear that these keywords really mean "un**Sucessfully**Evaluated*Thing*" rather than just "unEvaluated*Thing*".

handrews:
@Relequestual 
> Edit: waiting for reply on my in code comment

Let's make it a new issue to get this merged and then I can reply to that new issue in the next few days.

handrews:
Per conversation with @Relequestual on chat I've file #998 and am going to go ahead and merge this.

PR #984, opened 2020-09-06
Fix ambuguity in RFC 3339 reference.

### LINKED ISSUE #959: JSON Schema Validation, section 7.3.1 - clarification needed
https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.1  
(emphasis mine)

> Date and time format names are derived from **RFC 3339, section 5.6**. The duration format is from the ISO 8601 ABNF as given in **Appendix A of RFC 3339**.
> ...
> Implementations MAY support additional attributes using the other production names defined in **that section**.

In the last paragraph, which of the previously mentioned sections of RFC 3339 does "that section" refer to - [section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) or [Appendix&nbsp;A](https://tools.ietf.org/html/rfc3339#appendix-A)? I guess it refers to section 5.6 but would appreciate an official confirmation.

handrews:
Fixes #959.

Don't say "that section" when we referenced two sections.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -732,7 +732,7 @@
                     </t>
                     <t>
                         Implementations MAY support additional attributes using the other
-                        production names defined in that section.  If "full-date" or "full-time"
+                        production names defined anywhere in that RFC.  If "full-date" or "full-time"
                         are implemented, the corresponding short form ("date" or "time"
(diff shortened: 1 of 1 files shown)


Relequestual (approved):

Relequestual:
This fixes an ambiguity so I'm not hanging around to merge this.

PR #987, opened 2020-09-06
Collect unknown keywords as annotations

### LINKED ISSUE #698: Should unknown keywords be collected as annotations?
Assuming you support annotations (which is related to what output format(s) you support, now that I think of it), should unknown keywords be collected as annotations instead of ignored?

I think we will see more annotations (we've had several requested that don't clearly meet the significance criteria to be added to the standard), and it's an easy default behavior to add.

I don't feel strongly about this, just wondering if this is a sensible default option.

handrews:
And try to avoid memory exhaustion attacks, which were possible
even with out this change but potentially more likely to happen
by accident with it (e.g. unrecognized applicator with a very
large subschema).

Fixes #698.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -150,9 +150,9 @@
             <t>
                 JSON Schema can be extended either by defining additional vocabularies,
                 or less formally by defining additional keywords outside of any vocabulary.
-                Unrecognized individual keywords are ignored, while the behavior with respect
-                to an unrecognized vocabulary can be controlled when declaring which
-                vocabularies are in use.
+                Unrecognized individual keywords simply have their values collected as annotations,
+                while the behavior with respect to an unrecognized vocabulary can be controlled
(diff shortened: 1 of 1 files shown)


karenetheridge:
Thank you, this resolves an inconsistency in the 2019-09 specification document which indicates that the 'format' vocabulary should default to disabled, yet we should still collect the keyword value as an annotation (which requires implementations to special-case 'format' evaluation as "always evaluate this keyword, even if the vocabulary is false, but now the vocabulary boolean specifies only whether the validation check should be performed").

handrews:
@karenetheridge yeah, if `format` is "unknown" (and "vaguely known but not entirely supported" can be lumped in with "unknown", collecting it now matches the default behavior.  There's still the awkward "best effort" validation support option but we're stuck with that for now as it's what is realistically implemented today and there aren't yet viable alternatives.

Relequestual (approved):

PR #992, opened 2020-09-16
Boolean operators are boolean operators

### LINKED ISSUE #910: Section 9.2.1 (*Of keywords) is non-specific about sequencing
This may be irrelevant to the [test case](https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/359) that highlighted the issue to me, but it seems that the spec should say whether sequence of evaluation is important.

handrews:
This has been reported as a point of confusion.

Fixes #910

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2134,6 +2134,10 @@
                         keyword to be applied to an instance location with different values.
                         Annotation keywords define their own rules for combining such values.
                     </t>
+                    <t>
+                        These keywords have the same properties as the corresponding logical
+                        operators, e.g. commutativity and associativity.
+                    </t>
                     <section title="allOf" anchor="allOf">
                         <t>
                             This keyword's value MUST be a non-empty array.


Relequestual:
I don't have any problems with the PR, although I'm not familiar with those terms.
I wasn't convinced the associated issue actually required any changes.
I still feel like the words we have were enough to convey the requirements, but we didn't get a subsequent response from @gregsdennis on the issue.

handrews:
@Relequestual which terms- commutativity and associativity?

https://en.wikipedia.org/wiki/Commutative_property
https://en.wikipedia.org/wiki/Associative_property

Relequestual:
Yeah. I mean I used google and they made sense. I obviously didn't study that aspect of maths I guess.

handrews:
@Relequestual to be fair, the grammatical forms `commutativity` and `associativity` are probably not the most recognizable forms.  But they do come up easily in a google search and I like that they are precisely defined.

gregsdennis:
I don't think associativity is needed.  It describes how terms are grouped, but subschemas aren't grouped at all within a `*Of`, they're just listed.

Commutativity (ordering of terms) should be sufficient.

handrews:
I still don't even understand why I have to explain commutativity so I'd rather just cover all the bases I guess.

gregsdennis:
For reference:

Associativity:

```
(a + b) + c = a + (b + c)
```

Commutativity:

```
a + b = b + a
```

Relequestual:
@gregsdennis FYI still waiting to hear re: https://github.com/json-schema-org/json-schema-spec/issues/910#issuecomment-660594604

handrews:
> @gregsdennis FYI still waiting to hear re: #910 (comment)

Yeah I'm also happy to just drop this PR.

What I don't want to do is write something that facilitates the expectation that things are ordered unless explicitly stated otherwise.  These are boolean operators.  They act like boolean operators and that's more important than JSON syntax.

gregsdennis:
Happy to drop

handrews:
@gregsdennis cool, thanks!

PR #997, opened 2020-09-26
Fix 988 - Inconsistencies with schema uri and empty fragment URIs

### LINKED ISSUE #988: What is the $schema tag for the new 2019-09 draft?
Can I just confirm the identifier used for draft 2019-09, I've seen the a few variations, which is correct?

I'm assuming that its this one, although this breaks with the conventions used in the older drafts using https and having no empty fragment
https://json-schema.org/draft/2019-09/schema - seen [here](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.10.4) 

I've seen this in a few places
- http://json-schema.org/draft/2019-09/schema# - seen [here](https://json-schema.org/understanding-json-schema/UnderstandingJSONSchema.pdf) (4.13.1) 

And the http:// versions of both have cropped up in other resources.

Thanks

about-code:
Fixes #988 and other inconsistencies regarding empty fragments.

@Relequestual regarding the issue and empty fragments you [wrote](https://github.com/json-schema-org/json-schema-spec/issues/988#issuecomment-688483240)

> The spec says "MUST NOT have an empty fragment".

The exact wording of the [spec](https://github.com/json-schema-org/json-schema-spec/blob/2747ab4a506af937117a92cf35acce3880438a53/jsonschema-core.xml#L1330) with respect to `$id` (not `$schema`) currently is

>  "$id" MUST NOT contain a non-empty fragment, and SHOULD NOT contain an empty fragment.

I guess you referred to this part of the spec since I could not find any other mentioning of empty fragments. I read this as

- `http://example.com/my-schema-id#non-empty` must not be used
- `http://example.com/my-schema-id#` should not be used

I am only writing this because this PR picks up on your original comment and also removes some empty fragments which I assume are very likely unintended while not being totally against the spec itself, though.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1916,7 +1916,7 @@
                         <figure>
                             <artwork>
         <![CDATA[
-        Link: <https://example.com/my-hyper-schema#>; rel="describedby"
+        Link: <https://example.com/my-hyper-schema>; rel="describedby"
         ]]>
                             </artwork>
                         </figure>
@@ -1962,7 +1962,7 @@
                             <artwork>
         <![CDATA[
         Content-Type: application/schema-instance+json;
-                  schema="https://example.com/my-hyper-schema#"
+                  schema="https://example.com/my-hyper-schema"
         ]]>
                             </artwork>
                         </figure>
--- a/jsonschema-hyperschema.xml
+++ b/jsonschema-hyperschema.xml
@@ -282,7 +282,7 @@
         <section title="Meta-Schemas and Output Schema">
             <t>
                 The current URI for the JSON Hyper-Schema meta-schema is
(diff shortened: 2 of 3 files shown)


Relequestual:
Ah right, yes, that's correct. We left it as SHOULD to keep backwards compat rather than nit picking in this instance, iirc.

Relequestual (approved):

Relequestual:
Thank you Andreas for your contribution! Every little helps!

about-code:
You're welcome. I very much appreciate the hard work and time you and your collaborators dedicate to this project and

> Every little helps!

was exactly what I thought would be a good way to show my gratitude. :-)

Relequestual:
My first PR was for jQuery docs! =]

PR #999, opened 2020-09-26
Clarify what "absolute" means for keyword absolute location

handrews:
It is like absolute filesystem paths, not like absolute-URI
from RFC 3986.  In fact, an absolute keyword location can
never be an RFC 3986 absolute-URI because a fragment must always
be present to identify the keyword.

_**NOTE**: this PR is **not** for any discussion of renaming fields in the output format, since someone always wants to rename something.  If you want to rename something, get consensus around it and file it as a new issue/PR.  Do not raise it here._

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2705,11 +2705,18 @@
                 <section title="Keyword Absolute Location">
                     <t>
                         The absolute, dereferenced location of the validating keyword.  The value MUST
-                        be expressed as an absolute URI using the canonical URI of the relevant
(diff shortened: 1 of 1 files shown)


Relequestual (approved):

karenetheridge:
*extremely* minor nit - the absoluteKeywordLocation for the error generated by the schema `false` won't have a non-empty fragment.

(That's the only time this can happen, I believe, except for implementation-specific errors generated outside of keyword evaluation.. for example hitting a maximum recursion limit right after an `$id` changed the canonical uri)

Issue #847, opened 2020-02-05
Proofread schema document vs resource for consistency and correctness.


handrews:
It's not entirely consistent as it was a late change.

In most cases, we want to be talking about schema resources, and the root object within a schema resource.  We only want to talk about documents when we mean the actual text (or octet stream, I suppose is what it technically is), no matter how many resources are contained in it.

We should also make sure that keywords confined to a root object (`$schema`, `$recursiveAnchor`, `$vocabulary`) are confined to the root of the correct thing.  I _think_  that should be the resource root, not the document root, even if (as is currently the case) we note that switching the `$schema` value within the same document does not have well-defined behavior (that's in a CREF).

handrews:
We want to do this after other major work on this draft is complete, but it should be part of a holistic wording review.

ssilverman:
Looking through this now, using the following definitions:
1. A "schema document" is the whole resource, per the "actual text" definition above, and
2. A "schema resource" is an object having an $id.

ssilverman:
I went through every "schema document" and every "schema resource" search result and made sure they all made sense.

Relequestual:
Super. In that case, we can close out this issue! 🚀
Many thanks!

PR #1002, opened 2020-10-01
Add vocabulary files and keyword dependencies.

### LINKED ISSUE #996: Minimalist vocabulary file for expressing keyword dependencies.
_**NOTE:** It is 2020.  The world has gone to shit, and I'm barely functioning as a human being.  But we have a draft to get out and a vague deadline in the form of OAS 3.1.  So I'm trying to get this done.  **HOWEVER** I do not have the capacity to debate things in detail.  The only things I'm looking for here are:_
  * _Will it work?_
  * _Will it make our lives difficult in a significant way in the future?_

_If you think you have a substantially better idea, please file it as its own issue and if it is better we will do that.  Feel free to cut-and-paste this issue and just change whatever you want to change if it's not completely different._

_I am specifically **not up for a debate on naming.** If a group of people want to go off on slack and reach a consensus that includes at minimum @Relequestual and @gregsdennis on a single proposal of alternative names and bring that back, we'll switch to those.  Just don't debate it here._

_I am also **not interested in what else we could do with a vocabulary file.** I have a much more complex idea in my head that would even allow a generic implementation to automatically implement certain classes of keyword straight from the file.  But now is not the time for that._

_And I am **very much not at all interested** in revisiting the question of per-vocabulary meta-schemas.  I understand that there are redundancies.  I understand that not everyone "gets" why I set them up that way.  But that discussion is out of scope for this issue, and really for this whole draft._

_Yes, I'm being unreasonably dictatorial.  None of us are getting what we want this year, are we?_

-----

This is an alternative to #995 about a plugin architecture appendix.  This approach would support the plugin architecture sketched out there but is more precise.

The format (as I have long said would be the case) is ***not*** a schema.  There are two top-level keywords:

* `vocabulary`: this MUST match the URI used in the meta-schema's `$vocabulary` keyword
* `keywords`: the object of keywords defined by this vocabulary.

Within the `keywords` object, for each schema keyword there is an object with the following vocabulary file keywords:

* `inPlaceApplicator`: if `true`, this keyword is an in-place applicator, and MUST run before any other keyword that depends on annotations collected through adjacent in-place applicators _defaults to `false`_
* `dependsOn`: list of annotation names (which are also keyword names) on which this keyword depends; there is no way to distinguish which vocabulary provides the annotation, and as always (yes this is in the spec) combining vocabularies with conflicting semantics produces undefined behavior so just don't _defaults to an empty list `[]`_
* `throughInPlaceApplicators`: if `true`, the annotations in `dependsOn` need to be examined both as adjacent keywords, and as annotation propagating up through adjacent in-place applicators; if `false` only adjacent keywords noted in `dependsOn` are to be considered. _defaults to `false`_
* `dependsOnValidity`: object of keywords (although really it should only be one but that's should and not SHOULD) that MUST be present and MUST produce the given boolean value assertion result in order for this keyword to be processed

All the usual stuff about `$vocabulary` key names / `vocabulary` values being URIs and not URLs etc. etc. is the same as for all of the other stuff like this.  Furthermore, implementations are welcome to hardcode behavior for well-known vocabularies (e.g. the ones in our spec documents).

This is intended to manage keyword-level dependencies.  I am **NOT OPEN** to vocabulary-level dependencies.  Vocabularies are semantic units of convenience that facilitate a plugin architecture.  Vocabularies are "aware" of keywords, keywords are not aware of vocabularies.  Because reasons.

Since we're documenting dependencies, I wanted to handle `if`/`then`/`else`, and realized I want annotation dependencies and validation dependencies to be handled separately.  There's no concept of validation results interacting in a special way with in-place applicators, they just work at the immediately adjacent level.  You could in theory depend on both validity and annotations but I wouldn't.

I am open to ***ONE*** naming question and only this one: if folks would rather put a `$` in front of _all_ of the vocabulary file keywords I'd be OK with that.  But this format is not intended to be extensible so idk it just seemed simpler to not bother.

I'm not 100% sure I got all of the keywords here but you get the idea and it shows the dependency stuff.
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/core",
  "keywords": {
    "$schema": {},
    "$vocabulary": {},
    "$id": {},
    "$anchor": {},
    "$ref": {
      "inPlaceApplicator": true
    },
    "$dynamicAnchor": {},
    "$dynamicRef": {
      "inPlaceApplicator": true
    },
    "$defs": {},
    "$comment": {}
  }
}
```
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/applicator",
  "keywords": {
    "allOf": {
      "inPlaceApplicator": true
    },
    "anyOf": {
      "inPlaceApplicator": true
    },
    "oneOf": {
      "inPlaceApplicator": true
    },
    "not": {
      "inPlaceApplicator": true
    },
    "if": {
      "inPlaceApplicator": true
    },
    "then": {
      "inPlaceApplicator": true,
      "dependsOnValidity": {"if": true}
    },
    "else": {
      "inPlaceApplicator": true,
      "dependsOnValidity": {"if": false}
    },
    "dependentSchemas": {
      "inPlaceApplicator": true
    }, 
   "prefixItems": {},
    "items": {
      "dependsOn": ["prefixItems"],
      "throughInPlaceApplicators": false
    },
    "contains": {},
    "properties": {},
    "patternProperties": {},
    "additionalProperties": {
      "dependsOn": ["properties", "patternProperties"],
      "throughInPlaceApplicators": false
    },
    "propertyNames": {}
  }
}
```
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/unevaluated",
  "keywords": {
    "unevaluatedProperties": {
      "dependsOn": ["properties", "patternProperties", "additionalProperties", "unevaluatedProperties"],
      "throughInPlaceApplicators": true
    },
    "unevaluatedItems": {
      "dependsOn": ["prefixItems", "items", "contains", "unevaluatedItems"],
      "throughInPlaceApplicators": true
    }
  }
}
```
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/validation",
  "keywords": {
    "type": {},
    "enum": {},
    "const": {},
    "minimum": {},
    "maximum": {},
    "exclusiveMinimum": {},
    "exclusiveMaximum": {},
    "multipleOf": {},
    "minLength": {},
    "maxLength": {},
    "pattern": {},
    "minItems": {},
    "maxItems": {},
    "minContains": {
      "dependsOn": ["contains"]
    },
    "maxContains": {
      "dependsOn": ["contains"]
    },
    "required": {},
    "dependentRequired": {},
    "minProperties": {},
    "maxProperties": {}
  }
}
```
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/meta-data",
  "keywords": {
    "title": {},
    "description": {},
    "readOnly": {},
    "writeOnly": {},
    "deprecated": {},
    "examples": {},
    "default": {}
  }
}
```
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/format",
  "keywords": {
    "format": {}
  }
}
```
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/content",
  "keywords": {
    "contentMediaType": {},
    "contentEncoding": {},
    "contentSchema": {}
  }
}
```
```JSON
{
  "vocabulary": "https://json-schema.org/2019-09/vocab/hyper-schema",
  "keywords": {
    "base": {},
    "links": {}
  }
}
```

handrews:
This just covers the Core spec, the rest will be done
as a follow-on.  Fixes #996.

***Please read these instructions before commenting!***

The following topics are **in bounds** for PR comments:
* Does this follow as expected from issue #996
* Is it clear?
* Is it complete?
* Are there whole sections, paragraphs or sentences that should be moved or just dropped?

For smaller-scale sentence tweaking or wording preferences, please make notes in #847 but I don't want to get into that here.  If it's clear enough to be understood, that's good enough for now.

The following topics are **out of bounds** for comments on this PR, **but** if you can build consensus on slack or wherever and come back here with something that has consensus behind it, I will gladly switch to the consensus view.  I just don't want to debate alternatives myself (for reasons explained in detail on slack).  If an alternative emerges with broad support, as long as I don't see a major problem I'll just accept it:
* Other things that could be added
* Different names for things
* Completely different approaches

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -687,14 +687,74 @@
             </section>
             <section title="Keyword Interactions">
                 <t>
-                    Keyword behavior MAY be defined in terms of the annotation results
-                    of <xref target="root">subschemas</xref> and/or adjacent keywords
-                    (keywords within the same schema object) and their subschemas.
+                    Keyword behavior MAY be defined in terms of the annotation or
+                    validation results of adjacent keywords (keywords within the same
+                    schema object) and their <xref target="root">subschemas</xref>.
                     Such keywords MUST NOT result in a circular dependency.
                     Keywords MAY modify their behavior based on the presence or absence
                     of another keyword in the same
                     <xref target="schema-document">schema object</xref>.
                 </t>
+                <t>
+                    Keywords MAY specify dependencies on keywords in other vocabularies.
+                    However, dependencies are purely in terms of keyword/annotation name.
+                    While a keyword's description SHOULD indicate the vocabulary that is
+                    expected to provide the dependency, this is not an enforceable dependency.
+                    As with overlapping keyword names in general, the behavior of using
+                    vocabularies with incompatible semantics is undefined.
+                </t>
+                <t>
+                    In addition to depending on specific adjacent keywords, a keyword MAY
+                    depend on specific annotations collected through any and all in-place applicators,
+                    without needing to list individual in-place applicator keywords.  This allows
+                    for continued correct behavior when some other vocabulary adds previously unknown
+                    in-place applicators.  An annotation that has been collected through in-place
+                    applicators will have the same instance location as that which is currently
+                    being processed.
+                    <cref>
+                        At this time, the only class of keywords that can be depended upon like this
+                        are in-place applicators.  To date, no use cases have been found for other
+                        keyword class dependencies.  This could change in the future.
+                    </cref>
+                </t>
+                <t>
+                    Keywords defining interactions through these mechanisms SHOULD be supported by any
+                    vocabulary plugin architecture; interactions of this sort involving adjacent
+                    keywords and in-place applicators are expressed in machine-readable form in the
+                    <xref target="vocab-file">vocabulary file</xref>.
+                </t>
+                <t>
+                    Vocabulary authors MUST NOT assume that vocabulary plug-in architectures
+                    will allow any other form of interaction.  Save for explicit agreement,
+                    schema authors SHALL NOT expect these keywords depending on additional
+                    interactions to be supported by peer implementations.
+                </t>
+                <section title="Implementing keyword interactions">
+                    <t>
+                        While keyword interactions MUST be defined as described above in order
+                        to be implementable in any conforming plugin architecture, implementations
+                        MAY choose other ways to implement specific keyword dependencies for
+                        reasons of performance or simplicity.
+                    </t>
+                    <t>
+                        Because annotation collection can add significant cost in terms of both
+                        computation and memory, implementations MAY opt out of this feature.
+                        Keywords that are specified in terms of collected annotations SHOULD
+                        describe reasonable alternate approaches when appropriate.
+                        This approach is demonstrated by the
+                        "<xref target="items" format="title"/>" and
+                        "<xref target="additionalProperties" format="title"/>" keywords in this
+                        document.  However, "<xref target="unevaluatedItems" format="title" />"
+                        and "<xref target="unevaluatedProperties" format="title" />" do not
+                        specify alternate approaches, as the complexity of those keywords
+                        is not conducive to simple alternatives.
+                    </t>
+                    <t>
+                        Note that when no such alternate approach is possible for a keyword,
+                        implementations that do not support annotation collection will not
+                        be able to support those keywords or vocabularies that contain them.
+                    </t>
+                </section>
             </section>
             <section title="Default Behaviors">
                 <t>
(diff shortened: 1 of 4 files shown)


karenetheridge:
I see this was closed.. but let's revisit next year for the next draft, as there are lots of gems in here that I don't want to lose.

gregsdennis:
`@handrews` wanted to do this for draft 07, but we never got to a place where we could define WHAT a keyword was.  There's another issue floating around (I think @ssilverman or @jdesrosiers opened it?) that goes into some ideas around it that Henry summarily dismissed.

Relequestual:
Ok. Open another issue and reference this PR and that issue or PR and we will see what we can do! =D

gregsdennis:
Definitely not for this draft. Needs more discussion.

Issue #867, opened 2020-03-01
"missing" annotation keyword


awwright:
The "default" keyword is one of the most misused & abused keywords due to consequences of how JSON Schema works.

It is frequently assumed that if a property in an instancec document is missing, the "default" keyword lets you create that property and fill in that value as if it has the same behavior.

But this is not actually the case. The biggest reason is the "default" keyword does not produce an annotation unless the property exists in the instance in the first place... defeating the point.

The "default" keyword is mostly useful for user interface and IDE work, where a user indicates they want to create a value, and so the "default" keyword can provide a sensible initial default in these cases. For example:

- A user creates a new record in a document database (like a MongoDB collection). The interface creates an instance of the schema, reading the "default" keyword to provide a value; instead of creating a blank document (which is invalid JSON).

- A user in an IDE is typing `{ "name": ` and then tab-completes in the default value, an empty string (as opposed to a number).

However, all the time, users seem to think you can substitute in the "default" value for a missing one:

* https://github.com/json-schema-org/json-schema-spec/issues/858
* https://stackoverflow.com/questions/60433669/how-to-add-default-values-from-json-schema-to-json-document-in-java/60443802 (see [my answer](https://stackoverflow.com/questions/60433669/how-to-add-default-values-from-json-schema-to-json-document-in-java/60443802#60443802) for more details on why "default" probably doesn't do what you think it does)

This also seemed sensible to me until I realized that "default" doesn't produce an annotation if the instance is absent.

I'm proposing a keyword that does mean exactly this: it _means_ "if the instance is an object, and it is missing any of the given properties, the behavior will be the same as if it were defined with the specified value".

Example schema:

```javascript
{
  "default": {
    "port": 80
  },
  "properties": {
    "port": { "type":"integer" }
  }
}
``` 

This would allow implementations to infer that this:

```
{ }
```

will behave the same as this:

```
{ "port": 80 }
```

and some implementations or tools might even offer a way to return an instance with these values filled in. This would be useful for applications that want to both validate user input, and fill in defaults; instead of having to perform these as separate operations.

handrews:
That is a good point about `default` never actually annotating an instance!  I'll have to think more on the rest.

awwright:
Alternative names could be e.g. "undefinedValues", "defaultValues" or simply "defaults" (plural).

We may also want to consider an array form, for tuples, or function arguments.

[...]

PR #1006, opened 2020-10-22
provide defaults for all keywords and subschemas that lacked them

karenetheridge:
For vocabulary metaschemas, the default is just 'true'; for
the 'dependentRequired' and 'dependentSchemas' keywords, the default is the
empty object.

### DIFF
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -50,5 +50,6 @@
             "minItems": 1,
             "items": { "$dynamicRef": "#meta" }
         }
-    }
+    },
+    "default": true
 }
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -17,19 +17,16 @@
         },
         "properties": {
             "type": "object",
-            "additionalProperties": { "$dynamicRef": "#meta" },
-            "default": {}
+            "additionalProperties": { "$dynamicRef": "#meta" }
         },
         "patternProperties": {
             "type": "object",
             "additionalProperties": { "$dynamicRef": "#meta" },
-            "propertyNames": { "format": "regex" },
-            "default": {}
+            "propertyNames": { "format": "regex" }
         },
         "dependentSchemas": {
             "type": "object",
-            "additionalProperties": { "$dynamicRef": "#meta" },
-            "default": {}
+            "additionalProperties": { "$dynamicRef": "#meta" }
         },
         "propertyNames": { 
             "$dynamicRef": "#meta"
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -12,25 +12,24 @@
         "prefixItems": { "$ref": "#/$defs/schemaArray" },
         "items": { "$dynamicRef": "#meta" },
         "contains": { "$dynamicRef": "#meta" },
-        "additionalProperties": {
-            "$dynamicRef": "#meta"
-        },
+        "additionalProperties": { "$dynamicRef": "#meta" },
         "properties": {
             "type": "object",
-            "additionalProperties": { "$dynamicRef": "#meta" }
+            "additionalProperties": { "$dynamicRef": "#meta" },
+            "$comment": "default: {}"
         },
         "patternProperties": {
             "type": "object",
             "additionalProperties": { "$dynamicRef": "#meta" },
-            "propertyNames": { "format": "regex" }
+            "propertyNames": { "format": "regex" },
+            "$comment": "default: {}"
         },
         "dependentSchemas": {
             "type": "object",
-            "additionalProperties": { "$dynamicRef": "#meta" }
+            "additionalProperties": { "$dynamicRef": "#meta" },
+            "$comment": "default: {}"
         },
-        "propertyNames": { 
-            "$dynamicRef": "#meta"
-        },
+        "propertyNames": { "$dynamicRef": "#meta" },
         "if": { "$dynamicRef": "#meta" },
         "then": { "$dynamicRef": "#meta" },
         "else": { "$dynamicRef": "#meta" },
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -16,18 +16,18 @@
         "properties": {
             "type": "object",
             "additionalProperties": { "$dynamicRef": "#meta" },
-            "$comment": "default: {}"
+            "default": {}
         },
         "patternProperties": {
             "type": "object",
             "additionalProperties": { "$dynamicRef": "#meta" },
             "propertyNames": { "format": "regex" },
-            "$comment": "default: {}"
+            "default": {}
         },
         "dependentSchemas": {
             "type": "object",
             "additionalProperties": { "$dynamicRef": "#meta" },
-            "$comment": "default: {}"
+            "default": {}
         },
         "propertyNames": { "$dynamicRef": "#meta" },
         "if": { "$dynamicRef": "#meta" },
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -48,6 +48,5 @@
             "minItems": 1,
             "items": { "$dynamicRef": "#meta" }
         }
-    },
-    "default": true
+    }
 }
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -13,8 +13,7 @@
         "items": { "$dynamicRef": "#meta" },
         "contains": { "$dynamicRef": "#meta" },
         "additionalProperties": {
-            "$dynamicRef": "#meta",
-            "default": {}
+            "$dynamicRef": "#meta"
         },
         "properties": {
             "type": "object",
@@ -33,8 +32,7 @@
             "default": {}
         },
         "propertyNames": { 
-            "$dynamicRef": "#meta",
-            "default": {}
+            "$dynamicRef": "#meta"
         },
         "if": { "$dynamicRef": "#meta" },
         "then": { "$dynamicRef": "#meta" },
--- a/meta/core.json
+++ b/meta/core.json
@@ -48,6 +48,5 @@
             "type": "string",
             "format": "uri-reference"
         }
-    },
-    "default": true
+    }
 }
--- a/meta/core.json
+++ b/meta/core.json
@@ -34,9 +34,6 @@
             "additionalProperties": { "$dynamicRef": "#meta" }
         }
     },
-    "default": {
-        "$schema": "https://json-schema.org/draft/2020-12/schema"
-    },
     "$defs": {
         "anchorString": {
             "type": "string",
--- a/meta/core.json
+++ b/meta/core.json
@@ -31,9 +31,11 @@
         },
         "$defs": {
             "type": "object",
-            "additionalProperties": { "$dynamicRef": "#meta" },
-            "default": {}
+            "additionalProperties": { "$dynamicRef": "#meta" }
         }
+    },
+    "default": {
+        "$schema": "https://json-schema.org/draft/2020-12/schema"
     },
     "$defs": {
         "anchorString": {
--- a/meta/core.json
+++ b/meta/core.json
@@ -48,5 +48,6 @@
             "type": "string",
             "format": "uri-reference"
         }
-    }
+    },
+    "default": true
 }
--- a/meta/validation.json
+++ b/meta/validation.json
@@ -50,13 +50,11 @@
         "maxItems": { "$ref": "#/$defs/nonNegativeInteger" },
         "minItems": { "$ref": "#/$defs/nonNegativeIntegerDefault0" },
         "uniqueItems": {
-            "type": "boolean",
-            "default": false
+            "type": "boolean"
         },
         "maxContains": { "$ref": "#/$defs/nonNegativeInteger" },
         "minContains": {
-            "$ref": "#/$defs/nonNegativeInteger",
-            "default": 1
+            "$ref": "#/$defs/nonNegativeInteger"
         },
         "maxProperties": { "$ref": "#/$defs/nonNegativeInteger" },
         "minProperties": { "$ref": "#/$defs/nonNegativeIntegerDefault0" },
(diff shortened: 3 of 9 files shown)


awwright:
What's a concrete example of functionality this provides?

My interpretation of "default" is a suitable initial value for the instance when it's created; so at the very least, I think `{}` may be a preferable value over `true`. See also #867

karenetheridge:
This is to restore consistency to the metaschemas. In most other places, a `default` keyword is provided to indicate the intended behaviour when the keyword is absent from a subschema -- this just fills in the places where the default was missing.

karenetheridge:
Is this good to include in draft2020-xx?

karenetheridge:
rebased

Relequestual (approved):

Relequestual (approved):

gregsdennis:
Hm.. I didn't see this when I did #1057.  May have duplication.

Relequestual:
Given that we don't provide defaults in the same schema object which uses a ref or dynamic ref, we should remove that found in the applicator vocab schema on line 17, right?

https://github.com/json-schema-org/json-schema-spec/blob/77686deef8ec6363b76272c065b83e34c455e835/meta/applicator.json#L15-L18

Relequestual:
Can this be rebased one more time please?

karenetheridge:
rebased!

meta/applicator.json
```diff
@@ -13,8 +13,7 @@
         "items": { "$dynamicRef": "#meta" },
         "contains": { "$dynamicRef": "#meta" },
         "additionalProperties": {
-            "$dynamicRef": "#meta",
-            "default": {}
```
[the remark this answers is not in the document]
gregsdennis: From the[ spec on `additionalProperties`](https://github.com/json-schema-org/json-schema-spec/blob/master/jsonschema-core.xml#L2427):

> Omitting this keyword has the same assertion behavior as an empty schema.

gregsdennis (changes requested):

meta/core.json
```diff
@@ -48,5 +48,6 @@
             "type": "string",
             "format": "uri-reference"
         }
-    }
+    },
+    "default": true
```
gregsdennis, line 52: I don't agree with having defaults in the roots of the meta-schemas.  What's the motivation for this?

jdesrosiers: I get that the motivation is to avoid putting `"default": true` everywhere we reference a schema, but the default isn't always "true" for all references.

Traditionally, in our meta-schemas, `default` has meant something like: if this doesn't exist, the validation result would be the same as if it did exist and had this default value. For example, the `default` of `properties` is `{}`. Therefore, `{}` has the same result as `{ "properties": {} }`.

So, now consider the `not` keyword. It references the top-level meta-schema. If the top-level meta-schema has `"default": true` that says the `default` of `not` is `true`. Therefore, `{}` has the same result as `{ "not": true }`, which is not correct. That means each time something references the top-level meta-schema it needs to define a `default` that makes sense in its context.

... or just get rid of `default` in meta-schemas altogether because it doesn't provide any real value.

gregsdennis: @jdesrosiers I _think_ we're arguing the same side of the coin.

jdesrosiers: Yes, we are. Bottom line is that `default` doesn't belong at the root of meta-schemas.

karenetheridge: > So, now consider the not keyword. It references the top-level meta-schema. If the top-level meta-schema has "default": true that says the default of not is true. Therefore, {} has the same result as { "not": true }, which is not correct. That means each time something references the top-level meta-schema it needs to define a default that makes sense in its context.

Or, we can add a default for not -- which should be false.  that is, `..., "not": { "$dynamicRef": "#meta", "default": false }, ...`

So, how about we remove all uses of default where it's just the obvious thing (`{}` for objects, or `true`or `{}` for schemas), and keep them only when they're non-standard -- which is, I believe, just "not", "deprecated", "readOnly", "writeOnly", "uniqueItems", and "minContains".)

If that's amenable to all I can amend this PR.

karenetheridge:
In the end, this is the diff I get against master. is this what we want?

```
diff --git a/meta/applicator.json b/meta/applicator.json
index ff2cfc2..8652779 100644
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -13,8 +13,7 @@
         "items": { "$dynamicRef": "#meta" },
         "contains": { "$dynamicRef": "#meta" },
         "additionalProperties": {
-            "$dynamicRef": "#meta",
-            "default": {}
+            "$dynamicRef": "#meta"
         },
         "properties": {
             "type": "object",
@@ -33,8 +32,7 @@
             "default": {}
         },
         "propertyNames": { 
-            "$dynamicRef": "#meta",
-            "default": {}
+            "$dynamicRef": "#meta"
         },
         "if": { "$dynamicRef": "#meta" },
         "then": { "$dynamicRef": "#meta" },
diff --git a/meta/validation.json b/meta/validation.json
index 606b87b..1432ad8 100644
--- a/meta/validation.json
+++ b/meta/validation.json
@@ -65,7 +65,8 @@
             "type": "object",
             "additionalProperties": {
                 "$ref": "#/$defs/stringArray"
-            }
+            },
+            "default": {}
         }
     },
     "$defs": {
```

karenetheridge:
Given that we have been using the default keyword wrong anyway (missing keywords cannot produce annotations), maybe something like this is more correct: https://gist.github.com/karenetheridge/1239b063611b937c38713f6ac6a6eaa6

jdesrosiers:
@karenetheridge 
> missing keywords cannot produce annotations

That has never occurred to me. We would only know the `default` value if the value exists, but the only reason to care about the `default` value is if the value doesn't exist.

> ```json
>    "default": {
>        "items": true,
>        "contains": true,
>        "additionalProperties": true,
>        "properties": {},
>        "patternProperties": {},
>        "dependentSchemas": {},
>        "propertyNames": true,
>        "if": true,
>        "then": true,
>        "else": true,
>        "not": false
>    }
> ```

If an entire schema is missing, I would expect this `default` to apply. However, if the schema does exist with some of these values missing, I wouldn't expect this default to those values. The `default` would be applied all or nothing. So, I don't think this makes sense either.

It seems `default` is completely useless, even as an annotation.

karenetheridge:
> If an entire schema is missing, I would expect this default to apply. However, if the schema does exist with some of these values missing, I wouldn't expect this default to those values. The default would be applied all or nothing. So, I don't think this makes sense either.

Well, the user can do whatever he likes with the annotations he gets back -- it doesn't have to be all or nothing. In some of my code (that is, the caller of the json schema evaluator - not the evaluator itself!), I retrieve that default annotation and add its values into the data instance for every property that is missing.

jdesrosiers:
Here's another approach we might take with where we use `default` in meta-schemas.

If a keyword isn't present it has no effect. There's no reason to declare a default value that also has no effect. The only reason to have a default value is for keywords that depend on other keywords for their results. In order to process `additionalProperties`, we need to know the value of `properties` and `patternProperties`. Therefore, `properties` and `patternProperties` should have a default value. Things like `allOf`/`anyOf`/`oneOf`/`not` don't need defaults because nothing depends on their presence.

`if` would be a weird one tho. I'd have to look it up, but I'm pretty sure `then` and `else` don't apply if `if` is not present. So, when you're looking at `then`, `if`'s default is false. If you're looking a `else`, `if`'s default true.

karenetheridge:
I think we can leave out properties and patternProperties too -- as the spec says what to do if those keywords are missing (no keyword -> no annotation -> nothing to consume when considering the result of additionalProperties, unevaluatedProperties).

We can leave out if/then/else and not as well -- no keyword -> nothing to do -> no false value injected into the result.

The only outliers really are "minContains", which defaults to 1 when absent, and the meta-data keywords which default to false. We should be able to drop all the rest.

jdesrosiers:
I think `properties` and `patternProperties` might be nice to have as a way of documenting what the spec says, but I agree they aren't necessary.

I agree `if`/`then`/`else` don't need `default`s and `minContains` should have one.

Others to consider are
1. `addtionalProperties` - `true` - depended on by `unevaluatedProperties`
2. `items` - `true` - depended on by `unevaluatedItems`
3. `prefixItem` - `[]` - depended on by `items`

I think it would be a nice-to-have to include these defaults, but not necessary.

karenetheridge:
I pushed another commit for consideration, which would be squashed with the other two if acceptable.

meta/core.json
```diff
@@ -31,10 +31,12 @@
         },
         "$defs": {
             "type": "object",
-            "additionalProperties": { "$dynamicRef": "#meta" },
-            "default": {}
+            "additionalProperties": { "$dynamicRef": "#meta" }
         }
     },
+    "default": {
+        "$schema": "https://json-schema.org/draft/2020-12/schema"
```
jdesrosiers, line 38: Wouldn't this mean that every schema, including sub-schemas, have a `$schema` keyword. `$schema` has no effect in sub-schemas, so this doesn't make sense to me.

jdesrosiers:
@karenetheridge
> the user can do whatever he likes with the annotations he gets back -- it doesn't have to be all or nothing

I agree that `default` can be used however we want, but I'm not a fan of this approach. It feels awkward and wrong to me. I'd rather use it the way we have been.

Not a hard objection, just my preference. I won't argue if others like it.

meta/core.json
karenetheridge: Yes, good point. So we'll have to leave it out.. as we have no way currently of specifying in the metaschema "this should apply for the top level schema but not for anything dynamically referenced lower down".

awwright:
> It seems `default` is completely useless, even as an annotation.

This is close to my conclusion too, see #867

meta/validation.json
```diff
@@ -68,14 +66,16 @@
             }
         }
     },
+    "default": {
+        "minContains": 1
+    },
```
gregsdennis, line 71: I'm not a fan of this format, and I don't think we should encourage it.  I think the default value should stay with the property declaration.

meta/validation.json
```diff
@@ -50,13 +50,11 @@
...
         "minItems": { "$ref": "#/$defs/nonNegativeIntegerDefault0" },
         "uniqueItems": {
-            "type": "boolean",
-            "default": false
+            "type": "boolean"
         },
         "maxContains": { "$ref": "#/$defs/nonNegativeInteger" },
         "minContains": {
-            "$ref": "#/$defs/nonNegativeInteger",
-            "default": 1
```
gregsdennis, line 59: this should stay

gregsdennis:
I agree with @jdesrosiers's idea of only including defaults for keywords that have other keywords being dependent upon them.

---

Going forward, maybe we need to declare the default values of keywords _in the context where they're depended upon_.  For example, the section for `contains` would define a default for `minContains` should it be absent, opposed to as it is now where the section for `minContains` declares its own default.  This also has two side effects:

- a keyword (or subschema) can have different defaults based on the context in which it's used,
- custom keywords can be written (in new vocabs) that utilize keywords with different defaults.

Note that this doesn't give the `default` keyword any more purpose than for humans.

meta/applicator.json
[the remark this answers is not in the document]
Relequestual: Given we're no longer looking to have `default` here, let's consider this resolved, of sorts.

Relequestual:
In agreement with changes requested by @gregsdennis and @jdesrosiers, including those in 
https://github.com/json-schema-org/json-schema-spec/pull/1006#issuecomment-764977292


    addtionalProperties - true - depended on by unevaluatedProperties
    items - true - depended on by unevaluatedItems
    prefixItem - [] - depended on by items

meta/validation.json
karenetheridge: we're okay with using "default" totally contradictory to the spec? If this is here purely for humans, it should be in a `$comment`.

karenetheridge:
>  For example, the section for contains would define a default for minContains should it be absent

I don't see how that can be done with current syntax, because at the position of the contains keyword we are defining things at the `.../contains/` position, not `.../`.

karenetheridge:
I have pushed one more commit for consideration.

meta/validation.json
jdesrosiers: I don't think this use is contradictory to the spec, because the spec defines `default` so broadly that there is no wrong way to use it. Since there is no right or wrong way to use it and it doesn't do anything anyway, I'd rather use it in the spirit of how it was intended to be used.

> If this is here purely for humans, it should be in a `$comment`.

I had the same thought at first, but I've come to the conclusion that just because `default` is not useful as an annotation, doesn't mean it's not useful. Consider that a documentation generator doesn't use instances and therefore doesn't work on annotations as they are defined by the spec. The documentation generator can still make good use of `default` because the limitations of using it as an annotation don't apply in that context.

Relequestual: I'm in agreement with this. We should consider other applications beyond just being used with an instance, the most obvious, to me, being writing schemas by hand or using UI tools (in terms of adding these for the meta schema).

gregsdennis:
> I don't see how that can be done with current syntax...

Yeah, it can't. Was just thinking out loud.

Relequestual:
So the net result of this PR is `dependencies` gets a `default`? Was that the intent?

Do we want to handle defaults for `prefixItem` and `items` in a separate PR?

karenetheridge:
The net result is that defaults have been removed for all subschemas that recursively-reference `#meta` (e.g. additionalProperties, propertyNames), and dependencies gets a default. borrrring!

prefixItems already has a default, via the schemaArray definition, and items is another `#meta` reference.

Relequestual (approved):

karenetheridge:
I can squash with a better commit message if we're good with this.

PR #1013, opened 2020-11-09
Fix ECMA-262 link

ssilverman:
According to the main ECMA-262 page, the link needs an "index.html" suffix.
There's no guarantee that linking to the raw "11.0/" directory will lead to
the correct place, even if it does. This updates to the correct well-known
link. Not all sites necessarily support a default page for an empty path
suffix; it is not correct to remove the last part.

Ref: https://www.ecma-international.org/publications/standards/Ecma-262.htm


Relequestual:
Thanks for your dedication!

PR #1017, opened 2020-11-13
Replace the phrase "feature set" with dialect

Relequestual:
Made it clearer that the value of $schema is a dialect identifier,

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1168,9 +1168,9 @@
                 </t>
                 <section title='The "$schema" Keyword' anchor="keyword-schema">
                     <t>
-                        The "$schema" keyword is both used as a JSON Schema feature set identifier and
+                        The "$schema" keyword is both used as a JSON Schema dialect identifier and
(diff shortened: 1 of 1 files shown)


ssilverman:
I don't love the word "dialect" because it, to me, implies a whole new interpretation of JSON schema. I much prefer "feature set". This is just my opinion and I'm not proposing you change it back if "dialect" is the final choice.

Relequestual:
It's analogous to SQL which has different dialects.

handrews:
"dialect" came from me explaining this stuff to various people, including on consulting contracts.  It was effective and seemed easily understood.  My sample size is not huge, but there was certainly zero resistance to it.  Folks noted that it fits well with "vocabulary."  And @Relequestual , SQL dialects are a good analogue for explaining this, and a well-established precedent.

ssilverman:
If it works, it works.

ssilverman (approved):
👍

handrews (approved):

PR #1022, opened 2020-11-19
added language to explicitly specify behavior when a vocab is known

### LINKED ISSUE #1019: `$vocabulary:{<vocab>:false}` when a vocab is recognized is not defined
The spec states

> If the value is true, then implementations that do not recognize the vocabulary MUST refuse to process any schemas that declare this meta-schema with "$schema".  If the value is false, implementations that do not recognize the vocabulary SHOULD proceed with processing such schemas.

This doesn't say what the behavior is when the value is false and the implementation _does_ recognize the vocabulary.  The implication is that keywords defined by such vocabularies should still be processed.

@karenetheridge disagrees with this implication in that she interprets this as having the vocab configured as `false` means that the vocab's keywords can be entirely skipped (collected as annotations) even when the vocab is known.

It should be explicitly stated.

---

I had extensive DM conversations around this keyword with @handrews when he was first defining it for 2019-09 (sadly those are now lost to the Free Slack Aether).  This wasn't a special case for format.  The intent is to process the keywords if the vocab is known, regardless of the value.  Having `<vocab>: false` simply means that knowledge of the vocab doesn't preclude the processability of a schema.

gregsdennis:


jsonschema-core.xml
```diff
@@ -1226,7 +1228,8 @@
                         the vocabulary MUST refuse to process any schemas that declare
                         this meta-schema with "$schema".  If the value is false, implementations
                         that do not recognize the vocabulary SHOULD proceed with processing
-                        such schemas.
+                        such schemas.  The value has no impact if the implementation
+                        recognizes the vocabulary.
```
Relequestual, line 1232:
```suggestion
                        understands the vocabulary.
```

Recognising a vocabulary isn't the same as understanding it.

Relequestual (changes requested):
Single change suggestion. Semantics! 🙄

Otherwise, this PR is good to go.
Thanks for your effort in making it happen =]

gregsdennis:
merge when ready

karenetheridge:
Is a footnote necessary, to indicate that the handling for the format vocabulary is special-cased?

gregsdennis:
@karenetheridge no.  Please see #1023 for changes to `format`, which remove the special-case treatment.  This change is purely for vocabularies.

karenetheridge (approved):

Relequestual (approved):

PR #1023, opened 2020-11-19
reworked 'format' to default to annotation except when configured for assertion

### LINKED ISSUE #1020: `format` should not change behavior based on its vocabulary value
> The true or false value of the vocabulary declaration governs the implementation requirements necessary to process a schema that uses "format", and the behaviors on which schema authors can rely.

This is just wrong.  It precludes an ability to assert format when supported but not require assertion when not supported.  The spec still allows for configuration of this, but the practical side of that configuration becomes quite difficult and confusing:

- For <vocab>: false, my configuration needs to say "format should be an assertion."
- For <vocab>: true, my configuration needs to say "format should be an annotation."

This is hard to do and confusing for clients (users of the implementation).  The configuration should always work one direction.  If my implementation offers a "format behavior" configuration with values of "assert" and "annotate," setting either value only works for one of the vocab cases.  The only way to get the desired behavior is to have a configuration that says "use the non-default behavior," which changes its meaning depending on the schema it's processing.

What we SHOULD have is format as an annotation ALWAYS, but configurable to be assertion.

I agree with @karenetheridge that it'd be nice to have a way for the schema itself to indicate how format should be processed, and I think that changing it based on the vocab value was an attempt at that.  But the vocab value and the behavior of format are orthogonal concerns.  The spec is conflating them unnecessarily.

gregsdennis:
There is also this in the "Custom format attributes" section below the edits:

> An implementation MUST NOT fail validation or cease processing due to an unknown format attribute.

What is an implementation to do when an unknown format is encountered other than fail?

---

Closes #1026.
Closes #1027.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -550,6 +550,11 @@
                     See the Implementation Requirements below for details.
                 </t>
                 <t>
+                    The <xref target="meta-schema">standard core and validation meta-schema</xref>
+                    includes this vocabulary in its "$vocabulary" keyword with a value of false
+                    since implementations are not required to provide support for this keyword.
+                </t>
+                <t>
                     The current URI for this vocabulary, known as the Format vocabulary, is:
                     &lt;https://json-schema.org/draft/2020-11/vocab/format&gt;.
                 </t>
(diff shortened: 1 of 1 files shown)


Relequestual:
I'm not in favour of this change at this time.
I have explained why here: https://github.com/json-schema-org/json-schema-spec/issues/1020#issuecomment-730387830

gregsdennis:
Alternative change in #1026.

PR #1024, opened 2020-11-19
Clearly define a dialect

### LINKED ISSUE #1015: Define dialect
Dialect is not clearly defined. It should be.

Relequestual:
Fixes #1015

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -103,7 +103,8 @@
                 This specification defines JSON Schema core terminology and mechanisms, including
                 pointing to another JSON Schema by reference,
                 dereferencing a JSON Schema reference,
-                specifying the vocabulary being used,
+                specifying the dialect being used,
+                specifying a dialect's vocabulary requirements,
                 and defining the expected output.
             </t>
             <t>
@@ -146,9 +147,10 @@
                 some sort of condition.
             </t>
             <t>
-                To facilitate re-use, keywords can be organized into vocabularies.  A vocabulary
+                To facilitate re-use, keywords can be organized into vocabularies. A vocabulary
                 consists of a list of keywords, together with their syntax and semantics.
-                A set of vocabularies identified by a meta-schema is known as a dialect.
+                A dialect is defined as a set of vocabularies and their required support
+                identified in a meta-schema.
             </t>
             <t>
                 JSON Schema can be extended either by defining additional vocabularies,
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -1198,7 +1200,8 @@
                         vocabularies available for use in schemas described by that meta-schema.
                         It is also used to indicate whether each vocabulary is required or optional,
                         in the sense that an implementation MUST understand the required vocabularies
-                        in order to successfully process the schema.
+                        in order to successfully process the schema. Together, this information forms
+                        a dialect.
```
ssilverman, line 1204: This is even more confusing to me now because the “$vocabulary” keyword is used to say which vocabularies are included, but now you’re saying that this collection of vocabularies is really a “dialect”. I think introducing “dialect” contradicts with “$vocabulary” as singular and also “$vocabulary” as a “collection of vocabularies”. The keyword should really be “$dialect” otherwise this feels much more messy.

Relequestual: `$schema` in a schema is a dialect identifier and the `$id` of the dialects meta-schema.

tbh, I always misread it as  `$vocabularies`, but you can justify it as`$vocabulary[ specification object]`.

99.99% of people won't ever use this. I'm not here to bikeshed keyword names today.

Also, we didn't think of using the phrase dialect when we worked on 2019-09.

ssilverman: I’m not suggesting a change to the keywords (because I’m imagining it’s too late). I just thought it was important to point out that it’s confusing to me that “$vocabulary” = “dialect definition, a collection of vocabularies”. I agree about your “$schema” = “dialect” statement. So is it true, then, that “meta-schema” = “dialect”? If my statements are true, then it might be useful to clarify in the text.

Relequestual: Well, the vocabulary is actually the specification of the keywords... the `$vocabulary` keyword just includes their identifying URIs...

Relequestual: As such, it follows, the meta-schema itself is not the dialect, but the specification document DEFINES the dialect, by specifying the default vocabulary support.

handrews: Ideas don't always come in the order that is ideal for getting the details perfect, sadly.  I would not object to it being `$dialect` but we've put so much PR effort into "vocabulary" that I question whether it's enough a mismatch to be a problem.

Whether dialects or vocabularies, specifications _define_ and meta-schemas _describe_.

Relequestual: Only people who write meta-schemas will care about dialects. That is few. How many people go about creating their own dialect of SQL? Not many.

gregsdennis (approved):

PR #1025, opened 2020-11-19
Attempt to justify and explain unevaluated keyword phrasing

### LINKED ISSUE #998: Defining "evaluation"
From @Relequestual:

> I've been thinking about this section and its phrasing.
It's the first time in the whole document we mention "evaluation", and we don't really define what it means.
>
> People (including myself previously) incorrectly assume that if a location has been evaluated (had something applied to it), then it counts as having been evaluated.
>
> I'm going to riff some ideas...
>
> We say here "successful evaluations", we mean if an application doesn't result in validation errors, and therefore the applied schema or keyword is correctly applicable to the instance location.
>
> For example, if you had a schema...
anyOf [ { title> unicycle, props > wheels > const > 1 }, { title: bicycle, props > wheels > const > 2 } ]
>
> If your instance was a unicycle, the first subschema item in the anyOf would be applied and successfully evaluated. Because it only has one wheel, we know it's a unicycle. It is not a bicycle, so that whole subschema is not successfully evaluated.
>
> The instance is successfully evaluated as (or determined to be) a unicycle.
>
> We can further justify this terminology by the fact that evaluation can be short-circuited, and therefore there's no expectation for a schema to be fully evaluated (although that would be a weak argument, and probably confuse things).
>
> We need to clearly define what successful evaluation means, and what unsuccessful evaluation means, with justification (IMHO). I'm not currently convinced I can do that.

Relequestual:
Fixes #998

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2577,6 +2577,14 @@
                 Only successful evaluations are considered.
             </t>
             <t>
+                If an item in an array or an object property is "successfully evaluated", it
+                is logically considered to be valid in terms of the representation of the
+                object or array that's expected. For example if a subschema represents a car,
+                which requires between 2-4 wheeles, and the value of "wheeles" is 6, the instance
+                object is not "evaluated" to be a car, and the "wheeles" property is considerd
+                "unevaluated (successfully as a known thing)", and does not retain any annotations.
+            </t>
+            <t>
                 Recall that adjacent keywords are keywords within the same schema object,
                 and that the dynamic-scope subschemas include reference targets as well as
                 lexical subschemas.
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2581,7 +2581,7 @@
                 is logically considered to be valid in terms of the representation of the
                 object or array that's expected. For example if a subschema represents a car,
                 which requires between 2-4 wheeles, and the value of "wheeles" is 6, the instance
-                object is not "evaluated" to be a car, and the "wheeles" property is considerd
+                object is not "evaluated" to be a car, and the "wheeles" property is considered
                 "unevaluated (successfully as a known thing)", and does not retain any annotations.
             </t>
             <t>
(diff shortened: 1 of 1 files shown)


Relequestual:
I tried to take a stab at this, but I'm not sure if this actually helps, or causes further confusion.
Please give me your thoughts!

jsonschema-core.xml
```diff
@@ -2576,6 +2576,14 @@
                 to contribute to whether or not the item or property has been evaluated.
                 Only successful evaluations are considered.
             </t>
+            <t>
+                If an item in an array or an object property is "successfully evaluated", it
+                is logically considered to be valid in terms of the representation of the
+                object or array that's expected. For example if a subschema represents a car,
+                which requires between 2-4 wheeles, and the value of "wheeles" is 6, the instance
```
handrews, line 2583: Is "wheeles" British spelling of American "wheels"?  I was not aware of this one, so it's a bit jarring to read.

However, this explanation is fantastic and I enthusiastically approve of this PR.

handrews (approved):

ssilverman (changes requested):
“wheeles” -> “wheels”

jsonschema-core.xml
Relequestual: Sadly, this is just my almost dyslexia coming into play

jsonschema-core.xml
```diff
@@ -2576,6 +2576,14 @@
...
+                object is not "evaluated" to be a car, and the "wheeles" property is considerd
```
karenetheridge, line 2584:
```suggestion
                object is not "evaluated" to be a car, and the "wheeles" property is considered
```

karenetheridge (changes requested):

gregsdennis (approved):

ssilverman (approved):

Relequestual:
🚢

PR #1027, opened 2020-11-21
`format` as a annotation by default - separate vocabs (option 3)

### LINKED ISSUE #1020: `format` should not change behavior based on its vocabulary value

gregsdennis:
There is also this in the "Custom format attributes" section below the edits:

> An implementation MUST NOT fail validation or cease processing due to an unknown format attribute.

What is an implementation to do when an unknown format is encountered other than fail?

The context may be around collecting annotations, but it says "validation," so I'm confused.  Maybe this needs to be clarified a bit too.

---

Resolves #1020.
Closes #1023.
Closes #1026.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -610,7 +610,7 @@
 
                 <section title="Format-Assertion Vocabulary">
                     <t>
-                        When the Format-Assertion vocabulary is declared with a value of false,
+                        When the Format-Assertion vocabulary is declared with a value of true,
                         implementations MUST provide full validation support for all of the formats
                         defined by this specificaion. Implementations that cannot provide full
                         validation support MUST refuse to process the schema.
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -547,7 +547,7 @@
                     In addition to the Format-Annotation vocabulary, a secondary vocabulary is available
                     for custom meta-schemas that defines "format" as an assertion. The URI for the
                     Format-Assertion vocabulary, is:
-                    &lt;https://json-schema.org/draft/2020-11/vocab/format-assertation&gt;.
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-assertion&gt;.
                 </t>
                 <t>
                     The current URI for the corresponding meta-schema is:
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -668,10 +668,9 @@
                     <t>
                         Implementations MAY support custom format attributes. Save for agreement between
                         parties, schema authors SHALL NOT expect a peer implementation to support such
-                        custom format attributes.  An implementation MUST NOT fail
-                        validation or cease processing due to an unknown format attribute.
-                        When treating "format" as an annotation, implementations SHOULD collect both
-                        known and unknown format attribute values.
+                        custom format attributes. An implementation MUST NOT fail to collect unknown formats
+                        as annotations.  When the Format-Assertion vocabulary is specified, implementations
+                        MUST fail upon encountering unknown formats.
                     </t>
                     <t>
                         Vocabularies do not support specifically declaring different value sets for keywords.
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -540,20 +540,19 @@
 
                 <t>
                     The current URI for this vocabulary, known as the Format-Annotation vocabulary, is:
-                    &lt;https://json-schema.org/draft/2020-11/vocab/format-annotation&gt;.  This vocabulary
-                    is required by this specification.
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-annotation&gt;. The current
+                    URI for the corresponding meta-schema is:
+                    <eref target="https://json-schema.org/draft/2020-11/meta/format-annotation"/>.
+                    Implementing support for this vocabulary is REQUIRED.
                 </t>
                 <t>
                     In addition to the Format-Annotation vocabulary, a secondary vocabulary is available
                     for custom meta-schemas that defines "format" as an assertion. The URI for the
                     Format-Assertion vocabulary, is:
-                    &lt;https://json-schema.org/draft/2020-11/vocab/format-assertion&gt;.
-                </t>
-                <t>
-                    The current URI for the corresponding meta-schema is:
-                    <eref target="https://json-schema.org/draft/2020-11/meta/format"/>. Because the
-                    syntactic requirements of "format" do not change between the annotation and assertion
-                    vocabularies, the meta-schema is shared between them.
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-assertion&gt;. The current
+                    URI for the corresponding meta-schema is:
+                    <eref target="https://json-schema.org/draft/2020-11/meta/format-assertion"/>.
+                    Implementing support for the Format-Assertion vocabulary is OPTIONAL.
                 </t>
                 <t>
                     Specifying both the Format-Annotation and the Format-Assertion vocabularies is functionally
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -514,16 +514,10 @@
             <section title="Foreword">
                 <t>
                     Structural validation alone may be insufficient to allow an application to correctly
-                    utilize certain values.  The "format" annotation keyword is defined to allow schema
+                    utilize certain values. The "format" annotation keyword is defined to allow schema
                     authors to convey semantic information for a fixed subset of values which are
                     accurately described by authoritative resources, be they RFCs or other external
                     specifications.
-                </t>
-
-                <t>
-                    Implementations MAY treat "format" as an assertion in addition to an annotation,
-                    and attempt to validate the value's conformance to the specified semantics.
-                    See the Implementation Requirements below for details.
                 </t>
 
                 <t>
@@ -536,8 +530,8 @@
                     <xref target="json-schema">core JSON Schema.</xref>
                     <cref>
                         Note that the "type" keyword in this specification defines an "integer" type
-                        which is not part of the data model.  Therefore a format attribute can be
-                        limited to numbers, but not specifically to integers.  However, a numeric
+                        which is not part of the data model. Therefore a format attribute can be
+                        limited to numbers, but not specifically to integers. However, a numeric
                         format can be used alongside the "type" keyword with a value of "integer",
                         or could be explicitly defined to always pass if the number is not an integer,
                         which produces essentially the same behavior as only applying to integers.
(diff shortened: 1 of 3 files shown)


Relequestual:
> 
> 
> There is also this in the "Custom format attributes" section below the edits:
> 
> > An implementation MUST NOT fail validation or cease processing due to an unknown format attribute.
> 
> What is an implementation to do when an unknown format is encountered other than fail?
> 
> The context may be around collecting annotations, but it says "validation," so I'm confused. Maybe this needs to be clarified a bit too.

For annotation vocab, I would say, collect it as an annotation as long as it's a string. If it's not a string, then fail.

For assertion vocab, FAIL. The set of allowed values is explicit.

jsonschema-validation.xml
```diff
@@ -536,84 +530,66 @@
...
-                    &lt;https://json-schema.org/draft/2020-11/vocab/format&gt;.
+                    In addition to the Format-Annotation vocabulary, a secondary vocabulary is available
+                    for custom meta-schemas that defines "format" as an assertion. The URI for the
+                    Format-Assertion vocabulary, is:
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-assertation&gt;.
                 </t>
                 <t>
                     The current URI for the corresponding meta-schema is:
-                    <eref target="https://json-schema.org/draft/2020-11/meta/format"/>.
+                    <eref target="https://json-schema.org/draft/2020-11/meta/format"/>. Because the
```
Relequestual, line 554: I would really much rather this was explicitly a new URI part such as, `.../format-assertion` to make it clear this is not their regular ol `format` vocab they used previously.

gregsdennis:
I'll add this into the update.

jsonschema-validation.xml
```diff
@@ -536,84 +530,66 @@
...
                         The value of format MUST be collected as an annotation, if the implementation
-                        supports annotation collection.  This enables application-level validation when
+                        supports annotation collection. This enables application-level validation when
                         schema validation is unavailable or inadequate.
                     </t>
                     <t>
-                        This requirement is not affected by the boolean value of the vocabulary
-                        declaration, nor by the configuration of "format"'s assertion
-                        behavior described in the next section.
+                        Implementations MAY still treat "format" as an assertion in addition to an
```
Relequestual, line 577: I was thinking we would do away with that now, so any "best-effort" validation is not our concern, and in stead is  the result of processing annotation results

Relequestual (changes requested):
I'm not sure we were totally on the same page but close.
I was suggesting we should do away with saying anything about an implementation option to change annotation `format` to an assertion. It's confusing and was meant to only be a stop-gap till we found a solution.

Either you get annotations (default), and an implementation MAY do post-processing with the annotation results, but there should be no impact on the JSON Schema validation results.

Or you get full semantic validation.

jsonschema-validation.xml
gregsdennis: This is the meta-schema, which is shared.  The vocabs do have their own URIs.

jsonschema-validation.xml
gregsdennis: I think I'm with Henry in that it bears mentioning simply due to its history.

jsonschema-validation.xml
Relequestual: Yes... I blame sleep depreivation.

karenetheridge:
> For annotation vocab, I would say, collect it as an annotation as long as it's a string. If it's not a string, then fail.

This can only be done since we know that all existing formats are for instance data that are strings, and no other type. Are we expecting to add additional formats to this vocabulary in the future, or will those have to be added as separate vocabularies?

karenetheridge:
Another thought -- the format-as-assertion vocabulary should have an `enum` that lists the allowable values of the format keyword. Any unrecognized value should be a failure, and the metaschema can be used to perform that check.

jsonschema-validation.xml
```diff
@@ -536,84 +530,66 @@
...
                     </cref>
                 </t>
 
                 <t>
-                    Meta-schemas that do not use "$vocabulary" SHOULD be considered to
-                    utilize this vocabulary as if its URI were present with a value of false.
-                    See the Implementation Requirements below for details.
+                    The current URI for this vocabulary, known as the Format-Annotation vocabulary, is:
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-annotation&gt;.  This vocabulary
+                    is required by this specification.
```
karenetheridge, line 544:
```suggestion
                    is required to be implemented by this specification.
```

jsonschema-validation.xml
```diff
@@ -536,84 +530,66 @@
...
+                    for custom meta-schemas that defines "format" as an assertion. The URI for the
+                    Format-Assertion vocabulary, is:
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-assertion&gt;.
                 </t>
                 <t>
                     The current URI for the corresponding meta-schema is:
-                    <eref target="https://json-schema.org/draft/2020-11/meta/format"/>.
+                    <eref target="https://json-schema.org/draft/2020-11/meta/format"/>. Because the
+                    syntactic requirements of "format" do not change between the annotation and assertion
+                    vocabularies, the meta-schema is shared between them.
```
karenetheridge, line 556: Can we do this, given that meta/* files contain a $vocabulary keyword which documents the $vocabulary URI used for this keyword?

karenetheridge (commented):
I'm not sure of some of the specific wording here, but I lean towards favouring the approach in this PR vs. the other alternatives.

jsonschema-validation.xml
gregsdennis: That language is odd (the spec isn't doing the implementing) and a little redundant (I would think it's implied that a requirement would need to be implemented).

gregsdennis:
> the format-as-assertion vocabulary should have an enum that lists the allowable values of the format keyword

This would conflict with the section on custom keywords.  We need to allow for support for these, and having an enum of "approved" formats would prevent that.

jsonschema-validation.xml
gregsdennis: 🤔 good call.  I didn't see that.  I'll amend this and split out the meta-schemas.

karenetheridge:
> This would conflict with the section on custom keywords.

I had gotten the impression that we didn't want to continue to support custom formats -- if a user wants a new format they should create a new vocabulary for it. It would definitely be in the spirit of "making format more sane and less hand-wavey".

gregsdennis:
That's understandable, but custom formats aren't the problem that this PR is trying to resolve.  It's easy enough to continue supporting them for now.

jsonschema-validation.xml
Relequestual: OK then

jsonschema-validation.xml
Relequestual: Agree with @gregsdennis although the `required` should be `REQUIRED`.

jsonschema-validation.xml
Relequestual: Agree on the first part, but not the second.

jsonschema-validation.xml
gregsdennis: I see @karenetheridge's point, though.  I'll clean this up and make the requirements explicit for each vocabulary (`REQUIRED` for annotation, `OPTIONAL` for assertion).

gregsdennis: ✔️

jsonschema-validation.xml
gregsdennis: ✔️

gregsdennis:
As a side note, even though I split the meta-schemas, I haven't updated the draft: they're still 2019-09, but they're aligned with the others.

meta/format-annotation.json
```diff
@@ -0,0 +1,14 @@
+{
+    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$id": "https://json-schema.org/draft/2019-09/meta/format-annotation",
+    "$vocabulary": {
+        "https://json-schema.org/draft/2019-09/vocab/format-annotation": true
+    },
+    "$dynamicAnchor": "meta",
+
+    "title": "Format vocabulary meta-schema",
```
Relequestual, line 9:
```suggestion
    "title": "Format vocabulary meta-schema for annotation results",
```

Relequestual (changes requested):
Last change, but noticed the title for the meta-schema was the same for both `format` related meta-schemas, and this could be confusing.

If you could apply the same suggestions to both, I'll then approve and merge this PR.

jsonschema-validation.xml
```diff
@@ -634,14 +606,24 @@
                             validation in the application, which is the recommended best practice.
                         </cref>
                     </t>
+                </section>
 
+                <section title="Format-Assertion Vocabulary">
+                    <t>
+                        When the Format-Assertion vocabulary is declared with a value of false,
```
Relequestual, line 613:
```suggestion
                        When the Format-Assertion vocabulary is declared with a value of true,
```
...

> implementations MUST provide full validation support for all of the formats
>                         defined by this specificaion. Implementations that cannot provide full
>                         validation support MUST refuse to process the schema.


I'm assuming this is just an oversight, but there are no requirements when the vocabulary is declared with a value of `false` (which is "support not required").

Relequestual (changes requested):

jsonschema-validation.xml
gregsdennis: ✔️

meta/format-annotation.json
gregsdennis: ✔️

Relequestual (approved):

Relequestual:
I'm so annoying, but can you 
- Modify the section title from `A Vocabulary for Semantic Content With "format"'` to `Vocabularies for...`
- Update the general purpose meta-schema to include the new vocabulary URI
- Add something to the changelog about this

THEN, it will merge. Honest.

Relequestual:
Agreed with Greg, I will do the work as required in the above comment after merging this PR.

Relequestual:
I have done the three items as listed above.

jsonschema-validation.xml
```diff
@@ -514,18 +514,12 @@
             <section title="Foreword">
                 <t>
                     Structural validation alone may be insufficient to allow an application to correctly
-                    utilize certain values.  The "format" annotation keyword is defined to allow schema
+                    utilize certain values. The "format" annotation keyword is defined to allow schema
```
karenetheridge, line 517: can we keep unrelated whitespace changes out of the PR? it adds confusion and often conflicts with other things in flight.

jsonschema-validation.xml
```diff
@@ -536,84 +530,65 @@
...
 
                 <t>
-                    Meta-schemas that do not use "$vocabulary" SHOULD be considered to
-                    utilize this vocabulary as if its URI were present with a value of false.
-                    See the Implementation Requirements below for details.
+                    The current URI for this vocabulary, known as the Format-Annotation vocabulary, is:
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-annotation&gt;. The current
+                    URI for the corresponding meta-schema is:
+                    <eref target="https://json-schema.org/draft/2020-11/meta/format-annotation"/>.
+                    Implementing support for this vocabulary is REQUIRED.
```
karenetheridge, line 546: suggested change (using the proper IETF language):
+                    Implementations MUST support this vocabulary.

jsonschema-validation.xml
```diff
@@ -536,84 +530,65 @@
...
                 </t>
                 <t>
-                    The current URI for this vocabulary, known as the Format vocabulary, is:
-                    &lt;https://json-schema.org/draft/2020-11/vocab/format&gt;.
+                    In addition to the Format-Annotation vocabulary, a secondary vocabulary is available
+                    for custom meta-schemas that defines "format" as an assertion. The URI for the
+                    Format-Assertion vocabulary, is:
```
karenetheridge, line 551: remove comma

jsonschema-validation.xml
```diff
@@ -536,84 +530,65 @@
...
+                    &lt;https://json-schema.org/draft/2020-11/vocab/format-assertion&gt;. The current
+                    URI for the corresponding meta-schema is:
+                    <eref target="https://json-schema.org/draft/2020-11/meta/format-assertion"/>.
+                    Implementing support for the Format-Assertion vocabulary is OPTIONAL.
```
karenetheridge, line 555: suggested change:
+                     Implementations MAY choose to support this vocabulary.

jsonschema-validation.xml
```diff
@@ -536,84 +530,65 @@
...
                 </t>
                 <t>
-                    The current URI for the corresponding meta-schema is:
-                    <eref target="https://json-schema.org/draft/2020-11/meta/format"/>.
+                    Specifying both the Format-Annotation and the Format-Assertion vocabularies is functionally
+                    equivalent to specifying only the Format-Assertion vocabulary since its requirements
+                    are a superset of the Format-Annotation vocabulary.
```
karenetheridge, line 560: This isn't really true -- if format validation fails, the result doesn't contain an annotation.  If one wants an annotation in *all* cases, then the user can specify both vocabularies.

jsonschema-validation.xml
```diff
@@ -536,84 +530,65 @@
...
+                        supports annotation collection. This enables application-level validation when
                         schema validation is unavailable or inadequate.
                     </t>
                     <t>
-                        This requirement is not affected by the boolean value of the vocabulary
-                        declaration, nor by the configuration of "format"'s assertion
-                        behavior described in the next section.
+                        Implementations MAY still treat "format" as an assertion in addition to an
+                        annotation and attempt to validate the value's conformance to the specified
+                        semantics. The implementation MUST provide options to enable and disable such
```
karenetheridge, line 578: Why MUST they, if it can also be done via $vocabulary? I would soften this to a SHOULD. Ideally, the fewer knobs needed in an implementation outside of specifying a `$schema`, the better.

jsonschema-validation.xml
```diff
@@ -634,14 +606,24 @@
                             validation in the application, which is the recommended best practice.
                         </cref>
                     </t>
+                </section>
 
+                <section title="Format-Assertion Vocabulary">
+                    <t>
+                        When the Format-Assertion vocabulary is declared with a value of true,
+                        implementations MUST provide full validation support for all of the formats
+                        defined by this specificaion. Implementations that cannot provide full
```
karenetheridge, line 615: typo "specificaion"

karenetheridge:
whoops, looks like I hit submit a bit too late...

jsonschema-validation.xml
gregsdennis: This is solely for when the annotation vocabulary is used.  _If_ an implementation supports assertion (even partial assertion) they MAY choose to treat the keyword as such.  If they do, they MUST allow this feature to be enabled and disabled.

The assertion vocabulary doesn't have this allowance, instead specifying that implementations MUST implement assertion fully in order to support the vocabulary.

jsonschema-validation.xml
gregsdennis: > if format validation fails, the result doesn't contain an annotation

If the format validation fails, then its parent schema will fail, and no annotations will be returned anyway.

jsonschema-validation.xml
gregsdennis: OPTIONAL _is_ proper language. [RFC 2119](https://tools.ietf.org/html/rfc2119) section 5.

jsonschema-validation.xml
gregsdennis: REQUIRED _is_ proper language. [RFC 2119](https://tools.ietf.org/html/rfc2119) section 1.

jsonschema-validation.xml
gregsdennis: This was requested by @Relequestual in Slack.

jsonschema-validation.xml
karenetheridge: ok fair.

jsonschema-validation.xml
karenetheridge: I'm still confused. It sounds like a schema (via the metaschema) can specify to use the format-annotation vocabulary, but it might still get assertion behaviour anyway (with implementations that supported validations)? That would force the setting of an out-of-band configuration "no, really, when I say annotations I mean only annotations" and there would be no other way to not get validation behaviour as well.  If I'm correct so far, then the configuration value MUST default to false so validation behaviour doesn't happen accidentally.

gregsdennis: The schema can say, "You're getting annotations."

But this allows the client/application to say, "I know I'm supposed to get annotations, but this library I'm using supports asserting formats, and that'd save me some work, so I'd like to use that."  It gives the choice to the client.

It also the opens the door for implementations to provide partial validation, which is a pain point that `@handrews` had to deal with (pressure from implementers) and what the previous wording of this section tried so hard to allow.

However, if the schema says, "You're getting assertions," then the implementation MUST be able to provide validation or refuse to process the schema.

gregsdennis: The last scenario is `<format-assertion>: false`.  This means, "If you support asserting these things, that's great, please do.  If not, it's cool; they'll be annotations (because you don't understand them)."

gregsdennis: In the end, we have

| Annotation<br>Vocab | Assertion<br>Vocab | Impl Assert<br>Support | App Config | Result |
|--|--|--|--|--|
| false/true | n/a | no | n/a | annotation |
| false/true | n/a | yes | annotation | annotation |
| false/true | n/a | yes | assertion | assertion |
| n/a | false | no | n/a | annotation |
| n/a | false | yes | annotation/assertion | assertion |
| n/a | true | no | n/a | fail |
| n/a | true | yes | annotation/assertion | assertion |

karenetheridge: > But this bit of the spec allows the client/application to say, "I know I'm supposed to get annotations, but this library I'm using supports asserting formats, and that'd save me some work, so I'd like to use that." It gives the choice to the client.

My main point here is that if the implementation can do this, the setting MUST default to NOT doing it. Users must explicitly opt in. Otherwise, there is *no* way of a schema being processed as annotation only (with any combination of $schema and $vocabulary settings) with an implementation that can also handle assertions.

gregsdennis: > the setting MUST default to NOT doing it

Absolutely.  That is covered on the next line: "... evaluation and MUST be disabled by default."

Relequestual:
Ah nuts sorry @karenetheridge - My internet was super flakey yesterday (and few days before that, and still is).
There are a few obvious fixes needed here. Feel free to open a new PR, or I'll try to cover all your comments in final edits on Friday. =] Many thanks

PR #1028, opened 2020-11-23
it is no longer valid to treat $comment as an unknown keyword

karenetheridge:
..because now, we collect unknown keywords as annotations, and this behaviour
is explicitly prohibited for $comment.


Relequestual:
Thanks!

PR #1029, opened 2020-11-24
Fix "dynamic" keywords in core vocabulary meta-schema

jdesrosiers:
I appears this was missed when updating the meta-schemas to use `$dynamicRef` and `$dynamicAnchor`.

### DIFF
--- a/meta/core.json
+++ b/meta/core.json
@@ -27,13 +27,13 @@
             "type": "string",
             "format": "uri-reference"
         },
-        "$recursiveRef": {
+        "$dynamicRef": {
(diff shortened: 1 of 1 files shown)


Relequestual (approved):
Yikes. Thanks.
I don't think we've been keeping the meta schema up to date and planned to release them AFTER the spec docs themselves.

Relequestual:
Ah that would be the general-purpose meta-schema I see. Much appreciated!
It will need a full going over to check it matches up with the vocabulary schemas post draft docs release.

PR #1038, opened 2020-11-28
improve some wording in draft 2020-xx rc0

karenetheridge:


Relequestual:
Thank you so much.
Meta Schema issue... oh dear. Good catch!

PR #1041, opened 2020-11-30
Fixes for dynamic scope and keywords

jdesrosiers:
This PR fixes errors in the spec for the dynamic and keywords and dynamic scope. The spec describes dynamic scope to only be resolvable when processing an instance. This wasn't true for the recursive keywords, and it isn't true for the dynamic keywords either. You need to know all the schemas and how they reference each other, but you don't need an instance. The vast majority of these changes cover that issue. The rest fixes minor typos and such.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -654,25 +654,17 @@
                     appear in a schema resource's root schema.
                 </t>
                 <t>
-                    Other keywords may take into account the dynamic scope that
-                    exists during the evaluation of a schema, typically together
-                    with an instance document.  The outermost dynamic scope is the
-                    root schema of the schema document in which processing begins.
-                    The path from this root schema to any particular keyword (that
-                    includes any "$ref" and "$dynamicRef" keywords that may have
-                    been resolved) is considered the keyword's "validation path."
-                    <cref>
-                        Or should this be the schema object at which processing
-                        begins, even if it is not a root?  This has some implications
-                        for the case where "$dynamicAnchor" is only allowed in the
-                        root schema but processing begins in a subschema.
-                    </cref>
+                    Other keywords may take into account the dynamic scope that exists during the
+                    evaluation of a schema.  The outermost dynamic scope is the root schema of the
+                    schema document in which processing begins.  The path from this root schema to
+                    any particular keyword (that includes any "$ref" and "$dynamicRef" keywords that
+                    may have been resolved) is considered the keyword's "validation path."
                 </t>
                 <t>
                     Lexical and dynamic scopes align until a reference keyword
                     is encountered.  While following the reference keyword moves processing
                     from one lexical scope into a different one, from the perspective
-                    of dynamic scope, following reference is no different from descending
+                    of dynamic scope, following a reference is no different from descending
                     into a subschema present as a value.  A keyword on the far side of
                     that reference that resolves information through the dynamic scope
                     will consider the originating side of the reference to be their
@@ -741,7 +733,7 @@
                 <t>
                     Canonical schema URIs MUST NOT change while processing an instance, but
                     keywords that affect URI-reference resolution MAY have behavior that
-                    is only fully determined at runtime.
+                    is only fully determined dynamically.
                 </t>
                 <t>
                     While custom identifier keywords are possible, vocabulary designers should
@@ -798,12 +790,10 @@
                         of an instance against a schema.
                     </t>
                     <t>
-                        For some by-reference applicators, such as
-                        <xref target="ref">"$ref"</xref>, the referenced schema can be determined
-                        by static analysis of the schema document's lexical scope.  Others,
-                        such as "$dynamicRef" (with "$dynamicAnchor"),  may make use of dynamic
-                        scoping, and therefore only be resolvable in the process of evaluating
-                        the schema with an instance.
+                        For some by-reference applicators, such as <xref target="ref">"$ref"</xref>, the
+                        referenced schema can be determined by static analysis of the schema document's
+                        lexical scope.  Others, such as "$dynamicRef" (with "$dynamicAnchor"), are only
+                        resolvable with knowledge of all the schemas in it's dynamic scope.
                     </t>
                 </section>
             </section>
@@ -1397,7 +1387,6 @@
                         The "$anchor" and "$dynamicAnchor" keywords are used to specify such
                         fragments.  They are identifier keywords that can only be used to create
                         plain name fragments, rather than absolute URIs as seen with "$id".
-                        The behavior of the created fragment is identical for both keywords.
                     </t>
                     <t>
                         The base URI to which the resulting fragment is appended is the canonical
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -654,25 +654,17 @@
...
                 </t>
                 <t>
-                    Other keywords may take into account the dynamic scope that
-                    exists during the evaluation of a schema, typically together
-                    with an instance document.  The outermost dynamic scope is the
-                    root schema of the schema document in which processing begins.
-                    The path from this root schema to any particular keyword (that
-                    includes any "$ref" and "$dynamicRef" keywords that may have
-                    been resolved) is considered the keyword's "validation path."
-                    <cref>
```
jdesrosiers, line 664: I removed the cref here because I'm mostly sure it's just left over from the recursive keywords and doesn't apply any more. At the least, it's not correct that, $dyanmicAnchor is only allowed in the root schema. If someone thinks there's still some value in this cref, let me know and we'll try to correct it instead of delete it.

jsonschema-core.xml
```diff
@@ -1471,51 +1460,41 @@
...
                             resource in the <xref target="scopes">dynamic scope</xref> that defines
                             an identically named fragment with "$dynamicAnchor".
-                            <cref>
-                                Requiring both the initial and final URI fragment to be defined
-                                by "$dynamicAnchor" ensures that the more common "$anchor"
-                                never unexpectedly changes the dynamic resolution process
-                                due to a naming conflict across resources.  Users of
-                                "$dynamicAnchor" are expected to be aware of the possibility
-                                of such name collisions, while users of "$anchor" are not.
-                            </cref>
```
jdesrosiers, line 1514: I removed this cref because it appears to be wrong. There's no way that an `$anchor` can change dynamic scope resolution and even if it could, it's not clear how bookending could possibly solve that problem. If anyone knows what this means, please share and I'll put it back in.

Relequestual:
Many thanks!

karenetheridge:
> The spec describes dynamic scope to only be resolvable when processing an instance. This wasn't true for the recursive keywords, and it isn't true for the dynamic keywords either.

How is it not true?  The dynamic path is affected by the instance data because it can change the path through applicator keywords, which can change which references are followed or not followed. Example:

```
{
  "$defs": {
    "yes": { ... something complicated containing $refs and $dynamicRefs... },
    "no": { ..something complicated containing other $refs and $dynamicRefs ... },
  },
  "if": { "const": "yes" },
  "then": { "$ref": "#/$defs/yes" },
  "else": { "$ref": "#/$defs/no" },
}
```

We cannot know the dynamic scope in a particular subschema without knowing how we got there, and how we got there depends on the evaluation result of the if/then/else applicator keywords.

jdesrosiers:
@karenetheridge If I'm wrong about this, it should be relatively easy to prove because it would mean my implementation has a bug. I resolve all URLs during the compile step before it has an instance. All we need is one example we can run on https://json-schema.hyperjump.io and get instant feedback. We can then use that example in the test suite as well.

Relequestual:
What $schema should it have?

jdesrosiers:
Sorry, I forgot to mention, use `"$schema": "https://json-schema.org/draft/future/schema"` on https://json-schema.hyperjump.io to get support for the dynamic keywords.

karenetheridge:
I implemented this first with recursiveRef, so I could test it locally:

schema:
```
{
    "$id": "main.json",
    "$defs": {
        "inner": {
            "$id": "inner.json",
            "$recursiveAnchor": true,
            "title": "inner",
            "additionalProperties": {
                "$recursiveRef": "#"
            }
        }
    },
    "if": { "propertyNames": { "pattern": "^[a-m]" } },
    "then": {
        "title": "any type of node",
        "$id": "anyLeafNode.json",
        "$recursiveAnchor": true,
        "$ref": "main.json#/$defs/inner"
    },
    "else": {
        "title": "integer node",
        "$id": "integerNode.json",
        "$recursiveAnchor": true,
        "type": [ "object", "integer" ],
        "$ref": "main.json#/$defs/inner"
    }
}
```

passing data: `{ "alpha": 1.1 }`

produces result:

```
{
  "annotations" : [
    {
      "absoluteKeywordLocation" : "inner.json#/title",
      "annotation" : "inner",
      "instanceLocation" : "/alpha",
      "keywordLocation" : "/then/$ref/additionalProperties/$recursiveRef/$ref/title"
    },
    {
      "absoluteKeywordLocation" : "anyLeafNode.json#/title",
      "annotation" : "any type of node",
      "instanceLocation" : "/alpha",
      "keywordLocation" : "/then/$ref/additionalProperties/$recursiveRef/title"
    },
    {
      "absoluteKeywordLocation" : "inner.json#/additionalProperties",
      "annotation" : [
        "alpha"
      ],
      "instanceLocation" : "",
      "keywordLocation" : "/then/$ref/additionalProperties"
    },
    {
      "absoluteKeywordLocation" : "inner.json#/title",
      "annotation" : "inner",
      "instanceLocation" : "",
      "keywordLocation" : "/then/$ref/title"
    },
    {
      "absoluteKeywordLocation" : "anyLeafNode.json#/title",
      "annotation" : "any type of node",
      "instanceLocation" : "",
      "keywordLocation" : "/then/title"
    }
  ],
  "valid" : true
}
```

failing data: `{ "november": 1.1 }`

produces results:
```
{
  "errors" : [
    {
      "absoluteKeywordLocation" : "integerNode.json#/type",
      "error" : "wrong type (expected one of object, integer)",
      "instanceLocation" : "/november",
      "keywordLocation" : "/else/$ref/additionalProperties/$recursiveRef/type"
    },
    {
      "absoluteKeywordLocation" : "inner.json#/additionalProperties",
      "error" : "not all additional properties are valid",
      "instanceLocation" : "",
      "keywordLocation" : "/else/$ref/additionalProperties"
    },
    {
      "absoluteKeywordLocation" : "main.json#/else",
      "error" : "subschema is not valid",
      "instanceLocation" : "",
      "keywordLocation" : "/else"
    }
  ],
  "valid" : false
}
```

I believe the equivalent schema, in draft2020-12, would be:

```
{
    "$id": "main.json",
    "$defs": {
        "inner": {
            "$id": "inner.json",
            "$dynamicAnchor": "node",
            "title": "inner",
            "additionalProperties": {
                "$dynamicRef": "#"
            }
        }
    },
    "if": { "propertyNames": { "pattern": "^[a-m]" } },
    "then": {
        "title": "any type of node",
        "$id": "anyLeafNode.json",
        "$dynamicAnchor": "node",
        "$ref": "main.json#/$defs/inner"
    },
    "else": {
        "title": "integer node",
        "$id": "integerNode.json",
        "$dynamicAnchor": "node",
        "type": [ "object", "integer" ],
        "$ref": "main.json#/$defs/inner"
    }
}
```

Before runtime, we cannot know the destination of the $dynamicRef because we do not know the path by which we got here, and anywhere in the higher scope could have declared a $dynamicAnchor -- and indeed both of the possible source paths (`/then` or `/else`) have one.

jdesrosiers:
Well, this is embarrassing, but it seems you found a bug not related to this discussion. https://json-schema.hyperjump.io doesn't give the right result for this schema either and there are no branches.

```json
{
    "$id": "main.json",
    "$defs": {
        "inner": {
            "$id": "inner.json",
            "$dynamicAnchor": "node",
            "title": "inner",
            "additionalProperties": {
                "$dynamicRef": "#node"
            }
        }
    },
    "allOf": [{
        "title": "integer node",
        "$id": "integerNode.json",
        "$dynamicAnchor": "node",
        "type": [ "object", "integer" ],
        "$ref": "main.json#/$defs/inner"
    }]
}
```

I think the problem is with the mutually recursive references. I'm working to understand the problem better so I can either fix it quickly or find a workaround so we can see if the schema exposes a bug related to this discussion.

jdesrosiers:
@karenetheridge, I found the bug and I've confirmed that your schema does trip up my validator. Let's definitely add it to the test suite.

Relequestual:
So does this mean @karenetheridge is correct and the contents of this PR is wrong in relation to...

> The spec describes dynamic scope to only be resolvable when processing an instance. This wasn't true for the recursive keywords, and it isn't true for the dynamic keywords either.

?

karenetheridge:
ok, test case added! https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/450

jdesrosiers:
Yes, @Relequestual, that means I was wrong. I'll submit up a PR reverting the relevant changes.

Relequestual:
Thanks! =]

PR #1046, opened 2020-12-01
rc1 revisions

karenetheridge:


jsonschema-core.xml
```diff
@@ -3639,11 +3653,11 @@ https://example.com/schemas/common#/$defs/count/minimum
                 </t>
                 <t>
                     The main example meta-schema also restricts the usage of the Applicator
-                    vocabulary by forbidding the keywords prefixed with "unevaluated", which
+                    vocabulary by forbidding the keywords prefixed with "additional", which
```
Relequestual, line 3645: I'd rather not change this example.
It is still valid and is different from understanding or not understanding the unevaluated vocabulary.

Relequestual (changes requested):
One change suggestion, otherwise these changes are clearly the result of deep reading and understanding. Much appreciated!

jsonschema-core.xml
karenetheridge: Ok. I changed it because unevaluated* is now in its own vocabulary, but it's still reasonable to use that namespace in an example. Reverting (and fixing one line because of the vocabulary separation).

karenetheridge:
There was a conflict with a particular paragraph re dynamic scope that changed in a recent commit, so this is rebased to the latest master with conflicts resolved.

Relequestual (approved):
Many thanks, yet again!

PR #1050, opened 2020-12-01
[1043] Clarify last paragraph in "contains" section

### LINKED ISSUE #1043: Missing apostrophe in “contains”; strange sentence
https://github.com/json-schema-org/json-schema-spec/blob/803ce7dd3e08e02596477814f49d6908602e77ad/jsonschema-core.xml#L2378

Suggested sentence changes (first pass):
If the instance array to which this keyword’s  subschema is applicable is empty, the annotation value MUST NOT be missing.

To clarify: is this supposed to be saying that a “contains” annotation MUST be there if the instance array is empty?

Suggested changes, second pass:
The annotation value MUST be present if the instance array to which this keyword’s subschema applies is empty.

ssilverman:
Also fixes the "validates" tense to match the first use.

Fixes #1043

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2352,10 +2352,10 @@
                         <t>
                             This keyword produces an annotation value which is an array of
                             the indexes to which this keyword validates successfully when applying
-                            its subschema, in ascending order. The value MAY be a boolean "true" if the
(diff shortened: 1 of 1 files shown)


Relequestual:
🎉 Thanks

PR #1053, opened 2020-12-03
Revert changes to dynamic scope regarding dependence on instance

jdesrosiers:
Since this is mostly reverting, I added https://github.com/jdesrosiers/json-schema-spec/commit/321d3841631adb2514277f03f5d112af98b681cf to show what the actual diff is from the original spec. It's mostly typo fixes and removing a couple crefs.

The original PR was https://github.com/json-schema-org/json-schema-spec/pull/1041

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -654,11 +654,13 @@
                     appear in a schema resource's root schema.
                 </t>
                 <t>
-                    Other keywords may take into account the dynamic scope that exists during the
-                    evaluation of a schema.  The outermost dynamic scope is the root schema of the
-                    schema document in which processing begins.  The path from this root schema to
-                    any particular keyword (that includes any "$ref" and "$dynamicRef" keywords that
-                    may have been resolved) is considered the keyword's "validation path."
+                    Other keywords may take into account the dynamic scope that
+                    exists during the evaluation of a schema, typically together
+                    with an instance document.  The outermost dynamic scope is the
+                    root schema of the schema document in which processing begins.
+                    The path from this root schema to any particular keyword (that
+                    includes any "$ref" and "$dynamicRef" keywords that may have
+                    been resolved) is considered the keyword's "validation path."
                 </t>
                 <t>
                     Lexical and dynamic scopes align until a reference keyword
@@ -733,7 +735,7 @@
                 <t>
                     Canonical schema URIs MUST NOT change while processing an instance, but
                     keywords that affect URI-reference resolution MAY have behavior that
-                    is only fully determined dynamically.
+                    is only fully determined at runtime.
                 </t>
                 <t>
                     While custom identifier keywords are possible, vocabulary designers should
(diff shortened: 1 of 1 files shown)


jdesrosiers:
Like I said in #1041, if either of those crefs makes sense to anyone, we can fix/clarify them and put them back in.

Relequestual (approved):

karenetheridge (approved):
I did `git diff 3cf7cb8 upstream/pull/1053`\* to compare the changes (there weren't too many other commits that went in in the meantime so they're easy to mentally filter out).

I'd keep the second CREF, regarding the initial and final URI fragment, but I don't feel super strongly about it (and haven't looked too closely yet at the proposed changes that remove the requirement for two anchors).  The first CREF doesn't seem applicable anymore.

\* with this entry in my .git/config:
```
[remote "upstream"]
    url = git@github.com:json-schema-org/json-schema-spec.git
    fetch = +refs/heads/*:refs/remotes/upstream/*
    fetch = +refs/pull/*/head:refs/remotes/upstream/pull/*
```

Relequestual:
Good spot @karenetheridge 
@jdesrosiers can you add that back in please? Then it looks like we're good to merge! =]

jdesrosiers:
I think there's some miscommunication here. I don't think including that cref as it is is an option. Right now it doesn't make sense. We can leave it in, but if we do we need to fix/clarify it. I've called this out multiple times now and no one has come forward with an explanation or even indicated that they know what it means. If no one understands how this cref makes sense, I think the only option is to leave it out. If we publish that and someone comes around asking what that cref means, we'd have to tell them, ¯\_(ツ)_/¯ nobody knows ¯\_(ツ)_/¯.

jdesrosiers:
Here's the cref that was removed and my comment from the previous PR.

> Requiring both the initial and final URI fragment to be defined
                                by "$dynamicAnchor" ensures that the more common "$anchor"
                                never unexpectedly changes the dynamic resolution process
                                due to a naming conflict across resources.  Users of
                                "$dynamicAnchor" are expected to be aware of the possibility
                                of such name collisions, while users of "$anchor" are not.

I removed this cref because it appears to be wrong. There's no way that an $anchor can change dynamic scope resolution and even if it could, it's not clear how bookending could possibly solve that problem. If anyone knows what this means, please share and I'll put it back in.

karenetheridge:
@handrews added that CREF in commit 940586a1cb790dd56faaea1521f632d3d124685a; perhaps he can comment?

But we can certainly sort this out next year when we start looking at revisions for the next draft.

Relequestual:
Mmm I'm of two minds on this, but it as we can't confidently justify it, let's leave it removed, and work out the original intent later.

Relequestual:
@karenetheridge Please do file an issue to review this referencing this PR =]

jdesrosiers:
I'll file the issue, I was going to suggest that approach as well.

PR #1055, opened 2020-12-08
Fix incorrect stipulation on keyword ordering for unevaluated* keywords.

### LINKED ISSUE #1047: concerns with order of evaluation with unevaluated* keywords
The spec (https://json-schema.org/draft/preview/2020-12-rc-1/jsonschema-core.html#rfc.section.11.1) says, regarding unevaluatedItems:

> This means that "prefixItems", "items", "contains", and all in-place applicators MUST be evaluated before this keyword can be evaluated. Authors of extension keywords MUST NOT define an in-place applicator that would need to be evaluated before this keyword. 

If I wanted to create a new applicator keyword "twoOf", and it is evaluated after unevaluatedItems, then any annotations collected by its subschemas will not be visible to unevaluatedItems. So therefore, shouldn't this wording be "..would need to be evaluated **after** this keyword"?

Relequestual:
Fixes #1047

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2549,7 +2549,7 @@
                     This means that "prefixItems", "items", "contains", and all in-place
                     applicators MUST be evaluated before this keyword can be evaluated.
                     Authors of extension keywords MUST NOT define an in-place applicator
-                    that would need to be evaluated before this keyword.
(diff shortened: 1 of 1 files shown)


ssilverman (approved):
👍

karenetheridge (approved):

PR #1057, opened 2020-12-09
Meta-schema 2020-12 update

gregsdennis:

### DIFF
--- a/meta/applicator.json
+++ b/meta/applicator.json
@@ -12,7 +12,10 @@
         "prefixItems": { "$ref": "#/$defs/schemaArray" },
         "items": { "$dynamicRef": "#meta" },
         "contains": { "$dynamicRef": "#meta" },
-        "additionalProperties": { "$dynamicRef": "#meta" },
+        "additionalProperties": {
+            "$dynamicRef": "#meta",
+            "default": {}
+        },
         "properties": {
             "type": "object",
             "additionalProperties": { "$dynamicRef": "#meta" },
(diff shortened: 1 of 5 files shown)


meta/applicator.json
```diff
@@ -12,7 +12,10 @@
         "prefixItems": { "$ref": "#/$defs/schemaArray" },
         "items": { "$dynamicRef": "#meta" },
         "contains": { "$dynamicRef": "#meta" },
-        "additionalProperties": { "$dynamicRef": "#meta" },
+        "additionalProperties": {
+            "$dynamicRef": "#meta",
+            "default": {}
```
karenetheridge, line 17: Instead of adding a default in every keyword that uses a subschema, I would just add it to the top level schema definition right at `schema.json#meta`. (And `true` might be a more intuitive default than `{}`, at any rate.)

gregsdennis: Right, but the new spec (and probably the old because this likely didn't change) says:

> Omitting this keyword has the same assertion behavior as an empty schema.

Thus the default needs to be "an empty schema" which is _technically_ different from (although equivalent to) `true`.

karenetheridge:
Hah, okay. I will pursue changing that language to "..has the same assertion behaviour as boolean true" for the next spec revision :)

Relequestual (approved):
All these changes look good to me!
@karenetheridge OK to resolve the discussion you started here and give a review approval/rejection? =]

karenetheridge (approved):

karenetheridge:
@Relequestual I can't resolve conversations as I'm not a commiter for this repository.

PR #1066, opened 2021-01-17
added unevaluated meta-schema to allOf list in primary meta-schema

gregsdennis:
Missed this at first, but found it when updating my implementation.


Relequestual (approved):

karenetheridge:
nice, good catch!

Issue #59, opened 2016-09-18
IRI support (i.e. Unicode in URIs)


awwright:
Most modern Web/hypermedia formats support IRIs instead of just URIs that are 7bit ASCII. IRIs are a superset of URIs that support full Unicode. For standards that only support URIs, IRIs have to be converted/escaped into a URI-compatible format.

handrews:
+1 for IRI support.  It's 2016, let's be inclusive :-)

handrews:
@awwright what would need to be done for this?  Is it sufficient to just go through and change URI to IRI? (and IRI reference, etc. etc.).  I have not paid much attention to what sort of things come up with IRIs rather than URIs.

handrews:
Note that while there is no separate IRI Template specification, per RFC 6570:

>   Although the URI syntax is used for the result, the template string
   is allowed to contain the broader set of characters that can be found
   in Internationalized Resource Identifier (IRI) references [RFC3987].
   Therefore, a URI Template is also an IRI template, and the result of
   template processing can be transformed to an IRI by following the
   process defined in Section 3.2 of [RFC3987].

handrews:
@awwright @Relequestual @philsturgeon @gregsdennis any thoughts on this?

Is it just as simple as saying `$id` and `$ref` take IRI-references instead of URI-references?  Is that desirable?  It doesn't ever seem to come up, but IRIs will only get more important, I suspect.

gregsdennis:
Looks like [IRIs are already supported in .Net](https://docs.microsoft.com/en-us/dotnet/framework/network-programming/international-resource-identifier-support-in-system-uri) which means I'm already supporting it :tada:

I have no problems with this.

Relequestual:
Seems like the right move to me.
Some extra considerations from RFD: https://www.w3.org/TR/rdf11-concepts/#h3_section-IRIs

Should we note similar considerations or is it unnecessary?

jakubklimek:
I already use IRIs in `$ref` and `$id` as this is necessary for Czech schemas. The ids and refs would look terrible otherwise:
`https://ofn.gov.cz/turistické-cíle/2020-07-01/schémata/turistický-cíl.json`
vs.
`https://ofn.gov.cz/turistick%C3%A9-c%C3%ADle/2020-07-01/sch%C3%A9mata/turistick%C3%BD-c%C3%ADl.json`

[...]

PR #1068, opened 2021-01-25
Support for IRI references (#59)

jakubklimek:
This would add support for IRI references in `$id`, `$ref` and `$dynamicRef` as per https://github.com/json-schema-org/json-schema-spec/issues/59

### DIFF
--- a/meta/core.json
+++ b/meta/core.json
@@ -10,14 +10,14 @@
     "type": ["object", "boolean"],
     "properties": {
         "$id": {
-            "$ref": "#/$defs/uriReferenceString",
+            "$ref": "#/$defs/iriReferenceString",
             "$comment": "Non-empty fragments not allowed.",
             "pattern": "^[^#]*#?$"
         },
         "$schema": { "$ref": "#/$defs/uriString" },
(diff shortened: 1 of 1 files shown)


Relequestual:
Please DO NOT MERGE this yet. 2020-12 meta-schema needs to be published.

Relequestual:
Additionally, the specification documents would need updating, either first (in a separate PR), or at the same time.

karenetheridge:
`$vocabulary` and `$schema` use uri - should this be changed to iri?

jakubklimek:
@karenetheridge yes, we want the references to be IRIs as well so that they are also human readable

handrews:
@jakubklimek I'm going to convert this to a draft as there seems to still be discussion in the issue.

jakubklimek:
@Relequestual @handrews I have pushed commits changing URIs to IRIs in the spec as well where necessary. I left references to RFC 3986 (URIs) wherever resolving base URIs is mentioned, as this is also applicable to IRIs and RFC 3987 actually refers to RFC 3986 in this manner.

I would therefore like to resume discussions about merging this.

Issue #1080, opened 2021-03-16
Verbiage for `type` could be improved.


gregsdennis:
[Core Section 6.1.1](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.1.1)

> An instance validates if and only if the instance is in any of the sets listed for this keyword.

Here a "set" is intended to be a collection, indicating that the value should be contained within that collection.  However "in any of the sets" seems to imply there are multiple collections.

I think that the intent here was to say that the instance is valid if it is a _subset_ of the set of allowable values (which are defined in the preceding sentence), but it still reads weird.

I suggest

> An instance validates successfully if one of the following is true
> - it is a single value from the set above
> - it is an array of distinct values from the array above

This is a bit more verbose but significantly clearer.

karenetheridge:
What does "an instance ... is an array of distinct values from the array above" mean? That makes it sound like this data: `[ 1, true, null ]` would successfully validate against the schema `{"type": [ "integer", "boolean", "null" ]}`.

I suggest:

> An instance validates successfully if one of the following is true:
> * if the value of "type" is a string, and the instance is of the singular type indicated by the "type" keyword, or
> * if the value of "type" is an array, the instance is of one of the types listed by that array

gregsdennis:
oh... yeah 🤦‍♂️

jdesrosiers:
:+1:  Let's fix this!

jdesrosiers:
I'll write this up within the next day or two if there are no objections.

PR #1083, opened 2021-03-25
Improve wording for the type keyword in the spec

jdesrosiers:
Fixes https://github.com/json-schema-org/json-schema-spec/issues/1080

I adjusted the wording suggested in the issue, so let me know if anyone thinks it doesn't read clearly enough.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -225,8 +225,11 @@
                         or "integer" which matches any number with a zero fractional part.
                     </t>
                     <t>
-                        An instance validates if and only if the instance is in any of the sets listed
-                        for this keyword.
+                        If the value of "type" is a string, then an instance validates successfully if
+                        its type matches the type represented by the value of the string.
+
+                        If the value of "type" is an array, then an instance validates successfully if
+                        its type matches any of the types indicated by the strings in the array.
                     </t>
                 </section>
 


jsonschema-validation.xml
```diff
@@ -225,8 +225,11 @@
                         or "integer" which matches any number with a zero fractional part.
                     </t>
                     <t>
-                        An instance validates if and only if the instance is in any of the sets listed
-                        for this keyword.
+                        If the value of "type" is a string, then an instance validates successfully if
+                        it's type matches the type represented by the value of the string.
+
+                        If the value of "type" is an array, then an instance validates successfully if
+                        it's type matches any of the types indicated by the strings in the array.
```
karenetheridge, line 232:
```suggestion
                        its type matches the type represented by the value of the string.

                        If the value of "type" is an array, then an instance validates successfully if
                        its type matches any of the types indicated by the strings in the array.
```

karenetheridge (changes requested):

jsonschema-validation.xml
karenetheridge: ..otherwise, LGTM.

karenetheridge (changes requested):

jsonschema-validation.xml
jdesrosiers: Where's the embarrassment emoji :facepalm:

gregsdennis (approved):
This is clearer to me and correct.

awwright (approved):
Much improve, wow

Issue #937, opened 2020-05-23
Clarify whether plain name fragment URIs are canonical, and what it means if they are (or aren't)


handrews:
There's a little confusion and ambiguity on this topic.

* Absolute URIs (with no fragment, not even an empty one) are definitely canonical, and identify the complete schema resource.
    * This implies that the same URI with an empty JSON Pointer fragment is non-canonical, but it should work, and it would be bad if an implementation did not support it.
* URIs with non-empty JSON Pointer fragment URIs, relative to the nearest base URI, are considered canonical.
* URIs with non-empty JSON Pointer fragment URIs, relative to a _more distant_ base URI, are considered non-canonical _and are not guaranteed to work_.
    * This was the main point of the canonical URI stuff in 2019-09.
* URIs with  plain name fragments aren't discussed as canonical or non-canonical
    * However, they only work with the nearest base URI so there's never any ambiguity
    * Therefore, plain name fragment URIs should always work.

Things to figure out:

* Is "canonical" the right term here?  IIRC @awwright thought it was not, but I had a vaguely coherent reason as to why it was (which I should look up) and I was burnt out and needed to publish it so we did that.  Probably not the best process.
* Should we be framing the requirement around what URIs MUST be supported and what MAY be supported differently, in order to capture the possibilities better?
* Are there other use cases I'm missing here?

karenetheridge:
> Is "canonical" the right term here?

Depends on how the term is going to be used :)  I can give one example where a canonical URI that contains a plain-name fragment is *not* the most useful uri to be slinging around: when generating errors or annotations.  I just ran into this today while implementing `$anchor`. Let me show an example that I'm using in my tests:

```
data: 1
schema:
{
  "$id": "http://localhost:1234",
  "$defs": {
    "def1": {
      "$id": "def1.json",
      "$refL": "#/$defs/myint",
      "type": "integer",
      "maximum": -1,
      "minimum": 5,
    },
    "myint": "{
      "$id": "def2.json",
      "multipleOf": 5,
      "exclusiveMaximum": 1,
    },
  },
  "$ref": "#/$defs/def1",
}

result:
  {
    "valid": false
    errors": [
      {
        "instanceLocation": "#",
        "keywordLocation": "#/$ref/$ref/multipleOf",
        "absoluteKeywordLocation": "http://localhost:1234/def2.json#/multipleOf",
        "error": "value is not a multiple of 5",
      },
      {
        "instanceLocation": "#",
[...]

PR #1085, opened 2021-03-30
further specify the format of uri-references throughout

karenetheridge:
- $id cannot have a fragment at all, so separate it from the common definiiion
- $ref, $dynamicRef, $recursiveRef must be a uri-reference to a schema
location: when the fragment is non-empty, it must refer to an $anchor or a
json-pointer
- absoluteKeywordLocation in result outputs must use canonical
URIs/uri-references: either there is no fragment, or it is non-empty and
encodes a json pointer

### DIFF
--- a/meta/core.json
+++ b/meta/core.json
@@ -10,14 +10,15 @@
     "type": ["object", "boolean"],
     "properties": {
         "$id": {
-            "$ref": "#/$defs/uriReferenceString",
+            "type": "string",
+            "format": "uri-reference",
             "$comment": "Non-empty fragments not allowed.",
             "pattern": "^[^#]*#?$"
         },
         "$schema": { "$ref": "#/$defs/uriString" },
-        "$ref": { "$ref": "#/$defs/uriReferenceString" },
+        "$ref": { "$ref": "#/$defs/uriReferenceToSchemaString" },
         "$anchor": { "$ref": "#/$defs/anchorString" },
-        "$dynamicRef": { "$ref": "#/$defs/uriReferenceString" },
+        "$dynamicRef": { "$ref": "#/$defs/uriReferenceToSchemaString" },
         "$dynamicAnchor": { "$ref": "#/$defs/anchorString" },
         "$vocabulary": {
             "type": "object",
@@ -43,9 +44,11 @@
             "type": "string",
             "format": "uri"
         },
-        "uriReferenceString": {
+        "uriReferenceToSchemaString": {
             "type": "string",
-            "format": "uri-reference"
+            "format": "uri-reference",
+            "$comment": "any fragment must be empty, or match anchor or json-pointer syntax",
+            "pattern": "^[^#]*(|#([A-Za-z_][-A-Za-z0-9.:_]*|/([^~]|~[01])*))$"
         }
     }
 }
--- a/output/schema.json
+++ b/output/schema.json
@@ -19,7 +19,9 @@
         },
         "absoluteKeywordLocation": {
           "type": "string",
-          "format": "uri"
+          "format": "uri",
+          "$comment": "any fragment must be non-empty, and use json-pointer syntax",
+          "pattern": "^[^#]*(#/([^~]|~[01])*)$"
         },
         "instanceLocation": {
           "type": "string",
--- a/schema.json
+++ b/schema.json
@@ -51,7 +51,7 @@
         },
         "$recursiveRef": {
             "$comment": "\"$recursiveRef\" has been replaced by \"$dynamicRef\".",
-            "$ref": "meta/core#/$defs/uriReferenceString",
+            "$ref": "meta/core#/$defs/uriReferenceToSchemaString",
             "deprecated": true
         }
     }


schema.json
```diff
@@ -51,7 +51,7 @@
         },
         "$recursiveRef": {
             "$comment": "\"$recursiveRef\" has been replaced by \"$dynamicRef\".",
-            "$ref": "meta/core#/$defs/uriReferenceString",
+            "$ref": "meta/core#/$defs/uriReferenceToSchemaString",
```
jdesrosiers, line 54: The only defined value for `$recursiveRef` is `#`, so this could be `"const": "#"`. On the other hand, if there is no `$recursiveAnchor`, then it's supposed to behave like a `$ref` which allows any URI, so ¯\\\_(ツ)\_/¯.

jdesrosiers (commented):
I'm not sure this change is necessary, but I'm ok with it. I assume we aren't talking about a change to the already released meta-schema, but rather that this would be part of the next draft?

karenetheridge:
> I'm not sure this change is necessary, but I'm ok with it. I assume we aren't talking about a change to the already released meta-schema, but rather that this would be part of the next draft?

There's no reason why this can't go into the current draft's metaschema, because it is not adding an additional restriction that wasn't already there, but rather making the metaschema more conformant to existing language.

schema.json
karenetheridge: > The only defined value for $recursiveRef is #

Not really. The 201909 spec leaves the behaviour for other uri-references somewhat open -- it doesn't actually restrict (i.e. with a MUST) to the `#` value. I've gotten other $recursiveRef values to work just fine in my implementation, with no extra code.

jdesrosiers:
> There's no reason why this can't go into the current draft's metaschema

We generally don't consider meta-schemas living documents. Once they are released, we don't change them any more than we would change the spec. We have made exceptions for fixing bugs in meta-schemas, but this isn't fixing anything, it's just making it more strict. I don't think we have a rule written down anywhere about how to deal with this kind of thing (we probably should), but historical precedent seems to indicate that this should go in the next draft.

karenetheridge:
> We generally don't consider meta-schemas living documents.

We have in the past, however, fixed inconsistencies between the written spec and the metaschema, where the written spec prevails (because it very definitely is versioned and cannot be revised before the next release).

jdesrosiers:
> We have in the past, however, fixed inconsistencies between the written spec and the metaschema, where the written spec prevails 

Right, which is essentially bug fixes. This is not a bug fix. I don't feel strongly enough about this to try to block it, but I do feel that it's important for more voices to express opinions on whether we should be making this kind of change for meta-schemas before this gets approved. Even better would be if we decide on guidelines for when we want to allow updating meta-schemas and document that decision.

handrews:
@karenetheridge @jdesrosiers it would probably be a good idea to start a discussion in the community repo on the meta-schema process.  It can complement my spec release process discussion (https://github.com/json-schema-org/community/discussions/7) but **please start a new discussion!**  The meta-schema process should be separate.  I do not have an opinion on it (or rather, we tried my opinion, but no one liked it- which I say with humor, it's just time to try someone else's ideas).

In the meantime let's put this on hold while we sort out the process concerns.

handrews:
@karenetheridge I think this would need to be reworked now that JSON Schema uses IRIs rather than URIs, and I'm not quite sure how the regexes would work in that case.  Along with the branch changes, perhaps this is best closed and a new PR submitted?

karenetheridge:
I have rebased against the updates to `main`.

gregsdennis:
Good reminder that I need to update the output schema to the new structure. 👍

This should go through first, then I'll do that update.

meta/core.json
```diff
@@ -43,9 +44,11 @@
             "type": "string",
             "format": "iri"
         },
-        "iriReferenceString": {
+        "iriReferenceToSchemaString": {
             "type": "string",
-            "format": "iri-reference"
+            "format": "iri-reference",
+            "$comment": "any fragment must be empty, or match anchor or json-pointer syntax",
+            "pattern": "^[^#]*(#([A-Za-z_][-A-Za-z0-9.:_]*|/([^~]|~[01])*))?$"
```
notEthan, line 51:
```suggestion
            "pattern": "^[^#]*(#([A-Za-z_][-A-Za-z0-9.:_]*|(/([^~/]|~[01])*)*))?$"
```

The second half of the fragment match, for JSON Pointer, won't match the empty pointer (`$ref: "#"`, `$ref: "foo#"`). The suggested change follows the [JSON Pointer ABNF](https://www.rfc-editor.org/rfc/rfc6901.html#section-3).

This is still incomplete though. Fragments may contain %xx encoded characters. In a pointer their use may be necessary, but in either a pointer or anchor they are expected to be treated like the characters they represent, even when those characters don't actually need to be %-encoded.

```suggestion
            "pattern": "^[^#]*(#(([A-Za-z_]|%[A-Fa-f0-9]{2})([-A-Za-z0-9.:_]|%[A-Fa-f0-9]{2})*|((/|%2[fF])([^~/%]|~[01]|%[A-Fa-f0-9]{2})*)*))?$"
```

Not too pretty, but I think it allows anything that's valid. It won't reject some invalid anchors or pointers if they're %-encoded - it would be possible to express the same character ranges as allowed combinations of %-encoded digits, but I'm not masochistic enough to do it.

I'm a little uncertain describing a uri fragment with a regex isn't more trouble than it's worth.

output/schema.json
```diff
@@ -19,7 +19,9 @@
         },
         "absoluteKeywordLocation": {
           "type": "string",
-          "format": "uri"
+          "format": "iri",
+          "$comment": "any fragment must be non-empty, and use json-pointer syntax",
+          "pattern": "^[^#]*(#(/([^~]|~[01])*)*)?$"
```
notEthan, line 24: The comment says must be non-empty, but this does allow an empty fragment.

```suggestion
          "pattern": "^[^#]*(#(/([^~/]|~[01])*)+)?$"
```

The change to + requires a non-empty pointer; disallowing the / from the reference-token character range is better consistent with its spec's ABNF (and iriReferenceToSchemaString above), though it doesn't ultimately change what the regex matches.

And of course %-encoding needs fixing here too.

gregsdennis:
@karenetheridge please address @notEthan's regex concerns, but I think this is good.

output/schema.json
karenetheridge: actually I think we need to allow empty fragments, so it's the comment that is wrong here.

handrews: @karenetheridge Agreed, you need to be able to represent the root with a JSON Pointer fragment, which would be an empty fragment.

schema.json
unavailable10: https://github.com/json-schema-org/json-schema-spec/issues/937

gregsdennis:
@karenetheridge happy to bring this stuff in, but it needs to be rebased and reworked since the output was extracted to its own spec.

Otherwise, I'll probably just include it in some other PR.

jdesrosiers:
I'm going to close this due to it inactivity. But, feel free to update it for the new meta-schema and reopen or create a new PR.

Issue #1077, opened 2021-02-21
Object-based contains


gregsdennis:
Inspired by [this SO question](https://stackoverflow.com/q/66304520/878701).

We have `contains` to verify that at least one item in an array matches a subschema.  And through `additionalProperties` and `patternProperties` we recognize that objects can have dynamic properties.

It seems like we should have a way to specify that at least one property matches a subschema.

It could make sense to extend `contains` to do this when it receives an object, or we could define a new keyword.  If we extend `contains`, then `minContains` and `maxContains` would naturally follow.  `format` (in all its glory) already works on multiple value types, so it's not something we haven't had before.

jdesrosiers:
I'm sure this has come up before, but I did a search in the vocabularies repo and didn't find anything.

I should point out that there is a way to express object-contains-at-least-one without a new keyword. It would be sugar for,

```json
{
  "not": {
    "additionalProperties": {
      "not": { ... schema ... }
  }
}
```

Just like this is sugar for array-contains-at-least-one (aka: `contains`),

```json
{
  "not": {
    "items": {
      "not": { ... schema ... }
  }
}
```

There is no workaround for `minContains`/`maxContains` for arrays or objects.

I wouldn't be opposed to adding a keyword or extending the scope of `contains` to cover objects.

This has come up twice in the last week and I vaguely remember seeing a similar issue not long ago as well. This indicates that there is need for something like this. While there is a workaround, unless they're a JSON Schema expert they're unlikely to figure it out on their own. I also feel like it fits a hole in JSON Schema, if we support a contains operation on arrays, why not objects.

Generally, I don't like the idea of having a keyword that does more than one thing, but the semantics of `contains` for array and object really are the same, so I wouldn't be opposed.

gregsdennis:
> ```
> {
>   "not": {
>     "additionalProperties": {
>       "not": { ... schema ... }
>   }
> }
> ```

I don't follow how this (or the array one) works like `contains`.

jdesrosiers:
@gregsdennis Let's look at the array one first because I think it's a little easier to follow.

```json
{
  "items": {
    "not": { "required": ["foo"] }
  }
}
```

This schema asserts that none of the items in the array can have a property "foo". If an instance validates to false against this schema, then we know it must have at least one item with a "foo" property. Since that false result is the result we actually want, we negate it with `not`.

```json
{
  "not": {
    "items": {
      "not": { "required": ["foo"] }
    }
  }
}
```

[...]

PR #1092, opened 2021-04-19
Allow contains to apply to objects as well as arrays

jdesrosiers:
Resolves https://github.com/json-schema-org/json-schema-spec/issues/1077

The two ways to go with this were to have `contains`/`minContains`/`maxContains` apply to objects, or to introduce three new keywords that apply to objects. I went with the first option because that was the only one anyone expressed a preference for and no one objected.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -410,13 +410,13 @@
                         then this keyword has no effect.
                     </t>
                     <t>
-                        An instance array is valid against "maxContains" in two ways, depending on
-                        the form of the annotation result of an adjacent
+                        An instance array or object is valid against "maxContains" in two ways,
+                        depending on the form of the annotation result of an adjacent
                         <xref target="json-schema">"contains"</xref> keyword. The first way is if
                         the annotation result is an array and the length of that array is less than
                         or equal to the "maxContains" value. The second way is if the annotation
-                        result is a boolean "true" and the instance array length is less than or
-                        equal to the "maxContains" value.
+                        result is a boolean "true" and the instance length is less than or equal to
+                        the "maxContains" value.
                     </t>
                 </section>
 
@@ -429,13 +429,13 @@
                         then this keyword has no effect.
                     </t>
                     <t>
-                        An instance array is valid against "minContains" in two ways, depending on
-                        the form of the annotation result of an adjacent
+                        An instance array or object is valid against "minContains" in two ways,
+                        depending on the form of the annotation result of an adjacent
                         <xref target="json-schema">"contains"</xref> keyword. The first way is if
                         the annotation result is an array and the length of that array is greater
                         than or equal to the "minContains" value. The second way is if the
-                        annotation result is a boolean "true" and the instance array length is
-                        greater than or equal to the "minContains" value.
+                        annotation result is a boolean "true" and the instance length is greater
+                        than or equal to the "minContains" value.
                     </t>
                     <t>
                         A value of 0 is allowed, but is only useful for setting a range
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -415,8 +415,8 @@
                         <xref target="json-schema">"contains"</xref> keyword. The first way is if
                         the annotation result is an array and the length of that array is less than
                         or equal to the "maxContains" value. The second way is if the annotation
-                        result is a boolean "true" and the instance length is less than or equal to
-                        the "maxContains" value.
+                        result is a boolean "true" and the instance length (number of items or
+                        properties) is less than or equal to the "maxContains" value.
                     </t>
                 </section>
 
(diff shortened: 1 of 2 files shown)


jsonschema-validation.xml
```diff
@@ -410,13 +410,13 @@
...
-                        An instance array is valid against "maxContains" in two ways, depending on
-                        the form of the annotation result of an adjacent
+                        An instance array or object is valid against "maxContains" in two ways,
+                        depending on the form of the annotation result of an adjacent
                         <xref target="json-schema">"contains"</xref> keyword. The first way is if
                         the annotation result is an array and the length of that array is less than
                         or equal to the "maxContains" value. The second way is if the annotation
-                        result is a boolean "true" and the instance array length is less than or
-                        equal to the "maxContains" value.
+                        result is a boolean "true" and the instance length is less than or equal to
```
gregsdennis, line 418: Can "length" be applied to objects?  Maybe we keep "instance array length" and add "or property count"?

jsonschema-validation.xml
```diff
@@ -429,13 +429,13 @@
...
-                        An instance array is valid against "minContains" in two ways, depending on
-                        the form of the annotation result of an adjacent
+                        An instance array or object is valid against "minContains" in two ways,
+                        depending on the form of the annotation result of an adjacent
                         <xref target="json-schema">"contains"</xref> keyword. The first way is if
                         the annotation result is an array and the length of that array is greater
                         than or equal to the "minContains" value. The second way is if the
-                        annotation result is a boolean "true" and the instance array length is
-                        greater than or equal to the "minContains" value.
+                        annotation result is a boolean "true" and the instance length is greater
```
gregsdennis, line 437: Same here.

jsonschema-validation.xml
jdesrosiers: I think "length" makes sense. It may be an object, but in this context, it's being treated like a map, which does have a concept of length. But I you think it's questionable, others will be confused, so we should be more explicit.

karenetheridge:
I think this keyword should be moved out of this section entirely -- as it is presently located in "Keywords for Applying Subschemas to Arrays". It can be moved to a third section after "Keywords for Applying Subschemas to Objects", called "Keywords for Applying Subschemas to Arrays or Objects", or even just "Other Keywords for Applying Subschemas".

jsonschema-validation.xml
karenetheridge: "number of properties" is explicit and doesn't depend on a language/framework's particular data model.

awwright:
What happens if someone wants to specify that an array to contain an item, OR that an object to contain a property value?

I would like to suggest these should be different keywords, e.g. `containsItem` or `containsValue`/`containsProperty`

gregsdennis:
> What happens if someone wants to specify that an array to contain an item, OR that an object to contain a property value? - @awwright 

If they only wanted one or the other, they would need to also specify `type`, which is the current solution to that need anyway.

jdesrosiers:
@awwright If you think this needs more discussion, please comment on https://github.com/json-schema-org/json-schema-spec/issues/1077. I'm happy to put this PR on hold until consensus is achieved, but that discussion should take place on the issue, not here.

Relequestual (approved):
Seems reasonable to me.
Want to see all PRs adding to the changelog to avoid the mess of making a changelog at the end 😅

- [ ]  Added to changelog

jdesrosiers:
Good thinking about adding to the change log! It sounds like this PR might get closed as preference seems to be shifting toward using new keywords. When it's decided, I will make sure whichever one we go with is included in the changelog.

handrews:
While it sounds like this PR is a bit in-limbo, if that changes and it's re-approved, it would be good to wait to merge until we decide whether to [branch for the next major draft](https://github.com/json-schema-org/community/discussions/7).

jdesrosiers:
It appears we have general consensus on this direction, so I'm removing "draft" status to open this up for review. However, @handrews, I will not merge until we decide what's happening with the bug-fix draft. This should not be included in the bug-fix draft.

jdesrosiers:
I had the create a new section for the change log because it doesn't exist yet. I used `draft-bhutton-json-schema-next` as the draft identifier because it's not certain which draft we will be on when this gets released if there are one or more bug-fix drafts. It will have to be updated when the next full draft release is near.

handrews (approved):
Excellent work on reaching this solution!

gregsdennis (approved):

jdesrosiers:
I just want to include a quick summary of the discussion around adding this keyword.

There was no opposition to adding `contains`-like functionality that applies to objects. This functionality was already possible in JSON Schema using the following pattern.

```json
{
  "not": {
    "patternProperties": {
      "": { "not": { ... schema ... } }
    }
  }
}
```
Note: I used `patternProperties` instead of `additionalProperties` to make it explicit that it should not be affected by defined `properties`.

The only point of contention was whether `contains` should apply to objects as well as arrays, or there should be new keywords that only apply to objects. Using the same keyword could result in backwards incompatible behavior in very rare (and poorly designed) situations. But it can also be useful such as describing a value that can be an object or an array, but either way, it must contain some value. However, the main benefit of using the same keywords is that it's easy for schema authors to understand and allows us to avoid adding three new keywords.

In the end, the only people expressing strong preferences were for overloading `contains` and there was only one person with a slight preference for new keywords. So, went with overloading `contains`.

karenetheridge:
I noticed there were no changes to meta/applicator.json. Shall we do those in a subsequent PR?

jdesrosiers:
I don't think there are any meta-schema changes needed for this. Am I missing something?

PR #1103, opened 2021-05-11
Vocabulary definitions need not be formal or published depending on usage.

### LINKED ISSUE #1058: Degree of formality for third-party vocabularies
*This may be more appropriate for https://github.com/json-schema-org/json-schema-vocabularies, but I'm adding it here because this is where we define vocab requirements.  Feel free to move it.*

Do we expect that third-party vocabularies need to be written as formal specifications (like this spec is), or will any simple, published document (e.g. a Google doc or a blog post) suffice as long as it's specific enough for an implementor to write behavior for its keywords?

handrews:
Fixes #1058.  I think there is agreement on this?  It was faster to write a PR than ping the issue.  I'm not attached to this wording if folks prefer something else.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -399,6 +399,13 @@
                         of any vocabulary, there is no analogous mechanism to indicate individual
                         keyword usage.
                     </t>
+                    <t>
+                        A schema vocabulary can be defined by anything from an informal description
+                        to a standards proposal, depending on the audience and interoperability
+                        expectations.  In particular, in order to facilitate vocabulary use within
+                        non-public organizations, a vocabulary specification need not be published
+                        outside of its scope of use.
+                    </t>
                 </section>
                 <section title="Meta-Schemas">
                     <t>


karenetheridge:
I'd like to add that the URI used to represent the vocabulary (in the $vocabulary keyword) SHOULD resolve to a document (not a schema, but any sort of plain text document that describes the vocabulary - ranging from a freeform README right up to a formal specification), that need not be publicly be resolvable, but SHOULD be resolvable in the context where it is intended to be used -- for example, if a company develops a vocabulary for its private use, then the vocabulary URI need not be publicly available, but SHOULD be available on its own private network as a description of that vocabulary. Note "SHOULD" not "MUST" -- this is just a recommended best practice but there may be reasons why it can't be done in all circumstances.

I think this was the intent, as discussed earlier, but it never made it into the spec.

handrews:
@karenetheridge 

> I'd like to add that the URI used to represent the vocabulary (in the $vocabulary keyword) SHOULD resolve to a document (not a schema, but

That's outside the scope of this PR, although you're welcome to file an issue for it.  The sort of wording you propose was discussed and rejected.  I don't recall why (or even what side I was on), but the place to sort that out is in a new issue.

karenetheridge:
Ok, I created #1105.

jdesrosiers (approved):

Issue #1059, opened 2021-01-07
prohibit the use of fragments in $id and $schema


karenetheridge:
We permit empty fragments in `$id`, and say nothing at all about fragments in `$schema`. IMO we should prohibit all kinds of fragments in both keywords' URIs.

cross-reference: https://github.com/OAI/OpenAPI-Specification/issues/2433#issuecomment-756389903

jdesrosiers:
I'd rather go the other direction and make `$id` and `$schema` consistent with `$ref`. The absolute-URI portion determines the document fetched from the schema-store and the fragment portion determines which part of the document to start evaluation when processing the schema.

For `$schema`, this would mean that all of the following would identify the same dialect, but might resolve to different meta-schemas.
- `https://json-schema.org/draft/future`
- `https://json-schema.org/draft/future#`
- `https://json-schema.org/draft/future#strict`
- `https://json-schema.org/draft/future#uber-strict`

For `$id`, this would mean that embedded schema evaluation would need to take into account the fragment to determine where to begin evaluation.

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "$id": "/schema/bundle#/$defs/foo",
      "$defs": {
        "string": { "type": "string" },
        "foo": { "$ref": "#/defs/string" }
      }
    }
  }
}
```

In this example, `/properties/foo` resolves to `{ "type": "string" }`. It might seem strange when we look at it embedded like this, but this is exactly what is happening when we resolve a `$ref`. We begin validation at the location identified by the fragment, but the whole document has to be in scope in order to resolve references.

If we change that example slightly, I think it's a little easier to see how the `$id` with fragment is the same as `$ref` with fragment.

```json
{
  "type": "object",
  "properties": {
    "foo": { "$ref": "/schema/bundle#/$defs/foo" }
  },
  "$defs": {
    "": {
      "$id": "/schema/bundle",
      "$defs": {
        "string": { "type": "string" },
        "foo": { "$ref": "#/defs/string" }
      }
    }
  }
}
```

It's certainly more intuitive written like this, but I don't think aesthetics are good enough reason to forbid using fragments in `$id`.

notEthan:
allowing pointer fragments in $id was an oversight to begin with. disallowing it resolved a confusing/undefined situation where a schema could identify itself as being at one pointer, while actually being at a different pointer in the resource or document.

[...]

PR #1104, opened 2021-05-14
Clarify various things about canonical URIs

handrews:
Fixes issue #937, clarifying a number of other things along the way.
While it touches a fair number of lines, I'm fairly sure that it
doesn't anything about conformance.

_**[NOTE:** This **does not** address the question of whether "canonical" is what we want - that is too large of a change for a [patch release](https://github.com/json-schema-org/community/discussions/7) which is what this PR is targeted for.  That topic should get its own issue if someone wants to propose something about it.  It also **does not** deal with #1059, although it clarifies some things based on discussions there.]_

After spending more time reading various writings on the concept
of the "canonical" URI for a resource, and reviewing our language,
I came to the following conclusions:

* canonical URIs only make sense at the whole-resource scope
* A URI with a fragment is neither canonical nor non-canonical
* It makes more sense to talk about fragments w.r.t. canonical URIs
* Our language was sufficiently confusing that going this way seems fine.

As part of this, I fixed an outright incorrect statement that
identifier keywords set canonical URIs.  Since there is only
one canonical URI and a single schema object could contain
three ($id, $anchor, $dynamicAnchor) or more identifier keywords,
this statement is clearly a bug.  These keywords assign URIs,
but only $id assigns a canonical one.

I revamped a lot of wording in descriptions and examples to
hopefully be more precise.  I separated the discussion of
the empty fragment in $id from the main paragraph of its
functionality, and clarified that this is talking about a
media-type-specific semantic equivalence, and is not asserting
that RFC 3986 normalization applies to fragments (this has
been a point of confusion).

_If someone wants me to split this up and sees a way to do it, I would be happy to do that._

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3131,20 +3131,6 @@ https://example.com/schemas/common#/$defs/count/minimum
                         <t>Subtype name: schema+json</t>
                         <t>Required parameters: N/A</t>
                         <t>
-                            Optional parameters:
-                            <list style="hanging">
-                                <t hangText="schema:">
-                                    A non-empty list of space-separated URIs, each identifying
-                                    a JSON Schema resource.  The instance SHOULD successfully
-                                    validate against at least one of these meta-schemas.
-                                    Non-validating meta-schemas MAY be included for purposes such
-                                    as allowing clients to make use of older versions of
-                                    a meta-schema as long as the runtime instance validates
-                                    against that older version.
-                                </t>
-                            </list>
-                        </t>
-                        <t>
                             Encoding considerations: Encoding considerations are
                             identical to those specified for the "application/json"
                             media type.  See <xref target="RFC8259">JSON</xref>.
@@ -3174,20 +3160,7 @@ https://example.com/schemas/common#/$defs/count/minimum
                     <list>
                         <t>Type name: application</t>
                         <t>Subtype name: schema-instance+json</t>
-                        <t>
-                            Required parameters:
-                            <list style="hanging">
-                                <t hangText="schema:">
-                                    A non-empty list of space-separated URIs, each identifying
-                                    a JSON Schema resource.  The instance SHOULD successfully
-                                    validate against at least one of these schemas.
-                                    Non-validating schemas MAY be included for purposes such
-                                    as allowing clients to make use of older versions of a schema
-                                    as long as the runtime instance validates against that
-                                    older version.
-                                </t>
-                            </list>
-                        </t>
+                        <t>Required parameters: N/A</t>
                         <t>
                             Encoding considerations: Encoding considerations are
                             identical to those specified for the "application/json"


jsonschema-core.xml
```diff
@@ -1803,24 +1815,25 @@
 ]]>
                         </artwork>
                         <postamble>
-                            Here we see that the canonical URI for that "additionalProperties"
-                            subschema is still valid, while the non-canonical URI with the fragment
-                            beginning with "#/items/$ref" now resolves to nothing.
+                            Here we see that the URI for the "additionalProperties" schema object
+                            that is relative to its resource's canonical URI is still valid,
+                            while the URI relative to the "items" schema object's URI no longer
+                            resolves to anything.
```
jdesrosiers, line 1821: I found this sentence difficult to parse. I think it would be more clear if the identifiers in question were explicit.

```suggestion
                            Here we see that `https://example.com/bar#/additionalProperties`
                            is still valid, while `https://example.com/foo#/items/additionalProperties`
                            no longer resolves to anything.
```

jdesrosiers (commented):
I have a couple suggestions, but I think this cleans up a lot of the concerns I had about how the term canonical is used in the spec.

jsonschema-core.xml
```diff
@@ -1333,26 +1335,31 @@
                     <t>
                         If present, the value for this keyword MUST be a string, and MUST represent a
                         valid <xref target="RFC3986">URI-reference</xref>.  This URI-reference
-                        SHOULD be normalized, and MUST resolve to an
-                        <xref target="RFC3986">absolute-URI</xref> (without a fragment).  Therefore,
-                        "$id" MUST NOT contain a non-empty fragment, and SHOULD NOT contain an
-                        empty fragment.
+                        SHOULD be normalized, and MUST be semantically equivalent to an
+                        <xref target="RFC3986">absolute-URI</xref> (without a fragment).
```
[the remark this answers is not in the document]
handrews: @jdesrosiers it might be better to keep the original wording, yeah.  The reason I changed it is that media type-specific fragment transformations aren't really a part of URI "resolution."  But as @karenetheridge notes, "semantically equivalent" might be a little obscure.  It just means that the underlying meaning of the URI is the same.  Although really even that might be debatable as without a fragment identifies the resource, and with a fragment identifies the root object, which happens to effectively be the same thing.

Meh.  Yeah, maybe "resolve to" is more clear and as close to correct as anything?

jsonschema-core.xml
handrews: @jdesrosiers I generally agree, let me think a bit on wording, I feel like something about the reasoning is lost in the change, but agree that wording closer to your suggestion will be better.

jsonschema-core.xml
```diff
@@ -419,7 +419,9 @@
                     <t>
                         A JSON Schema resource is a schema which is
                         <xref target="RFC6596">canonically</xref> identified by an
-                        <xref target="RFC3986">absolute URI</xref>.
+                        <xref target="RFC3986">absolute URI</xref>.  Schema resources MAY
+                        also be identified by URIs including fragments.  Any such URIs
```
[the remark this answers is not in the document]
handrews: @karenetheridge I agree with your change, although now I'm skeptical of having added the statement at all.  Let me look into that, but if I keep the change I'll commit your suggestion.

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: Addressed in comment to Jason above.

jsonschema-core.xml
```diff
@@ -1333,26 +1335,31 @@
...
+                            An empty fragment is currently allowed because older meta-schemas have
+                            an empty fragment in their $id (or previously, id).
+                            A future draft may outright forbid even empty fragments in "$id".
                         </cref>
                     </t>
                     <t>
-                        This URI also serves as the base URI for relative URI-references in keywords
-                        within the schema resource, in accordance with
+                        The absolute-URI also serves as the base URI for relative URI-references
+                        in keywords within the schema resource, in accordance with
```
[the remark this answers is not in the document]
handrews: Except that those are all the non-canonical URIs that we want to discourage, so I'd rather not call attention to them.

jsonschema-core.xml
```diff
@@ -1743,7 +1750,7 @@
                         Since JSON Pointer URI fragments are constructed based on the structure
                         of the schema document, an embedded schema resource and its subschemas
                         can be identified by JSON Pointer fragments relative to either its own
-                        canonical URI, or relative to the containing resource's URI.
+                        canonical URI, or relative to a containing resource's URI.
```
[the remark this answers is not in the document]
handrews: Yeah, I agree.  Will fix.

jsonschema-core.xml
```diff
@@ -1803,24 +1815,25 @@
...
                     </t>
                     <t>
-                        An implementation MAY choose not to support addressing schemas
-                        by non-canonical URIs. As such, it is RECOMMENDED that schema authors only
-                        use canonical URIs, as using non-canonical URIs may reduce
-                        schema interoperability.
+                        An implementation MAY choose not to support addressing schema resource
+                        contents by URIs using a base other than the resource's canonical URI,
+                        plus a JSON Pointer fragment relative to that base.  Therefore, schema
+                        authors SHOULD NOT rely on such URIs, as using them may reduce interoperability.
```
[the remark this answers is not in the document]
handrews: Yes, very much outside the scope, please have this discussion elsewhere (you filed an issue, IIRC?)  It would be best not to bring up these comments in PRs because often they touch of big discussions despite repeated "this is outside the scope" attempts to close it down.  Although obviously there's no problem with that on this comment right now :-)

GUANLONGH (approved):

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: I tried to sort this out better, not sure if I succeeded or not.

jsonschema-core.xml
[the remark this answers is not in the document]
handrews: I'm going to stick with not bringing those other possible bases up here.  They are explained in detail elsewhere, and are really an odd confusing case that we'd rather people just not think about at all.

jsonschema-core.xml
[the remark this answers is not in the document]
jdesrosiers: You definitely succeeded. The new text says what it needs to say and is easy to read.

jdesrosiers (approved):
Approved. I think this is a big improvement.

jsonschema-core.xml
```diff
@@ -419,13 +419,25 @@
...
                         A JSON Schema resource is a schema which is
                         <xref target="RFC6596">canonically</xref> identified by an
-                        <xref target="RFC3986">absolute URI</xref>.
+                        <xref target="RFC3986">absolute URI</xref>.  Schema resources MAY
+                        also be identified by URIs, including URIs with fragments,
+                        if the resulting secondary resource (as defined by
+                        <xref target="RFC3986">section 3.5 of RFC 3986</xref>) is identical
+                        to the primary resource.  This can occur with the empty fragment,
+                        or when one schema resource is embedded in another.  Any such URIs
+                        with fragments are considered to be non-canonical.
```
Relequestual, line 428: You say...

> A URI with a fragment is neither canonical nor non-canonical

And yet this block says otherwise.

Relequestual (changes requested):
One question but otherwise looks good!

jsonschema-core.xml
karenetheridge: I would suggest rephrasing too "a URI with a fragment is **not necessarily** canonical or non-canonical" -- canonical state is a boolean state (this isn't Schroedinger's schema!), but we can use neither "has a fragment" nor "doesn't have a fragment" as an indicator of canonical state.

It might be worth stating more clearly how to determine the canonical URI of any particular location in a schema:
- if there is an "$id" keyword at this level, then the canonical URI is the $id URI after being fully resolved against its base (and there is never a fragment)
- otherwise, the canonical URI is the canonical URI of the closest $id (or however you want to say "the innermost schema that contains this one, that has an "$id"), with a fragment that contains the json pointer from that $id location to here; OR, if there is no $id in the schema at all, then we use the base URI of the root schema itself, plus a fragment of the json pointer from the root.

Relequestual: > but we can use neither "has a fragment" nor "doesn't have a fragment" as an indicator of canonical state

We can't? `$id` can have a fragment, but the resulting resolved URI will also then have a fragment, and this isn't allowed...

> If present, the value for this keyword MUST be a string, and MUST represent a valid URI-reference. This URI-reference SHOULD be normalized, and MUST resolve to an absolute-URI (without a fragment). Therefore, "$id" MUST NOT contain a non-empty fragment, and SHOULD NOT contain an empty fragment.

I've had this open for hours and I can't remember if I was going to say anything else.

Am I missing something?

karenetheridge: We're not talking just about `$id`s here, but canonical URIs everywhere.  e.g.:

```
{
  "$id": "https://myschema",
  "properties": {
    "foo": {
      "allOf": [
        { "type": "string" }
      ]
  }
}
```

The canonical URI of the subschema at `type" is "https://myschema#/properties/foo/allOf/0/type". It has a fragment, and this URI is canonical because it's using the base of the nearest `$id`. But we can't tell that it's canonical just from looking at the URI itself - we need to inspect the document it's in to determine that.

Relequestual: I agree with all of the above you've said, however upon re-reading I do have one question...

> ...unless the fragment is a plain-name fragment, i.e. derived from an $anchor, because $anchor URIs are never canonical.

If `$anchor` producing URIs are non-canonical, does that mean the keyword doesn't need to be supported?

Obviously that's not the intent here, but that's how I read it logically.

Do we need to amend this section to include plain name fragments?

> An implementation MAY choose not to support addressing schema resource
>                         contents by URIs using a base other than the resource's canonical URI,
>                         plus a JSON Pointer fragment relative to that base.  Therefore, schema
>                         authors SHOULD NOT rely on such URIs, as using them may reduce interoperability.

https://github.com/json-schema-org/json-schema-spec/pull/1104/files#diff-bda51aab56b4b2e23fc9e7249e80fe6e262dcd552f35c5de69fd4a6bf3e96b50R1847-R1850

karenetheridge: Yes, that's a much nicer way of putting that.  The intent is surely just to say "you don't have to support non-canonical URIs that involve json pointers".

Personally I find handling these sorts of URIs to be no trouble at all, so I wouldn't mind if this entire paragraph went away, but I gather that others store schemas internally in a way that makes it difficult to use a json pointer that crosses an $id boundary.

jsonschema-core.xml
```diff
@@ -419,13 +419,25 @@
                     <t>
                         A JSON Schema resource is a schema which is
                         <xref target="RFC6596">canonically</xref> identified by an
-                        <xref target="RFC3986">absolute URI</xref>.
+                        <xref target="RFC3986">absolute URI</xref>.  Schema resources MAY
+                        also be identified by URIs, including URIs with fragments,
+                        if the resulting secondary resource (as defined by
+                        <xref target="RFC3986">section 3.5 of RFC 3986</xref>) is identical
+                        to the primary resource.  This can occur with the empty fragment,
+                        or when one schema resource is embedded in another.  Any such URIs
```
karenetheridge, line 427: This can also happen if there is a `$dynamicAnchor` and/or `$anchor` at this location, but I'm not sure if it's worth mentioning that here.

Relequestual: Let's break this into a new issue if you feel strongly enough about it =]

jsonschema-core.xml
Relequestual: OK, I think that sounds like an agreement.

Relequestual:
Given the current absense of @handrews, I'll look to create a new superseding PR using these commits, but with one additional change as per the [above not yet resolved comment thread](https://github.com/json-schema-org/json-schema-spec/pull/1104#discussion_r757454253).

jsonschema-core.xml
Relequestual: > We're not talking just about $ids here, but canonical URIs everywhere. e.g. in:... - @karenetheridge 

I'm not sure that's correct. In re-reading every use of "canonical" in the spec, we either say "canonical URI" or "canonical URI plus pointer fragment".

Specifically, in unmodified (in terms of the patch branch) sections of the spec from 2020-12: https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.appendix.A

Relequestual: > Yes, that's a much nicer way of putting that. - @karenetheridge 

To what were you referencing, if you can recall?

jdesrosiers: > If `$anchor` producing URIs are non-canonical, does that mean the keyword doesn't need to be supported?
>
> Obviously that's not the intent here, but that's how I read it logically.
>
> Do we need to amend this section to include plain name fragments?
>
> >  An implementation MAY choose not to support addressing schema resource
    contents by URIs using a base other than the resource's canonical URI,
    plus a JSON Pointer fragment relative to that base. Therefore, schema
    authors SHOULD NOT rely on such URIs, as using them may reduce interoperability.

Yes, I think this unintentionally excludes anchor fragments and a few other things that should be ok. I think that putting this in terms of the canonical URI is not correct. Here are a few examples to show what I mean.

```json
{
  "$id": "/root",
  "type": "object",
  "properties": {
    "aaa": { "$ref": "/embedded" },
    "bbb": { "$ref": "#/$defs/embedded" },
    "ccc": { "$ref": "/embedded#/$defs/abc" },
    "ddd": { "$ref": "/embedded#xxx" },
    "eee": { "$ref": "#/$defs/embedded/$defs/abc" },
    "fff": { "$ref": "#xxx" }
  },
  "$defs": {
    "embedded": {
      "$id": "/embedded",
      "type": "number",
      "$defs": {
        "abc": {
          "$anchor": "xxx",
          "type": "string"
        }
      }
    }
  }
}
```

The references at "aaa" and "ccc" represent what MUST be supported. "bbb" and "ddd" show non-canonical references that the clause above technically excludes from being required to support, but there's no reason not to support them. "eee" is the only case the above clause should be referring to. (I'm not sure why it's even allowed. It should be forbidden. But, that's off topic and nothing we can change in a patch release.) I'm not even sure what the spec technically says about "fff", but it should be invalid because there is no "xxx" anchor in the "/root" schema resource. It's in the "/embedded" schema resource.

So, I'd remove mention of canonical URIs from that clause and say something like, the result of following a URI with a fragment that points to a location outside the current schema resource is undefined. Then I'd get rid of all mention of canonical URIs because it's no longer necessary to describe any feature of JSON Schema.

Relequestual: I don't THINK I dissagree.

"xxx" does not resolve for the reasons you've stated.

"eee" is the target of what isn't required to be supported.
I think people had a hard time conceptually with multiple URIs being valid, specifically across `$id` boundaries.

For every `$id` boundary you cross, you increase the number of potential resolveable URI locations by n^2.

I guess that could start to become a problem, but I haven't seen anyone raise any such issue.

Not being able to remember why this was done, I did some digging and found https://github.com/json-schema-org/json-schema-spec/commit/a93b605d2f5d72b0de824081567e435782c32ce2

Specifically...

> While URIs for subschemas with "$id" and their chidren can be
constructed using the base URI from a parent schema, these
URIs are non-canonical, and their behavior is undefined.  This is
on the grounds that switching between embedding and referencing
schema resources should behave essentially identically.

Essentially, taking @jdesrosiers above example, if we assume it's a bundled "compound schema" document, it works. However if the document was unbundled back into individual resources, "eee": `"$ref": "#/$defs/embedded/$defs/abc"` would no longer be resolveable.

This is because the reference at "eee" crosses resource boundary (aka a schema location with `$id` at the non-root location.

We called references such as "eee" "shadowed JSON Pointers" in discussions, it would seem.

karenetheridge: > > Yes, that's a much nicer way of putting that. - @karenetheridge
> 
> To what were you referencing, if you can recall?

The statement from you immediately above, where you propose new wording.

karenetheridge: Regarding plain name fragments, canonical uris, where they appear and what is needed to be supported \[copied from slack\]:

If the spec says (paraphrased) "you only MUST support references to canonical uris", and canonical uris can't have fragments, then it's impossible to have a reference to any location that doesn't have an $id. e.g. "$ref": "#/$defs/foo".  I think it's more correct to say that every position in the document has one and exactly one canonical URI that describes it, and zero or more other URIs that MAY also be used to refer to it (which some implementations may or may not choose to support), with the exception of plain-name fragments (generated via $anchor) which MUST also be supported.

jdesrosiers: I agree that @karenetheridge's summary is what the spec is trying to say. I still think it's unnecessarily complicated, but that's something that would need to be addressed elsewhere.

> zero or more other URIs that MAY also be used to refer to it (which some implementations may or may not choose to support)

I dislike this phrasing. I think it's backwards. It makes it sound like this is an optional feature that some will choose to support and others will not. It's not a feature. It's always wrong. Just because it MAY happen to work in some implementations doesn't mean it's not always wrong.

It's like duplicate properties in JSON. Given, `{ "a": 1, "a": 2 }`, if I expect `/a` to be `2` and I get `2`, then the implementation appears to be working as I expect. I might think that this implementation "supports" the second property definition overriding the second. But that's not a real feature and not a reliable behavior.

The spec should say that the behavior of referencing a location using a non-canonical URI is undefined. Your schema might work the way you expected in your favorite implementation, but your schema is still wrong.

karenetheridge: > I dislike this phrasing. I think it's backwards. It makes it sound like this is an optional feature that some will choose to support and others will not. It's not a feature. It's always wrong.

It's not wrong. We even give examples of this usage in the spec: https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.appendix.A -- see "non-canonical URI with fragment relative to root.json"

The only reason to discourage the usage of non-canonical URIs is because they might break if the document is broken up into a different form, e.g. some parts bundled elsewhere.  There are some usecases where referencing locations relative to the root of the document, rather than relative to the nearest `$id`, might even be the most convenient implementation.

jdesrosiers: > We even give examples of this usage in the spec

That explains what is considered canonical and what isn't. It's not an example of valid usage. It's there specifically to point out what isn't valid.

> The only reason to discourage the usage of non-canonical URIs is because they might break if the document is broken up into a different form, e.g. some parts bundled elsewhere.

Yeah, that's a pretty important reason.

> There are some usecases where referencing locations relative to the root of the document, rather than relative to the nearest `$id`, might even be the most convenient implementation.

I'm going to need proof for that assertion. I don't believe such a case exists.

karenetheridge: "pretty important" is subjective. Bundling can be completely irrelevant to some usecases, and quite useful to others. In any usecase where bundling is unimportant, non-canonical URIs can be valid and useful. Whether you believe in them or not is irrelevant.

> The schemas at the following URI-encoded JSON Pointers (relative to the root schema) have the following base URIs, **and are identifiable by any listed URI in accordance with sections 5 and 9.2.1 above**. 

Section 9.2.1 says:

> Since URIs involving JSON Pointer fragments relative to the parent schema resource's URI cease to be valid when the embedded schema is moved to a separate document and referenced, applications and schemas SHOULD NOT use such URIs to identify embedded schema resources or locations within them. 

Note: "should not", not "must not".

Why is this still being discussed? If you want to make this usecase prohibited, submit a proposal for the next draft.

jdesrosiers: > Why is this still being discussed? If you want to make this usecase prohibited, submit a proposal for the next draft.

I did veer off course a little. I'm sorry about that and the confusion that caused. But right now, I'm not arguing to prohibit anything, nor am I arguing to change anything the spec says. I'm saying that the behavior of referencing a location using a non-canonical URI is undefined, not that it's prohibited. You can do it, but there's no guarantee what behavior you will get. The spec would still describe the same things, it's just framed in different way.

Perhaps the cause of the confusion is my use of the word "wrong". When I say "wrong" I don't mean invalid. I mean it in the same way that duplicate properties in JSON is "wrong". It's not an invalid document, but it's conceptually a "wrong" way to use JSON. It's syntactically valid, but semantically nonsensical.

The words you used are more in line with the way the spec is currently written. Obviously I dislike the way it's framed, but if changing that is too much for this PR, that's fine. The scope of this PR was never to fix everything, it was just to try to make a few small improvements.

Relequestual:
I've re-evaluated my position and thinking on this, and I believe it's now more aligned with what Henry put in his first comment on this PR. I THINK the associated issue could be resolved with a much smaller set of changes, so will proceed to try that.

jdesrosiers:
Remember this issue is just intended to be minimal clean up to fix glaring problems. It doesn't have to solve everything.

Relequestual:
> Remember this issue is just intended to be minimal clean up to fix glaring problems. It doesn't have to solve everything.

Indeed. I'm going to propose a super minimal change (far smaller than this) based on my [comment](https://github.com/json-schema-org/json-schema-spec/issues/937#issuecomment-1016543486) from yesterday.

Relequestual:
Closing in favour of:
- https://github.com/json-schema-org/json-schema-spec/pull/1192
- https://github.com/json-schema-org/json-schema-spec/issues/1183

jdesrosiers:
Why was this closed? Even if the goal is to remove the "canonical" language in the next release, I'd still like to have these fixes for the patch release if possible. This fixes the vast majority of the problems with the way "canonical" is used. If we didn't already have this, I'd say don't bother if the goal is to remove "canonical" anyway. But, the work is already done and it's an improvement over what we have, so why not use it?

Relequestual:
OK. Reopened.

I closed it because it seems like there are outstanding questions which we aren't going to get answered, and I feel my alternate PR does enough for now to fix the immediate and biggest problem.

I'll re-evaluate this PR with new eyes.
How would you like to proceed? Make a new PR, cerrypicking those commits, and make additional commits to form the PR?

jdesrosiers:
> it seems like there are outstanding questions which we aren't going to get answered

What questions are you referring to? We don't have to solve everything for the patch release. If anything is controversial, we don't need to change it now.

> How would you like to proceed?

Merge it as it is. As it is now, it addresses problems and doesn't introduce any new problems. Any additional things we feel needs to be addressed for the patch release can be a separate PR.

jsonschema-core.xml
Relequestual: I made this initial comment as I was confused about the seeming contradiciton between the first comment and this section in the PR, however after review, I no longer believe they are contradictory, and so I'll resolve it.

Relequestual (approved):

jsonschema-core.xml
```diff
@@ -419,13 +419,25 @@
...
                         The root schema is the schema that comprises the entire JSON document
                         in question.  The root schema is always a schema resource, where the
                         URI is determined as described in section
                         <xref target="initial-base" format="counter"></xref>.
+                        <cref>
+                            Note that documents that embed schemas in another format will not
+                            have a root schema resource in this sense.  Exactly how such usages
+                            fit with the JSON Schema document and resource concepts will be
+                            clarified in a future draft.
+                        </cref>
```
karenetheridge, line 440: I'm not sure what the purpose is of this paragraph. What does "embed schemas in another format" mean?

karenetheridge (approved):

jsonschema-core.xml
jdesrosiers: This is referring to something like a JSON Schema in an OpenAPI document.

Relequestual:
Hey all. I screwed up, and force-pushed to try undo a merge commit (which happened when trying to resolve a conflict using the GitHub UI... mistake!).
I now do not have the permissions to force-push and re-introduce the commits.
As such, I'll do so on a new branch and create a new PR, referencing this one.
I'll then resolve the conflicts and merge, as there were three approvals on this PR.

karenetheridge:
> I now do not have the permissions to force-push and re-introduce the commits.

You have admin privileges on the repository, so you can temporarily give yourself permission to force-push master and then remove the privileges again. (don't tell anyone but it's been known to happen before.) ;)

Issue #1074, opened 2021-02-08
The value of the examples annotation should be an object or an array


K-Adam:
Json schema has [examples annotation](https://github.com/json-schema-org/json-schema-spec/blob/6c5b6729ce7482072404b09c02533291e9c528c7/jsonschema-validation.xml#L1223), which comes handy, when I would like to run a mock server for my api.

Unfortunately the spec states, that the value of `examples` keyword MUST be an array:

```json
{
  "title" : "Match anything",
  "description" : "This is a schema that matches anything.",
  "default" : "Default value",
  "examples" : [
    "Anything",
    4035
  ]
}
```

I suggest to enable objects as well, so each example could be referenced by name: 

```json
{
  "title" : "Match anything",
  "description" : "This is a schema that matches anything.",
  "default" : "Default value",
  "examples" : {
    "stringValue": "Anything",
    "integerValue": 4035
  }
}
```

This change could make the construction of complex examples ( like a list ) much simpler.

handrews:
Another option would be to add a `namedExamples` or `exampleObject` keyword to make things a bit easier on the statically typed language implementations and not break existing implementation assumptions.

karenetheridge:
Now that dependencies has split into dependentRequired and dependentSchemas, and items into prefixItems and items, do we have any polytyped keywords remaining? ..other than examples and const, which can be any type at all...

gregsdennis:
> to make things a bit easier on the statically typed language implementations - @handrews 

This isn't an issue in my .net implementation.  I just save the JSON model which can represent the appropriate types.

handrews:
@karenetheridge `type` is the other one.

The rationale for splitting `dependencies` wasn't about the types, though, but about separating the applicator and assertion functions.  Since the two syntaxes corresponded to two different sorts of functionality it made a lot of sense to split.  `type` really is doing the same in both syntaxes, it's just allowing you to leave off the array markers if there's only a single type.

handrews:
(idk why I felt compelled to go all history less on `dependencies` 🤷 )

gregsdennis:
We expect nothing less from you @handrews.

K-Adam:
@handrews `namedExamples` would be perfect

gregsdennis:
In 2020-12, we changed the behavior of unknown keywords to be included as annotations rather than being ignored.

This means that you can add `namedExamples` to any 2020-12 schema, and it should be included in the output as an annotation along with its value.

[...]

PR #1114, opened 2021-07-07
Add namedExamples

K-Adam:
#1074

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -1241,6 +1241,24 @@
                     an additional example.  If "examples" is absent, "default"
                     MAY still be used in this manner.
                 </t>
+                <t>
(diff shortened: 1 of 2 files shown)


gregsdennis (commented):
Thanks for preempting this change. I think that it needs a bit more discussion, though. We don't typically add keywords to the base schema lightly.

I'll leave further comments in the issue.

handrews:
Even if accepted, this also needs to be changed to target the `draft-next` branch rather than `master`.

K-Adam:
@handrews The [Guidelines for contributing](https://github.com/json-schema-org/json-schema-spec/blob/draft-next/CONTRIBUTING.md) mentioned, that "Pull requests should be made to master"

handrews:
@K-Adam oops, thanks, it's a recent change, forgot to update the file.

PR #1137, opened 2021-10-15
Support for IRI references

jakubklimek:
This would add support for IRI references in $id, $ref and $dynamicRef as per #59.
This PR fixes history of #1068.


jakubklimek:
@Relequestual @handrews I have pushed commits changing URIs to IRIs in the spec as well where necessary. I left references to RFC 3986 (URIs) wherever resolving base URIs is mentioned, as this is also applicable to IRIs and RFC 3987 actually refers to RFC 3986 in this manner.

I would therefore like to resume discussions about merging this.

@jdesrosiers I fixed the commit history, which closed #1068.

Issue #1064, opened 2021-01-15
Remove $dynamicRef bookending requirement


jdesrosiers:
The current process for resolving a `$dynamicRef` requires that both ends of the dynamic scope have a `$dynamicAnchor`. This bookending requirement complicates `$dynamicRef` resolution, makes some use cases impossible, and doesn't seem to serve any useful purpose as far as I can tell.

The only explanation for the bookending requirement in the spec comes from this cref.

>    Requiring both the initial and final URI fragment to be defined
    by "$dynamicAnchor" ensures that the more common "$anchor"
    never unexpectedly changes the dynamic resolution process
    due to a naming conflict across resources. Users of
    "$dynamicAnchor" are expected to be aware of the possibility
    of such name collisions, while users of "$anchor" are not.

But, that doesn't make sense. I don't see how `$anchor` could effect the dynamic resolution process nor what effect bookending could have in stabilizing the process.

I'll follow up shortly with a couple examples of how the bookending requirement is harmful.

(This was originally reported in https://github.com/json-schema-org/json-schema-spec/issues/1030 and split off here)

jdesrosiers:
Here's an example of a non-recursive use for `$dynamicAnchor`/`$dynamicRef` where it doesn't make sense to require a `$dynamicAnchor` at the initially resolved starting point.

Hyper-schemers will recognize this pattern. Let's say I have several resources and for each resource, I want two schemas: one for a single instance and another for a list of instances. The list of instances follows a common pattern that looks something like this.

```json
{
  "$id": "https://json-schema.hyperjump.io/schema/list",
  "$schema": "https://json-schema.org/draft/2020-12/schema",

  "type": "object",
  "properties": {
    "list": { "type": "array" },
    "nextPage": { "type": "integer" },
    "previousPage": { "type": "integer" },
    "perPage": { "type": "integer" },
    "page": { "type": "integer" }
  }
}
```

I can then compose this schema with another to create a list schema for a specific resource.

```json
{
  "$id": "https://json-schema.hyperjump.io/schema/foo-list",
  "$schema": "https://json-schema.org/draft/2020-12/schema",

  "$ref": "/schema/list",

  "properties": {
    "list": {
      "items": { "$ref": "/schema/foo" }
    }
  }
}
```

[...]

PR #1139, opened 2021-10-19
Remove bookending requirement for dynamicRef

jdesrosiers:
Resolves https://github.com/json-schema-org/json-schema-spec/issues/1064

Discussion about changes to `$dynamicRef` behavior are still on going, but I want to capture the part that we did get consensus on, which is to remove the bookending requirement.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3858,6 +3858,7 @@ https://example.com/schemas/common#/$defs/count/minimum
                     <t hangText="draft-bhutton-json-schema-next">
                         <list style="symbols">
                             <t>"contains" now applies to objects as well as arrays</t>
+                            <t>Remove bookending requirement for "$dynamicRef"</t>
                         </list>
                     </t>
                     <t hangText="draft-bhutton-json-schema-00">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1483,27 +1483,22 @@
                         <t>
                             Together with "$dynamicAnchor", "$dynamicRef" implements a cooperative
                             extension mechanism that is primarily useful with recursive schemas
-                            (schemas that reference themselves).  Both the extension point and the
-                            runtime-determined extension target are defined with "$dynamicAnchor",
-                            and only exhibit runtime dynamic behavior when referenced with
-                            "$dynamicRef".
+                            (schemas that reference themselves).  The extension point is defined with
+                            "$dynamicAnchor" and only exhibits runtime dynamic behavior when referenced
+                            with "$dynamicRef".
                         </t>
                         <t>
-                            The value of the "$dynamicRef" property MUST be a string which is
-                            a IRI-Reference.  Resolved against the current IRI base, it produces
-                            the IRI used as the starting point for runtime resolution.  This initial
-                            resolution is safe to perform on schema load.
+                            The value of the "$dynamicRef" property MUST be a string which is a
+                            IRI-Reference that contains a valid <xref target="anchor">plain name
+                            fragment</xref>.  Resolved against the current IRI base, it indicates
+                            the schema resource used as the starting point for runtime resolution.
+                            This initial resolution is safe to perform on schema load.
                         </t>
                         <t>
-                            If the initially resolved starting point IRI includes a fragment that
-                            was created by the "$dynamicAnchor" keyword, the initial IRI MUST be
-                            replaced by the IRI (including the fragment) for the outermost schema
-                            resource in the <xref target="scopes">dynamic scope</xref> that defines
-                            an identically named fragment with "$dynamicAnchor".
-                        </t>
-                        <t>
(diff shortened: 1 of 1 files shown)


jdesrosiers:
Force pushed to fix conflicts caused by merging IRI changes.

jsonschema-core.xml
```diff
@@ -1483,27 +1483,22 @@
...
-                            If the initially resolved starting point IRI includes a fragment that
-                            was created by the "$dynamicAnchor" keyword, the initial IRI MUST be
-                            replaced by the IRI (including the fragment) for the outermost schema
-                            resource in the <xref target="scopes">dynamic scope</xref> that defines
-                            an identically named fragment with "$dynamicAnchor".
-                        </t>
-                        <t>
-                            Otherwise, its behavior is identical to "$ref", and no runtime
-                            resolution is needed.
+                            The schema to apply is the outermost schema resource in the
```
erosb, line 1498: The word "outermost" is confusing for me. Is it outermost no a sense that it is the closest (by nesting level) to the root schema? Or is it the $dynamicAnchor that was found first on the dynamic path?

jsonschema-core.xml
```diff
@@ -1483,27 +1483,22 @@
...
                         </t>
                         <t>
-                            The value of the "$dynamicRef" property MUST be a string which is
-                            a IRI-Reference.  Resolved against the current IRI base, it produces
-                            the IRI used as the starting point for runtime resolution.  This initial
-                            resolution is safe to perform on schema load.
+                            The value of the "$dynamicRef" property MUST be a string which is a
+                            IRI-Reference that contains a valid <xref target="anchor">plain name
+                            fragment</xref>.  Resolved against the current IRI base, it indicates
+                            the schema resource used as the starting point for runtime resolution.
```
erosb, line 1494: This is unclear. So the $dynamicRef value gets resolved against the base IRI and we get an absolute IRI. This absolute IRI can be resolved to a schema resource (= object?). What is it a starting point for?

Instead of "starting point" , is it accurate to say that the dynamic resolution should use this initially resolved schema as the *default* product of the $dynamicRef lookup in cases when there is no matching $dynamicAnchor on the dynamic path? If yes, then please use this wording, would be much more understandable.

jsonschema-core.xml
```diff
@@ -1483,27 +1483,22 @@
                         <t>
                             Together with "$dynamicAnchor", "$dynamicRef" implements a cooperative
                             extension mechanism that is primarily useful with recursive schemas
-                            (schemas that reference themselves).  Both the extension point and the
-                            runtime-determined extension target are defined with "$dynamicAnchor",
-                            and only exhibit runtime dynamic behavior when referenced with
-                            "$dynamicRef".
+                            (schemas that reference themselves).  The extension point is defined with
```
erosb, line 1486: From my perspective the "extension point" phrase is misleading. In my understading the extension point is the $dynamicRef, since that is the point where a customizable/abstract schema makes itself open for extension. The $dynamicAnchor is the place where this extension point gets utilized, and the abstraction (the $dynamicRef attribute's value, which doesn't have semantics on its own) gets bound to a concrete value (the schema supplied by $dynamicAnchor).

I hope this makes sense. Please consider re-wording this section accordingly.

jdesrosiers: I think the "extension point" language came from `$recursiveRef` when the only use case was extending recursive schemas. I think you're right that "extension point" doesn't make sense, at least not anymore. But, this has nothing to do with the change in this PR. I suggest a separate issue/PR to discuss improvements to how we express this feature. We can even get those improvements into the patch release ahead of this one.

jsonschema-core.xml
jdesrosiers: I think this is more clear with a little more familiarity with some of the newer concepts in the spec. Here are a few sections you should (re-)read.

> A JSON Schema resource is a schema which is canonically identified by an absolute URI. - https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.4.3.5

> The outermost dynamic scope is the schema object at which processing begins - https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.7.1

I always thought "starting point" doesn't make sense. It's more like the "ending point". But again, this isn't relevant to the change in this PR. I suggest a new issue/PR to discuss improving the wording here.

jsonschema-core.xml
jdesrosiers: > The outermost dynamic scope is the schema object at which processing begins - https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.7.1

Relequestual (approved):

jsonschema-core.xml
Relequestual: I think it does still make sense if you consider an "extension point" is the point at which you're extending the schema.

The anchor is the location you wish to make extensible. It feels odd because you reference it, but you're really extending that anchor location's schema.

PR #1142, opened 2021-10-21
Drop the initial resolve step for dynamic references

### LINKED ISSUE #1140: Proposal: Remove initial resolution step for dynamic references
This issue is a spin-off of https://github.com/json-schema-org/json-schema-spec/issues/1064. That conversation drifted a bit, so I'm moving some of the proposed amendments to that proposal into a separate issue.

Removing the bookending requirement for dynamic references opens the door for some additional considerations. `$recrusiveRef` didn't allow non-fragment URI parts. `$dynamicRef` added that in order to allow using `$dynamicRef` in a non-recursive schema. However, without the bookending requirement, the initial resolution step is no longer necessary (see https://github.com/json-schema-org/json-schema-spec/issues/1064#issuecomment-761118502). Since it's not necessary, there exists the possibility that a `$dynamicRef` doesn't need to be a URI at all. It makes sense for it to just be the dynamic anchor name. There's no reason to encode it into a URI. (Credit to @ssilverman for this idea) (See also https://github.com/json-schema-org/json-schema-spec/issues/1064#issuecomment-807717544).

We could take this a step further and make dynamic anchors URIs instead of plain name identifiers. Because dynamic anchors aren't scoped to the local document like traditional anchors, there's a possibility of name conflicts between schemas, so using URI identifiers would be safer. It's a similar concept to using URIs for link relations (for those who are familiar with those).

jdesrosiers:
Fixes #1140 

Without bookending or initial resolution, describing how dynamic references work becomes much more straightforward.

This is provided as a draft PR just to show what the change might look like if agreed upon. Any discussion about whether or not to make this change should happen in #1140.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1510,22 +1510,15 @@
                         <t>
                             Together with "$dynamicAnchor", "$dynamicRef" implements a cooperative
                             extension mechanism that is primarily useful with recursive schemas
-                            (schemas that reference themselves).  The extension point is defined with
-                            "$dynamicAnchor" and only exhibits runtime dynamic behavior when referenced
-                            with "$dynamicRef".
+                            (schemas that reference themselves).  The extension point is defined
+                            with "$dynamicAnchor", and only exhibits runtime dynamic behavior when
+                            referenced with "$dynamicRef".
                         </t>
                         <t>
-                            The value of the "$dynamicRef" property MUST be a string which is a
-                            IRI-Reference that contains a valid <xref target="anchor">plain name
-                            fragment</xref>.  Resolved against the current IRI base, it indicates
-                            the schema resource used as the starting point for runtime resolution.
-                            This initial resolution is safe to perform on schema load.
-                        </t>
-                        <t>
-                            The schema to apply is the outermost schema resource in the
-                            <xref target="scopes">dynamic scope</xref> that defines a
-                            "$dynamicAnchor" that matches the plain name fragment in the initially
-                            resolved IRI.
+                            The value of the "$dynamicRef" property MUST be a valid
+                            <xref target="anchor">plain name fragment</xref>. The schema to apply is
+                            the outermost schema resource in the <xref target="scopes">dynamic scope</xref>
+                            that defines a "$dynamicAnchor" matching the value of this keyword.
                         </t>
                         <t>
                             For a full example using these keyword, see appendix


jdesrosiers:

jsonschema-core.xml
```diff
@@ -1510,22 +1510,15 @@
...
-                            "$dynamicAnchor" and only exhibits runtime dynamic behavior when referenced
-                            with "$dynamicRef".
+                            (schemas that reference themselves).  The extension point is defined
+                            with "$dynamicAnchor", and only exhibits runtime dynamic behavior when
+                            referenced with "$dynamicRef".
                         </t>
                         <t>
-                            The value of the "$dynamicRef" property MUST be a string which is a
-                            IRI-Reference that contains a valid <xref target="anchor">plain name
-                            fragment</xref>.  Resolved against the current IRI base, it indicates
```
gregsdennis, line 1520: We still need the first sentence, which describes the value form.

jsonschema-core.xml
```diff
@@ -1510,22 +1510,15 @@
                         <t>
                             Together with "$dynamicAnchor", "$dynamicRef" implements a cooperative
                             extension mechanism that is primarily useful with recursive schemas
-                            (schemas that reference themselves).  The extension point is defined with
-                            "$dynamicAnchor" and only exhibits runtime dynamic behavior when referenced
-                            with "$dynamicRef".
+                            (schemas that reference themselves).  The extension point is defined
+                            with "$dynamicAnchor", and only exhibits runtime dynamic behavior when
```
gregsdennis, line 1514: The comma isn't necessary in the current form.  If you want to include a comma, the second half of the sentence needs to be a complete sentence.

```suggestion
                            with "$dynamicAnchor", and it only exhibits runtime dynamic behavior when
```

jsonschema-core.xml
```diff
@@ -1510,22 +1510,15 @@
...
+                            referenced with "$dynamicRef".
```
gregsdennis, line 1515: In order to remove the "`$dynamicAnchor` produces a `$ref`-able anchor" behavior we, need to adjust this sentence.  We'll likely need to make changes in the `$dynamicAnchor` section as well.

gregsdennis:
@jdesrosiers do you want to pick this up again?

jdesrosiers:
Yes, this needs to get done before we can release, but I don't have the capacity to work on this at the moment.

gregsdennis:
Closing in favor of ☝️

Issue #1112, opened 2021-06-06
Suggestion: optional keyword to mirror required


Altreus:
I don't write objects with optional properties most of the time. I tend to prefer to use polymorphism to allow one of several data structures. When I do have an optional field, it's often something like "comments" or "extra info", and even then I don't have any philosophical objection to simply making it required but allowing it to contain the empty string.

An `optional` keyword to complement `required` would greatly benefit me (and hopefully others), without treading on the toes of people who are writing schemata with many optional values.

My proposition is thus:

- Either `required` or `optional` may be provided
- It is an error to provide both
- The default remains as it is (everything optional), which is the same as providing `required` with an empty array
- If `optional` is provided, anything *not* mentioned is now required
- Passing an empty array to `optional` means everything is required

Hope that's both clear and a useful idea.

gregsdennis:
This seems like a duplicate/variant of #1096.

jdesrosiers:
I can see how this could work. It would work similarly to `additionalProperties` in that it would collect the property names declared by `properties` and use that to make its assertion. The only change I would make to the proposal is to not forbid `required` when using `optional`. That would make a good linting rule, but it's unnecessary from JSON Schema architecture point of view.

Something like this has been requested for a long time, but this is the first proposal of this type that I think fits well into the JSON Schema architecture and I think we should consider it. Requiring properties by default (#1096) and `requiredProperties` (#846) are problematic because they assert more than one thing: (1) a property schema and (2) that the property is required. This proposal is better because it deals only with which properties are required.

Although I hate to add more keywords that break keyword independence, I think this is a good solution. In particular, this wouldn't have the same kind of problems with combining schemas that `additionalProperties` has.

karenetheridge:
> The only change I would make to the proposal is to not forbid required when using optional. That would make a good linting rule, but it's unnecessary from JSON Schema architecture point of view.

Agreed. But this also means we need to work out what happens when they are used together: for example if one schema pulls in another with `$ref`, and the reffed schema contains one of these keywords and the first schema contains the other. Would 'required' always overrule 'options'? Or the other way around? or should the innermost keyword always overrule the outermost (akin to a subclass overriding a superclass)?

jdesrosiers:
> we need to work out what happens when they are used together

[...]

Issue #846, opened 2020-02-04
"requiredProperties" keyword


awwright:
When accepting JSON documents for submission, it is frequently the case the documents are objects that have at least one property that is required, such as an ID or name field.

JSON Schema should have a "requiredProperties" keyword that is similar to `"properties"` in that it defines a sub-schema to apply to the respective properties; but in addition, each of those properties must exist, similar to the `"required"` keyword.

This:

```javascript
{
  "requiredProperties": {
    "name": { "type": "string" }
  }
}
```

Would be the same as this:

```javascript
{
  "required": ["name"],
  "properties": {
    "name": { "type": "string" }
  }
}
```

Alternatively, this functionality could be built into the "required" keyword (so that "required" accepts an array or an object). This may even be preferable, since older validators will (usually) bail if they see an object where an array is expected:

```javascript
{
  "required": {
    "name": { "type": "string" }
  }
}
```

This issue is intended to restate and close #659 #681 #734

gregsdennis:
Is this really the right direction?

I started working with JSON Schema at draft 4, so I don't have any details, but does anyone know why `required` was removed from the property schema into the `required` property when moving from draft 3 -> 4?

Some insight into that decision would probably be helpful here.

awwright:
@gregsdennis Boolean "required" didn't work the same way that other keywords worked; it only had meaning in a subschema in certain keywords, mostly "properties" and "items". It was nonsensical if used inside e.g. "patternProperties", "additionalItems", "additionalProperties", and there's even more considerations like what if you use it inside "oneOf" or "not"?

It also made it more difficult to create composable schemas; referenced a schema that had `required: true` it could cause a little bit of a problem depending on where the $ref was located.

I actually argued against it at the time because it was a nice authoring convenience compared to the array form; but I think this is superior to boolean "required" by far.

(rewritten from https://github.com/json-schema-org/json-schema-spec/issues/734#issuecomment-487906654 )

handrews:
While I'd be strongly against changing `required` to take an object form, I'm at least tentatively open to the idea of `requiredProperties` as a new keyword.  Not sold, but keeping an open mind.

I want to keep the keyword taxonomy clear, in terms of which keywords behave as applicators, assertions, or annotations.

[...]

PR #1144, opened 2021-11-01
Add requireAllExcept keyword

jdesrosiers:
Resolves https://github.com/json-schema-org/json-schema-spec/issues/1112

Adds the `requireAllExcept` keyword.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -493,6 +493,18 @@
                     </t>
                 </section>
 
+                <section title="requiredAllExcept">
+                    <t>
+                        The value of this keyword MUST be an array.
+                        Elements of this array, if any, MUST be strings, and MUST be unique.
+                    </t>
+                    <t>
+                        An object instance is valid against this keyword if every property name
+                        declared in "properties" except those that appear in the array is the name
+                        of a property in the instance.
+                    </t>
+                </section>
+
                 <section title="dependentRequired">
                     <t>
                         The value of this keyword MUST be an object.  Properties in
@@ -1445,6 +1457,7 @@
                     <t hangText="draft-next">
                         <list style="symbols">
                             <t>Use IRIs instead of URIs</t>
+                            <t>Add the "requireAllExcept" keyword</t>
                         </list>
                     </t>
                     <t hangText="draft-bhutton-json-schema-validation-01">
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -500,8 +500,8 @@
                     </t>
                     <t>
                         An object instance is valid against this keyword if every property name
-                        declared in "properties" except those that appear in the array is the name
-                        of a property in the instance.
+                        declared in "properties" is the name of a property in the instance, with
+                        the exception of the property names that appear in this keyword's array.
                     </t>
                 </section>
 
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -500,8 +500,9 @@
                     </t>
                     <t>
                         An object instance is valid against this keyword if every property name
-                        declared in "properties" is the name of a property in the instance, with
-                        the exception of the property names that appear in this keyword's array.
+                        declared in "properties" within the same schema object is the name of
+                        a property in the instance, with the exception of the property names
+                        that appear in this keyword's array.
                     </t>
                 </section>
 
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -225,7 +225,11 @@
                     <list>
                         <t>
                             "requireAllExcept", whose behavior is defined in terms of "properties"
+                        </t>
+                        <t>
                             "minContains", whose behavior is defined in terms of "contains"
+                        </t>
+                        <t>
                             "maxContains", whose behavior is defined in terms of "contains"
                         </t>
                     </list>
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -225,6 +225,8 @@
                     <list>
                         <t>
                             "requiredAllExcept", whose behavior is defined in terms of "properties"
+                            "minContains", whose behavior is defined in terms of "contains"
+                            "maxContains", whose behavior is defined in terms of "contains"
                         </t>
                     </list>
                 </t>
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -214,6 +214,22 @@
                 <eref target="https://json-schema.org/draft/next/meta/validation"/>.
             </t>
 
+            <section title="Keyword Independence">
+                <t>
+                    Schema keywords typically operate independently, without
+                    affecting each other's outcomes.
+                </t>
+                <t>
+                    For schema author convenience, there are some exceptions among the
+                    keywords in this vocabulary:
+                    <list>
+                        <t>
+                            "requiredAllExcept", whose behavior is defined in terms of "properties"
+                        </t>
+                    </list>
+                </t>
+            </section>
+
             <section title="Validation Keywords for Any Instance Type" anchor="general">
                 <section title="type">
                     <t>
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -224,7 +224,7 @@
                     keywords in this vocabulary:
                     <list>
                         <t>
-                            "requiredAllExcept", whose behavior is defined in terms of "properties"
+                            "requireAllExcept", whose behavior is defined in terms of "properties"
                             "minContains", whose behavior is defined in terms of "contains"
                             "maxContains", whose behavior is defined in terms of "contains"
                         </t>
@@ -511,7 +511,7 @@
                     </t>
                 </section>
 
-                <section title="requiredAllExcept">
+                <section title="requireAllExcept">
                     <t>
                         The value of this keyword MUST be an array.
                         Elements of this array, if any, MUST be strings, and MUST be unique.
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -515,10 +515,12 @@
                         Elements of this array, if any, MUST be strings, and MUST be unique.
                     </t>
                     <t>
-                        An object instance is valid against this keyword if every property name
-                        declared in "properties" within the same schema object is the name of
-                        a property in the instance, with the exception of the property names
-                        that appear in this keyword's array.
+                        This keyword specifies properties that are required based on the property
+                        names declared in a "properties" keyword within the same schema object.
+                        The set of required properites is the set of property names declared in
+                        "properties" minus the set of property names declared in this keyword's
+                        array. An object instance is valid against this keyword if it has defined
+                        properties for each of the property names in this set of required properties.
                     </t>
                 </section>
 
--- a/meta/validation.json
+++ b/meta/validation.json
@@ -61,6 +61,7 @@
         "maxProperties": { "$ref": "#/$defs/nonNegativeInteger" },
         "minProperties": { "$ref": "#/$defs/nonNegativeIntegerDefault0" },
         "required": { "$ref": "#/$defs/stringArray" },
+        "allRequiredExcept": { "$ref": "#/$defs/stringArray" },
         "dependentRequired": {
             "type": "object",
             "additionalProperties": {
--- a/meta/validation.json
+++ b/meta/validation.json
@@ -61,7 +61,7 @@
         "maxProperties": { "$ref": "#/$defs/nonNegativeInteger" },
         "minProperties": { "$ref": "#/$defs/nonNegativeIntegerDefault0" },
         "required": { "$ref": "#/$defs/stringArray" },
-        "allRequiredExcept": { "$ref": "#/$defs/stringArray" },
+        "requireAllExcept": { "$ref": "#/$defs/stringArray" },
         "dependentRequired": {
             "type": "object",
             "additionalProperties": {


jsonschema-validation.xml
```diff
@@ -491,6 +491,18 @@
...
 
+                <section title="requiredAllExcept">
+                    <t>
+                        The value of this keyword MUST be an array.
+                        Elements of this array, if any, MUST be strings, and MUST be unique.
+                    </t>
+                    <t>
+                        An object instance is valid against this keyword if every property name
+                        declared in "properties" except those that appear in the array is the name
+                        of a property in the instance.
```
karenetheridge, line 502: This wording could be improved:

An object instance is valid against this keyword if every property name
declared in "properties" is the name of a property in the instance, with
the exception of those that appear in this keyword's array.

karenetheridge (changes requested):

jsonschema-validation.xml
karenetheridge: Perhaps we could also mention one usecase that could become a common shortcut: `"requiredAllExcept": []` indicates that all properties in `properties` are required.

karenetheridge (changes requested):

jsonschema-validation.xml
jdesrosiers: Yeah, that wording sounds better. I'll update it.

I'd rather not include usage patterns in the spec. I don't think we do that anywhere else, so it would be a little out of place. It should definitely be mentioned in UJS tho.

karenetheridge:
> I don't think we do that anywhere else, so it would be a little out of place.

Fair enough!

Relequestual (approved):

awwright:
I find this keyword somewhat difficult to comprehend. It's behavior is defined in terms of another keyword, which I think we should avoid. It becomes non-intuitive when mixing "allOf", because this keyword does not "see into" sub-schemas.

What would be some cases where this keyword is superior to ["requiredProperties"](https://github.com/json-schema-org/json-schema-spec/issues/846)?

Relequestual:
> I find this keyword somewhat difficult to comprehend. It's behavior is defined in terms of another keyword, which I think we should avoid. It becomes non-intuitive when mixing "allOf", because this keyword does not "see into" sub-schemas.

I don't feel this is a problem. Many keywords have dependencies. That's one of the reason for the annotations system.

> What would be some cases where this keyword is superior to ["requiredProperties"](https://github.com/json-schema-org/json-schema-spec/issues/846)?

@awwright 
[Earlier](https://github.com/json-schema-org/json-schema-spec/issues/1112#issuecomment-864522338) you said...
> The problem would be if keywords do multiple things and there's no way around it.

By my understanding of your proposal #846, schema authors could use both `properties` and `requiredProperties` to define properties. `requiredProperties` would be doing multiple things... defining subschemas to apply, and specifying they are required.

Generally we've moved away from keywords doing multiple things, splitting them up, mainly because it was confusing. `requiredProperties` as you proposed in #846 feels like a step backwards.

awwright:
@Relequestual By "... and there's no way around it", I mean: It's fine if a keyword performs multiple things as an author convenience. But we have to be able to break it apart.

e.g. `{"type": "integer"}` is short for `{ "type": "number", "multipleOf": 1 }`.

Likewise, `{"requiredProperties": {"foo": bar}}` could be short for `{ "required": ["foo"], "properties": { "foo": bar } }`.

This keyword, by contrast, has interactions with other keywords. What happens if I use:

```javascript
{
"requireAllExcept": ["foo"],
"allOf": [ { "foo": false } ],
"patternProperties": { "f": true },
"additionalProperties": false
}
```

What's the behavior here? There probably is one—but I have to think about it. And I can't really take an educated guess. I can't even come up with a good example, maybe if I replaced `false` with `true` here, or vice-versa, the example would illustrate my point better.

It's true that we have other keywords like this (additionalProperties), but they tend to consume a disproportionate amount of descriptive effort we have to do.

> Generally we've moved away from keywords doing multiple things, splitting them up, mainly because it was confusing. 

This feels like _exactly_ the kind of keyword that does multiple things you're talking about.

jsonschema-validation.xml
```diff
@@ -491,6 +491,18 @@
...
 
+                <section title="requiredAllExcept">
+                    <t>
+                        The value of this keyword MUST be an array.
+                        Elements of this array, if any, MUST be strings, and MUST be unique.
+                    </t>
+                    <t>
+                        An object instance is valid against this keyword if every property name
+                        declared in "properties" is the name of a property in the instance, with
+                        the exception of the property names that appear in this keyword's array.
```
Relequestual, line 502: I think we need to clarify WHICH declaration of `properties` this relates to.

```suggestion
                        An object instance is valid against this keyword if every property name
                        declared in "properties" within the same schema object is the name of 
                        a property in the instance, with the exception of the property names 
                        that appear in this keyword's array.
```

Relequestual (changes requested):
In reviewing this further, I think some clarification is required.
I've left a suggested change.

awwright (changes requested):
I overall object to this because it "reads" the value of other keywords, which is something we try to avoid, see "[Keyword independence](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.1)".

However if this keyword _absolutely must_ be added, we would need to list this in said "Keyword independence" section. (Or rather, since this keyword does nothing if in a schema all by itself, it may be more consistent to say the "properties" keyword reads this one.)

gregsdennis:
@awwright what about `additionalProperties`, `unevaluated*`, `minCount`, `maxCount`?  These are all keywords that consider other keywords.  While we have keyword independence, it's not a hard requirement for keywords.

awwright:
@gregsdennis Those keywords are specifically listed as exceptions... I mean, yes, we never really went into detail as to _why_ they're exceptions... but as an exercise, write an alternative where the only information available to the keyword is its own value.

The best I can do is something like

```javascript
{
   "contains": {
      "needle": "foo",
      "min": 1,
      "max": 4
   }
}
```

While this nicely emphasizes how "minContains" doesn't do anything by itself—that it's technically an argument to "contains"—this is obviously clunky to write, which is why I think an "argument keyword" is acceptable as an authoring convenience.

In contrast, the single-keyword alternative to "requireAllExcept" combines the functionality of "required" and "properties" into a single keyword, entirely replacing the need for "required"; all the while "properties" can maintain its usual function for listing optional properties that "requireAllExcept" keyword is doing.

This idea that some keywords can be decomposed into more complicated but equal schemas should be familiar, e.g. `type: "integer"` and "dependentRequired".



※ A keyword "doesn't anything by itself" if the single-keyword schema `{keyword: foo}` is indistinguishable from `{}`, for all  values of foo. (As for "additionalProperties", this does do something by itself, but perhaps it should have had the semantics that "unevaluatedProperties" has had from the beginning... or maybe something else entirely, I'm still researching this.)

gregsdennis:
> the single-keyword alternative to "requireAllExcept" combines the functionality of "required" and "properties" into a single keyword, entirely replacing the need for "required"

I don't follow this.

The new keyword is an array of property names that are optional. The presence of the keyword implies that all properties in `properties` are required except the ones in the array.

It's not combining `properties` and `required`; it's the inversion of `required`.

gregsdennis:
With this approach you also need to update `additionalProperties` and `unevaluatedProperties` to look into the new `requiredProperties`.

awwright:
@gregsdennis Yes— this is sort of implied when I say "requiredProperties" would 'behave the same way as using "required" and "properties"'—but for consistency you are correct, it would be a good idea to update the definitions of those keywords.

awwright:
Also, to bring this back to my original point, "requireAllExcept" also needs to update some other paragraphs, specifically the section(s) that discuss keyword independence, so it can be added as one of the exceptions.

Relequestual:
Reading the more recent comments, I think we need to re-evaluate our approach here.

I know we're not all happy with the annotations system as defined, however I don't think that means we intend to throw it out. I feel that we can still use it, better defined, and any "keyword interactions" should rely on the premise annotation collection.

I'm VERY strongly opposed to adding another way to list properties which effects `additionalProperties` and `unevaluatedProperties`. I feel it will make things all the more confusing. I don't feel the trade-off is worth it.

Adding new constraint base keywords without effecting existing keywords feels preferable to me.

awwright:
> I feel that we can still use it, better defined, and any "keyword interactions" should rely on the premise annotation collection.

Can you detail what this would look like?

> I'm VERY strongly opposed to adding another way to list properties which effects additionalProperties and unevaluatedProperties. I feel it will make things all the more confusing. I don't feel the trade-off is worth it.

We're dealing with an authoring convenience, i.e. how to combine common patterns of keywords into a single keyword. There's going to be inherent complexity in that, regardless of the solution.

There's also some amount of subjectivity. We're going to have to balance what new people would expect, with what keeps the cognitive requirements low on very large schemas. (Also, these things might be the same.) I think all authoring conveniences will introduce some "surprise", but we can minimize that; the real purpose of an authoring convenience is it lets you build more complicated schemas on the same amount of brain power (it is probably more difficult to work with an if/then statement than to think "_x_ property means _y_ becomes required" — and so we have "dependentRequired").

We may want to put out a survey that tries to measure these two properties (expected/surprising behavior, and scalable/non-scalable for authors).

> ... without effecting existing keywords feels preferable to me.

I don't think there's a way to do this.

You can think of "requireAllExcept" as a keyword that first reads the value of "properties", then returns a validation result; or you can think of it as an argument to "properties" (which currently reads no arguments); these two interpretations are logically indistinguishable. However I'm inclined to say we should think of it only an argument to "properties", because the single-keyword schema `{ "requireAllExcept": anything }` has no behavior.

(An alternative name for these "argument keywords" could be "interacts with"... because even though "minContains" is an argument keyword to "contains", it still makes sense to talk about "minContains" as the source of validation errors. It just won't do so when it's the only keyword in the schema.)

Relequestual:
> Can you detail what this would look like?

Yes, but I'll have to get back to you. I'm overflowing with half done work right now 😭

jdesrosiers:
> @awwright we would need to list this in said "Keyword independence" section.

Agreed. This was an oversight. I updated the PR.

jsonschema-validation.xml
```diff
@@ -213,6 +213,22 @@
...
+                <t>
+                    Schema keywords typically operate independently, without
+                    affecting each other's outcomes.
+                </t>
+                <t>
+                    For schema author convenience, there are some exceptions among the
+                    keywords in this vocabulary:
+                    <list>
+                        <t>
+                            "requiredAllExcept", whose behavior is defined in terms of "properties"
```
gregsdennis, line 226: If we're going to have a list, we should probably add `minContains` and `contains`, `additionalProperties`, and any others that have dependencies.

jsonschema-validation.xml
```diff
@@ -491,6 +507,19 @@
                     </t>
                 </section>
 
+                <section title="requiredAllExcept">
+                    <t>
+                        The value of this keyword MUST be an array.
+                        Elements of this array, if any, MUST be strings, and MUST be unique.
```
gregsdennis, line 513: Why not require a minimum of 1, like `required`?

jdesrosiers:
@awwright 
> "minContains" doesn't do anything by itself [] it's technically an argument to "contains"

This is not the way `minContains` and `maxContains` are defined. They assert independently. For example, if you have schema `{ "contains": { "type": "number" }, "maxContains": 1 }` and instance `[1, 2]`, then `contains` will pass and `maxContains` will fail. Personally, I'd prefer if `minContains` and `maxContains` were just arguments to `contains`, but that's not the way it was defined.

gregsdennis (changes requested):

jsonschema-validation.xml
jdesrosiers: Currently there is "Keyword Independence" section for each vocabulary that includes keyword independence exceptions. The validation vocabulary didn't have one yet, so I added it. All the other keyword exceptions are mentioned in their respective vocabularies.

jsonschema-validation.xml
jdesrosiers: An empty array is useful because it means all properties are required. If we have a minimum of 1, then we couldn't express that case.

jsonschema-validation.xml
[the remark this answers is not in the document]
jdesrosiers: I see your point. I'll try to reword.

jdesrosiers:
@awwright 
> "requireAllExcept" is a "keyword" but it's not a validation keyword per se, it's an argument keyword: It's modifying the behavior of another (validation) keyword.

This isn't correct. `requireAllExcept` does it's own assertion. It does not affect the validation result of `properties`. You could say that `properties` is an argument of `requireAllExcept`, but not the other way around.

awwright:
> `requireAllExcept` does it's own assertion

@jdesrosiers I addressed this at the end of https://github.com/json-schema-org/json-schema-spec/pull/1144#issuecomment-1111430170; by "per se," I mean "by itself." Since JSON Schema is declarative, there's multiple equivalent ways validation could be performed in actuality, (1) the way I'm thinking about it where you perform the validation at the moment you're iterating through the keywords in "properties" and making sure that each property is either in the instance, or listed in "requireAllExcept"; or (2) your implementation where you validate it at the time the "requireAllExcept" keyword is encountered.

((3) You can also write a validator that performs all the validations _at the same time_, and you can do so deterministically, proving that "where" the validation occurs truly doesn't matter.)

What makes a keyword an "argument keyword" is that a schema consisting only of e.g. `"requireAllExcept"` will never be invalid.

awwright:
Aside: By "at the same time", I mean you can compile a validator for a good number of schemas down to a finite state machine; since common states are factored out, it doesn't make sense to consider a failure to be "from" one keyword or another (the best you can do is say this failure occurs because of the _existence_ of a keyword in the schema, and would have been valid if not for its existence).

jsonschema-validation.xml
karenetheridge: required` hasn't required an item since draft4.

karenetheridge:
I've seen implementations that evaluate all object-based keywords together, all array- together, etc. I'm not a fan of this approach, and I chose to implement each keyword separately in its own function, because it made it easier to selectively enable/disable individual keywords depending on which version of the specification was active -- but what matters is the outcome. As long as the validation result and emitted errors/annotations are correct, do whatever makes sense for your mental model and your choice of language/architecture.

This is, FWIW, why I think we have such divisive arguments about evaluation behaviour sometimes -- the implementation choices and mental models vary quite widely, and this informs our beliefs about how new features ought to work.

jsonschema-validation.xml
gregsdennis: How does this fit in, then? Are there no other dependent keywords in this vocabulary?

karenetheridge: `minContains` and `maxContains` are in this vocabulary, so they ought to be listed here.

jsonschema-validation.xml
```diff
@@ -491,6 +507,19 @@
...
+                    </t>
+                    <t>
+                        An object instance is valid against this keyword if every property name
+                        declared in "properties" within the same schema object is the name of
```
karenetheridge, line 517: We should consider the advantages/disadvantages of having this keyword use annotations from `properties`, which would allow it to use property definitions from other schemas via composition.

I could see this pattern being quite useful:

```
  $defs:
    model.user:
      type: object
      properties:
        id:
          type: string
        name:
          type: string
        type:
          enum: [foo, bar]
        comment:
          type: [string, 'null']
    create_operation:
      $ref: '#/$defs/model.user'
      requireAllExcept: [id, comment]
    get_operation:
      $ref: '#/$defs/model.user'
      requireAllExcept: []
```

meta/validation.json
```diff
@@ -61,6 +61,7 @@
         "maxProperties": { "$ref": "#/$defs/nonNegativeInteger" },
         "minProperties": { "$ref": "#/$defs/nonNegativeIntegerDefault0" },
         "required": { "$ref": "#/$defs/stringArray" },
+        "allRequiredExcept": { "$ref": "#/$defs/stringArray" },
```
karenetheridge, line 64:
```suggestion
        "requiredAllExcept": { "$ref": "#/$defs/stringArray" },
```

jsonschema-validation.xml
handrews: _[briefly de-lurks]_
@karenetheridge it's worth considering that these might be two separate use cases.  "require all of these properties right here" is a straightforward thing that would become hard if this always used annotations from applicators in subschemas.  And trying to make it optional within the keyword would introduce more complex syntax.
_[re-lurks]_

jsonschema-validation.xml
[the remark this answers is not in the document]
jdesrosiers: I pushed some alternate wording.

jsonschema-validation.xml
jdesrosiers: `minContains` and `maxContains` are now included as well.

meta/validation.json
jdesrosiers: Good catch. Thanks.

jdesrosiers:
@awwright 
> What makes a keyword an "argument keyword" is that a schema consisting only of e.g. `"requireAllExcept"` will never be invalid.

I see what you're trying to say, but I don't think `requireAllExcept` has that behavior because it's an argument keyword. I think it's an accidental convergence of behavior. By your definition, `additionalProperties` would not be an argument keyword and `requireAllExcept` would despite that the keywords work exactly the same way.

I think it would be fair to call something an argument keyword if it has no meaning if not in the presence of another keyword (Examples: `minContains`, `then`). This is not the case for `requireAllExcept`. Using it in a schema by itself still has meaning, just not a very useful meaning. It's equivalent to `{ "required": [] }`. The schema would always validate, but not because it's ignored. There just happens to not be a value that will make the assertion false.

---

Let me take a step back and acknowledge that I share your concerns about adding a new keyword that breaks keyword independence. In fact, I hate it. I'd rather be finding ways to remove or redefine all such keywords, but weighing the pros and cons, I think `requireAllExcept` is the least bad option right now. You can look at the same set of tradeoffs and come to a different conclusion just by weighing things differently than I do. That's fine. Here's a list of the things I've taken into account before championing this approach (I'm sure I'm forgetting some, but this should be the most important points).

1. The problem of maintaining schemas with a long list of required properties is one of the most common criticisms we get about JSON Schema.
2. `requireAllExcept` doesn't solve the problem of duplicating property names declared in `properties` in all cases, but it does reduce the burden. `requiredProperties` doesn't require duplicating property names.
4. `requiredProperties` would create two ways to define properties.
5. `requiredProperties` does more than one thing and is both an applicator and an assertion.
6. `requiredProperties` would require changing the definitions of keywords that depend on `properties` such as `additionalProperties` while `requireAllExcept` can be added without modifying the behavior of any existing keywords.
7. `requireAllExcept` breaks keyword independence, but does so in a very familiar way (works just like `additionalProperties`).

jsonschema-validation.xml
```diff
@@ -213,6 +213,24 @@
...
+                    affecting each other's outcomes.
+                </t>
+                <t>
+                    For schema author convenience, there are some exceptions among the
+                    keywords in this vocabulary:
+                    <list>
+                        <t>
+                            "requireAllExcept", whose behavior is defined in terms of "properties"
+                            "minContains", whose behavior is defined in terms of "contains"
+                            "maxContains", whose behavior is defined in terms of "contains"
```
notEthan, line 228: these should be 3 separate `<t>`s, I think? the list is rather incomplete as well. a complete list may be beyond the scope of this issue.

jdesrosiers: Thanks. I fixed the `<t>`s. The list includes all the everything relevant to the validation vocabulary. A similar list appears in the other vocabularies that have keywords that break keyword independence.

jdesrosiers:

gregsdennis:
The problem I have with how `requireAllExcept` is being described comes when trying to fit this into @handrews' behaviors model, which aims to prevent keywords from "looking into" other keywords.  `properties` generates an annotation of the names of the properties it evaluates (which is the the intersection of the keys in its value and the keys in the instance).

(from @awwright's examples above)
```json
{
   properties: {
      "name": {type: "string"},
      "comment": {type: "string"}
   },
   requireAllExcept: [
      "comment"
   ]
}
```

That means that for `{ "name": "foo", "bar": "baz" }`, `properties` in this schema would generate `[ "name" ]`.  This means that `requireAllExcept` must look into `properties` in order to see which properties are defined that aren't in the annotation.

A possible solution to this would be to have `properties` emit a different form of annotation that includes all of the defined property names _plus_ information on which ones were found.  Something like `{ "name": true, "comment": false }` would work, but  other forms could do.  Then `requireAllExcept` would just need to look at the annotation of `properties`, find all of the `false` values, and verify that none of those are in its own list.  (`additionalProperties`, etc, would only need to look at the keys of this new output from `properties`.)  This, then, fits into the model where annotations are the sole communication lines between keywords.

Also with this approach `additionalProperties`, etc, would need to be updated for the new `properties` annotation shape, but I think this is a minimal change.

jdesrosiers:
@gregsdennis You're right, this should probably be defined based on annotations (although I still think that approach needs to be revisited) and the current annotation behavior of `properties` is insufficient for `requireAllExcept` to make it's assertion.

From a classification point of view, this keyword should work exactly like `additionalProperties` except that it's not an applicator.

gregsdennis:
This PR needs to be rewritten as a proposal document.  See https://github.com/json-schema-org/json-schema-spec/pull/1450 for an example.

Closing this.

PR #1147, opened 2021-11-09
Defines what happens when relative JSON pointer starts with zero

### LINKED ISSUE #1121: Specify what happens when relative json pointer is `0`
Unless I'm missing something, we don't actualy specify what happens when a Relative JSON Pointer is `0` in the Relative JSON Pointer specification.

We hint and show examples, but it's not strictly specified.

We say...

>  If this value is more than zero, then the following steps are repeated that number of times...

But we do not say what happens when the value is zero.

This has resulted in a question being raised on our slack.

Relequestual:
Resolves #1121

### DIFF
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -125,7 +125,7 @@
                 Evaluation begins by processing the non-negative-integer prefix.
                 This can be found by taking the longest continuous sequence of decimal
                 digits available, starting from the beginning of the string, taking
-                the decimal numerical value. If the value is zero, the following stepps
+                the decimal numerical value. If the value is zero, the following steps
                 are skipped. If this value is more than zero, then the following steps
                 are repeated that number of times:
                 <list>
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -125,8 +125,9 @@
                 Evaluation begins by processing the non-negative-integer prefix.
                 This can be found by taking the longest continuous sequence of decimal
                 digits available, starting from the beginning of the string, taking
-                the decimal numerical value.  If this value is more than zero, then
-                the following steps are repeated that number of times:
+                the decimal numerical value. If the value is zero, the following stepps
+                are skipped. If this value is more than zero, then the following steps
+                are repeated that number of times:
                 <list>
                     <t>
                         If the current referenced value is the root of the document, then


gregsdennis:
https://tools.ietf.org/id/draft-handrews-relative-json-pointer-00.html#rfc.section.3

The spec already says that relative JSON pointers must start with positive integers.

Maybe we just remove "non-negative".

Relequestual:
> The spec already says that relative JSON pointers must start with positive integers.
> 
> Maybe we just remove "non-negative".

It quantifies in the ABNF...

> "0", or digits without a leading "0"

This is also shown in the proceeding examples...

> Starting from the value "baz" (inside "foo"), the following JSON strings evaluate to the accompanying values:
>                   "0"                         "baz"
>                   ...

It's essentially a no-op, but is valid.
The use is when apending `#` to get the key or index of the location.

(FYI latest is https://datatracker.ietf.org/doc/html/draft-bhutton-relative-json-pointer-00)

gregsdennis:
Then we should either change the section I linked or change the ANBF.  They conflict.  Personally, I'm happy with either.

Relequestual:
You do not believe zero is a non-negative integer?

relative-json-pointer.xml
```diff
@@ -125,8 +125,9 @@
                 Evaluation begins by processing the non-negative-integer prefix.
                 This can be found by taking the longest continuous sequence of decimal
                 digits available, starting from the beginning of the string, taking
-                the decimal numerical value.  If this value is more than zero, then
-                the following steps are repeated that number of times:
+                the decimal numerical value. If the value is zero, the following stepps
```
notEthan, line 128: typo "stepps"

might be better state the general case, which applies to zero as well, then note the behavior on zero. "The following steps are repeated the number of times indicated as the prefix value. If this value is zero, the steps are not performed."

gregsdennis:
My concern was alignment. I didn't care which it was. Positive and non-negative are both fine as long as the spec is consistent.

Relequestual:
It is consistent in the current version. You were viewing the previous 😅

relative-json-pointer.xml
Relequestual:
```suggestion
                the decimal numerical value. If the value is zero, the following steps
```

Relequestual:
As there has been little interest, I'm removing this PR and associated issue from the milestone.
It can be added back again should anyone show interest, but it shouldn't be a blocker.

handrews (approved):
While I find the phrasing slightly awkward, everything else I can think of is also slightly awkward.  This is definitely correct on a technical level, should we just go ahead and merge this?

PR #1154, opened 2021-12-06
resolve contradictions in the described behaviour of "additionalProperties" and "items"

karenetheridge:
Wording has been reverted to that of draft7, before annotations existed, and
the behaviour was clearly in terms of the presence of adjacent "properties"
and "patternProperties" keywords, not their evaluation results.

The described alternate behaviour when annotations are not collected is
contradictory to when annotations are collected, due to annotations only
existing when evaluation is *successful*. Producing errors at these
properties/items will never change the overall evaluation result, as errors
will ahve already been produced from the earlier keywords. Since this is a
changed behaviour between draft7 and draft2019-09, and no one had noticed this
inconsistency before now, it is likely that implementations have continued
with the previous behaviour, so correcting the language now will not actually
result in noticable changes in behaviour.

The inconsistencies in the behaviour of unevaluatedProperties and
unevaluatedItems with respect to evaluated-but-not-successfully
properties/items is greater, so will have to be addressed in a major revision
of the specification.

reference: https://github.com/json-schema-org/community/discussions/57

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2302,10 +2302,7 @@
                         <t>
                             This keyword applies its subschema to all instance elements
                             at indexes greater than the length of the "prefixItems" array
-                            in the same schema object, as reported by the annotation result
-                            of that "prefixItems" keyword.  If no such annotation
-                            result exists, "items" applies its subschema to all instance
-                            array elements.
+                            in the same schema object.
                             <cref>
                                 Note that the behavior of "items" without "prefixItems" is
                                 identical to that of the schema form of "items" in prior drafts.
@@ -2322,12 +2319,6 @@
                         <t>
                             Omitting this keyword has the same assertion behavior as
                             an empty schema.
-                        </t>
-                        <t>
-                            Implementations MAY choose to implement or optimize this keyword
-                            in another way that produces the same effect, such as by directly
-                            checking for the presence and size of a "prefixItems" array.
-                            Implementations that do not support annotation collection MUST do so.
                         </t>
                     </section>
 
@@ -2408,12 +2399,12 @@
                             The value of "additionalProperties" MUST be a valid JSON Schema.
                         </t>
                         <t>
-                            The behavior of this keyword depends on the presence and
-                            annotation results of "properties" and "patternProperties"
+                            The behavior of this keyword depends on the presence
+                            of "properties" and "patternProperties"
                             within the same schema object.
                             Validation with "additionalProperties" applies only to the child
-                            values of instance names that do not appear in the annotation
-                            results of either "properties" or "patternProperties".
+                            values of instance names that do not match any names in "properties",
+                            and do not match any regular expression in "patternProperties".
                         </t>
                         <t>
                             For all such properties, validation succeeds if the child instance
@@ -2426,13 +2417,6 @@
                         <t>
                             Omitting this keyword has the same assertion behavior as
                             an empty schema.
-                        </t>
-                        <t>
-                            Implementations MAY choose to implement or optimize this keyword
-                            in another way that produces the same effect, such as by directly
-                            checking the names in "properties" and the patterns in
-                            "patternProperties" against the instance property set.
-                            Implementations that do not support annotation collection MUST do so.
                         </t>
                     </section>
 


jdesrosiers (approved):

karenetheridge:
Note: I didn't add anything to the changelog.

gregsdennis:
> Since this is a changed behaviour between draft7 and draft2019-09, and no one had noticed this inconsistency before now, it is likely that implementations have continued with the previous behaviour, so correcting the language now will not actually result in noticable changes in behaviour.

The assumption that implementations don't use the prescribed behavior in 2019-09+ is wrong. Mine uses the annotation behavior; something I had to recently account for when processing a draft 6/7 schema.

I think 2019-09 & 2020-12 need to stay as they are, and this should be corrected (and `unevaluated*` aligned/resolved) in the next major version.

Relequestual:
I'm not sure we actually NEED to change the spec.
I think we read it wrong... See https://github.com/json-schema-org/json-schema-spec/issues/1172#issuecomment-1049686478

gregsdennis:
Are we sure that we want to push this for the 2020-12 patch?  It's a change to a behavior, not a clarification.

My library already works the way it's written.  I think this change is better for the next version.

gregsdennis (changes requested):
From the discussion in Slack with @Relequestual I agree with the reinterpretation of the spec that annotations are passed down at the schema level rather than the keyword level.  This means that _all_ annotations are maintained within the context of a (sub)schema evaluation, but upon leaving that context, annotation are dropped if that (sub)schema fails validation.

This resolves the issues that I (and likely many other implementors) have been receiving of confusing behavior by causing `additionalProperties` to behave as it did in draft 7 _without_ having to revert the language.

I think we need a more thorough clarification on how annotations are passed rather than changing this keyword.

(Note that this still means that I have to update my library.)

Relequestual:
@gregsdennis Given your second comment, do you still feel your previous comment stands?

> Are we sure that we want to push this for the 2020-12 patch? It's a change to a behavior, not a clarification.

gregsdennis:
I still think that this reversion is a change to behavior, yes.  It may be similar, but it's different than what is currently prescribed.

Relequestual:
Sorry, I didn't qualify my question.

THIS PR would be a change in behavior, sure.
Do you however believe that a clarification of interpritations (as per https://github.com/json-schema-org/json-schema-spec/pull/1154#issuecomment-1049687143) would be a change to behavior or not?

gregsdennis:
It's a change in behavior for my implementation because I implemented dropping annotations at the keyword level.

Based on the premise of the opening comment of this PR

> Since this is a changed behaviour between draft7 and draft2019-09, and no one had noticed this inconsistency before now, it is likely that implementations have continued with the previous behaviour, so correcting the language now will not actually result in noticable changes in behaviour.

I think a consensus of the reinterpretation should suffice, and changes represented in this PR are not needed.

karenetheridge:
I agree, if we go with the reinterpretation that @Relequestual discovered, then this PR is moot, but perhaps different wording would be needed to clarify the situation (and we'll need to update that one 'contains' test in the test suite too).

karenetheridge:
Okay, given the discussion in https://github.com/json-schema-org/json-schema-spec/issues/1172, I think this PR can be replaced with a different patch that clarifies the annotation behaviour of additionalProperties and unevaluatedProperties.  Whether or not it is more useful to always produce a `true` annotation (indicating that it evaluated at all properties, regardless of whether the results were valid or invalid) would be worthy of discussion.

I think we also need an amendment in the spec in the section on output formats: https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.12.3.4 implies that a keyword can only produce an error or annotations, but not both, but we can see that some keywords such as `properties` or `allOf` don't *produce* both, but their subschemas can produce annotations even if the keyword itself produces an error.

gregsdennis:
Having both errors and annotations is not supported with the current output structure since annotations are included as nested nodes.  I think this will need to wait until the next draft and once we've discussed and decided on those changes.

Relequestual:
Right. I'd really like to keep on topic as possible, opening a new issue for the next feature draft where we want to discuss further other topics.
As an aside, the verbose output structure does sort of support showing "all" annotations and validation results...

> The primary difference between [verbose] and the "Detailed" structure is that all results are returned. This includes sub-schema validation results that would otherwise be removed (e.g. annotations for failed validations, successful validations inside a `not` keyword, etc.). 
> https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.12.4.4

Relequestual:
> Okay, given the discussion in #1172, I think this PR can be replaced with a different patch that clarifies the annotation behaviour of additionalProperties and unevaluatedProperties...

@karenetheridge As such, would you be happy for this PR to be removed from the current patch draft milestone?

gregsdennis:
Given recent conversations around keyword behaviors & classifications, I'm unsure this PR needs to remain open.  I believe the overall intent is to keep with the annotation behavior (though the specifics are still being ironed out).

Can we close this now?

handrews:
@karenetheridge I'm seconding @gregsdennis 's request that we close this — it seems like the concerns have been addressed elsewhere, or we've shifted enough in our thinking that a new PR would be needed anyway?

karenetheridge:
could this be tagged with something like "reconsider for next draft"?

PR #1155, opened 2021-12-06
wording fixes for the interim release

karenetheridge:
This is an amended and abridged version of the changes proposed in #1110. This only contains clarifications; any changes to behaviour must be reserved for a future specification revision.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3855,7 +3855,7 @@ https://example.com/schemas/common#/$defs/count/minimum
                             <t>"$schema" MAY change for embedded resources</t>
                             <t>Array-value "items" functionality is now "prefixItems"</t>
                             <t>"items" subsumes the old function of "additionalItems"</t>
-                            <t>"contains" and "unevaluatedItems" interactions now specified</t>
+                            <t>"contains" annotation behavior, and "contains" and "unevaluatedItems" interactions now specified</t>
                             <t>Rename $recursive* to $dynamic*</t>
                             <t>$dynamicAnchor defines a fragment like $anchor</t>
                             <t>$dynamic* (previously $recursive) no longer use runtime base URI determination</t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2343,7 +2343,7 @@
                             its elements is valid against the given schema,
                             except when "minContains" is present and has a value of 0, in which
                             case an array instance MUST be considered valid against the "contains" keyword,
-                            even if NONE of its elements is valid against the given schema.
+                            even if none of its elements is valid against the given schema.
                         </t>
                         <t>
                             This keyword produces an annotation value which is an array of
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2318,6 +2318,8 @@
                             positions within the instance array, it produces an
                             annotation result of boolean true, indicating that all remaining array
                             elements have been evaluated against this keyword's subschema.
+                            This annotation affects the behavior of "unevaluatedItems" in the
+                            Unevaluated vocabulary.
                         </t>
                         <t>
                             Omitting this keyword has the same assertion behavior as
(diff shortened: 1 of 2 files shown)


Relequestual:
Saw too many commits because the target `draft-next` rather than `master. Changing PR target to `master` resolved the "woah that's a lot of commits from ages ago" issue. =]

jsonschema-core.xml
```diff
@@ -2337,15 +2339,18 @@
...
+                            its elements is valid against the given schema, provided that
+                            "minContains" is not present or is a value greater than 0.
                         </t>
                         <t>
-                            Logically, the validation result of applying the value subschema to each
-                            item in the array MUST be ORed with "false", resulting in an overall
-                            validation result.
+                            If "minContains" is present and has a value of 0, then an array instance
+                            is valid against this keyword only if NONE of its elements is valid
+                            against the given schema.
```
jdesrosiers, line 2348: I think this is backwards. If `"minContains": 0` then `contains` should always valid no matter what items are in the array.

jdesrosiers (changes requested):
Definitely an improvement. Just one thing that doesn't look quite right.

jsonschema-core.xml
Relequestual: I would agree. If `minContains` were 1, it would be a no-op. If `minContains` were 2, at least 2 of the items must match `contains`. If you say "At least N of the items in the array must match", then when N is 0, the logical reading is 0 of the items must match `contains`.

For none of the items, you would have to do `not > contains`.

Why would anyone want to do N = 0?
It would be an odd use case seening as we don't see many applications that use annotations at current, but it could be a means to add annotations to items in an array without causing validation assertions if there are none that match the `contains` schema value.

Relequestual: In fact, later we say...

> A value of 0 causes "contains" to always pass validation (but validation can still fail against a "maxContains" keyword).

In the validation spec doc.

Relequestual, line 2348:
```suggestion
                            If "minContains" is present and has a value of 0, then an array instance
                            MUST always be valid against this keyword.
```

jsonschema-core.xml
```diff
@@ -2337,15 +2339,18 @@
                         </t>
                         <t>
                             An array instance is valid against "contains" if at least one of
-                            its elements is valid against the given schema. The subschema MUST be
-                            applied to every array element even after the first match has
```
karenetheridge, line 2341: I just noticed these lines got lost inadvertently -- they were added (because unevaluatedItems now considers results from contains) after my original PR. I'll fix.

jsonschema-core.xml
karenetheridge: > I think this is backwards

Yes, I agree. I don't remember why I put this in here; `contains` should behave the same way no matter what `minContains` and `maxContains` say. Any additional constraints added by those keywords should result in errors being generated from those keywords, not `contains` itself.

karenetheridge: I am incorporating this edit into some changes I am about to push.

karenetheridge:
I've reworked the "contains" section from scratch and hopefully it is more clear (and with fewer falsehoods) now.

jdesrosiers (approved):

Relequestual (changes requested):
Apologies, I left a review unsubmitted for a while!
Few minor suggestions, but otherwise looks ideal.

jsonschema-core.xml
```diff
@@ -2337,15 +2340,10 @@
...
-                            This is to ensure that all possible annotations are collected.
-                        </t>
-                        <t>
-                            Logically, the validation result of applying the value subschema to each
-                            item in the array MUST be ORed with "false", resulting in an overall
-                            validation result.
+                            its elements is valid against the given schema,
+                            except when "minContains" is present and has a value of 0, in which
+                            case an array instance MUST be valid against the "contains" keyword,
+                            even if NONE of its elements is valid against the given schema.
```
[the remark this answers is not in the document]
karenetheridge: agreed.

Relequestual (approved):

Issue #1160, opened 2021-12-25
explicit licence needed in schema files and all repositories


karenetheridge:
I sent a request to the debian-perl folks to package my implementation for Debian. They are concerned that I bundle the metaschema files but there is no associated licence; additionally the tests bundle the test files from JSON-Schema-Test-Suite and there is no licence there either.

I would suggest:
- adding a licence in a `$comment` field in the metaschemas,
- add a LICENCE file in every repository (github will link to this when it exists, for extra visiblity).

cc @Relequestual

karenetheridge:
this can help - https://choosealicense.com/licenses/

karenetheridge:
It would also be useful to add a paragraph somewhere giving explicit guidance regarding the use of metaschemas, test files etc and what limitations, if any, are imparted on any software that uses these resources (e.g. what licences those works can or cannot use).

karenetheridge:
Note that for some packagers, the mere mention of the licence type/name is not enough - the full licence itself must be included in a LICENSE file.

gregsdennis:
Why do we need the license in every file?  That seems excessive.  (I suppose it depends on the license we choose.)

karenetheridge:
A single LICENSE file in the repository is likely sufficient (which would be included when the metaschemas are bundled elsewhere), but a comment in the metaschemas can't hurt, and is more clear.

gregsdennis:
😏 Do we need to at a `$license` keyword to the meta vocab?

awwright:
As far as my contributions go, the meta-schemas are public domain, and need neither a license nor attribution.

The idea that code is unusable unless it contains a license is a faulty assumption on their part (code can be public domain, and copyrighted code can be licensed separately from what it says), but to that end, there's https://unlicense.org/

I think putting that in a file next to the meta-schemas should be sufficient.

Also, I tend to doubt that most schemas have enough creative content to be copyrightable to begin with. At least in the US, objective facts & data generated by computer programs are generally not copyrightable, and outside the US, it doesn't make sense to me to cater to every broken copyright regime out there.

gregoa:
Hi,

one member of the Debian Perl Group who started with these questions here :)

What we need are the typical permissions of FLOSS (use, modify, distribute, etc.) for us as a project and for our users. That's typically covered in a license. It's not important how this is declared (LICENSE file, in a README, in each file, …) as long as it's clear and conistent.

[...]

PR #1173, opened 2022-01-14
Schemas and other data are public domain

awwright:
A remedy for #1160.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ If the pull request would solve a particular issue, reference the issue in the p
 
 Changes that would affect implementation behavior should typically be opened as an issue first.
 
-Pull requests should be made to master.
+Pull requests should be made to the default branch, which may be `draft-patch`, `draft-next`, or `master` depending on where we are in the [release process](https://github.com/json-schema-org/community/discussions/7).
 
 Most PRs, including all PRs that impact implementation behavior, will be left open for a minimum of 14 days.  Minor wording fixes may be merged more quickly once approved by a project member.
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Welcome to JSON Schema
-[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](https://github.com/json-schema-org/.github/blob/main/CODE_OF_CONDUCT.md) [![Financial Contributors on Open Collective](https://opencollective.com/json-schema/all/badge.svg?label=financial+contributors)](https://opencollective.com/json-schema)
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](https://github.com/json-schema-org/.github/blob/main/CODE_OF_CONDUCT.md) [![Project Status: Active – The project has reached a stable, usable state and is being actively developed.](https://www.repostatus.org/badges/latest/active.svg)](https://www.repostatus.org/#active) [![Financial Contributors on Open Collective](https://opencollective.com/json-schema/all/badge.svg?label=financial+contributors)](https://opencollective.com/json-schema)
 
 JSON Schema is a vocabulary that allows you to validate, annotate, and manipulate JSON documents.
 
--- a/README.md
+++ b/README.md
@@ -86,4 +86,6 @@ Here are our top sponsors. You could be next! [[Become a sponsor](https://openco
 
 ## License
 
-The source material in this repository is licensed under the AFL or BSD license.
+Prose and other content in this repository may be distributed under the terms of the AFL (Academic Free License) or BSD (simplified 2-clause) license.
+
+Sample code and schemas, including content in `meta/`, is public domain (as described by UNLICENSE).
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Welcome to JSON Schema
-[![Financial Contributors on Open Collective](https://opencollective.com/json-schema/all/badge.svg?label=financial+contributors)](https://opencollective.com/json-schema)
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](https://github.com/json-schema-org/.github/blob/main/CODE_OF_CONDUCT.md) [![Financial Contributors on Open Collective](https://opencollective.com/json-schema/all/badge.svg?label=financial+contributors)](https://opencollective.com/json-schema)
 
 JSON Schema is a vocabulary that allows you to validate, annotate, and manipulate JSON documents.
 
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3851,6 +3851,10 @@ https://example.com/schemas/common#/$defs/count/minimum
             </t>
             <t>
                 <list style="hanging">
+                    <t hangText="draft-bhutton-json-schema-next">
+                        <list style="symbols">
+                        </list>
+                    </t>
                     <t hangText="draft-bhutton-json-schema-00">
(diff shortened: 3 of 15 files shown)


karenetheridge:
- does the filename UNLICENSE have any special meaning in that it is not a LICENSE?
- I would put this file in the root of the repository, rather than in meta/, as the main schema file is in the root as `schema.json`. It will also be much more visible there to casual browsers of the repo.

awwright:
> does the filename UNLICENSE have any special meaning in that it is not a LICENSE?

I think the name is to emphasize that, to the extent it is acting as a license, that it's not intending to be one. I've found that GitHub recognizes the filename. But it doesn't matter either way.

> I would put this file in the root of the repository

To me, I would read this, and it would imply it's authoritative for all the content in the entire repository, but this isn't correct.

karenetheridge:
> To me, I would read this, and it would imply it's authoritative for all the content in the entire repository, but this isn't correct.

The file should either say explicitly which files it applies to, or be rewritten so it does apply to all files.

I feel that if the file is not in the root it will not be seen, and certainly not be understood to apply to /schema.json.

Relequestual:
I think this is fine, but let's add it to the agenda for our upcoming Open Community Working Meeting call (for which I have yet to create an issue... on it now and will update).

Relequestual:
@karenetheridge do you use schemas in the `output` directory too?
If so, we will need the licence and readme there also.

awwright:
> The file should either say explicitly which files it applies to, or be rewritten so it does apply to all files.

It's uncommon to modify the actual contents of the license file, and probably discouraged for possible legal complications that none of us can foresee. I think it's best if README.md describe which files are provided under which license. We can then include the text of each license to remove any remaining ambiguity.

> I feel that if the file is not in the root it will not be seen, and certainly not be understood to apply to /schema.json.

I think it can be buried somewhat if that helps clarify its scope; it's more important that we be unambiguous. We can also specify: "As attached in the `meta/UNLICENSE` file in this repository." (Or "As attached in `LICENSE.Unlicense`", or whatever the filename is.)

karenetheridge:
> I think it's best if README.md describe which files are provided under which license

Yup, it can't hurt to be explicit :)

> I think it can be buried somewhat if that helps clarify its scope

Yes, but part of my point was that burying a licence file in meta/ would be wrong if we intend it to apply to /schema.json as well.

> @karenetheridge do you use schemas in the output directory too?

I would assume that all the schema files would apply under the same licence.

Perhaps we should move them all to a common subdirectory. Then there can be a LICENSE file in that subdirectory, along with a README that clearly states that the licence is intended to apply to those files.

Julian:
#1255 covers the non-data related portion of #1160 (and so is related).

handrews:
@awwright @Julian is this an alternative to #1255 or is it complementary?
Austin, can you rebase/recreate/whatever this onto main so the diff is readable without having to dig out the individual commit?

Julian:
This one was before we got guidance from OpenJS legal. It's still possibly relevant though if we want an even more liberal licence for the schemas than for the rest of the repo.

karenetheridge:
I just looked at this again. This bit causes me concern:

> Anyone is free to copy, modify, publish, use, compile, sell, or distribute this software

How would we feel if someone took the main metaschema, modified it (say to change the semantics for a keyword they didn't like), and then published it in their implementation without changing the `$id`? That would cause confusion among any who used that implementation, and surely is something we would ask the author to stop doing. Can we avoid that by spelling out that the files can be redistributed but not altered?

ljharb:
That's how open source works; nothing stops someone from doing that - but in practice, who would actually care about the knockoff?

handrews:
> Can we avoid that by spelling out that the files can be redistributed but not altered?

I think one of the Creative Commons variations might do that.

gregsdennis:
> who would actually care about the knockoff?

If the knockoff is an implementation like ajv (which already does some roguish things), the impact could be significant.

handrews:
> If the knockoff is an implementation like ajv (which already does some roguish things), the impact could be significant.

That's a good point.  We should probably be better about defining what it means to claim you implement JSON Schema.  Which is not a spec thing, but is a thing the JSON Schema Org should think about.  At this point, what AJV implements by default (without having to reconfigure it) is not JSON Schema despite the claims, and it's a problem.

ljharb:
If a deviation is sufficiently useful that it gains adoption, then perhaps that’s a strong signal that the spec should incorporate it.

handrews:
@ljharb not if the deviations results in a constant stream of people showing up asking us why their schema that ought to work doesn't work.  AJV is one implementation, no matter how popular.  We don't get people showing up in the slack asking for more implementations to act like it, but we do get a lot of people who are confused and annoyed.

handrews:
We've adopted quite a few innovations from the community, including AJV (which played a large role in the adoption of `if`/`then`/`else`).  But those adoptions are due to broader demand beyond any one implementation.

handrews:
@Julian the language would be tied to the `$id` value.  Other people shouldn't publish `json-schema.org` identifiers.

Julian:
I still think such a thing would cause issues personally. Any custom language typically does for e.g. corporate users.

handrews:
> Any custom language

1. That's overly vague, let's talk about specifics
2. There are many licenses covering many usage patterns, it need not be "custom" (we shouldn't be writing custom legal language without a lawyer)

The things to consider are:

* Can you use a `json-schema.org`-identified schema (JSOIS) in a for-profit application?  ***Yes***
* Can you redistribute an unmodified JSOIS in a for-profit application?  ***Yes***
* Can you create a derivation of a JSOIS that does ***not*** use `json-schema.org` in its `$id` and do anything you want with it?  ***Yes***
* Can you non-functionally modify (e.g. minify or otherwise reformat) a JSOIS and use/redistribute it just as the original?  ***Yes***
* Can you functionally modify a JSOIS and re-publish it with the same `$id`?  ***No***
* Can you create a new schema with a `json-schema.org` `$id` and publish it?  ***No***

Julian:
I've shared my opinion. Feel free to have your own or disregard mine, but I stand by what I said.

handrews:
I'd be interested in hearing from any FOSS compliance people on whether the approach I sketched out would cause any problems.

I'm not dead set on these restrictions, but it's worth considering, and perhaps checking to see if the OpenJS Foundation has any resources or advice to consider.

Julian:
Yep that definitely seems reasonable to me too. (Obviously they will ask about what license the existing ones are under, which I'd suspect would be presumed to be BSD like the documents).

gregoa:
On Sun, 28 Aug 2022 20:46:38 -0700, Karen Etheridge wrote:

> > Anyone is free to copy, modify, publish, use, compile, sell, or distribute this software
> 
> How would we feel if someone took the main metaschema, modified it
> (say to change the semantics for a keyword they didn't like), and
> then published it in their implementation without changing the
> `$id`? That would cause confusion among any who used that
> implementation, and surely is something we would ask the author to
> stop doing. Can we avoid that by spelling out that the files can be
> redistributed but not altered?

If the license forbids modification, the software would fail any
FLOSS definition, from the Debian Free Software Guidelines to OSI's
Open Source Definition, etc.

For Debian that would mean that we can't include it in Debian "main"
(i.e. the "proper" Debian archive), we would have to put it into the
"non-free" archive area (which is not enabled by default and not
available for other packages etc.).


Cheers,
gregor, Debian Perl Group member

-- 
 .''`.  https://info.comodo.priv.at -- Debian Developer https://www.debian.org
 : :' : OpenPGP fingerprint D1E1 316E 93A7 60A8 104D  85FA BB3A 6801 8649 AA06
 `. `'  Member VIBE!AT & SPI Inc. -- Supporter Free Software Foundation Europe
   `-

PR #1181, opened 2022-02-03
Remove remaining media-type param references

jdesrosiers:
The "schema" media type parameter got removed for the 2020-12 release, but some of it was missed. This removes the remaining bits.

Reference: https://github.com/json-schema-org/json-schema-spec/issues/832

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3115,20 +3115,6 @@ https://example.com/schemas/common#/$defs/count/minimum
                         <t>Subtype name: schema+json</t>
                         <t>Required parameters: N/A</t>
                         <t>
-                            Optional parameters:
-                            <list style="hanging">
-                                <t hangText="schema:">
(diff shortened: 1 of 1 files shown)


Relequestual:
Wondering if this should actually target draft-next. Thoughts?

jdesrosiers:
@Relequestual No, this is for the patch release. The media type parameter was removed for 2020-12, but the removal was incomplete. This just removes the piece that was missed. In it's current state, we declare the media type parameter, but the section of the spec that describes how to use it is gone. We can bring it back in some form in draft-next, but for the patch we need to fix the inconsistency.

Relequestual:
OK. Please add to changelog and merge.

jdesrosiers:
It's already in the change log from when it was originally removed.

> * Remove media type parameters

https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.appendix.G

PR #1192, opened 2022-02-22
Clarify that plain name fragments are neither canonical or non-canonical

### LINKED ISSUE #937: Clarify whether plain name fragment URIs are canonical, and what it means if they are (or aren't)

Relequestual:
Attempt to resolve #937

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3400,7 +3400,7 @@ https://example.com/schemas/common#/$defs/count/minimum
                 rather, the base URI used (as part of the full URI with any fragment) is what
                 determines the canonical nature of the resutling full URI.
                 <cref>
-                    Multiple "canonical" URIs? We Acknowledge this is potetnially confusing, and
+                    Multiple "canonical" URIs? We Acknowledge this is potentially confusing, and
                     direct you to read the CREF located in the
                     <xref target="embedded">JSON Pointer fragments and embedded schema resources</xref>
                     section for futher comments.
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -50,7 +50,7 @@
             </address>
         </author>
 
-        <date year="2021"/>
+        <date year="2022"/>
         <workgroup>Internet Engineering Task Force</workgroup>
         <keyword>JSON</keyword>
         <keyword>Schema</keyword>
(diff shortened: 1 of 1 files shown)


jdesrosiers (approved):

jsonschema-core.xml
```diff
@@ -3390,6 +3395,18 @@ https://example.com/schemas/common#/$defs/count/minimum
                     </t>
                 </list>
             </t>
+            <t>
+                Note: The fragment part of the URI does not make it canonical or non-canonical,
+                rather, the base URI used (as part of the full URI with any fragment) is what
+                determines the canonical nature of the resutling full URI.
```
karenetheridge, line 3401:
```suggestion
                determines the canonical nature of the resulting full URI.
```

jsonschema-core.xml
```diff
@@ -3390,6 +3395,18 @@ https://example.com/schemas/common#/$defs/count/minimum
...
+                <cref>
+                    Multiple "canonical" URIs? We Acknowledge this is potetnially confusing, and
```
karenetheridge, line 3403:
```suggestion
                    Multiple "canonical" URIs? We Acknowledge this is potentially confusing, and
```

karenetheridge (approved):
Looks good, modulo a few spelling errors.

PR #1194, opened 2022-02-23
Clarify that we are interested in the data model.

ioggstream:
## This PR

- We have "JSON text" in notational conventions, 
   so there's no need to reference the application/json media type
- Since the document references an equivalence relation between JSON text, JSON value and JSON document
  now  JSON Document references the Data Model section below, so we don't need to mess with "sequence of bytes"

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -181,11 +181,8 @@
 
             <section title="JSON Document">
                 <t>
-                    A JSON document is an information resource (series of octets) described by the
-                    application/json media type.
-                </t>
-                <t>
-                    In JSON Schema, the terms "JSON document", "JSON text", and "JSON value" are
+                    A JSON document represents the data underlying a given "JSON text" or "JSON value".
+                    Thus in JSON Schema, the terms "JSON document", "JSON text", and "JSON value" are
                     interchangeable because of the data model it defines.
                 </t>
                 <t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3851,6 +3851,10 @@ https://example.com/schemas/common#/$defs/count/minimum
             </t>
             <t>
                 <list style="hanging">
+                    <t hangText="draft-bhutton-json-schema-next">
+                        <list style="symbols">
+                        </list>
+                    </t>
                     <t hangText="draft-bhutton-json-schema-00">
                         <list style="symbols">
                             <t>"$schema" MAY change for embedded resources</t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -188,7 +188,7 @@
                 <t>
                     JSON Schema is only defined over JSON documents. However, any document or memory
                     structure that can be parsed into or processed according to the JSON Schema data
-                    model can be interpreted against a JSON Schema, including media types like
+                    model can be interpreted against a JSON Schema, including data formats like
                     <xref target="RFC7049">CBOR</xref>.
                 </t>
             </section>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1483,27 +1483,22 @@
                         <t>
                             Together with "$dynamicAnchor", "$dynamicRef" implements a cooperative
                             extension mechanism that is primarily useful with recursive schemas
-                            (schemas that reference themselves).  Both the extension point and the
-                            runtime-determined extension target are defined with "$dynamicAnchor",
-                            and only exhibit runtime dynamic behavior when referenced with
-                            "$dynamicRef".
+                            (schemas that reference themselves).  The extension point is defined with
+                            "$dynamicAnchor" and only exhibits runtime dynamic behavior when referenced
+                            with "$dynamicRef".
                         </t>
                         <t>
-                            The value of the "$dynamicRef" property MUST be a string which is
-                            a IRI-Reference.  Resolved against the current IRI base, it produces
-                            the IRI used as the starting point for runtime resolution.  This initial
-                            resolution is safe to perform on schema load.
+                            The value of the "$dynamicRef" property MUST be a string which is a
+                            IRI-Reference that contains a valid <xref target="anchor">plain name
+                            fragment</xref>.  Resolved against the current IRI base, it indicates
+                            the schema resource used as the starting point for runtime resolution.
+                            This initial resolution is safe to perform on schema load.
                         </t>
                         <t>
-                            If the initially resolved starting point IRI includes a fragment that
-                            was created by the "$dynamicAnchor" keyword, the initial IRI MUST be
-                            replaced by the IRI (including the fragment) for the outermost schema
-                            resource in the <xref target="scopes">dynamic scope</xref> that defines
-                            an identically named fragment with "$dynamicAnchor".
-                        </t>
-                        <t>
-                            Otherwise, its behavior is identical to "$ref", and no runtime
-                            resolution is needed.
+                            The schema to apply is the outermost schema resource in the
+                            <xref target="scopes">dynamic scope</xref> that defines a
+                            "$dynamicAnchor" that matches the plain name fragment in the initially
+                            resolved IRI.
                         </t>
                         <t>
                             For a full example using these keyword, see appendix
@@ -3863,6 +3858,7 @@ https://example.com/schemas/common#/$defs/count/minimum
                     <t hangText="draft-bhutton-json-schema-next">
                         <list style="symbols">
                             <t>"contains" now applies to objects as well as arrays</t>
+                            <t>Remove bookending requirement for "$dynamicRef"</t>
                         </list>
                     </t>
                     <t hangText="draft-bhutton-json-schema-00">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2343,17 +2343,25 @@
                             This is to ensure that all possible annotations are collected.
                         </t>
                         <t>
-                            Logically, the validation result of applying the value subschema to each
-                            item in the array MUST be ORed with "false", resulting in an overall
-                            validation result.
+                            An object instance is valid against "contains" if at least one of
+                            its property values is valid against the given schema. The subschema
+                            MUST be applied to every property value even after the first match has
+                            been found, in order to collect annotations for use by other keywords.
+                            This is to ensure that all possible annotations are collected.
                         </t>
                         <t>
-                            This keyword produces an annotation value which is an array of
-                            the indexes to which this keyword validates successfully when applying
-                            its subschema, in ascending order. The value MAY be a boolean "true" if
-                            the subschema validates successfully when applied to every index of the
-                            instance. The annotation MUST be present if the instance array to which
-                            this keyword's schema applies is empty.
+                            Logically, the validation result of applying the value subschema to each
+                            item in the array or property in the object MUST be ORed with "false",
+                            resulting in an overall validation result.
+                        </t>
+                        <t>
+                            This keyword produces an annotation value which is an array of the
+                            indexes or property names to which this keyword validates successfully
+                            when applying its subschema, in ascending order. The value MAY be a
+                            boolean "true" if the subschema validates successfully when applied to
+                            every index or property value of the instance. The annotation MUST be
+                            present if the instance array or object to which this keyword's schema
+                            applies is empty.
                         </t>
                     </section>
                 </section>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2,6 +2,7 @@
 <!DOCTYPE rfc SYSTEM "rfc2629.dtd" [
 <!ENTITY RFC2119 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.2119.xml">
 <!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
+<!ENTITY RFC3987 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3987.xml">
 <!ENTITY RFC6596 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6596.xml">
 <!ENTITY RFC6839 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6839.xml">
 <!ENTITY RFC6901 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6901.xml">
@@ -205,7 +206,7 @@
                 </t>
                 <t>
                     Among these, this specification defines the "application/schema-instance+json"
-                    media type which defines handling for fragments in the URI.
+                    media type which defines handling for fragments in the IRI.
                 </t>
 
                 <section title="Instance Data Model">
@@ -316,7 +317,7 @@
                         <list style="hanging">
                             <t hangText="identifiers:">
                                 control schema identification through setting the schema's
-                                canonical URI and/or changing how the base URI is determined
+                                canonical IRI and/or changing how the base IRI is determined
                             </t>
                             <t hangText="assertions:">
                                 produce a boolean result when applied to an instance
@@ -393,7 +394,7 @@
                     <t>
                         Vocabularies are the primary unit of re-use in JSON Schema, as schema
                         authors can indicate what vocabularies are required or optional in
-                        order to process the schema.  Since vocabularies are identified by URIs
+                        order to process the schema.  Since vocabularies are identified by IRIs
                         in the meta-schema, generic implementations can load extensions to support
                         previously unknown vocabularies.  While keywords can be supported outside
                         of any vocabulary, there is no analogous mechanism to indicate individual
@@ -419,12 +420,12 @@
                     <t>
                         A JSON Schema resource is a schema which is
                         <xref target="RFC6596">canonically</xref> identified by an
-                        <xref target="RFC3986">absolute URI</xref>.
+                        <xref target="RFC3987">absolute IRI</xref>.
                     </t>
                     <t>
                         The root schema is the schema that comprises the entire JSON document
                         in question.  The root schema is always a schema resource, where the
-                        URI is determined as described in section
+                        IRI is determined as described in section
                         <xref target="initial-base" format="counter"></xref>.
                     </t>
                     <t>
@@ -482,7 +483,7 @@
                 fragment identifier structure: JSON Pointers.
             </t>
             <t>
-                The use of JSON Pointers as URI fragment identifiers is described in
+                The use of JSON Pointers as IRI fragment identifiers is described in
                 <xref target="RFC6901">RFC 6901</xref>.
                 For "application/schema+json", which supports two fragment identifier syntaxes,
                 fragment identifiers matching the JSON Pointer syntax, including the empty string,
@@ -638,9 +639,9 @@
                     schema object with no subschemas.
                 </t>
                 <t>
-                    Keywords MAY be defined with a partial value, such as a URI-reference,
+                    Keywords MAY be defined with a partial value, such as a IRI-reference,
                     which must be resolved against another value, such as another
-                    URI-reference or a full URI, which is found through the lexical
+                    IRI-reference or a full IRI, which is found through the lexical
                     structure of the JSON document.  The "$id", "$ref", and
                     "$dynamicRef" core keywords, and the "base" JSON Hyper-Schema
                     keyword, are examples of this sort of behavior.
@@ -725,20 +726,20 @@
             </section>
             <section title="Identifiers" anchor="identifiers">
                 <t>
-                    Identifiers set the canonical URI of a schema, or affect how such URIs are
+                    Identifiers set the canonical IRI of a schema, or affect how such IRIs are
                     resolved in <xref target="references">references</xref>, or both.
                     The Core vocabulary defined in this document defines several
                     identifying keywords, most notably "$id".
                 </t>
                 <t>
-                    Canonical schema URIs MUST NOT change while processing an instance, but
-                    keywords that affect URI-reference resolution MAY have behavior that
+                    Canonical schema IRIs MUST NOT change while processing an instance, but
+                    keywords that affect IRI-reference resolution MAY have behavior that
                     is only fully determined at runtime.
                 </t>
                 <t>
                     While custom identifier keywords are possible, vocabulary designers should
                     take care not to disrupt the functioning of core keywords. For example,
-                    the "$dynamicAnchor" keyword in this specification limits its URI resolution
+                    the "$dynamicAnchor" keyword in this specification limits its IRI resolution
                     effects to the matching "$dynamicRef" keyword, leaving the behavior
                     of "$ref" undisturbed.
                 </t>
@@ -913,7 +914,7 @@
                                 such as "$ref" were followed to reach the absolute schema location.
                             </t>
                             <t>
-                                The absolute schema location of the attaching keyword, as a URI.
+                                The absolute schema location of the attaching keyword, as a IRI.
                                 This MAY be omitted if it is the same as the schema location path
                                 from above.
                             </t>
@@ -1117,14 +1118,14 @@
             </t>
             <t>
                 Meta-schemas that do not use "$vocabulary" MUST be considered to
-                require the Core vocabulary as if its URI were present with a value of true.
+                require the Core vocabulary as if its IRI were present with a value of true.
             </t>
             <t>
-                The current URI for the Core vocabulary is:
+                The current IRI for the Core vocabulary is:
                 &lt;https://json-schema.org/draft/2020-12/vocab/core&gt;.
             </t>
             <t>
-                The current URI for the corresponding meta-schema is:
+                The current IRI for the corresponding meta-schema is:
                 <eref target="https://json-schema.org/draft/2020-12/meta/core"/>.
             </t>
             <t>
@@ -1176,12 +1177,12 @@
                         set of valid schemas written for this particular dialect.
                     </t>
                     <t>
-                        The value of this keyword MUST be a <xref target="RFC3986">URI</xref>
-                        (containing a scheme) and this URI MUST be normalized.
-                        The current schema MUST be valid against the meta-schema identified by this URI.
+                        The value of this keyword MUST be a <xref target="RFC3987">IRI</xref>
+                        (containing a scheme) and this IRI MUST be normalized.
+                        The current schema MUST be valid against the meta-schema identified by this IRI.
                     </t>
                     <t>
-                        If this URI identifies a retrievable resource, that resource SHOULD be of
+                        If this IRI identifies a retrievable resource, that resource SHOULD be of
                         media type "application/schema+json".
                     </t>
                     <t>
@@ -1208,15 +1209,15 @@
                     </t>
                     <t>
                         The value of this keyword MUST be an object.  The property names in the
-                        object MUST be URIs (containing a scheme) and this URI MUST be normalized.
-                        Each URI that appears as a property name identifies a specific set of
+                        object MUST be IRIs (containing a scheme) and this IRI MUST be normalized.
+                        Each IRI that appears as a property name identifies a specific set of
                         keywords and their semantics.
                     </t>
                     <t>
-                        The URI MAY be a URL, but the nature of the retrievable resource is
+                        The IRI MAY be a URL, but the nature of the retrievable resource is
                         currently undefined, and reserved for future use.  Vocabulary authors
                         MAY use the URL of the vocabulary specification, in a human-readable
-                        media type such as text/html or text/plain, as the vocabulary URI.
+                        media type such as text/html or text/plain, as the vocabulary IRI.
                         <cref>
                             Vocabulary documents may be added in forthcoming drafts.
                             For now, identifying the keyword set is deemed sufficient as that,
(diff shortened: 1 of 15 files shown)


jsonschema-core.xml
```diff
@@ -181,17 +181,14 @@
...
-                    In JSON Schema, the terms "JSON document", "JSON text", and "JSON value" are
+                    A JSON document represents the data underlying a given "JSON text" or "JSON value".
+                    Thus in JSON Schema, the terms "JSON document", "JSON text", and "JSON value" are
                     interchangeable because of the data model it defines.
                 </t>
                 <t>
                     JSON Schema is only defined over JSON documents. However, any document or memory
                     structure that can be parsed into or processed according to the JSON Schema data
-                    model can be interpreted against a JSON Schema, including media types like
+                    model can be interpreted against a JSON Schema, including data formats like
```
ioggstream, line 191: CBOR is a data format. It can have different media types.

jsonschema-core.xml
```diff
@@ -181,17 +181,14 @@
 
             <section title="JSON Document">
                 <t>
-                    A JSON document is an information resource (series of octets) described by the
-                    application/json media type.
```
ioggstream, line 185: iiuc this means "JSON documents" := "JSON text", thus resulting in a redundant definition.

jdesrosiers: I think you're right. I don't think there's much value in including that sentence.

jsonschema-core.xml
junioryo: Good job

jsonschema-core.xml
junioryo: Damn

awwright:
The JSON Schema spec is being submitted as an Internet specification, or at least must be compatible with Internet standards. Saying "application/json media type" is intended in the first part, as it's a well-defined technical term, and using it implies something specific for interoperability on the Internet.

It's saying more than simply "JSON document := JSON text." (Where "JSON text" is the ABNF production.) A JSON document is not merely any series of bytes that matches the JSON text grammar. It also has to be marked as being a JSON document. A Web browser cannot parse a "text/plain" document as JSON just because it matches the JSON grammar.

As for using "media types" vs. "data formats", I don't care one way or the other, we don't actually need the well-defined term here. But CBOR _is_ a media type, it's content-type registration is `application/cbor`.

ioggstream:
Hi @awwright, let me clarify + forgive the nitpicking

> The JSON Schema spec is being submitted as an Internet specification, or at least must be compatible with Internet standards. 

Ok

> Saying "application/json media type" is intended in the first part, as it's a well-defined technical term,
> and using it implies something specific for interoperability on the Internet.

My understanding of "The media type for JSON text is application/json" from RFC8259 is that they imply the same thing.

> A JSON document is not merely any series of bytes that matches the JSON text grammar. 
> It also has to be marked as being a JSON document. 
> A Web browser cannot parse a "text/plain" document as JSON just because it matches the JSON grammar.

Your point is clear now. Thinking twice, imho a `text/plain` document is not a "JSON text" since according to RFC8259, the media type for JSON text is `application/json`. But I may be wrong.

> As for using "media types" vs. "data formats", I don't care one way or the other,
> we don't actually need the well-defined term here.

To progress with the draft, in my experience these are the kind of clarifications I was asked to resolve by the IETF community  for a couple of specs I started working, so I'm just trying to share the lesson learnt :) While I try to do my best, I can still make mistakes. Moreover I understand that this whole discussion may seem pointless, but I think it is not from the perspective of a first time reader.

> But CBOR _is_ a media type, it's content-type registration is `application/cbor`.

See above wrt "pointless nitpicking": my understanding from reading [CBOR](https://www.rfc-editor.org/rfc/rfc8949.html#abstract) is that it is a data format.
The registered media type for a single encoded CBOR data item  is `application/cbor`.

awwright:
As far as I'm concerned, nit-picking is worthwhile and encouraged.

> My understanding of "The media type for JSON text is application/json" from RFC8259 is that they imply the same thing.

I think this is an unfortunate case where RFC 8259 is using "JSON text" to refer to both the ABNF production (the set of all strings that form valid JSON) and a JSON document (a string that's valid JSON that's known to be `application/json`). Whereas I see these as different things and deserving of different terms. You're probably right, that I may be making _too much_ of a distinction where it's not needed.

I'm not sure why RFC 8259 avoids the term "JSON document" since that's the terminology in other media types ("HTML document", "PDF document", etc). I didn't previously notice this.

> See above wrt "pointless nitpicking": my understanding from reading [CBOR](https://www.rfc-editor.org/rfc/rfc8949.html#abstract) is that it is a data format.

I think it's fair to call CBOR either a data type or a media type; since media types I would describe as standardized data formats.

junioryo (changes requested):
Nice

jdesrosiers:
> The JSON Schema spec is being submitted as an Internet specification

Just to be clear, the only thing we plan to submit as an internet standard are the JSON Schema media types, which are just relatively minor extension of `application/json`. There are no plans to submit any particular dialect defined in the past or the future as an internet standard.

awwright:
@jdesrosiers I would call JSON Schema a little more than a "relatively minor extension of `application/json`"... What do you mean by that? The namespace is in the standards tree, which means it must be "of general interest to the Internet community", have some sort of permanent document that completely defines its behavior, and it receives some sort of expert review.

jdesrosiers:
@awwright We want `application/schema+json` to be a general purpose media type. Any given dialect is a profile of that media type. So, `application/schema+json` is just `application/json` plus fragment semantics and dialect identification using a media-type parameter or the `$schema` keyword. Once we get the media type registered with IETF, the JSON Schema spec can be dedicated entirely to defining a dialect. JSON Schema can just be a language that can be serialized as a `application/schema+json` document, if you need such a thing.

awwright:
@jdesrosiers I'm not sure what a "general purpose" media type means in this context, or "dialect". We can define an extension mechanism, but there has to be a certain amount of core (default) functionality. For example, if I specify the schema `{}`, how do we know what that means? If I specify the schema `{"type":[]}` or `{"not":{}}`, how do I know what that means?
Is this being asked in some other thread, do you know?

jdesrosiers:
> I'm not sure what a "general purpose" media type means in this context

XML is an example of a general purpose media type. It's a way to serialize generic data. There's no semantics associated with that data. XHTML is a domain specific media type. It's XML, but its tags have semantics relevant to the domain of presenting a document visually. I want the JSON Schema media type to be like XML, not XHTML.

> I'm not sure what a ["dialect"] means in this context

A JSON Schema dialect is any distinct flavor of JSON Schema. Each release we put out is a dialect. OpenAPI has defined three dialects over their various versions. MongoDB has it's own dialect. The concept has been documented in the spec since 2019-09.

> how do I know what [a schema] means?

The media type defines how to declare what dialect a schema is using. The dialect defines the semantics of the keywords. How the dialect is specified is up to each dialect's author.

awwright:
> I want the JSON Schema media type to be like XML, not XHTML.

This sounds backwards. This suggests that authors can write JSON Schema documents that have no meaning. But they _have_ to have a meaning. `{}` (the shortest possible JSON Schema that's still syntactically valid) means "the set of all JSON documents." `{"type":[]}` means "the empty set." Etc.

Are you proposing an _extension mechanism_? That sounds perfectly reasonable. XHTML can be extended with XML namespaces, for example MathML or SVG.

jdesrosiers:
> This suggests that authors can write JSON Schema documents that have no meaning.

I've said multiple times now that the meaning comes from the dialect declaration. Does this not make sense? Do you see this as insufficient? Why?

> Are you proposing an _extension mechanism_?

Sure, you can think of dialect declaration as an extension mechanism. The document has no intrinsic semantics. You extend those semantics by declaring a dialect. The media type defines how to declare the dialect of the schema using the `$schema` keyword or the `schema` media type parameter. That declaration is like `doctype` declaration in an HTML document that tells the user-agent what version of the HTML specification to use to interpret the rest of the document. The dialect declaration tells the JSON Schema implementation what version of the specification to use to interpret the schema.

But, yes, if the document doesn't declare a dialect, the document would have no meaning that a user-agent could act on. All meaning and behavior is pushed to user-space. Other than JSON Pointer fragment identification, it would be just like working with a JSON document. It wouldn't function as a schema, but that doesn't mean it's useless in some other context. You could also use this media type to declare a dialect that has nothing to do with schemas. It could have some other purpose entirely. Of course the media type isn't specifically designed for or intended to be used in these ways, but it's possible.

awwright:
> the meaning comes from the dialect declaration

 I asked earlier: If I specify the schema {"type":[]} or {"not":{}}, how do I know what that means? More specifically, how do I determine what the "dialect" is?

awwright:
> the document would have no meaning that a user-agent could act on

> it would be just like working with a JSON document. It wouldn't function as a schema

It sounds like what you're proposing is just JSON with Namespaces. There's similar projects that do this, like JSON-LD. I'm not necessarily against namespaces or a new media type for them, but this misses the point of JSON Schema, which is to have a _single_, standard vocabulary for describing assertions, annotations, and sets of JSON.

And also, most people are not interested in declaring a meta-schema or dialect for day-to-day use of JSON Schema. Different implementations doing something different for even simple statements like `{"minimum":0}` would be fatal.

And by the way, I see a good idea related to what you're proposing; what would it take to encode a meta-schema as a JSON-LD document, and write a validator that supports these JSON-LD schemas?

jdesrosiers:
> More specifically, how do I determine what the "dialect" is?

Sorry, I didn't realize that was the question you were asking. Dialect identification works the way it always has. See [here](https://github.com/ietf-wg-httpapi/mediatypes/blob/main/draft-ietf-httpapi-rest-api-mediatypes.md#json-schema-media-types) for a summary. If no dialect is declared, behavior is undefined. Implementations can assume a default dialect, or use heuristics to guess a dialect or, throw an error, or proceed with no dialect (all properties treated as unknown keywords). I'd rather it not be undefined, but it has to be compatible with older dialects and existing implementations.

> It sounds like what you're proposing is just JSON with Namespaces.

If you think of a dialect as a namespace, I guess you could say that, but that's not really the point. The point is just to define something flexible enough to support any past or future JSON Schema dialect.

> the point of JSON Schema [] is to have a single, standard vocabulary for describing assertions, annotations, and sets of JSON.

I disagree that that's the goal, but even if is, the reality is that we have many dialects of JSON Schema in use today and this media type needs to support all of the dialects people are using in the wild. Then you also have to take into account the third-party dialects out there for things like code-gen, form-gen, and databases. All of these should be able to use the media-type as well. That's why the media-type needs to be flexible.

awwright:
> Dialect identification works the way it always has

I'm not sure what you're referring to, "dialect" is new term to me. Usually we just called this the meta-schema.

> the reality is that we have many dialects of JSON Schema in use today

Can you specifically name some of these dialects?

JSON Schema is not (or should not be) any meaningfully different from HTML... over time newer functions (versions) have emerged, but there is only one HTML "dialect." (In HTML 4.01 you could declare usage of one of a few DTDs, but these have all been superseded.) And if these features are not sufficient, you can use a different XML namespace.

awwright:
My understanding is that the schema identifier is provided as a convenience, and it doesn't allow someone to unilaterally redefine a keyword to be (significantly) backwards-incompatible. But it can be used as a hint to suggest that there's custom keywords in use, or that there's new keywords that the validator doesn't understand (which is effectively the same thing from the validator's point of view).

There's a few reasons I understand it this way:

First, the meta-schema is not authoritative, just informative. The spec defines the behavior and permitted values, the meta-schema tries to describe that as closely as possible; which is why we can issue revisions of a meta-schema. _Or_, at least it used to work this way.

Second, the [8.1. Meta-Schemas and Vocabularies](https://json-schema.org/draft/2020-12/json-schema-core.html#vocabulary) section begins by listing just two purposes, which makes the following section  (8.1.1. The "$schema" Keyword) seem much less ambitious than I gather from your description.

Third, version identifiers are frequently ignored. Maybe, if we absolutely had to, we could use $schema to backtrack on functionality that we decided was a bad idea. (Similar to User-Agent sniffing, how TLS uses its version identifiers, or "quirks mode" in HTML.) But using a version identifier for something more than this is generally a bad idea.

Finally, as a general Internet principle, undefined and implementation-specific behavior should be avoided, and the idea that an `application/schema+json` document would have an undefined meaning until we define the dialect in use is somewhat unprecedented. There's some similar examples but I don't think they match up:

(a) In XHTML, you had to specify the xmlns=, but this isn't quite the same thing because it's the same regardless of the HTML version, and if you omit the namespace declaration, the `application/xhtml+xml` document would fail to validate entirely.

(b) I sort of anticipated the doctype argument—it used to exist, but it turns out it didn't do much and it was eventually discontinued. The elements were always supposed to mean the same thing, and the behavior for all the various DTDs was always included in the respective media type definition.

There's a lot of implementations out there that will never implement $vocabulary or even $schema because there's no need to. The core functionality can all be implemented, with reverse-compatibility supported all the way back to the beginning (with very niche exceptions probably not encountered in the wild).

> Some examples include draft-04, draft-07, 2020-12, Open API 2 JSON Schema, Open API 3.0 JSON Schema, Open API 3.1 JSON Schema, and Mongo DB JSON Schema.

Thanks, this is helpful. I don't see a need to identify these as separate dialects. draft-04 through 2020-12 are just different releases of the meta-schema: They're a machine-readable version of the same media type specification at different points in time.

Open API 2 JSON Schema etc. don't need their own dialects either, unless they're defining custom keywords. You can implement a subset of JSON Schema functionality as long as you bail out if a schema tries to use the unsupported functionality.

jdesrosiers:
I really like this vision. It's very similar to where I want JSON Schema to end up. Where we disagree is that your interpretation can be applied to the current and historical state of JSON Schema. Every release (with the possible exception of draft-07) has had backwards incompatible changes. This makes it necessary to know which dialect defines the intended semantics for a given schema. `id` isn't just a deprecated version of `$id`. `id` doesn't have any effect at past draft-04. `$ref` works differently starting in 2019-09. The only way to know how to properly interpret it is to know what draft the schema was written for. In HTML you can evaluate newer HTML against an implementation written for older HTML with just some features missing. You can also evaluate old HTML against an implementation written for the latest HTML. The same isn't true for JSON Schema. That's why we need dialects.

In the relatively near future, I'd like to declare something like `https://json-schema.org/schema` as the last dialect we'll ever create, commit to backwards and forwards compatibility, and continue as a living spec like HTML. But, we're not there yet and even if we were, we would still need to support the older drafts somehow, not to mention dialects for things other than validation.

> Finally, as a general Internet principle, undefined and implementation-specific behavior should be avoided, and the idea that an `application/schema+json` document would have an undefined meaning until we define the dialect in use is somewhat unprecedented.

You're thinking about this differently than what is intended. The data in an `application/json` document has no semantics. It's just data. It's meaning is out-of-band knowledge baked into whatever application is using it. An `application/schema+json` document without a dialect would be exactly the same. In `application/xml` it would be the same as an XML document with no namespace, just raw values. Adding a namespace in that XML to give those values meaning is similar to declaring dialect. That's precedent we're building on. But, again, no one is expected to use `application/schema+json` this way. Technically you could use it without a dialect, but that's not what it's designed for.

awwright:
Not to get this PR too off-topic, I filed #1202. Sorry I didn't realize I had a somewhat different interpretation than everyone.

> Every release (with the possible exception of draft-07) has had backwards incompatible changes. This makes it necessary to know which dialect defines the intended semantics for a given schema. id isn't just a deprecated version of $id. id doesn't have any effect at past draft-04.

I don't think this has ever been necessary. With respect to changing spec'd behavior, my thought at the time was we would not introduce behavior that would be _incompatible_ with older drafts; implementations could continue to implement removed behavior, from the standpoint of not breaking reverse-compatibility in their own libraries.

Additionally, declaring (and being valid against) the newer meta-schema is a way of verifying that you don't use any deprecated functionality.

And the reason that we have different drafts described in the test suite is not to suggest the behavior can change outright, but so that implementations can test backwards compatibility. My own implementation passes every test in every draft (except a few of the newer features in 2019-09), yet it doesn't read "$schema" at all. Do you have any counter-examples? Are there any tests that flatly contradict themselves between drafts?

It occurs to me now, the best way to handle backwards compatibility is not to remove the functionality from the spec, but to move it into a separate section describing deprecated functionality. The downside is this is slightly more complicated to author.

Infrequently, there were also a few features that we retconned (for example: `"$ref"` only being permitted where a subschema is expected), that could potentially break someone in the wild, but we're not aware of anyone using it in a way that would break (or it would be an easy fix that could be withheld until a major version increment in an implementation), and the benefits outweighed the costs.

> The data in an application/json document has no semantics. It's just data. It's meaning is out-of-band knowledge baked into whatever application is using it. An application/schema+json document without a dialect would be exactly the same.

Normally I would agree `application/json` has no semantics, but with respect to this comparison, I would point out it sort of does: It represents a data model that can represent arrays, objects, etc. It would be strange if it did anything more than that! We sort of expect that it will gain additional meaning depending on the context it is found within.

In contrast, with JSON Schema, we do define semantics. And I wouldn't ever expect that "type" can possibly mean different things in different contexts; it's a standardized name for a standard rule.

Relequestual:
> Are there any tests that flatly contradict themselves between drafts?

Yes.

Take `$ref` for example. In draft-07 and prior, sibling keywords must be ignored.
https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/060caae0dd58e34af0449baec1103606a0ef4977/tests/draft7/ref.json#L146-L177

While in 2019-09 and above, they must NOT be ignored.
https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/060caae0dd58e34af0449baec1103606a0ef4977/tests/draft2019-09/ref.json#L146-L177

Unless a dialect is provided, how do you know the intent of the schema author?

Take `format` as another example. In draft-07 and prior, support was optional, but you could perform assertions if you wanted.
From 2019-09 and above, you strictly cannot provide assertions from the `format` keyword.

> My own implementation passes every test in every draft (except a few of the newer features in 2019-09), yet it doesn't read "$schema" at all.

Unless the behaviour of JSON Schema validation is predictable and reliable, it's useless.
That's the whole reason we split `format` into two vocabularies... predictability and reliability is paramount.

I'm all for discussing future drafts being backwards compatible without breaking changes (when we've reached acceptable stability), but that's future, and this is now.

---

> Open API 2 JSON Schema etc. don't need their own dialects either, unless they're defining custom keywords. You can implement a subset of JSON Schema functionality as long as you bail out if a schema tries to use the unsupported functionality.

That IS the exact reason they DO need their own dialect; OpenAPI defines their own custom keywords.

It's totally possible to create a dialect using the core and applicator vocabularies, but not the validation vocabulary.


---

@jdesrosiers @awwright I'm very strongly of the opinion that a "living standard" would be a huge mistake for JSON Schema.
It has already resulted in several serious security issues in terms of URLs.
(see https://json-schema.slack.com/archives/C8CQ81GKF/p1648542728065289?thread_ts=1648454348.004629&cid=C8CQ81GKF)

> Show notes/transcript (from page 7): https://www.grc.com/sn/SN-853-Notes.pdf
> The security paper/report: https://claroty.com/wp-content/uploads/2022/01/Exploiting-URL-Parsing-Confusion.pdf

I think "living standard" sort of works for WHATWG because the major vendors are at the table and commit to implement or reject and riase concerns over changes. The objective they have is to align the spec with the implementaitons, of which there are few, and over which they have control. Neither of those things are true for JSON Schema.

WHATWG reference: https://whatwg.org/faq#living-standard

I'm open to being convinced otherwise.

---

**Any implementation which attempts to mish-mash across dialects of JSON Schema is in for a hard time, and can't provide reliable and predictable validaiton. IMHO, it's a travesty. I can't imagine a situation where unreliable and unpredictable validation is even close to acceptable. This is why we have the test suite, and why JSON Schema isn't dead.**

awwright:
> Yes. Take $ref for example. In draft-07 and prior, sibling keywords must be ignored.

Ok, good point.

This is another example that I don't think is a problem in practice. In my implementation, I forgot, I do mask a few tests. It's one of those situations I described where it didn't really break anyone in the wild, and if it does, it can be implemented after a major version increment, and there's an easy fix. (In this case, if you had a draft-07 document with ignored keywords adjacent to a $ref keyword, you can rename/move them into a "$comment" keyword. This preserves the old behavior and makes it forward-compatible.)

Specifications sometimes do things like this. HTML has done this _a lot_. That doesn't mean suddenly there's different "dialects" of HTML.

> Unless a dialect is provided, how do you know the intent of the schema author?

I had the impression we made the change understanding that case doesn't (significantly) occur in the wild. And if anything, the new behavior was found in the wild more often (because of misunderstanding).

> That IS the exact reason they DO need their own dialect; OpenAPI defines their own custom keywords.

Sure, this is what I imagine as the use case for $schema. I think there's a strong case for "$schema".

In fact, I'll walk back what I said a little bit: If they want to define a "dialect" that is a _strict subset_ of JSON Schema, that would be another good case for `$schema`. Then you can verify you're conforming to that subset. I think this is analogous declaring the "HTML 4.01 Strict" DTD, it's not changing the semantics of HTML, it's just verifying you're not using deprecated behavior (e.g. frames, marquee, blink).

> I'm very strongly of the opinion that a "living standard" would be a huge mistake for JSON Schema.

Fully agreed!

I'm not citing HTML for this reason. It's just that, as the biggest hypermedia format, it's easy to cite for examples. (I don't even like that name... as opposed to what, a "dead standard?" It sounds like a snub).

awwright:
Also, if you want to sniff `"$schema"` as a hint for which behavior to use, I don't think that's wrong‚—that's perfectly fin, maybe sometimes necessary. But that doesn't mean that the behavior has to be "implementation defined". Acknowledging that some implementations don't support all the latest features is not the same thing as saying that the implemented version of the spec is implementation defined. "Implementation defined" means something more specific than that.

Relequestual:
> In this case, if you had a draft-07 document with ignored keywords adjacent to a $ref keyword, you can rename/move them into a "$comment" keyword. This preserves the old behavior and makes it forward-compatible.

Unequivocally no.

You've assumed only annotation keywords adjacent to $ref.
I feel you've made many assumptions here.

Part of the work we are looking to do allows us to evaluate implementations against the test suite by ourselves, and report that to users.

Any implementation which doesn't pass all the required tests will be "not recommended".

We actually recently refused to update the support listing of an implementation for deliberately going against the specified required behaviour.

Having proper interoperability is just so critical to JSON Schema. I can't see anyone being able to convince me otherwise.

Relequestual:
> It's one of those situations I described where it didn't really break anyone in the wild...

I consider that line of justification having no weight here.

We have a technical specification. It clearly defines requirements. We even have a test suite.

People should be able to rely on conformance to the spec. If they cannot, it's useless.

karenetheridge:
> My own implementation passes every test in every draft (except a few of the newer features in 2019-09), yet it doesn't read "$schema" at all. Do you have any counter-examples? Are there any tests that flatly contradict themselves between drafts?

In draft2020-12: "The value of "items" MUST be a valid JSON Schema. " But not so in earlier drafts -- an array was also valid (and indeed there are tests of this variant). So if your implementation accepts an array at the `items` keyword, it is acting in contradiction to the spec. (There is no test for this because at present there are no tests for invalid schemas.)

Another behaviour that is more limited than it used to be is the syntax of `$id`: a string like "#/foo/bar" was a valid `$id` in draft7, but later drafts prohibit that.

draft2019-09 also removed the `dependencies` and `definitions` keywords. Implementations conforming to a newer version should ignore those keywords, not evaluate it. And `$recursiveRef` and `$recursiveAnchor` were removed in draft2020-12.

karenetheridge:
> I never understood "$schema" to be how we differentiated between these behaviors. For example, what if you write your own meta-schema? Which $ref behavior gets used? 

The URI at `$schema` is first treated as a schema (either loaded from web/disk, if you support that, or it must previously be "known" to the evaluator). The document at that location must also have a `$schema` keyword. And so on, recursively, until and we must eventually get to a draft metaschema URI. That's where we determine what version of the spec we're using.

This is how the `$schema` keyword can be forward-compatible: we can define its meaning now, and promise that it won't be materially changed. We can release draft2022-04 tomorrow and evaluators that only support up to draft2020-12 will be able to properly give errors when it sees schemas containing `{"$schema":"https://json-schema.org/draft/2022-04/schema"}` (load that document, if we support that, and then discover that it references itself, therefore it must be a new unknown draft specification metaschema.)

So, if you want to support your own metaschema that's not based on any existing spec version, you need to build that logic into your evaluator, that is triggered when that schema URI is seen.

awwright:
> But not so in earlier drafts -- an array was also valid

But supporting the schema form is not mutually exclusive with supporting the array form.  Is there an example where supporting a current feature precludes supporting an older one?

The example of properties adjacent to $ref was given (previously they were ignored, now they're evaluated)— but my issues with this example are (1) "$ref" is a core keyword, it's behavior doesn't change with respect to "$schema", and (2) nobody was using keywords next to $ref and expecting them to _not_ do anything. We changed it because people were using them and they _were_ expecting it to do something.

> Implementations conforming to a newer version should ignore those keywords, not evaluate it.

No, it is perfectly legal to support features that have since been removed from the spec. This is how we support reverse compatibility with older drafts. We don't redefine behavior that's been specified in older drafts, it only becomes undefined. For example, we would not be able to redefine the array form of "items" with a different meaning. In any event, "definitions" was always technically ignored (it was merely reserved).

> The document at that location must also have a $schema keyword.

This isn't true. A meta-schema can provide it's own URI as it's schema, or omit it entirely.

karenetheridge:
> Is there an example where supporting a current feature precludes supporting an older one?

I gave you several -- are you not counting where earlier versions of the spec supported something and now it doesn't, even if it now says "MUST NOT"?  That's a bit disingenuous don't you think?  Are you suggesting that it is never possible to remove a keyword entirely, because there isn't a conflict if an implementation chooses to keep implementing it anyway? And some implementations continuing to support old keywords while others don't will lead to unpredictable behaviour in the wild, which is exactly what we're trying to prevent by having a specification document.

> "$ref" is a core keyword, its behavior doesn't change with respect to "$schema"

That's nothing to do with it being a core keyword. We've changed core keyword behaviours several times, and may do so again in the future. All we really need to lock down between versions is `$id` and `$schema`, and everything else flows from that.

> No, it is perfectly legal to support features that have since been removed from the spec.

Maybe, if the schema doesn't explicitly contain a `$schema` keyword, but strictly speaking no, you should pick one version of the spec (preferably the latest, unless the user specifically overrides that), and stick to it.

> In any event, "definitions" was always technically ignored

It wasn't, because schemas underneath `definitions` (and now `$defs`) needed to be examined for identifiers (`$id` and `$anchor` keywords) that other schemas elsewhere might want to reference.

> A meta-schema can provide its own URI as its schema

If it does, it better be one of the official draft specification metaschemas, or some other metaschema that is "known" to the implementation as the root source of keyword behaviours. Hyperschema is one - it has its own specification document, and cannot be implemented solely in terms of JSON Schema draft<whatever> on its own.

I suppose a metaschema could omit the `$schema` keyword entirely; the spec really should cover how this is handled. I would assume to default to the same behaviour as when a non-meta schema omits the keyword - i.e. generally default to the latest supported draft version.  But my point is that schemas referencing each other as metaschemas is a recursive process that has to end at an official spec metaspec eventually.

awwright:
> are you not counting where earlier versions of the spec supported something and now it doesn't, even if it now says "MUST NOT"?

The use of "MUST NOT" in the "$id" definition is a limitation on what authors & implementations may produce; this doesn't imply that validators must produce an error if they encounter the prohibited form. They can still support the older behavior, because it's not ambiguous.

As far as I know, none of your other examples mention MUST NOT.

>  Are you suggesting that it is never possible to remove a keyword entirely, because there isn't a conflict if an implementation chooses to keep implementing it anyway? And some implementations continuing to support old keywords while others don't will lead to unpredictable behaviour in the wild, which is exactly what we're trying to prevent by having a specification document.

Implementations can continue to support functionality removed from the spec, as a matter of not breaking reverse compatibility in their implementation.

This is not to imply that old behavior is cross-platform, of course. If we want to support deprecated functionality in a cross-platform way, we would have to write it into the spec.

> We've changed core keyword behaviours several times, and may do so again in the future.

Right, but there's nothing in the specification that says the behavior changes based on the $schema keyword.

> Maybe, if the schema doesn't explicitly contain a $schema keyword, but strictly speaking no, you should pick one version of the spec (preferably the latest, unless the user specifically overrides that), and stick to it.

This is the basis for my issue #1202. Your suggestion seems reasonable, but the spec currently suggests that validators encountering schemas without a "$schema" keyword may define any arbitrary behavior, including breaking behavior; and this sounds unreasonable to me.

> It wasn't, because schemas underneath definitions (and now $defs) needed to be examined for identifiers

Agreed, this is an important clarification.

> If it does, it better be one of the official draft specification metaschemas

This makes sense, but this behavior isn't specified in the spec.

awwright:
> Are you suggesting that it is never possible to remove a keyword entirely

Also, if by "entirely" you mean we have to pay attention to behavior long since removed—yes. This is generally the case. For example, HTTP is probably never going to redefine behavior for 402, 419, or 420.

jdesrosiers:

handrews:
@ioggstream could you please re-submit your changes as a new branch from main if you are still interested in having them reviewed and adopted?  Between the removal of draft-next and the very long tangential digression of the comments discussion (which is better covered in #1242 and should not be continued here), the PR in this form is too confusing to finalize.

I would recommend splitting the added xref (which seems straightforward as a PR) from the debate over what is or is not a data format, which could probably use clarification through discussion in an issue since there did not seem to be a consensus here.

ioggstream:
Hi, I think I could do it for September;) adding a note for that.

ioggstream:
@handrews since you only wanted the editorial parts, it was easy (see #1273  #1274 ). CBOR defines itself as a data format, so I think we can trust it :) https://cbor.io/

PR #1229, opened 2022-05-16
Describe syntax with regex.

ioggstream:
## I suggest

- describing syntax using regexp
- remove non-actionable informative content since it distracts the reader from the normative part (it can be added to a FAQ section)

## Note

The spec already references ECMA regexp

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ If the pull request would solve a particular issue, reference the issue in the p
 
 Changes that would affect implementation behavior should typically be opened as an issue first.
 
-Pull requests should be made to master.
+Pull requests should be made to the default branch, which may be `draft-patch`, `draft-next`, or `master` depending on where we are in the [release process](https://github.com/json-schema-org/community/discussions/7).
 
 Most PRs, including all PRs that impact implementation behavior, will be left open for a minimum of 14 days.  Minor wording fixes may be merged more quickly once approved by a project member.
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Welcome to JSON Schema
-[![Financial Contributors on Open Collective](https://opencollective.com/json-schema/all/badge.svg?label=financial+contributors)](https://opencollective.com/json-schema)
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](https://github.com/json-schema-org/.github/blob/main/CODE_OF_CONDUCT.md) [![Financial Contributors on Open Collective](https://opencollective.com/json-schema/all/badge.svg?label=financial+contributors)](https://opencollective.com/json-schema)
 
 JSON Schema is a vocabulary that allows you to validate, annotate, and manipulate JSON documents.
 
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -3851,6 +3851,10 @@ https://example.com/schemas/common#/$defs/count/minimum
             </t>
             <t>
                 <list style="hanging">
+                    <t hangText="draft-bhutton-json-schema-next">
+                        <list style="symbols">
+                        </list>
+                    </t>
                     <t hangText="draft-bhutton-json-schema-00">
                         <list style="symbols">
                             <t>"$schema" MAY change for embedded resources</t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2343,17 +2343,25 @@
                             This is to ensure that all possible annotations are collected.
                         </t>
                         <t>
-                            Logically, the validation result of applying the value subschema to each
-                            item in the array MUST be ORed with "false", resulting in an overall
-                            validation result.
+                            An object instance is valid against "contains" if at least one of
+                            its property values is valid against the given schema. The subschema
+                            MUST be applied to every property value even after the first match has
+                            been found, in order to collect annotations for use by other keywords.
+                            This is to ensure that all possible annotations are collected.
                         </t>
                         <t>
-                            This keyword produces an annotation value which is an array of
-                            the indexes to which this keyword validates successfully when applying
-                            its subschema, in ascending order. The value MAY be a boolean "true" if
-                            the subschema validates successfully when applied to every index of the
-                            instance. The annotation MUST be present if the instance array to which
-                            this keyword's schema applies is empty.
+                            Logically, the validation result of applying the value subschema to each
+                            item in the array or property in the object MUST be ORed with "false",
+                            resulting in an overall validation result.
+                        </t>
+                        <t>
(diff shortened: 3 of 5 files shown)


gregsdennis:
If we change this I think I'd prefer ABNF instead, as it's more common in specifications for describing syntax.

awwright:
What purpose does this achieve? Do you think this paragraph is actually confusing or could be improved? Could you elaborate on your experience a bit?

Like @gregsdennis points out, if the rationale is to make the description machine-readable, ABNF would be preferred. While we technically could use the ECMA RegExp syntax (they describe the same thing), the ABNF is more accessible to the standards community at large; and this usage of the ECMA syntax is too ambiguous (it lacks the slashes that quote the RegExp, and it lacks start and end anchors that are implicit in ABNF, but not in RegExp... we would want to write an actual line of ECMAScript).

And the implementation guidance should be left in-line; explanations like this are critical for helping readers make connections between related concepts _before_ they move on to the next section. In this case, readers will make the connection that the syntax is the same as in XML; even if they haven't read the XML spec, they're likely to have seen this pattern before.

karenetheridge:
This regex is not correct.

- it must be anchored with `^` and `$`
- "any number" can be zero, so `+` should be `*`
- literal hyphen characters must be at the start or the end of the character class, so they are not interpreted as a range.

Therefore: `^[A-Za-z_][A-Za-z0-9._-]*$`

ioggstream:
I don't have preferences between abnf and regexp.
Here and elsewhere in this document, both are ok  as long as it's not in prose :)

> we would want to write an actual line of ECMAScript

+1 Since this spec doesn't use ABNF, I used ecma to avoid adding normative references.

About the reference to xml, I think it does not help that much since the regexp is easy.

In general I find this document very long: this results in readers just skipping this and just look for secondary sources :)

handrews:
@ioggstream the point of the reference to XML is that the [WC3 Best Practices for Fragment Identifiers and Media Type Definitions](https://www.w3.org/TR/fragid-best-practices/#structures) (which is also cited elsewhere in JSON Schema Core) references the XML NCName production as part of their best practices for plain name frag ids (see the large quote below for details).

There is more going on in this section than just how to represent the syntax (I agree with @gregsdennis and @awwright that ABNF is preferable).  The point of this part of the spec is to tie JSON Schema's plain name fragments to the larger standards ecosystem of such things.  That could be done more clearly and concisely, but it should not be removed.

> Plain names are a common type of fragid structure. A plain name fragid is a [fragid](https://www.w3.org/TR/fragid-best-practices/#dfn-fragid) that is used to identify a named structure within a document, such as one identified by an @id attribute in HTML, a @xml:id attribute in XML or the name of a function within a Python program. These fragids are opaque to processors and as such they do not normally include punctuation characters, though this depends on the language: in XML, for example, they usually match the [NCName](http://www.w3.org/TR/xml-names11/#NT-NCName) production from [XML Namespaces](http://www.w3.org/TR/xml-names11/) [[XML-NAMES11](https://www.w3.org/TR/fragid-best-practices/#bib-XML-NAMES11)] which means they can contain hyphens (-) and periods (.).
>
> Plain name fragids are usually created by human authors but may also be generated by applications. They provide a good method of identifying content that is equivalent across content-negotiated variants of a document, for example paragraphs of text in French and Chinese that contain the same semantic content. Plain name fragids that do not identify a portion of a document are frequently used in Semantic Web applications as a way of providing an identifier for something described by the document.
>
> ***Best Practice 3: Reserve Plain Name Fragids***
>
> *If the media type includes structures that can be given local names or identifiers, [plain name fragids](https://www.w3.org/TR/fragid-best-practices/#dfn-plain-name-fragid) should be reserved for addressing those structures.*

ioggstream:
> ... e [WC3 Best Practices for Fragment Identifiers and Media Type Definitions](https://www.w3.org/TR/fragid-best-practices/#structures) (which is also cited elsewhere in JSON Schema Core) references the XML NCName

Ok, it wasn't clear to me when reading the section, and I think it should be clarified e.g. in a referenced  Appendix (e.g. JSON Schema and XML Schema ...) so that the reader can better identify normative parts.

handrews:
@ioggstream I don't think there's any need for an appendix here.  All this section needs to do is:

1. `xref` the Best Practices document to explain _why_ this syntax (XML's NCName) is used (not currently done in this section)
2. `xref` the XML Namesaces spec for the normative NCName production (currently done, just not clear why or whether it's normative)
3. Reproduce NCName ABNF in a simplified form for informative convenience (currently done in prose, and presented as normative)

awwright:
> 1. (not currently done in this section)

We do discuss it elsewhere though, in <https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.5>.

> just not clear why or whether it's normative

XML technically supports Unicode characters and for ease of getting the first draft out we opted not to support that. We can change that now.

Even if XML only specified the same ASCII set that we do, it might still be easier to explain "use this pattern ... which is selected to match other popular syntaxes including XML" rather than "refer to the XML definition."

handrews:
@awwright 

> We do discuss it elsewhere though

yes, that's why I said "in this section" 🙂 

> Even if XML only specified the same ASCII set that we do, it might still be easier to explain "use this pattern ... which is selected to match other popular syntaxes including XML" rather than "refer to the XML definition."

Again, the important reference is the W3C Best Practices document, specifically the section on plain name fragments.  It _just so happens_ that that spec references the XML Namespaces specification, so there is a connection there.  But the motivation here is ***not*** "align with XML."  It is "align with W3C Best Practices regarding fragments.  (I think it would be fine to expand that to Unicode because it's 2022 FFS).

The reference to XML Namespace's "NCName" production only makes sense in the context of "We are following the WC3 Best Practices for plain name fragments, which suggest using the NCName production from the XML namespace spec." (tidied up for clarity regarding what is normative, of course).

The reason to reference the XML Fragment spec is to make it clear what we are trying to do, so that if the ABNF in our spec does not match, someone will be able to tell that that was an unintentional error and not an intentional deviation.

handrews:
And given the number of times various people have messed with the plain name syntax description, keeping that external reference (both of them, really) as a correctness check feels valuable to me.

jdesrosiers:

gregsdennis:
I think for this particular change, I would prefer leaving the text but using a `pattern` (which uses regex) in the meta-schema.  If the syntax really _must_ be specified in some common format, I'd still prefer ABNF over regex, but I think it's overkill.

handrews:
@ioggstream this PR doesn't seem to have a consensus around what is needed, and the change in branching has made it difficult to review.  If you would still like something done here, could you please file an issue summarizing the options that have come up in this discussion so that we can debate it properly?

Since no one has spoken up in favor of a regex in the text, I'm going to go ahead and close this.  If, after discussion in an issue, we agree that we should use a regex for this, it can be re-submitted.

PR #1238, opened 2022-05-26
GSD tweaks for 2020 12 patch

### LINKED ISSUE #1221: 2020-12-patch-1-RC-0 wording issues
[`additionalItems` abiguity cref, L17](https://github.com/json-schema-org/json-schema-spec/compare/2020-12...draft-bhutton--01-rc0#diff-4df123eae0f368edcbd04439e500ef4ed52d8a7a2479fb3d303d1a49d49df3a7R17)

> When we changed the specification to use annotations as the context in which some keywords behave, we included a clause that implementations which didn't use annotations may implement or optimize the processing of `additionalProperties` in another way which produces the same effect as the prior behaviour.

This sentence takes several reads to understand.  Suggestion:

> When we changed the specification to use annotations as the context in which some keywords behave, we included a clause that allowed implementations which didn't use annotations to optimize the processing of `additionalProperties` in another way in order to produce the same effect as the prior behaviour.

---

[core, L1370](https://github.com/json-schema-org/json-schema-spec/compare/2020-12...draft-bhutton--01-rc0#diff-bda51aab56b4b2e23fc9e7249e80fe6e262dcd552f35c5de69fd4a6bf3e96b50R1370)

> Therefore, "$id" MUST NOT contain a non-empty fragment, and SHOULD NOT contain an empty fragment.

This _sounds_ contradictory.  It may not be, but it sounds like it is.  Can we add a couple examples of forbidden or unrecommended URIs?

I see that [lines 1875-1877](https://github.com/json-schema-org/json-schema-spec/compare/2020-12...draft-bhutton--01-rc0#diff-bda51aab56b4b2e23fc9e7249e80fe6e262dcd552f35c5de69fd4a6bf3e96b50R1875-R1877) refer to an appendix, but I don't see such invalid examples listed there.

---

[validation, L444-447](https://github.com/json-schema-org/json-schema-spec/compare/2020-12...draft-bhutton--01-rc0#diff-a975ce3eecafc70bc583c303965f3d4fce74d8652cd97c69d1464ba5b33f8e5bR444-R447)

Changed from 

> A value of 0 is allowed, but is only useful for setting a range of occurrences from 0 to the value of "maxContains".  A value of 0 with no "maxContains" causes "contains" to always pass validation.

to

> A value of 0 is allowed, but is only useful for setting a range of occurrences from 0 to the value of "maxContains". A value of 0 causes "minContains" to always pass validation (but validation can still fail against a "maxContains" keyword).

There is no longer any mention that `minContains : 0` causes `contains` to pass.  I think this is important as the instance may not contain an item that matches `contains`.

There _was_ mention in [core, L2375-2379](https://github.com/json-schema-org/json-schema-spec/compare/2020-12...draft-bhutton--01-rc0#diff-bda51aab56b4b2e23fc9e7249e80fe6e262dcd552f35c5de69fd4a6bf3e96b50R2375-R2379), but maybe it bears repeating here.

gregsdennis:
Resolves #1221 remaining issues.

### DIFF
--- a/adr/2022-04-08-cref-for-ambiguity-and-fix-later-gh-spec-issue-1172.md
+++ b/adr/2022-04-08-cref-for-ambiguity-and-fix-later-gh-spec-issue-1172.md
@@ -5,6 +5,7 @@
 * Date:  2022-04-08
 
 Related...
+
 Issue: https://github.com/json-schema-org/json-schema-spec/issues/1172
 
 Discussion: https://github.com/json-schema-org/community/discussions/57
--- a/adr/2022-04-08-cref-for-ambiguity-and-fix-later-gh-spec-issue-1172.md
+++ b/adr/2022-04-08-cref-for-ambiguity-and-fix-later-gh-spec-issue-1172.md
@@ -1,8 +1,8 @@
 # Acknowledge ambiguity in additionalProperties behaviour and fix after patch release
 
-* Status: proposed
+* Status: accepted
 * Deciders: @relequestual @gregsdennis, @jdesrosiers, @karenetheridge
-* Date:  2022-04-08
+* Date:  2022-05-19
 
 Related...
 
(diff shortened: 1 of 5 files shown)


gregsdennis:
ugh... what happened?  this was a clean pull of latest!  I changed a single line!

gregsdennis:
oh... it defaulted to draft-next.

PR #1244, opened 2022-06-14
Add some paragraphs to Overview to put some context to "vocabulary"

awwright:
The Overview is the first section of the document that describes _how_ JSON Schema accomplishes the goals laid out in the abstract and the introduction.

Currently, it jumps to talking about vocabularies and keywords before the definition can be meaningful to readers. This PR adds some context to that, and provides a good primer for the essential mechanics of how schemas are read, at a more "architectural" level.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -127,19 +127,36 @@
 
         <section title="Overview">
             <t>
-                This document proposes a new media type "application/schema+json" to identify a JSON
-                Schema for describing JSON data.
-                It also proposes a further optional media type, "application/schema-instance+json",
-                to provide additional integration features.
+                This document specifies the media type "application/schema+json" which
+                can describe a set of JSON documents, classify documents as instances of this set,
+                and to generate annotations about a given instance.
+                It also specifies "application/schema-instance+json",
+                to describe JSON documents known to be instances of a particular schema.
                 JSON Schemas are themselves JSON documents.
-                This, and related specifications, define keywords allowing authors to describe JSON
-                data in several ways.
             </t>
             <t>
-                JSON Schema uses keywords to assert constraints on JSON instances or annotate those
-                instances with additional information.  Additional keywords are used to apply
-                assertions and annotations to more complex JSON data structures, or based on
-                some sort of condition.
+                A JSON Schema document consists of a set of keywords,
+                typically encoded as properties in a JSON object.
+                The name of the keyword is used as the property name,
+                and any arguments to the keyword are provided as the value.
+            </t>
+            <t>
+                Each keyword provides an assertion and/or annotations about the instance.
+                Assertions add constraints that instances must conform to.
+                Given a schema and an instance, the schema "accepts" an instance whenever all the assertions are met,
+                and the schema "rejects" when any of the assertions fail.
+                Schemas may also be used to build a set of JSON documents:
+                The "valid set" of a schema consists of all instances that the schema accepts,
+                and the "invalid set" of a schema consists of all instances that the schema rejects.
+                In a schema without any assertion keywords, the set of all instances is the set of all JSON documents.
+            </t>
+            <t>
+                Schemas may also provide "annotations" to instances:
+                metadata that describes the instance.
+                For example, you can document the meaning of a property,
+                suggest a default value for new instances,
+                generate a list of hyperlinks from the instance,
+                or declare relationships between data.
             </t>
             <t>
                 To facilitate re-use, keywords can be organized into vocabularies. A vocabulary


jdesrosiers:

jsonschema-core.xml
```diff
@@ -135,13 +135,66 @@
...
+                A JSON Schema document is comprised of a set of keywords,
+                which are specified as properties in a JSON object.
+                The name of the keyword is used as the property name,
+                and any arguments to the keyword are provided as the value.
+                Keywords can be classified by their functions.
+                The most important classes are validation, annotation, and applicator keywords.
             </t>
             <t>
-                To facilitate re-use, keywords can be organized into vocabularies. A vocabulary
+                "Validation keywords" form the basis of JSON Schema.
```
Julian, line 146: Is it true that we consider validation to be any more important than other uses? I was under the impression we very much wanted to avoid making it seem that way, but this language would seem to conflict.

Also I think you're changing established vocabulary (no pun intended) here, unless some discussion has happened already that I wasn't privy to -- the original term is "assertion", so unless there's good reason to do so it seems wise to avoid churning terminology, and to not do so alongside a PR which is more about making the doc read smoother.

jsonschema-core.xml
```diff
@@ -135,13 +135,66 @@
...
+                Given a schema and an instance, the instance is "valid" if all of the keywords in
+                the schema are valid against the instance.
+                Validation keywords may be used to describe a set of JSON documents.
+                Without any validation keywords, the set of all valid instances is the set of all valid JSON documents.
```
Julian, line 150: Do people or specifications use the term "JSON document" to refer to non-object typed JSON? If not this is a bit imprecise.

jsonschema-core.xml
```diff
@@ -135,13 +135,66 @@
...
+                Typically, adding validation keywords creates a subset.
+                Keywords are typically defined to not be redundant with other keywords:
```
Julian, line 152: I don't know if redundancy is the right word for this (or whether putting it this early is needed at all). I'd probably not mention this level of detail at this point, it's more philosophical than really an overview, and it's also not immediately clear to me just how true it is, there's lots of keywords that "shrink" the set of valid instances across different types at once, like `const` or `enum`, or keywords added in a `not` schema, or ...

jsonschema-core.xml
```diff
@@ -135,13 +135,66 @@
...
+                For example, the "minimum" keyword only shrinks the valid set of numbers, but not strings or other
+                types&#x2014;this is what the "type" keyword does.
+            </t>
+            <t>
+                When an instance validates against a schema, "annotation keywords" may provide "annotations":
```
Julian, line 157: Regardless of the above (whether validation is more important), do we call the process of applying annotations "validation"? I think it's just an independent process one can do even without performing validation, no?

Julian (commented):
Some initial comments in case their helpful, I only read the first half or so carefully. Overall seems like a decent idea not to jump straight into vocabularies, but personally I'm not 100% sure we should go into quite as much detail as what's here, and I think there are a few things in here that don't necessarily fit the current model.

jsonschema-core.xml
awwright: I think it's fair to say validation is the basis of JSON Schema—there's no part of JSON Schema that's not touched by validation in some manner. E.g. The difference between `allOf` and `anyOf` is only validation. There's no similar keyword where the difference is only annotation. (Maybe, hypothetically, there's a keyword that passes through only the validation result, but filters out annotation results.)

For "assertion" vs. "validation", you're right, but we seem to be using it inconsistently throughout the document. Maybe we should open a separate issue for that.

jsonschema-core.xml
awwright: Maybe a different term should be used because JSON doesn't use the term "JSON document."

But "document" has an Internet meaning, so "JSON document" should imply "document with media type `application/json`"

awwright: Also, older versions of JSON required that documents be an array or an object... this restriction has since been lifted, a document can be simply `true` or `null` or a number.

jsonschema-core.xml
Julian: I don't think others will agree with your first paragraph, especially so considering all the relevant assertion keyword are in a separate vocabulary even. But let's see.

jsonschema-core.xml
awwright: The phrasing can be played with, but I think it's important to point out that `minLength` doesn't impact numbers and there's a good, predictable reason for that.

Julian: To me it's again not very important to point out at this point, but if we do so I don't have too much of an issue with the way we've described the concept so far, namely that keywords may apply to a specific type, and if so, generally will then ignore (consider valid) instances of other types.

jsonschema-core.xml
awwright: The language emphasizes that annotations are a byproduct of validation (an invalid instance doesn't produce annotations). Mostly, I haven't introduced the term "applying" yet, and I don't think "validation" is _wrong_.

jsonschema-core.xml
Julian: Right, as I say I just don't know that people use the word document to refer to those. Values maybe. But I could be wrong.

awwright:
That's exactly what I'm looking for. In most cases I'm not sure what language would improve on the issues you identified, maybe you can suggest something?

jsonschema-core.xml
Julian: I am saying I don't think that that's how we talk about annotations previously. One can "process" an instance, not validate anything, and collect some annotations as far as I know, and we support and encourage that use case.

jsonschema-core.xml
handrews: @awwright as you know, I won't have bandwidth to fully respond for another week, but a few quick things:

* While validation is foundational to many things, I agree with @Julian that we should not promote it as the leading case.  It is possible to use JSON Schema core, applicators, annotations, and locations (custom vocabularies aside from core) without any assertions.  I did this for a client to create a complex configuration system (the configurations could also be validated by JSON Schema, but they were not assertion-based themselves).
* "assertion" is the keyword classification, "validation" is the overall process which involves more than just assertions.
* In general, this is all directly overlapping and/or conflicting with my much-delayed (and, as you know, delayed again) keyword behavior presentation.  So I can't really give proper feedback until that's done, for the same reaons that I've declined to dribble out bits and pieces of the presentation.

jsonschema-core.xml
gregsdennis: The production of annotations requires validation because no annotations are returned if validation fails.

Conversely many validation keywords require annotation output from other keywords.

They are inextricably linked.

jsonschema-core.xml
awwright: What would you say forms the "basis of JSON Schema"? Recursively enumerating the value space of JSON?

jsonschema-core.xml
Julian: So I can't define a keyword with annotation result without saying it participates in some validation process? I guess I could have answered that by looking at what we say for title or description or default. I stand corrected then, especially since handrews in another comment also said we do indeed call the general process validation.

handrews: If your schema has no assertions (and you don't use applicators that convert a passing subschema to a validation failure, e.g. `not` and `oneOf`) it can't fail validation, therefore validation is only involved to a trivial degree.  And as I noted before, yes, I have used JSON Schema that way (sadly the project is not public).

jsonschema-core.xml
awwright: I changed it to say "assertion keyword", as for what forms the "basis" of JSON Schema that seems to have picked up below.

Julian:
> That's exactly what I'm looking for. In most cases I'm not sure what language would improve on the issues you identified, maybe you can suggest something?

Sorry, just saw this one too, and I assume it's in response to one of my comments above, will try to suggest something!

jsonschema-core.xml
Julian: Looking back again I'm honestly confused now by the bottom of this :D -- @handrews (I know you're not fully responsive so feel free to delay here obviously) -- you +1'ed my initial comment, but then Greg contradicted it pointing out it was wrong, and then your last comment seems ambiguous :D so I again am confused as to whether I'm wrong or not, so you'll forgive me for asking it again:

What do we call the process itself of applying (where I'm not using a technical term here by "applying") the schema `{"title" : "Cool Stuff", "default": 12}` to an instance `"foo"` in an implementation with no support whatsoever for the validation vocabulary, it's some completely different use of JSON Schema as it sounds like the project you're referencing may have done.

Do we still call that process "validation", just now in a vacuous sense?

jsonschema-core.xml
```diff
@@ -135,13 +135,67 @@
...
-                JSON Schema uses keywords to assert constraints on JSON instances or annotate those
-                instances with additional information.  Additional keywords are used to apply
-                assertions and annotations to more complex JSON data structures, or based on
-                some sort of condition.
+                A JSON Schema document is comprised of a set of keywords,
+                which are specified as properties in a JSON object.
+                The name of the keyword is used as the property name,
+                and any arguments to the keyword are provided as the value.
+                Keywords can be classified by their functions.
+                The most important classes are assertion, annotation, and applicator keywords.
```
gregsdennis, line 143: It may be important to recognize that not all keyword provide any of these functions (e.g. `$id`) and some provide more than one (e.g. `properties`).

jsonschema-core.xml
gregsdennis: I think "validation" is currently used, but I like the more generic "evaluation."  "Apply" or "application" may be construed as related to "applicator" keywords specifically.

jsonschema-core.xml
```diff
@@ -135,13 +135,67 @@
...
+            <t>
+                "Applicator keywords" are keywords that accept sub-schemas as part of their argument,
+                which are validated against the instance in some fashion.
+                Applicator keywords may use or modify the validation result of their sub-schemas,
+                for example, to be valid conditional on which sub-schemas are valid
+                (e.g. the "if", "then", "oneOf", or "anyOf" keywords).
+                They may also validate a schema against values within the instance,
+                for example, a specific property within the object (the "properties" keyword).
+                The "$ref" keyword is a special applicator keyword that takes a reference to a schema,
+                instead of a literal schema. This allows schemas to be defined recursively.
```
gregsdennis, line 173: it also allows schemas to be broken into manageable bits and promote code reuse.

jsonschema-core.xml
```diff
@@ -135,13 +135,67 @@
...
+            </t>
+            <t>
+                Since keywords are listed in an object by their name,
+                they may be used at most once per schema and sub-schema.
+                However, the effect of multiple keywords may be achieved by listing each keyword in its own sub-schema,
+                inside an applicator keyword designed for this purpose, such as the "allOf" keyword.
+                Additionally, many keywords can often be reduced to a single keyword;
+                for example, when using "minimum", only the highest value of many will be effective.
+                Schemas may also be a boolean value:
+                The booleans true/false represent a schema that's always valid/invalid, respectively.
```
gregsdennis, line 189: Does this statement deserve its own paragraph?  It seems unrelated to this paragraph.

jsonschema-core.xml
```diff
@@ -135,13 +135,67 @@
...
+            </t>
+            <t>
+                JSON Schema documents can themselves be described by a schema.
+                In this context, this "schema of the schema" is called the meta-schema.
+                In addition to a core vocabulary and a default validation vocabulary,
```
gregsdennis, line 194:
```suggestion
                In addition to a core vocabulary and a set of default vocabularies,
```

gregsdennis (changes requested):

jsonschema-core.xml
```diff
@@ -135,13 +135,67 @@
...
             </t>
             <t>
-                JSON Schema uses keywords to assert constraints on JSON instances or annotate those
-                instances with additional information.  Additional keywords are used to apply
-                assertions and annotations to more complex JSON data structures, or based on
-                some sort of condition.
+                A JSON Schema document is comprised of a set of keywords,
+                which are specified as properties in a JSON object.
+                The name of the keyword is used as the property name,
+                and any arguments to the keyword are provided as the value.
```
notEthan, line 141: I don't think the the spec uses the term "arguments" like this anywhere else. I'm not sure it's the best word for keyword values - it does make sense to me, but wording that's more consistent with the rest of the spec would be better.

jsonschema-core.xml
handrews: I'd go with "evaluation."  I think "apply" is also fine whether the application is because of an applicator or because of a top-level invocation, but to avoid confusion, "evaluate" is probably best (and I think we settled on `evaluationPath` in the revised output format?)

Whether or not you can have annotation without validation depends on how you conceptualize it:

1. "Annotations happen unless validation fails"
2. "Annotations only happen if validation succeeds"

A schema that does not contain any keywords that can produce a validation failure by definition cannot fail validation.  If you think of that as performing validation that succeeds, and only then having annotations, then technically you always have validation.

If you think of validation as a process that only occurs when it's possible to fail (meaning that something like `title` wouldn't really be performing validation because it can't cause a failure), then a schema with no keywords that can fail isn't doing validation but it does produce annotations.

I conceptualize it as option 1:  the schema `{"title": "foo"}` produces an annotation.  Since there are no assertions, and no applicators that can fail validation in the absence of assertions, then there is no validation going on that would drop those annotations.

But you can think of it the other way around and it works just as well.

And as @gregsdennis said, you have some assertions that require** annotations to function correctly, so you could also say that annotation is just as fundamental.  Or you could say that if none of your keywords produce annotations, then no annotation is happening, just validation.

Which I think proves that neither should be framed as the primary behavior.

All of this is addressed in _a lot_ more detail in the presentation that I should work on rather than answering all of this stuff, but answering short comments is easier right now.

**there's some wording to allow people to do something different in terms of actual code, but conceptually they require annotations

awwright:
I trimmed this down a bit. This meshes nicely with #1365 and should go in first, and it sets up for some changes that will likely be made in that issue.

jsonschema-core.xml
```diff
@@ -127,19 +127,36 @@
...
+            </t>
+            <t>
+                Each keyword provides an assertion and/or annotations about the instance.
+                Assertions add constraints that instances must conform to.
+                Given a schema and an instance, the schema "accepts" an instance whenever all the assertions are met,
+                and the schema "rejects" when any of the assertions fail.
+                Schemas may also be used to build a set of JSON documents:
+                The "valid set" of a schema consists of all instances that the schema accepts,
+                and the "invalid set" of a schema consists of all instances that the schema rejects.
+                In a schema without any assertion keywords, the set of all instances is the set of all JSON documents.
```
gregsdennis, line 151: How necessary is it to bifurcate JSON documents into sets (explicitly)?  It seems like this is an understood consequence of a schema passing or failing an instance.

jsonschema-core.xml
```diff
@@ -127,19 +127,36 @@
...
+                typically encoded as properties in a JSON object.
+                The name of the keyword is used as the property name,
+                and any arguments to the keyword are provided as the value.
+            </t>
+            <t>
+                Each keyword provides an assertion and/or annotations about the instance.
+                Assertions add constraints that instances must conform to.
+                Given a schema and an instance, the schema "accepts" an instance whenever all the assertions are met,
+                and the schema "rejects" when any of the assertions fail.
+                Schemas may also be used to build a set of JSON documents:
```
gregsdennis, line 148: "build a set" sounds like you're entering data generation territory, which is not something we've discussed anywhere. (Yeah, I have a library that does that, but it's my design, not really based on any conversation here.)

gregsdennis (changes requested):

jsonschema-core.xml
awwright: I intend to mean "build" as in "set-builder notation" that JSON Schema closely resembles, e.g. `{x | x is an even positive integer}`. Is there a better phrasing? I think it's worth it to point out that describing requirements also can be used as a description of a set of members meeting those requirements.

jsonschema-core.xml
awwright: Would it be better to take out "invalid set"? It might not warrant explaining, at least not in this section.

jsonschema-core.xml
awwright: I don't believe there's a formal term for these otherwise I would put that here... If we did have a term, I think it'd still be helpful to say it's an argument, the same way that functions have arguments. Do you have an idea for a better term?

jsonschema-core.xml
gregsdennis: I'm wondering why the sets are mentioned at all.  This statement doesn't seem to add anything beyond a definition.

awwright: I think the overview is a natural place to show how JSON Schema relates formal language theory to JSON. I can try to rework it so it doesn't sound like a definition, but I'd like to emphasize this is a logical consequence of how formal languages work, not a use case that we're meeting.

gregsdennis: I agree that the intro is the right place, but I question its necessity.

Schemas defining sets of JSON instances isn't a concept that's discussed anywhere else in the document.  As such, presenting this concept at all seems confusing.

If it were referenced somewhere else or provided a framework to explain some other idea, I could understand its inclusion, but not on its own.

jsonschema-core.xml
```diff
@@ -127,19 +127,36 @@
...
+                The "valid set" of a schema consists of all instances that the schema accepts,
+                and the "invalid set" of a schema consists of all instances that the schema rejects.
+                In a schema without any assertion keywords, the set of all instances is the set of all JSON documents.
+            </t>
+            <t>
+                Schemas may also provide "annotations" to instances:
+                metadata that describes the instance.
```
jdesrosiers, line 155: I find this sentence awkward and I'm not sure this is a correct use of `:`. It's clear what you're saying, but I had to read it twice and slowly to follow it. I suggest breaking it into two sentences to make it easier for readers.

```suggestion
                Schemas may also provide "annotations" to instances. Annotations are
                metadata that describe the instance.
```

jsonschema-core.xml
jdesrosiers, line 148: I think this should end with a `.` rather than a `:`.

```suggestion
                Schemas may also be used to build a set of JSON documents.
```

jdesrosiers (approved):
Looks good to me other than a couple editorial suggestions.

jsonschema-core.xml
awwright: You have a fair point, I'm trying to figure out if there's a better way to convey what this is intending to convey.

I'm not defining a concept that's going to be used by the rest of the document; rather it's describing a feature from formal languages that JSON Schema provides.

I have two alternate ideas I want to try...

jsonschema-core.xml
gregsdennis: I think maybe "define" or "denote" instead of "build" might be better?

Issue #1172, opened 2022-01-13
ambiguous behaviour of additionalProperties when invalid


ItsVeryWindy:
I was testing a schema and getting an additionalProperties error that didn't make much sense to me as the name of the property was in the schema.

So I put together a smaller schema to reproduce it.

```json
{
    "type": "object",
    "properties": {
        "test": {
            "type": "string"
        }
    },
    "additionalProperties": false
}
```

I tried testing it against this piece of json.

```json
{
  "test": 1
}
```

I tried it against a few of the online validators and depending on which implementation I tried the errors that I would get are different, sometimes I would just get the error about `test` not being a string, other times I would also get an error about no additional properties.

Reading http://json-schema.org/understanding-json-schema/reference/object.html#additional-properties

Specifically around `properties whose names are not listed in the properties keyword or match any of the regular expressions in the patternProperties keyword` implies that only the property name is important, not if it's valid or not.

I know that some of these implementations use annotations in order to determine if a property is additional or not, and those annotations will get dropped if any properties are invalid as referenced here, which is probably why I'm seeing what I am.

https://github.com/json-schema-org/json-schema-spec/issues/939

Is there any expectations as to how additionalProperties is meant to work in these instances?

karenetheridge:
That sounds like the issue raised in https://github.com/json-schema-org/community/discussions/57 and addressed for the next interim release in https://github.com/json-schema-org/community/discussions/67.

Relequestual:
I’m wondering if we read https://github.com/json-schema-org/community/discussions/57 incorrectly.

> Schema objects that produce a false assertion result MUST NOT produce any annotation results, whether from their own keywords or from keywords in subschemas.
https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.7.7.1.2


> The annotation result of [the properties] keyword is the set of instance property names matched by this keyword.
https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.2.1

>In addition to possibly defining annotation results of their own, applicator keywords aggregate the annotations collected in their subschema(s) or referenced schema(s).
https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.7.7.1.3

I read this as, the `properties` keyword makes its own annotations, which are not effected by the assertion results of the schema values of its object. The same is true for `patternProperties`.

jviotti:
[...]

PR #1249, opened 2022-06-27
Output nodes should be per subschema not per keyword

gregsdennis:
This is quite a change, so I'd be happy to walk anyone through it.

## Summary

Following the idea from [this thread](https://github.com/orgs/json-schema-org/discussions/63#discussioncomment-2960665) and [this thread](https://github.com/orgs/json-schema-org/discussions/63#discussioncomment-2968029), this PR changes the concept of the output unit from being based on individual output from keywords to being based on output from schemas.  The benefits of this are in the thread, so I won't rehash them here.  The primary difference is that errors and annotations are included in the output unit as properties (objects with keyword names for keys) rather than as nested units.

It also does a few other things:

- removes the `detailed` format
- renames the `verbose` format `hierarchical` (since there's only one now)
- adds clarification in some of the language
- adds passing examples to show how annotations are included

Doing this made the formats considerably more information-dense, which meant that the verbose examples now reasonably fit in the spec rather than needing to be defined in an external file.

All of this is pretty related, otherwise I would have broken it into multiple PRs.


gregsdennis:
I have implemented this in the [`schema/experiment-new-output-format` branch](https://github.com/gregsdennis/json-everything/tree/9d27047d1d1d49d9c4729f1e85cc819fc153c8f0) of `json-everything`

jdesrosiers:

karenetheridge:
I am strongly against parts of this PR but I am unable to spend much time going through it again in detail until the end of the week.

jsonschema-core.xml
```diff
@@ -2727,18 +2726,18 @@
...
-                        <t>
-                            Verbose - Provides validation information in an uncondensed hierarchical
-                            structure that matches the exact structure of the schema.
+                            Hierarchical - Provides validation information in a hierarchical
+                            structure that mimics the structure of the schema.
+                            <cref>
+                                To be precise, this structure follows the paths that a validator
+                                would traverse while processing the schema, including composition
+                                via by-reference keywords.
+                            </cref>
```
handrews, line 2735: Can we say "that matches the evaluation paths of the schema" or something like that?  Since we use `evaluationPath` as the name for identifying points within this hierarchical process.  Perhaps "evaluation structure"?  This would let you drop the CREF.

I also like the idea of talking about an "evaluation structure" because that provides context for the `evaluationPath` JSON Pointers.  There have been a few complaints about these pointers not corresponding to a document, which is not entirely true:  they do not correspond to a _persistent_ document, but they do correspond to the document structure traced by evaluation.  Naming that structure will make it easier to talk about that correspondence.

jsonschema-core.xml
```diff
@@ -2757,9 +2756,9 @@
                     Implementations MAY elect to provide additional information.
                 </t>
 
-                <section title="Keyword Relative Location">
+                <section title="Subschema Relative Location">
```
handrews, line 2759: I think it would make more sense to talk about the dynamic scope location, or evaluation path, or something.  For one thing, this includes root schemas as well as subschemas.  And for another, as I've been digging into keyword behaviors, I've really wanted to be more clear and consistent on the difference between a dynamic scope (the application of a schema to an instance, identified by `evaluationPath` + `instanceLocation`) and a schema (identified by a URI).

This is the distinction that the whole "relative location" vs "absolute location" was trying to get at, and talking about dynamic scopes vs schema locations is much more accurate.

jsonschema-core.xml
```diff
@@ -2757,9 +2756,9 @@
...
                     <t>
-                        The relative location of the validating keyword that follows the path
+                        The relative location of the validating subschema that follows the path
```
handrews, line 2761: I kind of want to take "relative" out of this altogether, particularly since we are dropping "absolute" from `schemaLocation`.  They only really made sense in contrast to each other.

jsonschema-core.xml
```diff
@@ -2777,22 +2776,20 @@
                         </artwork>
                     </figure>
                     <t>
-                    Note that this pointer may not be resolvable by the normal JSON Pointer process
-                    due to the inclusion of these by-reference applicator keywords.
+                        Note that this pointer may not be resolvable by the normal JSON Pointer process
+                        due to the inclusion of these by-reference applicator keywords.
```
handrews, line 2780: If we talk about the evaluation paths as a transient document, and `evaluationPath` as pointers for that document, then we can drop notes like this.  We may want to talk about that in the section on dynamic scopes rather than here, so that it is defined earlier in the spec.

jsonschema-core.xml
```diff
@@ -2777,22 +2776,20 @@
...
                     </t>
                     <t>
                         The JSON key for this information is "evaluationPath".
                     </t>
                 </section>
 
-                <section title="Keyword Absolute Location">
+                <section title="Subschema Absolute Location">
```
handrews, line 2787: Aren't we just making this "Schema Location"?  "absolute" and "relative" just ended up being confusing.

jsonschema-core.xml
```diff
@@ -2777,22 +2776,20 @@
...
                     <t>
-                        The absolute, dereferenced location of the validating keyword.  The value MUST
-                        be expressed as a full IRI using the canonical IRI of the relevant schema resource
-                        with a JSON Pointer fragment, and it MUST NOT include by-reference applicators
-                        such as "$ref" or "$dynamicRef" as non-terminal path components.
-                        It MAY end in such keywords if the error or annotation is for that
-                        keyword, such as an unresolvable reference.
+                        The absolute, dereferenced location of the validating subschema.  The value MUST
+                        be expressed as a full IRI using the canonical IRI of the relevant
+                        schema object, and it MUST NOT include by-reference applicators
```
handrews, line 2791: Can we mandate consistent use of JSON Pointer fragments instead?  I believe we changed the wording so that "canonical IRI" is only relevant to whole resources (no fragments).  So really it would be canonical IRI plus JSON Pointer fragment.  If this is too big of a topic I'm happy to take it back out to the larger discussion and let it be handled in a later PR.

jsonschema-core.xml
```diff
@@ -2827,32 +2820,61 @@ https://example.com/schemas/common#/$defs/count/minimum
...
+                        The JSON key for this information is "annotations".
                     </t>
                 </section>
 
                 <section title="Nested Results">
                     <t>
-                        For "basic", this property will appear only at the root node and will hold
-                        all errors or annotations in a list.
+                        Nested results are generated from applicator keywords: keywords with
+                        subschemas or collections of subschemas as values.  Keywords which
```
handrews, line 2861: I'd suggest changing "keywords with subschema or..." to "keywords that create dynamic scopes by applying schema(s) to the instance" (technically they're not always lexically subschemas).

jsonschema-core.xml
```diff
@@ -2953,52 +3076,127 @@ https://example.com/schemas/common#/$defs/count/minimum
...
+                        The root output unit contains "valid" for the overall result and "nested"
+                        for the list of specific results.  All other information is explicitly
+                        omitted from the root output unit.  If the root schema produces errors or
+                        annotations, then the output node for the root MUST be present within the
+                        root output unit's "nested" list with those errors or annotations.
+                    </t>
+                    <t>
+                        Output units which do not contain errors or annotations SHOULD be excluded
+                        from this format, however implementations MAY choose to include them for
+                        completeness.
```
handrews, line 3100: Is this intended to allow implementations to do things like raise an exception?

handrews (commented):
I've got various nitpicks and wording ideas (some of which may belong somewhere other than this PR), but overall this looks good to me.  I like how this gets rid of the duplication of the three location fields.

For the purposes of this PR I'm not worrying about whether "basic" becomes mandatory, or whether there are reasons for a more substantial difference between error and annotation output.

jsonschema-core.xml
gregsdennis: I'll leave this here until it gets moved there.

jsonschema-core.xml
gregsdennis: The title isn't the property name.  It's supposed to be a bit more descriptive.  The other title has "relative" in it.

I can update them both, but I don't think just the humanized keyword name is sufficient here.

jsonschema-core.xml
gregsdennis: Yeah, let's take this separately.

jsonschema-core.xml
gregsdennis: No, just to reduce the payload size.  If there are no errors or annotations, the output unit doesn't give much info (just locations and a validation result) and isn't very helpful.

jsonschema-core.xml
handrews: Yeah my main concern is to drop the absolute/relative thing as I think it was more confusing than helpful, and we can work up better language in terms of dynamic vs static/lexical schema identification (not necessarily _those_ words, but better words around those concepts).

Because really, this organizes output by dynamic scope more than anything else.  The same [sub]schema might appear may times, but it's a different dynamic scope each time.

jsonschema-core.xml
gregsdennis: I think I have a wording that might work for this, but it's still definitely worth discussing externally to this PR.

jsonschema-core.xml
gregsdennis: "dereferenced evaluation structure"?

gregsdennis:
@karenetheridge I welcome your feedback, but it's been over two weeks since you posted.

jsonschema-core.xml
handrews: > "dereferenced evaluation structure"?

That could work as long as it's defined somewhere.

Or maybe "dynamic scope document" which might not be as intuitive, but if defined clearly would fit with focusing on dynamic scopes?  idk, just tossing this out there, don't feel strongly about it.

jsonschema-core.xml
```diff
@@ -2757,11 +2756,11 @@
                     Implementations MAY elect to provide additional information.
                 </t>
 
-                <section title="Keyword Relative Location">
+                <section title="Evaluation path">
                     <t>
-                        The relative location of the validating keyword that follows the path
-                        traversed through the schema. The value MUST be expressed as a JSON
-                        Pointer, and it MUST include any by-reference applicators such as
+                        The location of the schema resource that produced the output unit, represented
```
handrews, line 2761: I like this a lot better, although I still have some thoughts :-)

1.  Should be "schema object" and not "schema resource" - that's an important distinction.

2.  Maybe instead of "location" this should just be "The path traversed to reach the schema object that produced the output unit"?  Because it's not really a location in the same sense as the other two, _unless_ we very clearly define something like "dereferenced evaluation structure" and then say "The location within the dereferenced evaluation structure".

jsonschema-core.xml
```diff
@@ -2777,22 +2776,21 @@
...
-                <section title="Keyword Absolute Location">
+                <section title="Subschema Location">
                     <t>
-                        The absolute, dereferenced location of the validating keyword.  The value MUST
-                        be expressed as a full IRI using the canonical IRI of the relevant schema resource
-                        with a JSON Pointer fragment, and it MUST NOT include by-reference applicators
-                        such as "$ref" or "$dynamicRef" as non-terminal path components.
-                        It MAY end in such keywords if the error or annotation is for that
-                        keyword, such as an unresolvable reference.
+                        The absolute, dereferenced location of the schema resource that produced
```
handrews, line 2789: schema resource -> schema object

jsonschema-core.xml
```diff
@@ -2777,22 +2776,21 @@
...
+                        the output unit.  The value MUST be expressed using the canonical IRI of the
+                        relevant schema resource plus a JSON Pointer fragment that indicates the contained
+                        subschema that produced the output.  It MUST NOT include by-reference applicators
```
handrews, line 2792: For consistency, can we replace "contained subschema" with "schema object" here as well?  In general I find myself using "schema object" instead of "subschema" as it works regardless of the location of the object.

jsonschema-core.xml
```diff
@@ -2827,32 +2821,62 @@ https://example.com/schemas/common#/$defs/count/minimum
...
+                        be included if the validation was unsuccessful.  The value
+                        for this property MUST be an object where the keys are the names of
+                        keywords and the values are the annotations produced by the
+                        associated keyword.
                     </t>
                     <t>
-                        The JSON key for failed validations is "error"; for successful validations
-                        it is "annotation".
+                        Each keyword defines its own annotation data type (e.g. "properties"
+                        produces a list of keywords, whereas "$title" produces a string).
```
handrews, line 2852: "$title" -> "title"

jsonschema-core.xml
```diff
@@ -2827,32 +2821,62 @@ https://example.com/schemas/common#/$defs/count/minimum
...
                 </section>
 
                 <section title="Nested Results">
                     <t>
-                        For "basic", this property will appear only at the root node and will hold
-                        all errors or annotations in a list.
+                        Nested results are generated from keywords which create a new dynamic
+                        scope by applying a subschema to the instance or a child of the instance.
+                        Keywords which have multiple subschemas (e.g. "anyOf") will generally
+                        generate an output unit for each subschema.  In order to accommodate
```
handrews, line 2864: This is an excellent use of "subschema", just to clarify that I don't mean eliminating the term everywhere :-)

jsonschema-core.xml
```diff
@@ -2872,73 +2896,174 @@ https://example.com/schemas/common#/$defs/count/minimum
                             failure
                         </t>
                         <t>
-                            "nested" - the collection of errors or annotations produced by a keyword
+                            "nested" - the collection of errors or annotations produced by a subschema
```
handrews, line 2899: Here I would use "schema object", because root schemas also produce output units.

jsonschema-core.xml
```diff
@@ -2872,73 +2896,174 @@ https://example.com/schemas/common#/$defs/count/minimum
...
                     </artwork>
                 </figure>
                 <t>
-                    This instance will fail validation and produce errors, but it's trivial to deduce
-                    examples for passing schemas that produce annotations.
+                    The failing instance will produce the following errors:
+                    <list>
+                        <t>
+                            The value at "/foo"
+                            validated at "/properties/foo/allOf/0"
```
handrews, line 2971: Can we say "evaluated at" or "evaluated by" instead of "validated at"?

jsonschema-core.xml
```diff
@@ -3007,142 +3207,185 @@ https://example.com/schemas/common#/$defs/count/minimum
...
                     <t>
-                        The "Detailed" structure is based on the schema and can be more readable
-                        for both humans and machines.  Having the structure organized this way makes
-                        associations between the errors more apparent.  For example, the fact that
-                        the missing "y" property and the extra "z" property both stem from the same
-                        location in the instance is not immediately obvious in the "Basic" structure.
-                        In a hierarchy, the correlation is more easily identified.
+                        The "Hierarchical" structure is a tree structure that follows the
+                        evaluation path during the validation process.  Typically, it will
+                        resemble the schema as if all referenced schemas were bundled in place
```
handrews, line 3214: This paragraph is great, but might I suggest "inlined" instead of "bundled"?  We've been using bundling specifically for the pattern of embedding resources under `$defs`.

jsonschema-core.xml
```diff
@@ -3007,142 +3207,185 @@ https://example.com/schemas/common#/$defs/count/minimum
...
-                            </t>
-                            <t>
-                                Nodes that have a single child are replaced by the child.
-                            </t>
-                        </list>
-                        Branch nodes do not require an error message or an annotation.
+                        All output units are included in this format.
+                    </t>
+                    <t>
+                        The location properties of the root output unit MAY be omitted.
```
handrews, line 3221: What does this mean exactly?  I would really rather not have to special-case any output unit, they should all always provide the same set of location information.

handrews (commented):
Thanks for the update!  While i made quite a few comments they're mostly tiny things or just an effort to standardize the terminology.  Overall this is looking really good!

jsonschema-core.xml
gregsdennis: For the `basic` output, results from the root schema are also an item in the list, which means that there is no need to have location properties (`evaluationPath`, `schemaLocation`, & `instanceLocation`) in the root output unit (and only the root).  Thus for `basic` it's just `valid` and `nested`.

This allowance permits these properties to also be omitted from the root of the `hierarchical` format.  The thinking behind this is that it aligns with `basic`, if that matters to implementors, but also that these location properties will _always_ be

```
evaluationPath: "" (empty pointer)
schemaLocation: "https://example.com/schema#"
instanceLocation: "" (empty pointer)
```

Having them isn't useful or really required.

jsonschema-core.xml
gregsdennis: I think going with 

> Provides validation information in a hierarchical structure that _follows the evaluation paths generated while processing the schema_.

Aligns better with my newest (marginally related) PR, #1264.

jsonschema-core.xml
gregsdennis: 1. 👍 
2. #1264 defines "evaluation path" as an actual term (see the first change, line 682).  I think this really just needs to echo that.  You're right, though, in that it's not a location, _per se_.  I'll make an update.

jsonschema-core.xml
gregsdennis: 🤦

jsonschema-core.xml
gregsdennis: "Subschema" is proper here.  The context is the `hierarchical` output.  In this case, `nested` holds output units from strictly subschemas.

jdesrosiers:
I'm still not entirely sure how I feel about the approach. I think it will work fine. I think my only concern is that we would be making a drastic change without really knowing how it's going to work out. Not that what we have now is well proven either, I just don't want to be implementing a completely new output format every release. I don't know that there's really a solution to that without a crystal ball, but I wanted to mention it anyway.

gregsdennis:
@jdesrosiers I understand and share that concern.  However, as I mentioned in [my blog post](https://json-schema.org/blog/posts/fixing-json-schema-output), the current design was not met with much acceptance and there were a lot of questions.

Secondly, I think this aligns better with the [discussion](https://github.com/json-schema-org/json-schema-spec/issues/1172) we had around how annotations are passed on / blocked.

It really _needs_ to change.  I really think that we're a lot closer to a final form.

Also, having implemented both formats, I can tell you that this is _so_ much simpler!

jdesrosiers:
I completely agree that it needs an update. Doing nothing is not an option. My main concern is maintaining a bunch of different versions of the output format as we iterate and experiment to figure out what's going to work best. I'd feel a lot better about this if it were it's own spec and we actually treated it like a draft where each draft replaces the previous and has no backwards/forwards compatibility guarantees until things stabilize. That way I always only have to maintain the latest version.

gregsdennis:
I agree with what you're saying.  For now, I think we can still make improvements in-place and then extract the output separately later.

handrews:
@jdesrosiers give that we quite likely won't even put out the next revision under the IETF process, I don't think we should worry too much about what is in which document at the moment.

gregsdennis:
Let's continue the output-in-a-new-spec discussion in [this thread](https://github.com/orgs/json-schema-org/discussions/63#discussioncomment-3329274).  For now, it's all in Core.  We'll open a new PR for further developments.  I think this one is complete.

jdesrosiers:
> given that we quite likely won't even put out the next revision under the IETF process, I don't think we should worry too much about what is in which document at the moment.

I don't think what process we are using makes a difference, unless your point is that it doesn't make sense to split things out until we know what process we're going to use moving forward. That makes sense, but I wasn't suggesting it be split out now, just that it's an important consideration to keep in mind. If we're going to make such a big change we need to consider the effect on implementors and that very much includes discussion of what the life-cycle of this feature will be.

I think it's a relevant an very important to thing to be thinking about now, but it's not a blocker for this PR, just an important thing to call out and follow up on later.

jsonschema-core.xml
handrews: This feels like a design smell from trying to make two things that aren't really the same pretend like they are the same.

jsonschema-core.xml
```diff
@@ -2826,7 +2826,8 @@ https://example.com/schemas/common#/$defs/count/minimum
                         be included if the validation was successful.  The value
                         for this property is an object where the keys are the names of
                         keywords and the values are the error message produced by the
-                        associated keyword.
+                        associated keyword.  If the subschema itself is producing the
+                        error, that error MUST be listed with an empty string key.
```
handrews, line 2826: What kind of errors do subschemas produce on their own?

gregsdennis: Maybe this can be more explicit, but something needs to be in place to handle the `false` schema.  For example:

```json
{
  "properties": {
    "foo": false
  }
}
```

You still need an error message for the `foo` property when it's included in the instance.  With the proposed format it'd be:

```json
{
   "valid": false,
   "nested": [
    {
      "valid": false,
      "instanceLocation": "/foo",
      "schemaLocation": "http://example/schema#/properties/foo"
      "evaluationPath": "/properties/foo",
      "errors": {
        "": "All values fail against the false schema"   // this
      }
    }
  ]
}
```

There's not a keyword that produces the error.  While someone technically _could_, I expect no one would actually create an empty-string keyword, so the empty string is sufficient to declare that this error came from the schema.

This is the immediate example I can think of.  I'm sure other people can conceive of other reasons to have a schema produce an error directly.  I worded it this way because I didn't want to limit to the `false` schema case.

handrews: @gregsdennis thanks, this makes sense.  I'm a little skeptical of other use cases but that's not a reason to discourage, much less forbid them.  How would you feel about noting that the boolean `false` schema is the most common (or some vaguely similar wording) use case for this?  I feel like it would make it more clear why this has to be there.

I think something that'd probably want to avoid would be some sort of automatic roll-up error that appears in every failed dynamic scope / output unit.  That was the sort of thing that had me concerned- where errors are repeated without really adding anything.

gregsdennis: Sure I can add a note to that effect.

handrews (approved):
I'm in favor of merging this.  The open points of discussion are either not things that I'd block the PR on anyway, or I'm confident that we'll continue talking about them.  What's here right now is a great improvement.

PR #1259, opened 2022-07-20
Add month to Relative JSON Pointer's spec <date/>

jviotti:
It is currently not possible to build the Relative JSON Pointer's spec.
`xml2rfc` will complain about the date tag not being the current year
while not having a month set either.

This is the error I get:

```
xml2rfc --html relative-json-pointer.xml -o relative-json-pointer.html
json-schema-spec/relative-json-pointer.xml(16): Error: Expected <date> to have the current year when month is missing, but found '2021'
Unable to complete processing relative-json-pointer.xml
```

This PR changes the `<date/>` year to the current year.

### DIFF
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -40,7 +40,7 @@
             </address>
         </author>
 
-        <date year="2021"/>
+        <date year="2022"/>
         <workgroup>Internet Engineering Task Force</workgroup>
         <keyword>JSON</keyword>
         <keyword>JavaScript</keyword>


Relequestual:
Relative JSON Pointer has not been updated in 2022, and so should not be built with that date.
It's an "error", but not a crash, so more like a warn.
I think the "correct" solution is to add the Month and Day, as required per https://datatracker.ietf.org/doc/html/rfc7991#section-2.17

Relequestual:
Your PR title says "add month" but you've actually changed the year to this year.

jviotti:
@Relequestual 

> I think the "correct" solution is to add the Month and Day, 

Sounds good. I'll change the dates to match https://json-schema.org/draft/2020-12/relative-json-pointer.html then.

> Your PR title says "add month" but you've actually changed the year to this year.

Oops! I was initially adding the month, then thought changing the year might be expected and forgot to update the description.

handrews:
Usually we just bump the year.

jviotti:
Done!

jviotti:
@handrews Yeah, that was my impression too, given that the latest draft's date is January 2020, but the current year on the XML file was already 2021.

jviotti:
Ping @Relequestual @handrews . How should we move forward here? This is a pre-requisite for reviving the CI on this repo

handrews:
Just update the year to 2022 and don't put a month, that's what we've always done.

jviotti:
Done!

Relequestual:
> Just update the year to 2022 and don't put a month, that's what we've always done.

Isn't it supposed to reference the year of publicaiton?

I guess it doesn't matter unless we publish that document anywhere, like on the website.

handrews (approved):
> Isn't it supposed to reference the year of publicaiton?

It's always the current year, and then when it's published that's the  year of publication.

We have literally done this every year with every spec we edited that year.

Relequestual:
I guess that's fine. I just don't want someone turning up asking what change for that spec when it didn't get any changes.

handrews:
@Relequestual it won't build without it, and build errors are a bad thing.

Relequestual:
I've clearly missed that we have done this every year regardless of the spec release process, so please ignore me and proceed =]

PR #1260, opened 2022-07-20
Track the xml2rfc dependency using a requirements.txt file

jviotti:
The standard practice for tracking dependencies on Python projects is to
declare the dependencies and the expected versions on a
`requirements.txt` file that is read by `pip`.

Right now, we request people to manually install the latest version of
`xml2rfc`. After this commit, we properly track `xml2rfc` on
`requirements.txt` and tweak the documentation a bit to show users how
to install it from there instead.


jviotti:
A bonus addition as part of the CI-revival work. To make sure we all (including CI) run on the same version of `xml2rfc`.

requirements.txt
```diff
@@ -0,0 +1 @@
+xml2rfc == 3.12.0
```
Julian, line 1: Generally a `requirements.txt` should have all transitive dependencies fully pinned too:

```suggestion
appdirs==1.4.4
certifi==2022.6.15
charset-normalizer==2.1.0
configargparse==1.5.3
google-i18n-address==2.5.2
html5lib==1.1
idna==3.3
intervaltree==3.1.0
jinja2==2.11.3
kitchen==1.2.6
lxml==4.9.1
markupsafe==2.1.1
pycountry==22.3.5
pyflakes==2.4.0
pyyaml==6.0
requests==2.28.1
six==1.16.0
sortedcontainers==2.4.0
urllib3==1.26.10
webencodings==0.5.1
xml2rfc==3.12.0
```

Otherwise lgtm.

jviotti: @Julian Ah, I didn't know this. Do you know how to create the `requirements.txt` automatically from the dependency using `pip`? I can't seem to figure it out using pip's help

Julian: So you do so "these days" by having a file called `requirements.in` (and put just the text `xml2rfc>=3.12.0` in it), and then you pass that to [`pip-tools`](https://pypi.org/project/pip-tools/) -- specifically you run `pip-compile -r requirements.in` and it will spit out the contents above (which is more or less how I got them). And then you commit both files.

I hesitated to recommend that (not for any real reason other than it means we also need to write the above in the README so people know how to update the file). But yeah I guess it's the right thing to do really.

jviotti: Done! I also included some brief instructions on the README on how to do this

Julian: Nice!

PR #1264, opened 2022-08-04
align path terminology in the text with the renamings made for the output

gregsdennis:
I noticed a few places where the spec is using several different phrasings to mean the same thing.  This PR aligns them to what we agreed it should be called in the output.


jdesrosiers (approved):

handrews (approved):
Awesome- thanks for catching these!

Issue #1262, opened 2022-07-22
Check on and fix CI checks on spec repo


Relequestual:
Action item from https://github.com/json-schema-org/community/issues/200 - @jviotti 

Our CI for this repo has not been running.
We should investigate what it was doing, define what we would like it to do, and (probably) resolve by using GitHub actions.

Related: https://github.com/json-schema-org/json-schema-spec/issues/618

Relequestual:
The merged PR did not work as expected.
@jviotti see https://github.com/json-schema-org/json-schema-spec/runs/7915793140?check_suite_focus=true#step:6:9

Relequestual:
Julian fixed it. Woop!

PR #1266, opened 2022-08-05
Replace defunct Travis CI integration with GitHub Actions

jviotti:
The GitHub Actions setup is equivalent to the Travis CI configuration
with two differences:

- The GitHub Action setup does not set the `TAG` environment variable,
  as it didn't seem like `xml2rfc` would consume that at all (unless I'm
  missing something)

- The GitHub Action preserves the spec output (txt and html) as
  artifacts that can be downloaded after a successful run (to see the
  changes without having to locally render the specs)

Fixes: https://github.com/json-schema-org/json-schema-spec/issues/1262

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,14 @@
+name: JSON Schema
+on:
+  - push
+jobs:
+  specs:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: pip install --requirement requirements.txt
+      - run: xml2rfc --version
+      - run: make all
+      - uses: actions/upload-artifact@v2
+        with:
+          path: *.(html|txt)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,4 +11,6 @@ jobs:
       - run: make all
       - uses: actions/upload-artifact@v2
         with:
-          path: *.(html|txt)
+          path: |
+              *.(html|txt)
+              !requirements.txt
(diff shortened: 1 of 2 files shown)


jviotti:
@Relequestual I think you might need to enable the repository to run GitHub Actions from forks to be able to test this out. Once it is merged, we should also make this check required.

gregsdennis:
What I've had to do in `json-everything` is merge fork PRs into a secondary branch, just to get all of the actions to run properly (even when you get them running, they don't always have all the permissions they need).  Then I have another PR to pull into the primary branch.

handrews (approved):
Looks good to me.

Relequestual:
I believe this will also upload `requirements.txt` to the artifact bundle.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,14 @@
...
+  specs:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: pip install --requirement requirements.txt
+      - run: xml2rfc --version
+      - run: make all
+      - uses: actions/upload-artifact@v2
+        with:
+          path: *.(html|txt)
```
Relequestual, line 14:
```suggestion
          path: |
              *.(html|txt)
              !requirements.txt
```

Relequestual (changes requested):

jviotti:
Good catch @Relequestual ! Fixed.

Relequestual (approved):

Relequestual:
I decided to merge this immediatly as our current CI simply doesn't work.

PR #1268, opened 2022-08-10
Update the xrefs, which appear now to have underscores

Julian:
Fixes the build errors -- I don't know how this stuff works and/or whether this has to do with upstream changes or some normalization that's happening, but this does seem to work here at least now.


handrews (approved):
Ah... I did not notice that when I checked the source documents.  Yeah, this gets checked against some sort of online database.

PR #1269, opened 2022-08-10
Ignore unknown keywords or treat as annotations

handrews:
This is just a bug fix for an accidental oversight.  I'm hoping it's non-controversial as I don't think any of us meant to allow arbitrary behaviors for unrecognized keywords.

When we added the "SHOULD collect as annotations" behavior for
unknown keywords, we forgot to retain the MUST directive to ignore
them if the SHOULD is not followed.  Which also ensures that
no other behavior than collecting as a annotations is performed.

This also consolidates the behavioral specification in one place
as it appeared in two places before.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -350,7 +350,7 @@
                         and its companions, are free to define other behaviors as well.
                     </t>
                     <t>
-                        A JSON Schema MAY contain properties which are not, or are not recognized as, schema keywords.
+                        A JSON Schema MAY contain properties which are not schema keywords or are not recognized as schema keywords.
                         The behavior of such keywords is governed by section
                         <xref target="unrecognized" format="counter"></xref>.
                     </t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -350,9 +350,9 @@
                         and its companions, are free to define other behaviors as well.
                     </t>
                     <t>
-                        A JSON Schema MAY contain properties which are not schema keywords.
-                        Unknown keywords SHOULD be treated as annotations, where the value
-                        of the keyword is the value of the annotation.
+                        A JSON Schema MAY contain properties which are not, or are not recognized as, schema keywords.
+                        The behavior of such keywords is governed by section
+                        <xref target="unrecognized" format="counter"></xref>.
                     </t>
                     <t>
                         An empty schema is a JSON Schema with no properties, or only unknown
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -350,9 +350,9 @@
                         and its companions, are free to define other behaviors as well.
                     </t>
                     <t>
-                        A JSON Schema MAY contain properties which are not schema keywords.
-                        Unknown keywords SHOULD be treated as annotations, where the value
-                        of the keyword is the value of the annotation.
+                        A JSON Schema MAY contain properties which are not, or are not recognized as, schema keywords.
```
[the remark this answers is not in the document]
handrews: @jdesrosiers Yeah your wording is better, although I'd put a comma between "are not schema keywords" and "orare not recognized as schema keywords."  Would that work for you?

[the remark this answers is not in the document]
jdesrosiers: Yep. That's fine.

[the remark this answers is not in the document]
gregsdennis: Grammatically, the comma should not be there as "or are not recognized..." is not a complete sentence.

gregsdennis (approved):

jsonschema-core.xml
```diff
@@ -350,9 +350,9 @@
                         and its companions, are free to define other behaviors as well.
                     </t>
                     <t>
-                        A JSON Schema MAY contain properties which are not schema keywords.
-                        Unknown keywords SHOULD be treated as annotations, where the value
-                        of the keyword is the value of the annotation.
+                        A JSON Schema MAY contain properties which are not schema keywords or are not recognized as schema keywords.
```
notEthan, line 353: I think it'd suffice to say "A JSON Schema MAY contain properties which are not recognized as schema keywords."

notEthan: oh, I see this was the subject of the resolved review. no need to relitigate that if it's been sufficiently addressed, disregard if so.

handrews: @notEthan I preserved the "not schema keywords" part from the original text because I don't remember why it was phrased that way and did not want to get into a debate about whether it is OK to remove it.  If you think it should be removed, we should probably discuss that in an issue.  I figured adding "or are not recognized as..." would cover the more common way this is thought about without removing anything previously allowed.

notEthan: sounds good to me

karenetheridge (approved):

Issue #1271, opened 2022-08-12
Consolidate language around duplicate schema IRIs


handrews:
The spec mentions duplicate IRIs in two places with slightly different requiremetns:

[§8.2.2 "Defining location-independent identifiers](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-defining-location-independe) (last paragraph):

> The effect of specifying the same fragment name multiple times within the same resource, using any combination of "$anchor" and/or "$dynamicAnchor", is undefined. Implementations MAY raise an error if such usage is detected.

[§9.1.2 "Loading a referenced schema"](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-9.1.2) (last paragraph):

> A schema MAY (and likely will) have multiple IRIs, but there is no way for an IRI to identify more than one schema. When multiple schemas try to identify as the same IRI, validators SHOULD raise an error condition.

It would be good to have 8.2.2 refer to 9.1.2 (as it is more general), so that the "SHOULD raise an error" requirement is consistent regardless of how the schemas came to have the same IRI.

PR #1272, opened 2022-08-13
Consolidate and clarify language around duplicate IRIs

handrews:
Fixes issue #1271 and addresses one point mentioned in https://github.com/json-schema-org/json-schema-spec/issues/1059#issuecomment-1212726799 (but please do not drag the rest of that issue into this PR, thanks!)

The only actual changed requirement is the forbidding of `"$id": "#"`, and `"$id": ""` which are confusing and are either pointless (in a document root where they resolve to the retrieval IRI, exactly as if `$id` was not present) or problematic (in an embedded resource root, where they produce duplicate IRIs for the embedded and containing resource).

This also consolidates the two different places where duplicate IRIs were addressed.  Since the more general of the two paragraphs stated that this SHOULD be an error, I kept the SHOULD rather than the MAY that addressed only a subset of the cases.  The SHOULD technically covered the subset as well anyway.

I have emphasized that "undefined" here means that the result, whatever it is, will not be interoperable (most likely, implementations will just return the last-seen schema under that IRI, but who knows).

Note that we could continue to allow `"$id": "#"` and `"$id": ""` and just assume that the SHOULD regarding duplicate IRIs is enough to discourage that.  If that is the consensus then I can remove the commit that handles that part of the change.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1353,8 +1353,8 @@
                     </t>
                     <t>
                         If present, the value for this keyword MUST be a string, and MUST represent a
-                        valid <xref target="RFC3987">IRI-reference</xref> with a non-empty non-fragment
-                        component.  This IRI-reference SHOULD be normalized, and MUST resolve to an
+                        valid <xref target="RFC3987">IRI-reference</xref>.  This IRI-reference
+                        SHOULD be normalized, and MUST resolve to an
                         <xref target="RFC3987">absolute-IRI</xref> (without a fragment),
                         or to a IRI with an empty fragment.
                     </t>
@@ -1392,7 +1392,10 @@
                         <xref target="RFC3986">RFC 3986 section 5.1.2</xref> regarding
                         encapsulating entities, if an "$id" in a subschema is a relative
                         IRI-reference, the base IRI for resolving that reference is the IRI of
-                        the parent schema resource.
+                        the parent schema resource.  Note that an "$id" consisting of an empty IRI or
+                        of the empty fragment only will result in the embedded resource having
+                        the same IRI as the encapsulating resource, which SHOULD be considered
+                        an error per section <xref target="duplicate-iris" format="counter"></xref>.
                     </t>
                     <t>
                         If no parent schema object explicitly identifies itself as a resource
@@ -1458,7 +1461,7 @@
                     </t>
                 </section>
 
-                <section title="Duplicate schema identifiers">
+                <section title="Duplicate schema identifiers" anchor="duplicate-iris">
                     <t>
                         A schema MAY (and likely will) have multiple IRIs, but there is no way
                         for an IRI to identify more than one schema. When multiple schemas
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1456,11 +1456,16 @@
                             fragment "#foo" when used in a IRI.  See below for full examples.
                         </cref>
                     </t>
+                </section>
+
+                <section title="Duplicate schema identifiers">
                     <t>
-                        The effect of specifying the same fragment name multiple times within
-                        the same resource, using any combination of "$anchor" and/or
-                        "$dynamicAnchor", is undefined.  Implementations MAY
-                        raise an error if such usage is detected.
+                        A schema MAY (and likely will) have multiple IRIs, but there is no way
+                        for an IRI to identify more than one schema. When multiple schemas
+                        attempt to identify as the same IRI through the use of "$id", "$anchor",
+                        "$dynamicAnchor", or any other mechanism, implementations SHOULD raise
+                        an error condition.  Otherwise the result is undefined, and even if
+                        documented will not be interoperable.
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -1353,8 +1353,8 @@
                     </t>
                     <t>
                         If present, the value for this keyword MUST be a string, and MUST represent a
-                        valid <xref target="RFC3987">IRI-reference</xref>.  This IRI-reference
-                        SHOULD be normalized, and MUST resolve to an
+                        valid <xref target="RFC3987">IRI-reference</xref> with a non-empty non-fragment
+                        component.  This IRI-reference SHOULD be normalized, and MUST resolve to an
```
jdesrosiers, line 1357: I don't think this change is necessary. The two valid options are presented as an absolute-IRI or an IRI with an empty fragment. Both absolute-IRI and an IRI are by definition non-relative, so we don't need to say "non-empty". It also says it "MUST resolve to", which also indicates a non-relative result.

jdesrosiers (changes requested):

jsonschema-core.xml
notEthan: @jdesrosiers: This doesn't sound right. $id can be a relative IRI. If that relative IRI is empty (apart from fragment), it resolves to the same absolute IRI as the parent schema (or the same as its retrieval IRI if it's at the root, but that part isn't a problem). Then a subschema has the same id as its parent. Forbidding the empty non-fragment portion disallows this id collision.

@handrews: The wording is awkward, "with a non-empty non-fragment component". Maybe a new sentence would be better:

> If present, the value for this keyword MUST be a string, and MUST represent a valid IRI-reference. [unchanged]
> This IRI-reference MUST NOT be empty or consist only of an empty fragment.
> or
> This IRI-reference MUST NOT have an empty non-fragment component.
>
> This IRI-reference SHOULD be normalized ...

handrews: @jdesrosiers I'm not sure I understand.  `"$id": "foo/bar"` is valid, both before and after this change.  It is a path-only relative IRI-reference that, resolved against a base IRI (say, `https://example.com/schemas/base`) produces an absolute-IRI (`https://example.com/foo/bar`).

@notEthan Thanks, I think it like "This IRI-reference MUST NOT be empty or consist only of an empty fragment." I checked RFC 3986 and it uses the term "empty" to describe relative references in [§4.4 Same-Document Reference](https://datatracker.ietf.org/doc/html/rfc3986#section-4.4).

jdesrosiers: I wasn't saying that `$id` can't be relative. I was talking about the result of resolving the `$id` against the base IRI. I thought this change was trying to protect against case where you don't have a baseURI and therefore the IRI can't be fully resolved. I was pointing out that the current wording already protects against that case.

> If that relative IRI is empty (apart from fragment), it resolves to the same absolute IRI as the parent schema

Oh, ok, now I understand where this change is coming from. But, I still don't think it's necessary because we talk about not allowing duplicate identifiers more generally elsewhere. IMO, it's already covered and doesn't need to be repeated here.

handrews: > I thought this change was trying to protect against case where you don't have a baseURI and therefore the IRI can't be fully resolved. I was pointing out that the current wording already protects against that case.

Ah, I see.  Yeah, that is handled and not improved (or made worse) by this change.

> we talk about not allowing duplicate identifiers more generally elsewhere. IMO, it's already covered and doesn't need to be repeated here.

Cool- yeah, that's the main question (for one of the commits in this PR- the rest of it remains valid as it clarifies that "already covered" part that, as currently written, kinda contradicts itself in a confusing way).

I'd like to leave this open and see if we get any more opinions from other @json-schema-org/spec-team folks (or anyone else, I just can't tag the whole universe).  I don't feel strongly on this point, so if no one advocates for the change I'll strip it out and we can go with just the clean-up of the duplicate IRI stuff.

handrews: One option might be to note that, while syntactically allowed, `""` and `"#"` effectively produce duplicate IRIs which are considered errors per §whatever:

> Note that while the empty IRI and the empty fragment-only IRI are legal values, they produce duplicate IRIs when used in an embedded resource, which is considered an error per 8.2.3.

That's probably enough to discourage the values without requiring people to detect and block them.  I'm currently leaning towards this option.

notEthan: it might be a little odd to include a MUST NOT with the intent of heading off a route to duplicate IRIs, when the language on duplicate IRIs itself isn't a MUST NOT ("implementations SHOULD raise an error condition").

handrews:
Based on feedback so far I have removed the forbidding of certain `$id` values (and changed the title of this PR to match that).  I did add extra wording in the section on subschemas with `$id` about how empty IRIs are problematic, as it seems useful to do that there.  It links to the new section on duplicates.

jdesrosiers (approved):

PR #1273, opened 2022-08-14
Add data-model anchor.

ioggstream:


handrews (approved):
This should stay open for a while so folks have a chance to see it, but seems like a straightforward improvement to me.

gregsdennis (approved):

handrews:
I think this is simple enough that 2 approvals across a few days is sufficient, so I'll merge it.

Issue #526, opened 2017-12-31
consider joining the WG


dret:
https://mailarchive.ietf.org/arch/msg/json/Ec78QgZqpvcqPd97fvGV3Nt2tVA makes the point that a JSON schema spec probably would be a worthwhile contribution to the JSON community as a whole. i am not in a position to make the call whether the spec developed here should make this step, but i pretty confident that this would help the spec development in terms of community support, and getting more momentum behind it.

akuckartz:
@dret Which WG? Currently there does not seem to exist a JSON WG within the IETF: https://datatracker.ietf.org/wg/

handrews:
@dret draft-08 is still dealing with some fairly fundamental issues that need to get resolved, as they are the same ones that essentially killed the project for three years or so after draft-04.

Once that is done, the main remaining question for core and validation is `$data`.  My intention was to work through that in draft-09 while getting feedback on draft-08, and then try to move to a working group.

I do want to get to a working group status, but previously when it's come up the consensus was that we weren't ready yet.

I will also admit that I am dreading working over listserv, which is a huge reason that I'm reluctant to pursue it.  As I understand it, that's an absolute requirement for going through a working group (corrections welcome if I'm wrong).

It's about to be 2018.  We have much better collaboration tools, and we can use them as we wish while we are doing I-Ds.  The volume and pace of discussions on this project are a challenge to keep up with when they're organized in github issues.  I cannot image dealing with this over email.

I have considered trying to get the core and validation specs to a decent point and then hoping that I can hand them off to someone who will be more comfortable with the IETF's technology choices and working style (unless they've modernized since the last time I checked, I suppose).  I could then stick with Hyper-Schema which will be in I-D territory for a while yet as it doesn't even have proper implementations at this point.

handrews:
> I cannot image dealing with this over email.

Or rather, I can, from other projects back when email was the best option, and I'd prefer to never repeat the experience.

dlax:
@akuckartz 
> Which WG?

The jsonbis one https://datatracker.ietf.org/wg/jsonbis/about/, I guess.

dlax:
> I will also admit that I am dreading working over listserv, which is a huge reason that I'm reluctant to pursue it. As I understand it, that's an absolute requirement for going through a working group (corrections welcome if I'm wrong).

[...]

PR #1277, opened 2022-08-17
Add ADR for decoupling from IETF

jdesrosiers:
Here's a fist draft for an ADR for decoupling from JSON Schema. I'm sure it needs some cleanup and fleshing out in places. I'll revise tomorrow.

The scope of this ADR is decoupling from from IETF. That doesn't include the decision to stop using "draft", but enables us to make that decision later.

### DIFF
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -30,7 +30,7 @@ from the IETF process.
   disbanding. In some cases, the RFC may be revisited and revised in the future,
   but this is rare and generally contains little more than clarifications and
   reference updates. JSON Schema is not like that. JSON Schema is more like a
-  programming language. When it stops evolving, it will loose it's relevance.
+  programming language. When it stops evolving, it will lose its relevance.
   When we finish a release of JSON Schema, we don't disband, we start work on
   the next release.
 * Every one of our releases is expected to be used in production and will be
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -62,7 +62,7 @@ Our decision is to go with Option 3 and decouple from IETF completely and come
 up with our own governance model. The idea of having to come up with our own
 governance model is daunting, but when the alternatives are to continue to abuse
 the I-D process or to conform to the IETF process that doesn't fit our needs,
-going our own way seems the bast option.
+going our own way seems the best option.
 
 ### Positive Consequences
 
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -59,7 +59,7 @@ from the IETF process.
 ## Decision Outcome
 
 Our decision is to go with Option 3 and decouple from IETF completely and come
-up with our own governance model. The idea of having the come up with our own
+up with our own governance model. The idea of having to come up with our own
 governance model is daunting, but when the alternatives are to continue to abuse
 the I-D process or to conform to the IETF process that doesn't fit our needs,
 going our own way seems the bast option.
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -15,10 +15,14 @@ Related Issues:
 ## Context and Problem Statement
 
 Currently JSON Schema loosely follows the IETF Internet-Draft (I-D) process for
-spec development and releases but isn't associated with IETF in any way. Our
-perceived involvement with IETF causes confusion and misunderstanding within our
-community in the cases were our practices and the realities of our situation
-deviate from the typical IETF I-D process.
+spec development and releases but isn't associated with any IETF working group.
+JSON Schema is an individual draft. That means it isn't on a standards track
+with IETF and IETF is not involved nor supports the spec in any way other than
+hosting the canonical version of our I-Ds. Our perceived involvement with IETF
+causes confusion and misunderstanding within our community in the cases were our
+practices and the realities of our situation deviate from a typical IETF I-D
+lifecycle. The thing that makes our situation different than a typical I-D is
+that our "drafts" are intended for use in production.
 
 ## Decision Drivers
 
@@ -46,13 +50,19 @@ deviate from the typical IETF I-D process.
 * Several members of the JSON Schema team have had poor interactions with IETF
   and don't feel that working with them would be productive. This is a
   relatively minor consideration. If we thought IETF was right for JSON Schema,
-  we could find ways to make those relationships work.
+  we could find ways to make those relationships work. We have a good
+  relationship with the relatively new HTTPAPIs working group and working with
+  them would be far more likely to be productive than the people/WG we were
+  previously in communication with.
 
 ## Considered Options
 
 1. Continue to submit I-Ds, while using our customized process with no intention
-  of pursing standards track RFC status.
-2. Go all-in with IETF and pursue a standards track RFC with the IETF.
+   of pursing standards track RFC status.
+2. Go all-in with IETF and pursue a standards track RFC with the IETF. The
+   approach would be to describe the essential characteristics of evaluating a
+   JSON Schema: the keywords that everybody is guaranteed to support, and any
+   extension mechanisms.
 3. Join W3C and pursue a standards track with them using their process.
 4. Decouple completely from any standards organization and come up with our own
    specification development lifecycle (SDLC) model inspired by well established
@@ -62,11 +72,44 @@ deviate from the typical IETF I-D process.
 
 Our decision is to go with Option 4 and decouple from standards organizations
 that don't fit our needs. We don't currently have a plan for what to replace
-IETF with. We are currently investigating how other established projects do
+IETF with, but we are currently investigating how other established projects do
 their SDLC and will likely choose one to emulate and adapt to our needs.
 Although we don't have a replacement solution in place yet, we are confident
 that continuing to abuse the IETF I-D process or conforming to a standards
 organization process that doesn't fit our needs is not the way to go.
+
+Option 2 and 3 are still on the table if we feel it makes sense when we get to a
+more stable place in the future. The main concern is the pain this process is
+causing while we are in this unusual phase of simultaneous unstable growth and
+production use. Standardization isn't out of the question, it's just not
+productive for us to be developing JSON Schema within the constraints of a
+standards organizations procedures.
+
+Option 1 was rejected because it ignores the problems we've been facing and
+provides no solution. No one wants this.
+
+Option 2 was rejected for several reasons. If we go all in with IETF, we would
+have to join a working group and treat JSON Schema like a normal I-D. That means
+we would have to start treating drafts as drafts, which means not recommending
+production use until we are ready for RFC and not releasing a new
+production-ready version of JSON Schema until we've reached RFC status. Most of
+the core contributors don't believe that we are close enough to an RFC-ready
+release that we want to commit to not being able to issue another release until
+that happens.
+
+There are other concerns including skepticism that even with an extension
+mechanism that the RFC wouldn't need regular updates, which is not normal
+practice for an RFC and would require significant effort to issue a replacing
+RFC. Without a concrete proposal on the scope of the RFC and the extension
+mechanisms, it's hard to commit to this path.
+
+Additionally, many of the core contributors have found working with the IETF
+unproductive and have concerns about JSON Schema getting deeper involved without
+compelling enough reason. Most agree that the reasons are not sufficiently
+compelling at this point.
+
+Option 3 was rejected because it has the same problems as Option 2 accept that
+we don't have the same unpleasant history with W3C than we do with IETF.
 
 ### Positive Consequences
 
@@ -79,8 +122,14 @@ organization process that doesn't fit our needs is not the way to go.
 
 * Not being associated with a recognized standards organization such as IETF,
   W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
+  However, people seem comfortable adopting OpenAPI without it being associated
+  with a standards organization, so we don't expect this to be a significant
+  issue for JSON Schema either.
 * If we don't go the standardization route with IETF or W3C, we lose access to
   their expert review process.
+* One of the benefits of an RFC is other standards can normatively reference it,
+  and use JSON Schema to define their JSON-based syntaxes. Independently
+  publishing a specification does not permit this.
 * Defining our own SLDC process will be a lot of work and none of us have
   expertise in defining such a process. However, we can take inspiration from
   existing well established projects and we would have the freedom to update our
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -128,6 +128,15 @@ compelling at this point.
 
 Option 3 was rejected because it has the same problems as Option 2 except that
 we don't have the same unpleasant history with W3C than we do with IETF.
+Additionally, the W3C is a "pay to play" standards organization. Organizations
+must pay to contribute to specifications the W3C publish, which doesn't match
+the JSON Schema Org's open ethos.
+
+Ben Hutton has had multiple calls with various individuals at different levels
+within the W3C, and has a friendly contact should we wish to investigate again
+at a later point. The W3C does have an "invited expert" solution for when
+a persons employer doesn't want to be a paying member, however this is supposed
+to be an exception to the rule, and not frequently used.
 
 ### Positive Consequences
 
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -1,0 +1,82 @@
+# Decoupling from IETF
+
+* Status: accepted
+* Deciders: @jdesrosiers @relequestual @awwright
+* Date: 2022-08-17
+
+Related Issues:
+* https://github.com/orgs/json-schema-org/discussions/69 - This issue is about
+  dropping the "draft" prefix from our releases. This ADR doesn't cover that,
+  but much of the discussion about whether or not to decouple from IETF is in
+  that discussion.
+* This topic has been discussed in many other places as well that are difficult
+  to link to here including Slack, Twitter, OCWMs, and conference discussions.
+
+## Context and Problem Statement
+
+Currently JSON Schema follows the IETF Internet-Draft (I-D) process for spec
+development and releases but aren't associated with IETF in any way. Our
+perceived involvement with IETF causes confusion and misunderstanding within our
+community in the cases were our needs and the realities of our situation deviate
+from the IETF process.
+
+## Decision Drivers
+
+* IETF's draft versioning system doesn't work for JSON Schema and we stopped
+  using it to version our releases a while ago. We now use a date based version
+  and even have more than one draft submission per release (the initial release
+  and the patch release).
+* The IETF process is optimized for working on a draft until it's done and then
+  disbanding. In some cases, the RFC may be revisited and revised in the future,
+  but this is rare and generally contains little more than clarifications and
+  reference updates. JSON Schema is not like that. JSON Schema is more like a
+  programming language. When it stops evolving, it will loose it's relevance.
+  When we finish a release of JSON Schema, we don't disband, we start work on
+  the next release.
+* Every one of our releases is expected to be used in production and will be
+  depended on for many years forward. This is not consistent with normal IETF
+  drafts. Even if we don't publicly use the term "draft", we're still using the
+  IETF I-D system in a way that's not intended.
+* Under IETF, JSON Schema fits under the category of "draft". The community has
+  repeatedly told us that they perceive this to meant that JSON Schema
+  "incomplete" and not "not ready for production use". This is the wrong message
+  for us to be sending as all of our releases are intended to be used in
+  production. This ADR doesn't decide whether or not to drop the "draft" from
+  our releases, but decoupling from IETF gives us that option.
+* Several members of the JSON Schema team have had poor interactions with IETF
+  and don't feel that working with them would be productive. This is a
+  relatively minor consideration. If we thought IETF was right for JSON Schema,
+  we could find ways to make those relationships work.
+
+## Considered Options
+
+1. Continue to submit I-Ds, while using our customized process with no intention
+  of pursing standards track RFC status.
+2. Get JSON Schema to a stable place and pursue a standards track RFC with the
+  IETF.
+3. Decouple from IETF completely and come up with our own governance model.
+
+## Decision Outcome
+
+Our decision is to go with Option 3 and decouple from IETF completely and come
+up with our own governance model. The idea of having the come up with our own
+governance model is daunting, but when the alternatives are to continue to abuse
+the I-D process or to conform to the IETF process that doesn't fit our needs,
+going our own way seems the bast option.
+
+### Positive Consequences
+
+* Decoupling from IETF allows us to distance ourselves from the assumptions that
+  people make about JSON Schema because they assume it works like a typical I-D.
+* Decoupling from IETF allows us to customize our governance model to what works
+  best for JSON Schema.
+
+### Negative Consequences
+
+* Choosing our own governance model means we aren't associated with a recognized
+  standard body which could reduce to credibility of JSON Schema in the eyes of
+  some.
+* If we don't go the standardization route with IETF, we lose access to their
+  expert review process.
+* Defining our own process will be a lot of work and none of use have expertise
+  in defining such a process.
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -37,23 +37,30 @@ that our "drafts" are intended for use in production.
   programming language. When it stops evolving, it will lose its relevance.
   When we finish a release of JSON Schema, we don't disband, we start work on
   the next release.
-* Every one of our releases is expected to be used in production and will be
-  depended on for many years forward. This is not consistent with normal IETF
-  drafts. Even if we don't publicly use the term "draft", we're still using the
-  IETF I-D system in a way that's not intended.
+* Since the project resumed activity after the gap following draft-04, every one
+  of our releases is expected to be used in production and will be depended on
+  for many years forward. This is not consistent with normal IETF drafts. Even
+  if we don't publicly use the term "draft", we're still using the IETF I-D
+  system in a way that's not intended.
 * Under IETF, JSON Schema fits under the category of "draft". The community has
   repeatedly told us that they perceive this to meant that JSON Schema
   "incomplete" and not "not ready for production use". This is the wrong message
   for us to be sending as all of our releases are intended to be used in
   production. This ADR doesn't decide whether or not to drop the "draft" from
   our releases, but decoupling from IETF gives us that option.
-* Several members of the JSON Schema team have had poor interactions with IETF
-  and don't feel that working with them would be productive. This is a
-  relatively minor consideration. If we thought IETF was right for JSON Schema,
-  we could find ways to make those relationships work. We have a good
-  relationship with the relatively new HTTPAPIs working group and working with
-  them would be far more likely to be productive than the people/WG we were
-  previously in communication with.
+* Several members of the JSON Schema team have interacted with JSON-related IETF
+  working groups. Some of these interactions demonstrated an indifference or
+  hostility to JSON Schema, and a preference for projects taking a different
+  approach. Equally important was a lack of any active interest or constructive
+  engagement. Finally, we were informed that any schema project for JSON would
+  not necessarily start from JSON Schema as a base, indicating that a "JSON
+  Schema" working group would quite likely not involve JSON Schema itself. This
+  impression has been reinforced by observing the amount of change introduced to
+  JSON Path as a consequence of its adoption by an IETF working group. While we
+  have a good relationship with the relatively new HTTPAPIs working group, the
+  combination of these other experiences with other mismatches between our
+  project and the IETF process contributes to our reluctance to move forward
+  through the iETF.
 
 ## Considered Options
 
@@ -77,6 +84,17 @@ their SDLC and will likely choose one to emulate and adapt to our needs.
 Although we don't have a replacement solution in place yet, we are confident
 that continuing to abuse the IETF I-D process or conforming to a standards
 organization process that doesn't fit our needs is not the way to go.
+
+However, we still plan to use the IETF process to register the media types
+defined by JSON Schema with IANA. This effort is currently in progress with the
+HTTPAPIs working group.
+
+The decision to not use IETF applies only to the main specification documents
+and not necessarily supporting components we have defined or will define in the
+future. Currently our only such component is Relative JSON Pointer, but there
+could be others in the future. These components will be examined on a case by
+case basis and we may choose an IETF standards path for those components if it
+makes sense.
 
 Option 2 and 3 are still on the table if we feel it makes sense when we get to a
 more stable place in the future. The main concern is the pain this process is
@@ -108,7 +126,7 @@ unproductive and have concerns about JSON Schema getting deeper involved without
 compelling enough reason. Most agree that the reasons are not sufficiently
 compelling at this point.
 
-Option 3 was rejected because it has the same problems as Option 2 accept that
+Option 3 was rejected because it has the same problems as Option 2 except that
 we don't have the same unpleasant history with W3C than we do with IETF.
 
 ### Positive Consequences
@@ -120,20 +138,21 @@ we don't have the same unpleasant history with W3C than we do with IETF.
 
 ### Negative Consequences
 
-* Not being associated with a recognized standards organization such as IETF,
-  W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
-  However, we have received feedback from people involved in standards
-  development that told us that they were comfortable referencing OpenAPI's self
-  published specification in their standards and that OpenAPI's membership with
-  the Linux Foundation was an important aspect of what makes them comfortable
-  doing so. JSON Schema is a member of the OpenJS Foundation, which is a
-  sub-group of the Linux Foundation, so we expect standards developers to be
-  just as comfortable referencing JSON Schema as they are referencing OpenAPI.
 * If we don't go the standardization route with IETF or W3C, we lose access to
   their expert review process.
+* Not being associated with a recognized standards organization such as IETF,
+  W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
+  However, we have received feedback that our membership with OpenJS/Linux
+  Foundation provides the credibility that we need.
 * One of the benefits of an RFC is other standards can normatively reference it,
-  and use JSON Schema to define their JSON-based syntaxes. Independently
-  publishing a specification does not permit this.
+  and use JSON Schema to define their JSON-based syntaxes. However, we have
+  received feedback from people involved in standards development that told us
+  that they were comfortable referencing OpenAPI's self published specification
+  in their standards and that OpenAPI's membership with the Linux Foundation was
+  an important aspect of what makes them comfortable doing so. JSON Schema is a
+  member of the OpenJS Foundation, which is a sub-group of the Linux Foundation,
+  so we expect standards developers to be just as comfortable referencing JSON
+  Schema as they are referencing OpenAPI.
 * Defining our own SLDC process will be a lot of work and none of us have
   expertise in defining such a process. However, we can take inspiration from
   existing well established projects and we would have the freedom to update our
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -1,7 +1,7 @@
 # Decoupling from IETF
 
 * Status: accepted
-* Deciders: @jdesrosiers @relequestual @awwright
+* Deciders: @jdesrosiers @relequestual @awwright @handrews
 * Date: 2022-08-17
 
 Related Issues:
@@ -123,9 +123,12 @@ we don't have the same unpleasant history with W3C than we do with IETF.
 * Not being associated with a recognized standards organization such as IETF,
   W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
   However, we have received feedback from people involved in standards
-  development that say they are comfortable adopting OpenAPI without it being
-  associated with a standards organization, so we don't expect this to be a
-  significant issue for JSON Schema either.
+  development that told us that they were comfortable referencing OpenAPI's self
+  published specification in their standards and that OpenAPI's membership with
+  the Linux Foundation was an important aspect of what makes them comfortable
+  doing so. JSON Schema is a member of the OpenJS Foundation, which is a
+  sub-group of the Linux Foundation, so we expect standards developers to be
+  just as comfortable referencing JSON Schema as they are referencing OpenAPI.
 * If we don't go the standardization route with IETF or W3C, we lose access to
   their expert review process.
 * One of the benefits of an RFC is other standards can normatively reference it,
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -122,9 +122,10 @@ we don't have the same unpleasant history with W3C than we do with IETF.
 
 * Not being associated with a recognized standards organization such as IETF,
   W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
-  However, people seem comfortable adopting OpenAPI without it being associated
-  with a standards organization, so we don't expect this to be a significant
-  issue for JSON Schema either.
+  However, we have received feedback from people involved in standards
+  development that say they are comfortable adopting OpenAPI without it being
+  associated with a standards organization, so we don't expect this to be a
+  significant issue for JSON Schema either.
 * If we don't go the standardization route with IETF or W3C, we lose access to
   their expert review process.
 * One of the benefits of an RFC is other standards can normatively reference it,
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -14,16 +14,16 @@ Related Issues:
 
 ## Context and Problem Statement
 
-Currently JSON Schema follows the IETF Internet-Draft (I-D) process for spec
-development and releases but aren't associated with IETF in any way. Our
+Currently JSON Schema loosely follows the IETF Internet-Draft (I-D) process for
+spec development and releases but isn't associated with IETF in any way. Our
 perceived involvement with IETF causes confusion and misunderstanding within our
-community in the cases were our needs and the realities of our situation deviate
-from the IETF process.
+community in the cases were our practices and the realities of our situation
+deviate from the typical IETF I-D process.
 
 ## Decision Drivers
 
 * IETF's draft versioning system doesn't work for JSON Schema and we stopped
-  using it to version our releases a while ago. We now use a date based version
+  using it to version our releases a while ago. We now use date based versioning
   and even have more than one draft submission per release (the initial release
   and the patch release).
 * The IETF process is optimized for working on a draft until it's done and then
@@ -52,31 +52,36 @@ from the IETF process.
 
 1. Continue to submit I-Ds, while using our customized process with no intention
   of pursing standards track RFC status.
-2. Get JSON Schema to a stable place and pursue a standards track RFC with the
-  IETF.
-3. Decouple from IETF completely and come up with our own governance model.
+2. Go all-in with IETF and pursue a standards track RFC with the IETF.
+3. Join W3C and pursue a standards track with them using their process.
+4. Decouple completely from any standards organization and come up with our own
+   specification development lifecycle (SDLC) model inspired by well established
+   projects with an SDLC that more closely meets or needs.
 
 ## Decision Outcome
 
-Our decision is to go with Option 3 and decouple from IETF completely and come
-up with our own governance model. The idea of having to come up with our own
-governance model is daunting, but when the alternatives are to continue to abuse
-the I-D process or to conform to the IETF process that doesn't fit our needs,
-going our own way seems the best option.
+Our decision is to go with Option 4 and decouple from standards organizations
+that don't fit our needs. We don't currently have a plan for what to replace
+IETF with. We are currently investigating how other established projects do
+their SDLC and will likely choose one to emulate and adapt to our needs.
+Although we don't have a replacement solution in place yet, we are confident
+that continuing to abuse the IETF I-D process or conforming to a standards
+organization process that doesn't fit our needs is not the way to go.
 
 ### Positive Consequences
 
 * Decoupling from IETF allows us to distance ourselves from the assumptions that
   people make about JSON Schema because they assume it works like a typical I-D.
-* Decoupling from IETF allows us to customize our governance model to what works
-  best for JSON Schema.
+* Decoupling from IETF allows us to customize our SDLC model to what works best
+  for JSON Schema.
 
 ### Negative Consequences
 
-* Choosing our own governance model means we aren't associated with a recognized
-  standard body which could reduce to credibility of JSON Schema in the eyes of
-  some.
-* If we don't go the standardization route with IETF, we lose access to their
-  expert review process.
-* Defining our own process will be a lot of work and none of use have expertise
-  in defining such a process.
+* Not being associated with a recognized standards organization such as IETF,
+  W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
+* If we don't go the standardization route with IETF or W3C, we lose access to
+  their expert review process.
+* Defining our own SLDC process will be a lot of work and none of us have
+  expertise in defining such a process. However, we can take inspiration from
+  existing well established projects and we would have the freedom to update our
+  process as we learn what works and what doesn't.
--- a/adr/2022-08-decouple-from-ietf.md
+++ b/adr/2022-08-decouple-from-ietf.md
@@ -1,7 +1,7 @@
 # Decoupling from IETF
 
 * Status: accepted
-* Deciders: @jdesrosiers @relequestual @awwright @handrews
+* Deciders: @jdesrosiers @relequestual @awwright @handrews @gregsdennis
 * Date: 2022-08-17
 
 Related Issues:


adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,82 @@
...
+* Decoupling from IETF allows us to distance ourselves from the assumptions that
+  people make about JSON Schema because they assume it works like a typical I-D.
+* Decoupling from IETF allows us to customize our governance model to what works
+  best for JSON Schema.
+
+### Negative Consequences
+
+* Choosing our own governance model means we aren't associated with a recognized
+  standard body which could reduce to credibility of JSON Schema in the eyes of
+  some.
```
handrews, line 78: This is where putting an emphasis on our involvement with the OpenJS Foundation / Linux Foundation is important, and why I think that decision belongs in this ADR.  The ADR will be a public statement of intent, and stating that we plan to work within that context is a more re-assuring one than stating that we're abandoning the IETF without a new plan.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,82 @@
...
+* If we don't go the standardization route with IETF, we lose access to their
+  expert review process.
+* Defining our own process will be a lot of work and none of use have expertise
+  in defining such a process.
```
handrews, line 82: Pointing at a project or two (whether that's OpenAPI or something else) and saying that we intend to do something similar would put us in a better position while also reassuring the community that we're not just winging it.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,82 @@
+# Decoupling from IETF
+
+* Status: accepted
+* Deciders: @jdesrosiers @relequestual @awwright
```
handrews, line 4: Has this decision already been made?  I guess I'm a bit confused about the ADR process.  If this is a done deal and all this PR is concerned with is the write-up, then my comments probably don't apply.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,82 @@
...
+
+* IETF's draft versioning system doesn't work for JSON Schema and we stopped
+  using it to version our releases a while ago. We now use a date based version
+  and even have more than one draft submission per release (the initial release
+  and the patch release).
+* The IETF process is optimized for working on a draft until it's done and then
+  disbanding. In some cases, the RFC may be revisited and revised in the future,
+  but this is rare and generally contains little more than clarifications and
+  reference updates. JSON Schema is not like that. JSON Schema is more like a
+  programming language. When it stops evolving, it will loose it's relevance.
```
notEthan, line 33:
```suggestion
  programming language. When it stops evolving, it will lose its relevance.
```

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,82 @@
...
+1. Continue to submit I-Ds, while using our customized process with no intention
+  of pursing standards track RFC status.
+2. Get JSON Schema to a stable place and pursue a standards track RFC with the
+  IETF.
+3. Decouple from IETF completely and come up with our own governance model.
+
+## Decision Outcome
+
+Our decision is to go with Option 3 and decouple from IETF completely and come
+up with our own governance model. The idea of having the come up with our own
```
notEthan, line 62:
```suggestion
up with our own governance model. The idea of having to come up with our own
```

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,82 @@
...
+governance model is daunting, but when the alternatives are to continue to abuse
+the I-D process or to conform to the IETF process that doesn't fit our needs,
+going our own way seems the bast option.
```
notEthan, line 65:
```suggestion
going our own way seems the best option.
```

handrews: > going our own way seems the bast option.

And here I was hoping it was the ancient Egyptian cat deity option... 🐈

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,82 @@
...
+  relatively minor consideration. If we thought IETF was right for JSON Schema,
+  we could find ways to make those relationships work.
+
+## Considered Options
+
+1. Continue to submit I-Ds, while using our customized process with no intention
+  of pursing standards track RFC status.
+2. Get JSON Schema to a stable place and pursue a standards track RFC with the
+  IETF.
+3. Decouple from IETF completely and come up with our own governance model.
```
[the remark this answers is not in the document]
jdesrosiers: Yes, good point about W3C. It has been discussed, I don't think we've dug into this as much as other options. I think we've mostly just assumed that it has the same kind of issues with "fit" as IETF, but I don't know the W3C processes as well as IETF so I can't really say that for sure.

I'm not really sure OpenJS/LinuxFoundation are relevant here. What we're talking about is the spec development lifecycle, which doesn't have anything to do with OpenJS. Whether we do something similar to OpenAPI/AsyncAPI or something different is still undecided, so I'm not sure bringing that up here makes sense.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: > stating that we're abandoning the IETF without a new plan.

I felt uncomfortable writing this up saying we're sure we don't want IETF without being able to say precisely (or even vaguely) what we do want. Maybe that means it's too early for this ADR?

> This is where putting an emphasis on our involvement with the OpenJS Foundation / Linux Foundation is important

I disagree. While OpenJS does provide credibility in some ways, I was referring to the credibility of being a standard recognized by recognized standards body (IETF, W3C, IEEE, etc). OpenJS is not standards body. I can reword to make that perspective more clear. We are already part of OpenJS and nothing we are deciding here changes that.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: I'd rather not put something in an ADR that says what we might do, but I think I can add some wording to the effect that we will be taking inspiration from how other well established projects do things.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: I wrote this up because @Relequestual asked me to, but I agree that it feels awkward in that I don't feel like there was ever a point where a decision was made.

I'm viewing this PR as a proposed resolution that isn't a "done deal" until it's merged. At this point, I think there's more than enough consensus to write up this PR, but not yet enough to merge it. The PR is a sort of "last call" for people to express support or dissent before a final decision is made.

jdesrosiers:
I pushed draft 2. This time I replaced the mention of "governance" with "specification development lifecycle (SDLC)". I don't think "governance" was the right word and its use confuses the purpose of the ADR.

adr/2022-08-decouple-from-ietf.md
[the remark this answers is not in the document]
handrews: > I'm not really sure OpenJS/LinuxFoundation are relevant here.

That's fair (here and in the other comment where you mention something similar).  I'm a little vague on exactly what OpenJS brings.  My main point is that people are comfortable with OpenAPI/AsyncAPI (whatever it is that they're doing), so while they don't have the endorsement of an Official Standards Organization™, it doesn't seem to have been a problem.

So I think there's some good precedent there.  I like pointing to them as a precedent more than, say, WHATWG.  I haven't heard complaints about the governance or process of OpenAPI, whereas WHATWG has provoked a non-trivial amount of dislike.  I really don't want to be a WHATWG and launch public broadsides against the IETF (or W3C) in our actual documents.

Negative experiences with some IETF working groups that happen to be the most relevant aside, I don't really have anything against the IETF (well, except their insistence on using incredibly outdated mailing list software, but that's not a _serious_ complaint).

adr/2022-08-decouple-from-ietf.md
handrews: > I felt uncomfortable writing this up saying we're sure we don't want IETF without being able to say precisely (or even vaguely) what we do want. Maybe that means it's too early for this ADR?

Maybe - if the purpose of the discussion here is to sort that out, then it's fine.  But yeah, I think having at least a vague statement of what we're doing instead is important, and if we're not at the point of figuring that out, we should focus on that.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -14,16 +14,16 @@ Related Issues:
 
 ## Context and Problem Statement
 
-Currently JSON Schema follows the IETF Internet-Draft (I-D) process for spec
-development and releases but aren't associated with IETF in any way. Our
+Currently JSON Schema loosely follows the IETF Internet-Draft (I-D) process for
+spec development and releases but isn't associated with IETF in any way. Our
 perceived involvement with IETF causes confusion and misunderstanding within our
```
handrews, line 19: It's actually common and correct for individual drafts (drafts with someone's name in the identifier) to not have any formal connection to the IETF.  You only get a formal connection when the I-D is adopted by a working group.  So this is not actually unusual or incorrect.  People thinking there is IETF support behind a draft-handrews or draft-bhutton is incorrect.  (sorry I didn't notice this last time)

adr/2022-08-decouple-from-ietf.md
```diff
@@ -14,16 +14,16 @@ Related Issues:
...
-community in the cases were our needs and the realities of our situation deviate
-from the IETF process.
+community in the cases were our practices and the realities of our situation
```
handrews, line 20: I'm pretty sure the only way our specification-writing practices deviate from anything is that we take too long between drafts, which expire after 6 months.  But that's actually pretty common, even if we have a relatively extreme case.

The "realities of our situation" part (with broad adoption for over a decade despite being an individual draft, and with the lack of suitability of what we've been told would happen if we joined a working group) are the real issues here.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -52,31 +52,36 @@ from the IETF process.
...
 1. Continue to submit I-Ds, while using our customized process with no intention
   of pursing standards track RFC status.
-2. Get JSON Schema to a stable place and pursue a standards track RFC with the
-  IETF.
-3. Decouple from IETF completely and come up with our own governance model.
+2. Go all-in with IETF and pursue a standards track RFC with the IETF.
+3. Join W3C and pursue a standards track with them using their process.
+4. Decouple completely from any standards organization and come up with our own
+   specification development lifecycle (SDLC) model inspired by well established
+   projects with an SDLC that more closely meets or needs.
```
handrews, line 59: I like this wording a lot, thanks!

adr/2022-08-decouple-from-ietf.md
```diff
@@ -52,31 +52,36 @@ from the IETF process.
...
 
-Our decision is to go with Option 3 and decouple from IETF completely and come
-up with our own governance model. The idea of having to come up with our own
-governance model is daunting, but when the alternatives are to continue to abuse
-the I-D process or to conform to the IETF process that doesn't fit our needs,
-going our own way seems the best option.
+Our decision is to go with Option 4 and decouple from standards organizations
+that don't fit our needs. We don't currently have a plan for what to replace
+IETF with. We are currently investigating how other established projects do
+their SDLC and will likely choose one to emulate and adapt to our needs.
```
handrews, line 66: I think it would be good to identify a few written SDLC processes (as opposed to just pointing to projects and saying "what they're doing", although some SDLC processes may be project-specific documents and still be worth referencing) that we are considering.  That would make this feel a little less like we're just kind of wandering off indefinitely.

I do not have links for SDLCs off the top of my head, I figure I'd see what the reaction to this idea was before doing any research.

adr/2022-08-decouple-from-ietf.md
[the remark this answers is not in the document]
jdesrosiers: > while they don't have the endorsement of an Official Standards Organization™, it doesn't seem to have been a problem.

Ah, yes, I see your point now and agree that it's worth mentioning.

awwright:
Hold up a bit... I feel like this misrepresents many things. This view seems to assume JSON Schema is only used to standardize validators, but we must also think about the larger standards community: One of the benefits of an RFC is other standards can normatively reference it, and use JSON Schema to define their JSON-based syntaxes. Independently publishing a specification does not permit this.

---

> IETF's draft versioning system doesn't work for JSON Schema and we stopped
  using it to version our releases a while ago. We now use date based versioning
  and even have more than one draft submission per release (the initial release
  and the patch release).

I'm not sure what this is referring to. We switched from serially numbering the meta-schemas to using dates, but this was unrelated to how the I-D system works. (We never published the meta-schemas through the I-D process, they were non-normative.)

> The IETF process is optimized for working on a draft until it's done and then
  disbanding. In some cases, the RFC may be revisited and revised in the future,
  but this is rare and generally contains little more than clarifications and
  reference updates. JSON Schema is not like that. JSON Schema is more like a
  programming language. When it stops evolving, it will lose its relevance.
  When we finish a release of JSON Schema, we don't disband, we start work on
  the next release.

This doesn't have to be true if you don't want it to be. An RFC is not the complete and closed description of everything the system is capable of doing. It's supposed to be able to evolve without updates to the specification text.

All an RFC has to do is describe is the essential characteristics of evaluating a JSON Schema: the keywords that everybody is guaranteed to support, and any extension mechanisms.

> Every one of our releases is expected to be used in production and will be
  depended on for many years forward. This is not consistent with normal IETF
  drafts. Even if we don't publicly use the term "draft", we're still using the
  IETF I-D system in a way that's not intended.

Expecting use in production would be a strong reason to publish an RFC.

> Under IETF, JSON Schema fits under the category of "draft". The community has
  repeatedly told us that they perceive this to meant that JSON Schema
  "incomplete" and not "not ready for production use". This is the wrong message
  for us to be sending as all of our releases are intended to be used in
  production. This ADR doesn't decide whether or not to drop the "draft" from
  our releases, but decoupling from IETF gives us that option.

This also seems like a good argument to get the specification published as an RFC.

> Several members of the JSON Schema team have had poor interactions with IETF
  and don't feel that working with them would be productive. This is a
  relatively minor consideration. If we thought IETF was right for JSON Schema,
  we could find ways to make those relationships work.

All of those interactions were with members of the general public.

handrews:
@awwright prominent members of the JSON working group told us (in email I can no longer find b/c their email software is such a mess) that they would likely throw away the current spec and start over.  And that appears to be the sort of thing that is going on with JSON Path.  You can't dismiss that as "members of the general public" because these were people who had just published the JSON RFC and have deep knowledge of the considerations involved.

I don't think most of us disagree with the idea that it would be great if we published JSON Schema as a proper standards-track RFC.  While I understand @jdesrosiers's concerns about evolution, I tend to agree with you that that could probably be sorted out somehow.

But I do not see a path for standardizing what we consider to be JSON Schema.  People who would have influence have directly told us that there is no such path.  The JSON Path process seems to reinforce that.

So why do you think we should pursue the IETF?  You are asking us to believe that such a process would go in a radically different direction than we have been told by knowledgeable people, and can see in a similar project.

I just can't take your recommendation seriously in the face of that.  You have not offered any explanation for why you think an IETF working group would go well for us, in the face of significant evidence that it would not.

admin-cimug:
Hi @handrews.  I'm subscribed to the watch list for the JSON schema spec.  I represent the U.S. as part of ANSI (i.e. the American National Standards Institute) and am an expert member of the U.S. National Committee on global standards for the Smart Grid / Power Systems industry.  In this capacity I serve on the IEC International Standards body which is an SDO that works closely with other groups such as ISO and IEEE.  Work we've undertaken over the last couple years would categorize us as a 'downstream project'.  Specifically, we have drafted new data exchange standards that are dependent upon work done as part of draft-2019-09 and later pivot to adjust for draft-2020-12 after later decisions were made with regards to 2019-09.  @Relequestual and I have chatted off and on early on at a few points over this time period. As implied by your comments...many standards bodies (including the IEC) lean strongly that any dependent "standards" themselves be based upon stable, normative standards from other recognized SDO's (e.g. IETF, W3C, ISO, IEEE, etc.)  Originally we were under the auspices that the direction of the JSON schema spec was as an eventual IETF RFC.  This generally provides the requisite assurance needed to proceed.  At present though we are in a holding pattern.  In the interim period we've gone ahead and implemented two early reference implementations against the draft publication.  One for potential use for data exchanges amongst E.U. power systems.  

So to summarize, my feedback is that we absolutely need some form of definitive processes in place to sanction what can be publicly recognized as a "normatively published standard" of the JSON schema specification.  This must carry a bit more weight and have more than just the perception of it as a "de facto standard" if you know what I mean.  

Anyway, thanks for engaging and soliciting outside thoughts/perspectives.  I commend the work you and the team have done and understand the at times rather painful and laborious process that drafting "standards" entails.

handrews:
@admin-cimug thanks so much for this comment, it's tremendously helpful.  I'd like to understand how a couple of possible scenarios might impact you.

Currently, @jdesrosiers (from our core team) and @ioggstream (as the spec editor - he is associate with the Italian government so I suspect he has some similar constraints as you) are working on [registering the JSON Schema media types](https://github.com/ietf-wg-httpapi/mediatypes) through an IETF [informational-track RFC](https://www.rfc-editor.org/search/rfc_search_detail.php?page=All&pubstatus[]=Informational&pub_date_type=any&sortkey=Number&sorting=ASC).  This effort is already under the aegis of the IETF's [HTTP API working group](https://datatracker.ietf.org/wg/httpapi/about/). (no one here is opposed to the IETF in general, or thinks that everyone involved with it is hostile to JSON Schema!)

Informational-track RFCs are what is used to document de-facto standards in a more official way, but it's not clear to me how well that satisfies your requirements.  Here's the most relevant thing [RFC 1796](https://datatracker.ietf.org/doc/html/rfc1796) says about informational RFCs:

>    The community is also well served by having access to specifications
   of which have been developed outside the IETF standards process,
   either because the protocols are experimental in nature, were
   developed privately, or failed to achieve the acquire _[sic]_ the degree of
   consensus required for elevation to the standards track.
>
>   The IAB believes that publication is better than ignorance.  If a
   particular specification ends up being used in products that are
   deployed over the Internet, we are better off if the specification is
   easy to retrieve as an RFC than if it is hidden in some private
   repository.

We are still working out exactly what the boundary between the media type specification and the main specification of JSON Schema is, but it at minimum needs to specify how to identify the version/dialect of JSON Schema, so that an implementation can figure out which document(s) govern further processing.

So, questions that arise:

1.  How does the use of an informational RFC (orange bar such as this [update to ECMAScript media types](https://datatracker.ietf.org/doc/html/rfc9239)], as opposed to standards-track ([blue bar](https://datatracker.ietf.org/doc/html/rfc6901) or [green bar](https://datatracker.ietf.org/doc/html/rfc8259)) impact your situation?
2.  If `application/schema+json` and `application/schema-instance+json` were registered through an informational RFC, which specified how to determine what other documents specified processing the contents of any resource with that media type, would you still need those other documents to have some sort of formal standards body endorsement?
3.  If the media type registration alone is insufficient, would it be sufficient to publish all of JSON Schema (not just the media type registration) as informational-track RFCs?

handrews:
@admin-cimug a few more questions:

* how do specifications like OpenAPI ([charter](https://www.openapis.org/participate/how-to-contribute/governance)) or AsyncAPI ([charter](https://github.com/asyncapi/community/blob/master/CHARTER.md)) which have their own governance processes and are affiliated with the Linux Foundation, fit into your work?  Are they considered usable in the way that you would need JSON Schema to be (JSON Schema is also associated with the Linux Foundation by way of the OpenJS Foundation)?
* while I hate to bring them up because I dislike how they've gone about their interactions with standards groups, how does [WHATWG](https://whatwg.org/faq) fit in?  I guess they sort-of worked things out with W3C eventually, but for a while there HTML was basically the subject of an unpleasant turf war (something I very much want to avoid - independent is one thing, hostile is another entirely), and as far as I know there's no resolution between WHATWG's URL "living standard" and RFC 3986.
* Are there any standards groups (other than IETF, W3C, IEEE, or ISO) that we should consider but might not know about?  IETF and W3C seem the most relevant.

jdesrosiers:
@awwright 

> Hold up a bit... I feel like this misrepresents many things.

I want to make sure this document accurately represents your side of the discussion even if you don't agree with the conclusion. It would not have been merged without your feedback.

> One of the benefits of an RFC is other standards can normatively reference it, and use JSON Schema to define their JSON-based syntaxes. Independently publishing a specification does not permit this.

Good point. I will add this to the document. And, thank you @admin-cimug for providing an example of when this might be needed.

> We switched from serially numbering the meta-schemas to using dates, but this was unrelated to how the I-D system works.

This is not a meta-schema issue. This is about how we represent releases of JSON Schema. People don't implement draft-handrews-json-schema-01 or draft-bhutton-json-schema-01, they implement draft-07 or 2020-12. We issue our releases as I-Ds, but the details are hidden behind our versioning system because the I-D versions don't fit our needs. For example, we have been releasing patch updates to clarify things or fix bugs. We don't want these non-functional updates to change the JSON Schema version because that would be confusing and unnecessary.

> An RFC is not the complete and closed description of everything the system is capable of doing. It's supposed to be able to evolve without updates to the specification text.

I'll add this perspective to the document. I agree that there is a possible path here, but it means we have to put up with the I-D process and it's incompatibility with our needs until we get there. I don't think we're particularly close to ready for an RFC release.

> Expecting use in production would be a strong reason to publish an RFC.

Not everything that is used in production needs to be recognized by a standards body. I'm not aware of single programming language that has an RFC with IETF or any other standard body. No one has any issue using those languages. Why should it be any different for us?

> All of those interactions were with members of the general public.

I wasn't involved, but first hand reports are telling me otherwise. I'm going the leave this in the document. I feel that it accurately represents the way many of the core contributors to the project feel about working with IETF even if there are potentially misunderstanding involved. However, I can add a reference to the relatively new HTTPAPI working group that is far more likely to be receptive to our project than the WG that we interacted with in the past.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: Yes, that was what I was trying to express. I'll see if I can make that more clear.

handrews:
> However, I can add a reference to the relatively new HTTPAPI working group that is far more likely to be receptive to our project than the WG that we interacted with in the past.

Yeah I think it would be good to show that we do not have a blanked negative reaction to or dislike of the IETF.  Our concerns come from specific past statements that getting JSON Schema adopted by a working group would more or less involve starting over.  If there turns out to be a working group that would take a different approach, that would be worth considering.

It is worth noting (here, not necessarily in the ADR text) that the HTTPAPI working group is producing informational, rather than standards-track RFCs, at least for the ones we're involved with.  The prior discussions were in the context of a standards-track working group/RFC.

jdesrosiers:
I just pushed a new draft incorporating feedback from @handrews and @awwright. @awwright I hope this gets closer to accurately representing your point of view.

jdesrosiers:
> It is worth noting [...] that the HTTPAPI working group is producing informational, rather than standards-track RFCs

I don't think that's true. There was discussion about "tracks" at one point and it was decided that the group would default to standards track unless there was reason not to. The links, problem+json, deprecation, and rate-limiting efforts are all standards track if I'm not mistaken.

handrews:
> There was discussion about "tracks" at one point and it was decided that the group would default to standards track unless there was reason not to.

Ah, this is good news!  @admin-cimug we may need to add the "media type registration is a standards-track RFC, with the rest of JSON Schema behavior described some other way" as an option.

@jdesrosiers I was going by [this setting](https://github.com/ietf-wg-httpapi/mediatypes/blob/main/draft-ietf-httpapi-rest-api-mediatypes.md?plain=1#L5) which produces "Intended Status: Informational" in the current published I-D.  But perhaps that is outdated?

jdesrosiers:
> I was going by [this setting](https://github.com/ietf-wg-httpapi/mediatypes/blob/main/draft-ietf-httpapi-rest-api-mediatypes.md?plain=1#L5) which produces "Intended Status: Informational" in the current published I-D. But perhaps that is outdated?

I'm not 100% sure of the timeline, but I'm pretty sure that the document was created before this was discussed and no one noticed that it still says "informational". I certainly didn't notice.

admin-cimug:
> @admin-cimug a few more questions:
> 
>     * how do specifications like OpenAPI ([charter](https://www.openapis.org/participate/how-to-contribute/governance)) or AsyncAPI ([charter](https://github.com/asyncapi/community/blob/master/CHARTER.md)) which have their own governance processes and are affiliated with the Linux Foundation, fit into your work?  Are they considered usable in the way that you would need JSON Schema to be (JSON Schema is also associated with the Linux Foundation by way of the OpenJS Foundation)?
> 
>     * while I hate to bring them up because I dislike how they've gone about their interactions with standards groups, how does [WHATWG](https://whatwg.org/faq) fit in?  I guess they sort-of worked things out with W3C eventually, but for a while there HTML was basically the subject of an unpleasant turf war (something I very much want to avoid - independent is one thing, hostile is another entirely), and as far as I know there's no resolution between WHATWG's URL "living standard" and RFC 3986.
> 
>     * Are there any standards groups (other than IETF, W3C, IEEE, or ISO) that we should consider but might not know about?  IETF and W3C seem the most relevant.

@handrews / @jdesrosiers :  
I apologize for the delay.  Thank you for your patience!  I had a scheduled standards meeting today that I added your questions as an agenda topic.  Other IEC expert members / European reps were attending so I was able to solicit their perspectives as well.  

First off...Henry you'd commented:

> Our concerns come from specific past statements that getting JSON Schema adopted by a working group would more or less involve starting over. If there turns out to be a working group that would take a different approach, that would be worth considering.

I completely agree and don't think that the JSON schema spec team should find themselves cornered into a position to start over if that's where you find yourself at.  There are many instances in my own involvement in standards where technologies have evolved/matured and have been introduced and accepted as the "baseline" for a standard from which future evolution can occur.  To give you one practical example is the OData (Open Data Protocol) standard.  It is an [ISO/IEC approved](https://www.oasis-open.org/news/pr/iso-iec-jtc-1-approves-oasis-odata-standard-for-open-data-exchange), [OASIS standard](https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=odata) that defines an option for a set of best practices for building and consuming RESTful APIs. I don't know the nuances or the day-to-day of the IETF RFC process as well as I do our own at the IEC but hopefully there is some flexibility on that front.  

- I would avoid [WHATWG](https://whatwg.org/faq)
- The An OpenAPI approach would be acceptable.  Within our IEC 61968-100 series standards (i.e. "Enterprise Integration Implementation Profiles" of which I'm a member of that sub-team) we are exploring OpenAPI as one of the profile approaches.  
- As for the Linux foundation.  There's a favorable view of this consortium and I'd say that this also may be a good avenue if you find it to be complementary to what the team wants.  As far as the orange, blue or green bar.  Any of these should be fine.  I wanted to confirm this before expressing this independently.  As an outcome of my call today I got confirmation from several other key people do reconfirm and they were in agreement.  So these two would be options. I thinks it the formality and the published state even if it's an "Informational" such as RFC9239.  In my original authoring of our JSON data exchange format we provide a strict syntactical mapping from contextual profiles based on the CIM (i.e. a global canonical information model for the Power Systems industry) to JSON schema.  As part of the normative references section of the draft you'll see several examples of IETF related references below.  It is difficult to include here a "JSON schema Specification Internet Draft (draft 2020-12) December 2020" type reference in the list.  Definitely a Linux Foundation or an IETF informational category doc would do be acceptable.

image

- Finally, I think anything with the word "draft" included in it is not preferable.  In this sense it is about perception when dealing from the SDO viewpoint.  So if you end up establishing a JSON schema specific charter try to do releases similar to OAS.  For example, the **OAS 3.1.0** Release from February is a release of the specification that I would have no problem including in a normative future published standard corresponding to [IEC 61968-100](https://webstore.iec.ch/publication/67766).  There would though be issue with **OAS 3.1.0-rc1** for example.  Even our own sanctioned profile tooling ([CIMTool](https://github.com/CIMug-org/CIMTool/releases/tag/1.10.0.RC3)) used by the IEC standards body and referenced within IEC published standards must be in a stable release for referencing normatively (I have a release today for this precise reason).

I hope this helps.  Please feel free to run past further questions or if I may have missed anthing.

~Todd

handrews:
@admin-cimug thank you so much for this information!  It is very helpful, especially since you were able to confer with a wider audience.

While none of this should be taken as a commitment (and I do not speak for the whole JSON Schema Org), I think this will all work out well.  While we have not made a final decision, there is [movement towards dropping the "draft" signifier](https://github.com/orgs/json-schema-org/discussions/69) with the next release.  It just means that we have to decide what our compatibility process is going forwards (a tricky topic) since we wouldn't be able to just wave our hands about breakages between drafts, and we need to by clear as to why people should rely on our publication if it is no longer an IETF draft.

Barring a truly unexpected development, there will be an IETF RFC on one the acceptable tracks registering the `application/schema+json` and `application/schema-instance+json` media types, which as noted is already under the aegis of the IETF HTTP APIs working group.  I don't know the timeline, but I think it mostly depends on us figuring out the boundary between that IETF spec and everything else, which we're working on right now.  I'm not aware of anything else holding it up, although @jdesrosiers would know better than I would (I think he's on vacation this week so he might not reply soon).

We are already listed as a Linux Foundation project (under [Web and Application Development](https://www.linuxfoundation.org/projects/#explore)) through our [affiliation](https://json-schema.org/blog/posts/json-schema-joins-the-openjsf) with the OpenJS Foundation, where we are listed as an [incubation-status project](https://openjsf.org/projects/#incubating).  Having only recently formally joined, we're still working through our checklist to exit incubation status.  See json-schema-org/community#129  for progress – I'm not sure if that's exactly the checklist for exiting incubation but it's stuff we are doing to finish onboarding, at least, so if anyone needs evidence that we're working on it, there it is in writing.

We need to get more of this information onto our web site, which is undergoing a redesign, but it sounds like if we officially publish a road map in this direction, it would reassure the formal standards-based community that we are making real progress towards being a properly cite-able specification.

Please let me know if you think of anything that would still be a concern.  We understand that things look a bit shaky right now, but we want folks to know that this is a transition rather than heading off into the wild west 🤠 🤓

jdesrosiers:
@admin-cimug Thank you for the feedback! I'll incorporate that into the document. If you are comfortable with OpenAPI, there's no reason you shouldn't be comfortable with what we end up with.

adr/2022-08-decouple-from-ietf.md
devinbost: If it's common for individual drafts to not have any formal connection with IETF, why are they saying that involvement with them would basically require starting over with JSON Schema?
It sounds like there were technical concerns. What were the issues?

jdesrosiers: > why are they saying that involvement with them would basically require starting over with JSON Schema?
It sounds like there were technical concerns. What were the issues?

@devinbost, that's a good question that maybe @handrews can provide more details about, but I'd prefer not to get into it in this PR. We don't expect that joining an IETF working group would mean starting over and the prospect was never a factor in this decision. If we were to join an IETF working group it would not be the one we talked to in the past, it would be the HTTPAPIs working group that we have a good relationship with today.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -122,9 +122,10 @@ we don't have the same unpleasant history with W3C than we do with IETF.
 
 * Not being associated with a recognized standards organization such as IETF,
   W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
-  However, people seem comfortable adopting OpenAPI without it being associated
-  with a standards organization, so we don't expect this to be a significant
-  issue for JSON Schema either.
+  However, we have received feedback from people involved in standards
+  development that say they are comfortable adopting OpenAPI without it being
+  associated with a standards organization, so we don't expect this to be a
```
handrews, line 127: This was said in the context of the connection with the Linux Foundation, not simply OpenAPI as a completely independent entity.  Which still works as we are also connected.  But it's important to keep the focus on connection to *some* sort of larger organization here.

jdesrosiers: It didn't appear to me that the answer to OpenAPI's approach being acceptable was related to or contingent on their connection with the Linux Foundation. Those appeared to be two separate answers to two separate questions. Considering that the Linux Foundation has nothing to do with standards development or publishing, it's hard to see how the two are related. However, I do see how association with the Linux Foundation provides credibility in the general sense and is therefore worth mentioning in this context.

handrews: >  However, I do see how association with the Linux Foundation provides credibility in the general sense and is therefore worth mentioning in this context.

Yes, that's what I was trying to get at, thank you for figuring that out despite my poor articulation!

admin-cimug: > This was said in the context of the connection with the Linux Foundation, not simply OpenAPI as a completely independent entity. Which still works as we are also connected. But it's important to keep the focus on connection to _some_ sort of larger organization here.

Agreed Henry.  Thanks for calling out this point.  It's an important aspect of how things will be perceived as JSON schemas is utilized as a dependency in other standards.

admin-cimug:
> While none of this should be taken as a commitment (and I do not speak for the whole JSON Schema Org), I think this will all work out well. While we have not made a final decision, there is [movement towards dropping the "draft" signifier](https://github.com/orgs/json-schema-org/discussions/69) with the next release. It just means that we have to decide what our compatibility process is going forwards (a tricky topic) since we wouldn't be able to just wave our hands about breakages between drafts, and we need to by clear as to why people should rely on our publication if it is no longer an IETF draft.

@handrews:  Thanks for this insight into the potential future direction of how the team me be releasing the spec.  I think this move of dropping the "draft" would be extremely beneficial for the folks in the standards/SDO spaces.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,137 @@
...
+  significant issue for JSON Schema either.
+* If we don't go the standardization route with IETF or W3C, we lose access to
+  their expert review process.
+* One of the benefits of an RFC is other standards can normatively reference it,
+  and use JSON Schema to define their JSON-based syntaxes. Independently
+  publishing a specification does not permit this.
+* Defining our own SLDC process will be a lot of work and none of us have
```
Relequestual, line 134: 🙃

```suggestion
* Defining our own SLDC process will be a lot of work and most of us have
```

Relequestual:
I see no problems with the ADR.
I'm to take an action from our last call to make some suggestions for updating and providing context around W3C considerations.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,137 @@
...
+  IETF I-D system in a way that's not intended.
+* Under IETF, JSON Schema fits under the category of "draft". The community has
+  repeatedly told us that they perceive this to meant that JSON Schema
+  "incomplete" and not "not ready for production use". This is the wrong message
+  for us to be sending as all of our releases are intended to be used in
+  production. This ADR doesn't decide whether or not to drop the "draft" from
+  our releases, but decoupling from IETF gives us that option.
+* Several members of the JSON Schema team have had poor interactions with IETF
+  and don't feel that working with them would be productive. This is a
+  relatively minor consideration. If we thought IETF was right for JSON Schema,
```
handrews, line 52: Since this is going to be a formal public document, I've been thinking about this wording.  I'm also going to explain some things here because they keep getting asked (most recently by @devinbost), so I spent a few hours tracking down the history.  Skip down to the last section for wording recs.

## Contacting the old JSON working group

When I contacted the JSON working group in January 2018 (see #526), [this quite helpful response](https://mailarchive.ietf.org/arch/msg/json/HSUXh5FAtToI1YiAbziHt1FyuNI/) from Paul Hoffman indicated some likely difficulties:

> There appears to be interest in something around schema for JSON; it is 
not clear if there is interest in starting from the specific spec at 
https://github.com/json-schema-org/json-schema-spec. I bet there is some 
interest in that, and some interest in something else. This is quite 
typical of requests for the IETF to adopt work that originated outside 
the IETF, with mixed results in any of of "adopt with the intention of 
making only minor changes", "adopt and make major changes", and "start 
from scratch while getting inspiration from the existing work".

While the above paragraph suggests that maybe we could have gotten it adopted, there were _zero_ encouraging responses or expressions of interest aside from Paul's response.  There were several non-productive criticisms and dismissals on such topics as [not needing `minimum` other than 0 or 1, and not needing `maximum` at all](https://mailarchive.ietf.org/arch/msg/json/uwAdA5i49ITIL6GuWePqRaOIGhI/).  Of course, individual poor-quality feedback is not necessarily representative, but it was the _lack of anything resembling constructive engagement_ was the more important concern.  Sadly I can't link to the absence of something.

In response to that discussion, someone else [brought up](https://mailarchive.ietf.org/arch/msg/json/zJDC6UtP0fIhVWeU-gF9Ig-RVYE/) a project called [JSON Content Rules](https://datatracker.ietf.org/doc/html/draft-newton-json-content-rules-09), which got much more interest (although notably, no further drafts were ever produced and the project was abandoned).  Somewhat later, the project that eventually became [JSON Type Definition](https://datatracker.ietf.org/doc/html/rfc8927) also got more attention (although note that it is an "experimental" RFC and [was never adopted by a working group](https://datatracker.ietf.org/doc/rfc8927/)).

The important point here is not that some people were grumpy and hostile (I'm... really not in a position to throw stones there), it's that ***no one was actively interested, and instead those who participated actively favored other projects.***

## The JSON Path experience

While I don't want to put anyone on the spot, we do know that JSON Path's adoption by a working group has led to more changes in that specification than were expected or wanted by at least some participants.  It essentially illustrates the fear discussed above:  that the momentum would be away from what works for our existing user community.

Of course, that doesn't meant the same thing would happen, but it is very relevant (and involves some overlap of people).

A perusal of email archives and GitHub issues and PRs also reveals a number of incompatible expectations in working styles and process.  Whether or not this qualifies as a "negative experience" depends on one's expectation, but it would probably be negative for us.

## Suggested wording

I would suggest replacing the "poor interactions" wording with something like:

> Several members of the JSON Schema team have interacted with JSON-related IETF working groups.  Some of these interactions demonstrated an indifference or hostility to JSON Schema, and a preference for projects taking a different approach.  Equally important was a lack of any active interest or constructive engagement.  Finally, we were informed that any schema project for JSON would not necessarily start from JSON Schema as a base, indicating that a "JSON Schema" working group would quite likely not involve JSON Schema itself.  This impression has been reinforced by observing the amount of change introduced to JSON Path as a consequence of its adoption by an IETF working group.  While we have a good relationship with the relatively new HTTPAPIs working group, the combination of these other experiences with other mismatches between our project and the IETF process contributes to our reluctance to move forward through the iETF.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,137 @@
...
+Currently JSON Schema loosely follows the IETF Internet-Draft (I-D) process for
+spec development and releases but isn't associated with any IETF working group.
+JSON Schema is an individual draft. That means it isn't on a standards track
+with IETF and IETF is not involved nor supports the spec in any way other than
+hosting the canonical version of our I-Ds. Our perceived involvement with IETF
+causes confusion and misunderstanding within our community in the cases were our
+practices and the realities of our situation deviate from a typical IETF I-D
+lifecycle. The thing that makes our situation different than a typical I-D is
+that our "drafts" are intended for use in production.
+
```
handrews, line 26: Can we add a note here that we will continue to pursue registering the JSON Schema media types through an RFC published by the the IETF HTTPAPIs working group?  I think including that up front in the "Context" will reassure folks that a.) we're not totally abandoning internet standards, and b.) we actually do have some experience with working groups and are not dismissing the approach blindly.

adr/2022-08-decouple-from-ietf.md
```diff
@@ -0,0 +1,137 @@
...
+  W3C, or IEEE reduces the credibility of JSON Schema in the eyes of some.
+  However, we have received feedback from people involved in standards
+  development that say they are comfortable adopting OpenAPI without it being
+  associated with a standards organization, so we don't expect this to be a
+  significant issue for JSON Schema either.
+* If we don't go the standardization route with IETF or W3C, we lose access to
+  their expert review process.
+* One of the benefits of an RFC is other standards can normatively reference it,
+  and use JSON Schema to define their JSON-based syntaxes. Independently
+  publishing a specification does not permit this.
```
handrews, line 136: We now have feedback that publishing independently in the context of a large industry organization will allow for this.

handrews (commented):
I just have a few minor suggestions - the giant text block is backstory for the folks who keep asking.  This looks ready to move out of draft status to me (I'm not marking "Approve" because I assume that should happen once it's out of draft?)

Relequestual:
@admin-cimug Thanks for joining and sharing your thoughts. Really appreciate you taking the time to not only answer questions, but put them to your wider group. This matters and will inform our decisions.

adr/2022-08-decouple-from-ietf.md
admin-cimug: > While the above paragraph suggests that maybe we could have gotten it adopted, there were _zero_ encouraging responses or expressions of interest aside from Paul's response. There were several non-productive criticisms and dismissals on such topics as [not needing `minimum` other than 0 or 1, and not needing `maximum` at all](https://mailarchive.ietf.org/arch/msg/json/uwAdA5i49ITIL6GuWePqRaOIGhI/). Of course, individual poor-quality feedback is not necessarily representative, but it was the _lack of anything resembling constructive engagement_ was the more important concern. Sadly I can't link to the absence of something.

@handrews , in reviewing the mail archives you provided and acquainting myself with the references to criticisms posed it only reinforces the decision to go in a different direction.  The criticisms of minimum or maximum, for example, I patently would not support.  In the IEC standards we're working on we have a distinct need for exactly such keywords to properly define the contextual model mappings into JSON schema syntax.  This is "real world stuff".  I don't think valuable time/energy should be spent arguing fundamental constructs such as these in the spec.

admin-cimug:
> @admin-cimug Thanks for joining and sharing your thoughts. Really appreciate you taking the time to not only answer questions, but put them to your wider group. This matters and will inform our decisions.

No problem Ben.  I particularly appreciate the level of engagement you, Henry and the spec team has demonstrated.  I much prefer to work in that type of collaborative dynamic.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: I'm using the suggestion. A first-hand account is much preferred to my second-hand account.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: Good catch. It's important that we include our decision about media type registration. However, I think it should be in the "Decision Outcome" section not the problem statement.

handrews:
Also pinging @davidgtonge who commented on #526 a while back with a perspective from the ISO.  If you don't have time to go through everything here, [this comment](https://github.com/json-schema-org/json-schema-spec/pull/1277#issuecomment-1229528807) summarizes the general direction of the most likely (but not committed) approach for us.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: I thought I had written that somewhere, but didn't see it the last time I did an update.

jdesrosiers: I also refactored a bit to address credibility and normative referencing separately.

adr/2022-08-decouple-from-ietf.md
jdesrosiers: Can you elaborate. We all have experience with what we've been doing with JSON Schema and IETF. Some of us are familiar with OpenAPI's process. Some of us have looked into what other specifications are doing. But, I wouldn't consider any of us experts beyond what we do and what IETF does. I think we have enough experience and are informed enough to take a stab at defining our own process, but we're not experts.

In case it was confusing, the "S" in SLDC stands for "specification", not "software". If it was "software", I would say we have the necessary expertise, but "specification" is a different animal.

jdesrosiers:
> This looks ready to move out of draft status to me

I'm going to wait until @Relequestual contributes his notes about W3C. Then we should be ready.

adr/2022-08-decouple-from-ietf.md
handrews: SpLDC? 🙃

handrews:
@jdesrosiers this is looking great now!  I left one or two threads "unresolved" but only because they have useful info from which I think any new visitors would benefit.  specifically [this thread with the account of the JSON working group history](https://github.com/json-schema-org/json-schema-spec/pull/1277#discussion_r965306366).  But please consider everything from me prior to this comment to be resolved and feel free to press the "resolve" buttons if needed to merge or un-draft.

Relequestual (approved):
I've added my comments regarding W3C. I'm happy with the rest of the document.
Please feel free to remove the draft status on this PR @jdesrosiers =]

awwright:
Editor hat on: There’s certain uses of the I-D process we can reevaluate, but the text here is essentially a change to the charter that's out of scope. However all of the problems here are fixable by an editing process, and with robust interoperability requirements. Let's try fixing these underlying problems first. (Editor hat off)

_What do interoperability requirements have to do with this?_

The feeling that the I-D system is unsuitable for JSON Schema, or that an RFC would mark the end of our work, comes primarily from the lack of interoperability requirements.

In the I-D process, new publications replace old ones in their entirety. This poses a problem when we’ve removed or changed features that validators have already implemented. Implementors (correctly) want to publish stable, production-ready packages that aren't going to break just because we published a new document. So in order to produce a stable package from an unstable (draft) source, they have to make some additional assumptions, out-of-spec, to provide stability.

This process often conflates the specification documents, the meta-schemas, and the media type. These are all different things, and they are "versioned" differently:

* The specification drafts use Datatracker identifiers. E.g. draft-bhutton-json-schema-01.
* The meta-schemas were serially numbered, now date numbered (to avoid confusion with the draft numbering, which over time did not line up). When discussing the proscribed behavior, since meta-schemas are non-normative, this name can also refer to the behavior that the meta-schemas are summarizing.
* The media type is unversioned. _No_ mechanism is provided to distinguish “which specification” was used to author a schema. When we publish a new document, it becomes the new authority on how to read a JSON Schema.

Again, it should be apparent that JSON Schema (as a whole) is unversioned (each document describes its own expiration policy). But since this is not suitable for stable software packages, they adopt $schema as a version identifier to detect use of obsolete (since removed) features. This is tolerable, not because the specification says to do it this way, but rather we expect that software packages to avoid breakage. (It was my intention that a different mechanism would be used for reverse compatibility, but I never wrote that in, so that’s not what ended up happening.)

We want to accommodate these implementations, so we’ve run with this. Each release is considered a new “dialect” among many. But when we treat old and new releases as equivalent, we can’t use the I-D system for its intended purpose, or publish simple revisions in response to feedback, and so the I-D process feels obsolete. 

By (effectively) publishing multiple versions in parallel, and a new concern arises: if just one is published as an RFC, wouldn’t we have to publish future changes as another RFC, subject to the whims of the working group? Wouldn’t development slow down because of that?

To avoid this possibility, the idea is to independently publish “vocabularies” or “dialects”. But this only helps software packages. Suppose you’re an API and you publish a schema, where Validator A only implements the latest JSON Schema Core & validation, and Validator B reads $schema and will ignore certain requirements in the current JSON Schema Core as necessary. In this environment, two validators can be “compliant” (at least with the tests) yet disagree on if an instance is valid/invalid, and this is bad. Or, Validator C errors on anything that’s not draft-4, and Validator D errors on anything that’s not draft2020-12, so no one schema will work for everyone. It defeats the point of a standard, which is interoperability.

Independent publishing does not solve these problems, it just covers them up. The RFC process is designed to promote interoperability, so it’s no wonder we’re having a difficult time with it. The solution to these problems is to define robust interoperability requirements that will be guaranteed well into the future.

What are the solutions?

(1) A strong reverse compatibility policy

This means we cannot remove behavior from the core specification, unless we’ve determined it can be implemented as a module, extension, or `OPTIONAL` requirement going forward, and dropping the requirement won’t be disruptive.

This is mostly my fault. In the past we've just yanked out text that we don't like, but the specification isn't supposed to describe what's good, as such, it's supposed to describe how to be interoperable.

By making backwards compatibility interoperable, implementations will no longer need to vary from the specification text in order to maintain stability. This relieves the pressure on us to maintain support for “dialects” of old and expired specifications, and allows us to reap the benefits of the I-D process.

(2) Implement changes as extensions.

We prototype changes to the specification as optional extensions. These extensions can then be tested within specialized ecosystems.

This would allow us to publish I-Ds that solicit feedback without locking the JSON Schema universe into a particular expectation.

Vocabularies are partial solution, but more fundamental changes to the core semantics still lack an extension mechanism.

(3) Robust, ubiquitous extension mechanisms

Solution parts (1) and (2) are procedural changes that assume there’s an extension mechanism. We still need to define this concrete extension mechanism.

In Email or HTTP, you want reliable messages that are delivered even if you don’t understand a part, so unrecognized headers are ignored. In JSON Schema, the functionality of JSON Schema is more analogous to a programming language: If you encounter an undefined function, you don’t want it ignored; you want an error. Yet we copy the Email behavior even though this doesn’t make sense in our particular case.

Another hint that ignoring unknown keywords is the wrong approach is that known keywords with a bad argument do error. (Email ignored malformed headers whenever possible.) Either bad arguments should be ignored, or unrecognized keywords should error. It doesn’t make sense to have it both ways.

To fix this, we need a concrete way to tell when there’s unsupported validation keywords in use, so that validators won’t return a determinate (valid/invalid) result. The most straightforward way to do this is to error on unrecognized keywords. Keywords can become "recognized" by any standard mechanism, including:

1. when an implementation implements a standard that defines a new keyword,
2. when a meta-schema enables some feature,
3. or even some out-of-band feature flag.

Finally, sometimes authors do want keywords to be ignored, especially annotation keywords, so there should be some way to flag keywords that serve no validation-related function, even if the rest of the semantics remains unknown.

This would allow new publications that are backwards compatible with older implementations, and it would allow functionality to be defined separately from the essential semantics.

By defining essential semantics separately, this would relieve the need to publish “dialects” as an RFC.

---

The PR notes

> 2. Go all-in with IETF and pursue a standards track RFC with the IETF. The approach would be to describe the essential characteristics of evaluating a JSON Schema: the keywords that everybody is guaranteed to support, and any extension mechanisms.
> 3. Join W3C and pursue a standards track with them using their process.

And

> Option 2 and 3 are still on the table if we feel it makes sense when we get to a more stable place in the future.

I think we can accomplish this in fairly short order, and in a way that expands our ability to make changes and experiment. I see two realistic paths (among others).

(1) Over the past couple months I’ve tossed around a couple of ideas for the HTTP WG and the HTTP APIs WG that have been well received, and I’m making it a personal goal to be at IETF 116 Yokohama and propose a BOF around at least one of those ideas. (BOF=“Birds of a Feather”, the most common way to form a WG or persuade one to take up work.) If that goes well, then I hope to have the JSON Schema Core document ready for a BOF at IETF 117 San Francisco. This is about 10 months away.

(2) We can form a Community Group at the W3C. This is simple to do and requires 5 people, including one Chair. These community groups can publish reports and submit them for adoption by the W3C.

(3) There’s other standards organizations that could publish a standard, e.g. ISO (recall that ISO also publishes JSON as ISO/IEC 21778:2017), but they are less open typically require some sort of corporate sponsorship. It’s possible the OpenJS Foundation could help here, but I don’t know enough.

The IETF process I think is better evolved and more open, though the W3C has more expertise in hypermedia formats.

gregsdennis (approved):
I'm happy with this.  It's in line with conversations I've had with other members of the core team.

handrews (approved):

Relequestual:
@awwright In agreeing and ratifying this ADR, the consensus of the JSON Schema Org rejects your poposals in your latest comment, explicitly 1: looking to form an IETF BOF, and 2: looking to form a W3C community group.

The ADR should stand and we should operate accordingly unless there is new information.
And, consider, it's hard to evaluate if the new process "works" till we actually try it. I feel like it is likely to take a few years and a few releases before we have enough data to evaluate any new information against.

admin-cimug:
> @awwright In agreeing and ratifying this ADR, the consensus of the JSON Schema Org rejects your poposals in your latest comment, explicitly 1: looking to form an IETF BOF, and 2: looking to form a W3C community group.
> 
> The ADR should stand and we should operate accordingly unless there is new information. And, consider, it's hard to evaluate if the new process "works" till we actually try it. I feel like it is likely to take a few years and a few releases before we have enough data to evaluate any new information against.


@Relequestual / @handrews / @awwright :  

I apologize that I've been checked out here for a few weeks.  I hosted our first return to in-person IEC Joint WG global standards meetings this past week so I was completely out of pocket.  As part of this I had an opportunity to meet in-person with some of our expert members from the National Committees of E.U. member states (we had representation from France, U.K., Hungary, Ireland, Norway, Sweden amongst others with some having to attending remotely due to corporate restrictions still in place for COVID).  

As I think I previously mentioned we have pending I.S. publications waiting on JSON schema due to it being the core dependency.  Given that we've collectively made significant investment in a reference implementation and preliminary work for use of the draft publication for use in European energy related data exchanges. this is a very high priority area with lots of visibility (also we have other standards in process that will, in turn, depend on the publication of this one.)  

During this past week's meetings we had several sessions that were held where questions came up as to when we will finally be able to push this to publication.  This has become definitely a critical topic.  I provided a status report of the discussions we've had over the past month.  

@awrights points on interoperability is really a central point that remains fundamentally important.  Even though there's a rejection of his IETF BOF and W3C related proposals.  

His point on APIs and interoperability is important:

> To avoid this possibility, the idea is to independently publish “vocabularies” or “dialects”. But this only helps software packages. Suppose you’re an API and you publish a schema, where Validator A only implements the latest JSON Schema Core & validation, and Validator B reads $schema and will ignore certain requirements in the current JSON Schema Core as necessary. In this environment, two validators can be “compliant” (at least with the tests) yet disagree on if an instance is valid/invalid, and this is bad. Or, Validator C errors on anything that’s not draft-4, and Validator D errors on anything that’s not draft2020-12, so no one schema will work for everyone. It defeats the point of a standard, which is interoperability.

When I originally presented my early standards work at the DOE's PNNL Research Lab 3 years ago what I presented got the green light as the result of what was perceived as clear intentions to undertake the IETF RFC process. Now, I understand the JSON spec teams' concerns so I support what you're trying to do as the team wrestle the process through. I just don't want to lose the key goals of interoperability along the way as @awrights is pointing out. 

Another concern I have is that we find a way forward that will prevent another three years to establish something from which other SDOs can build upon.  Finally, a key concern I have is that we're losing external confidence/credibility.  I have no doubt that the team is fully competent but its the optics that I'm trying to manage at this stage.  I have the E.U. National Committees looking for something definitive and I'm now cornered into trying to find a path forward to publication. I'm wrestling with whether to back off and publish our core standard as a Technical Specification as opposed to an I.S. (this is not ideal at all) or do I back off further and publish it as a Technical Recommendation (?)  I don't even consider that a descent option.  Anyway, I trust the team here will get my point.

Any possibility we can brainstorm together a short term path forward that minimally puts the "scaffolding" in place upon which to build the long term path you want to take?  This scaffolding being enough to manage the optics and build confidence in the path forward?

~Todd

jdesrosiers:
@admin-cimug We've been discussing all of the issues @awwright brings up here for a while and they have been well received. We're still discussing what our process will be going forward, but if the proposal on the table is adopted, all of your concerns about interoperability and compatibility will be addressed. The plan is to release a "stable" spec as soon as possible. I think releasing by the end of first quarter 2023 is achievable.

By "stable" I mean all the interoperability/compatibility concerns would be addressed. Once the stable version is released, all changes from then on would follow strict backwards and forwards compatibility rules. New features would go through a process similar to what EMCAScript uses where features have to be implemented and show real-world use and benefit before they can be considered part of the stable spec. Of course, in your situation, your spec would limit to only the stable JSON Schema spec because you lose interoperability guarantees when using unstable features.

You can view the proposal and join the conversation at, https://github.com/orgs/json-schema-org/discussions/234. And, again, thank you for your feedback. It's incredibly helpful for us.

jdesrosiers:
@awwright I'm going to keep this short because the decision has already been made and I don't want to drag this out. But, you put a lot of effort into your comment and you deserve a response.

> Editor hat on: There’s certain uses of the I-D process we can reevaluate, but the text here is essentially a change to the charter that's out of scope.

I've noticed you refer to yourself as "the Editor of JSON Schema" a few times recently. I don't know what you mean by that. We don't officially have such a role.

As far as I know we don't have a charter for JSON Schema. Maybe we should have one, but I'm quite sure we don't.

> What do interoperability requirements have to do with this?

As I've told you many times, your interoperability/compatibility concerns have been well received and have broad agreement. Addressing this issue is a top priority for the next release. Your explanation was good, but it wasn't necessary. You're preaching to the disciples at this point.

> (3) Robust, ubiquitous extension mechanisms
> I think we can accomplish this in fairly short order

I think your plan is reasonable, but it's a too vague in a lot of areas for me to have confidence. The biggest problem in my opinion is that I don't think we can create a stable extension mechanism in any reasonable amount of time, especially since you still haven't articulated what you have in mind. Currently our extension mechanism is the vocabulary system. We put it out knowing that it was an experiment and would require some real-world use to illicit feedback on what problems there were and how it can be improved. It's very much a work in progress. Whether your idea of an extension mechanism is based on the vocabulary system or something new entirely, I'm convinced it will take real use to vet and get right. That's going to take years. Your plan for an RFC core with extensions is dependent on having this extension mechanism in place. That means being a "draft" for quite a while longer, which is not something we're willing to do.

I'm still interested in what you have in mind regarding extension mechanisms. You've sold me on just about everything you've proposed lately other than IETF, so I expect I'd like what you have to say. But, let's create a new issue for that conversation and not in the context of IETF since this decision has already been made.

handrews:
@admin-cimug are the "optics" concerns not addressed by what we previously discussed regarding an IETF media type registration plus our own publication in the context of being an OpenJS Foundation / Linux Foundation project?  Particularly if we move out of OpenJS "incubation" status and into the next one (which I think is "impact" but I'm not sure)?

admin-cimug:
> @admin-cimug We've been discussing all of the issues @awwright brings up here for a while and they have been well received. We're still discussing what our process will be going forward, but if the proposal on the table is adopted, all of your concerns about interoperability and compatibility will be addressed. The plan is to release a "stable" spec as soon as possible. I think releasing by the end of first quarter 2023 is achievable.
> 
> By "stable" I mean all the interoperability/compatibility concerns would be addressed. Once the stable version is released, all changes from then on would follow strict backwards and forwards compatibility rules. New features would go through a process similar to what EMCAScript uses where features have to be implemented and show real-world use and benefit before they can be considered part of the stable spec. Of course, in your situation, your spec would limit to only the stable JSON Schema spec because you lose interoperability guarantees when using unstable features.
> 
> You can view the proposal and join the conversation at, https://github.com/orgs/json-schema-org/discussions/234. And, again, thank you for your feedback. It's incredibly helpful for us.

@jdesrosiers / @handrews  :  First, I wish I was more up to speed on all past discussions so I'm genuinely grateful for you and the team's patience as you graciously allow me to voice my questions and comments.  Thank you for confirming/clarifying the interoperability related concerns and that the team has discussed and has plans to address along the way.  It's a core tenet for the IEC standards we publish so was not totally clear from @awwright original post if this was going to be fully addressed (and thus my last response).  Now, I just reviewed the discussions link you sent.  This is **great**.  Had I caught it earlier I'm sure it would have helped.  

> @admin-cimug are the "optics" concerns not addressed by what we previously discussed regarding an IETF media type registration plus our own publication in the context of being an OpenJS Foundation / Linux Foundation project? Particularly if we move out of OpenJS "incubation" status and into the next one (which I think is "impact" but I'm not sure)?

@handrews I think we may be close but there's a final level of assurance I think I'm fishing for but it may be just be the need for a bit more discussion here.  First, after we talked and in lieu of my meetings last weeks the key question that came out of things is timing.  I did check out [OpenJS Foundation](https://openjsf.org/) as we discussed and see JSON Schema listed as an Incubation Project.  The question now is as to the timeline for completing the on-boarding checklist to join the foundation?  Do you see any obstacles or concerns that may present new roadblocks that might derail us as the IETF RFC processes did?  If I knew definitively that JSON schema was close to on-boarding I'd definitely have more peace of mind. This help address the "optics".  A charter would be nice to have but I think official acceptance with a charter to come later like we discussed would really help.  

I apologize that I'm jittery on this.  There's a two fold reason why.  The first is the original three year window.  THe second is that the most efficient/ideal standardization window I have is probably 12 months from the time I officially submit what we term an **IEC NWIP** (i.e. essentially a "work proposal" for a new standard.)  Our plan (i.e. myself and my European counterparts) has been to submit a nearly complete late stage draft of an I.S. (thereby shortening the review cycles).  Such a move is what we would term a "fast track" process.  This approach bought me time the past couple of years as we waited for the originally suggest IETF RFC process to progress for JSON schema.  That (for very understandable reasons that you've explained) stalled out.  Now at last weeks meeting the urgency has escalated. Hopefully that helps explain what remains. So I see a couple of open items for us:

1. Is it possible to gauge with any accuracy the timing of the transition from an Incubation Project?
2. I like what I saw in the "New Spec Development Lifecycle Model" proposal.  Open questions we have: 
- In trying to do a cursory overview of the discussion it seems a fair bit of consensus has been reached.  Can you confirm? Are there any major internal difference in opinions that would sidetrack the proposal?  
- Has there been any talk amongst the team on when this would start?  
- Would there be a stable Jan-2003 release that might fit into what is described?  (this question may be moot based on the previous answers)

Now, I don't have the bandwidth to catch myself up on all related and relevant discussion so if any of my above questions should be documented and obvious elsewhere than thank in advance for bearing with me.

Relequestual:
>The question now is as to the timeline for completing the on-boarding checklist to join the foundation? Do you see any obstacles or concerns that may present new roadblocks that might derail us as the IETF RFC processes did? If I knew definitively that JSON schema was close to on-boarding I'd definitely have more peace of mind. This help address the "optics". A charter would be nice to have but I think official acceptance with a charter to come later like we discussed would really help. - @admin-cimug 

Levelling up our status in the OpenJS Foundation is tracked here: https://github.com/json-schema-org/community/issues/129

I can't foresee any roadblocks. It has just been a matter of time and priority.
I had hoped to get back to working through the list again this year, but I've been franitcally trying to bring on new hires to take a load of work off my plate so I can focus. (Having said that, others are welcome to drive any of the outstanding items on the list).

>Is it possible to gauge with any accuracy the timing of the transition from an Incubation Project?

In truth, I can't say. What I will say is, even though we're "only" an incubation project, we are still considered proper members of the OpenJS Foundation. It's not a "half measure", but a signal of commitment to see it over the line. I made a personal promise to do everything in my power to see JSON Schema go beyond the incubation status.

> In trying to do a cursory overview of your discussion on that proposal it seems a fair bit of consensus has been reached. Can you confirm? Are there any major internal difference in opinions that would sidetrack the proposal?

We try to discuss as much as possible in public. I don't think we've discussed much about this proposal which hasn't later been made public, either there or during one of our open weekly meetings.

Sorry if I've not addressed every part of your comment, but I hope that's still useful. Thanks again for sharing and allowing us to hear your important feedback and comments... we want to see this more from yourself and others! 🙌

handrews:
@admin-cimug 

> In trying to do a cursory overview of your discussion on that proposal it seems a fair bit of consensus has been reached. Can you confirm? Are there any major internal difference in opinions that would sidetrack the proposal?

I'm probably the person with the most objections/divergence regarding the SDLC proposal.  However, I am (and have been) actively working on getting to "yes" with it, which is why there's not a competing proposal.  I also do not intend to drag it out: as I have told the rest of the team, if there is a strong majority sentiment to accept it that develops before I can articulate any sort of requested changes or alternatives, I will go with that majority sentiment.

So while I cannot give you an exact timeline or assurance, as the person who is the least on board of those who have commented to date, I can guarantee that it is moving towards a resolution.  My concerns at this point are mostly around the level at which we define stability and compatibility:  whether it is keyword-centric or feature-centric.  Which I realize is very vague but I am working on un-vague-ing it 😅 

> Has there been any talk amongst the team on when this would start?
> 
> Would there be a stable Jan-2023 release that might fit into what is described? (this question may be moot based on the previous answers)

It will start for the next release.  We would like to put one out in January 2023, although given the level of stability/compatibility guarantees needed to do this, it might be more like March 2023.  We would not wait until 2024 or shift the stability promotion month, we'd just have an abbreviated cycle and do the next round of promotions in January 2024 and continue from there.

Given that the team is partially volunteer (if anyone wants to sponsor my or other volunteer work with JSON Schema pls reach out to the org), and that the full time staff have a variety of responsibilities beyond just spec-writing (including OpenJS onboarding, community engagement, test suite development, etc.), we can't rush the next release but so much.  But January 2023 would be our ideal, and we'll get as close to it as we can.

admin-cimug:
> So while I cannot give you an exact timeline or assurance, as the person who is the least on board of those who have commented to date, I can guarantee that it is moving towards a resolution. My concerns at this point are mostly around the level at which we define stability and compatibility: whether it is keyword-centric or feature-centric. Which I realize is very vague but I am working on un-vague-ing it 😅

Thanks Henry for your response.  You got a grin out of me :grin: ... LOL.  You summary is just what I needed.  That Jan-2023 to Mar-2023 timeframe should actually work well.  If we accomplish what you've summarized it should clear away the concerns and I don't see any issues then having JSON schema specification included in our normative references and expressing it's association with the Linux Foundation (I believe this removes the whole "optics" thing).  Based on this out goal can be to update our draft I.S. during one of the review cycles align it with that initial stable release for Jan or Mar.

@Relequestual :  Ben, :slightly_smiling_face: thanks for pointing me to: https://github.com/json-schema-org/community/issues/129 that will be helpful for tracking puposes.

handrews:
@admin-cimug great!  Just to be clear, I cannot (on behalf of the org) _commit_ to the Jan-March 2023 timeframe. But we are targeting January so I think that range is realistic.

If we cannot hit that range, I would (personally, not an org commitment although we could probably discuss such a thing) expect that we would publish some sort of guidance as to what is going on.  A delay would presumably mean that we found an important compatibility problem with our process, and that would be the kind of thing we'd want to communicate.  Presumably, it would be something that anyone waiting on a stable spec would want to see resolved anyway,

awwright:
@jdesrosiers I apologize for not having a more concrete proposal sooner, there's many moving parts and I'm hesitant to propose changes that involve the vocabulary system, because that's so highly utilized. I hope some of my regular updates show that there's interesting developments in the space, though. I'm trying to figure out how I can devote more time to have a workable solution ready by March, when I would like to take other HTTP-related work to IETF 116.

Regarding "editor," that's how the document itself lists my title. This is ever since I asked the previous editors about publishing an update. As the scope has grown, I've turned over various aspects of editing and publishing to different people over time, particularly so that different people oversee different documents, which now list multiple editors and authors; but I still edit Core and oversee interoperability and Internet related issues. My intention was always that our organization would serve as a forum for implementors, and to promote the standard, but that nobody owns the specification, we just steward it until it can be adopted by a venue (or multiple venues) with experience in Internet standards. We don't have a formal "charter" as such but the I-D process is a written down process, and deciding to change the process or the venue is unusual.

handrews:
@awwright creating and publishing a formal charter is part of our OpenJS Foundation todo list: json-schema-org/community#29

Given that we are no longer simply continuing the approach that we inherited from the previous editorial team around 2016 (or whenever you first picked it up), I think that increases the importance and urgency of sorting out the charter.

mehdij55991 (approved):

PR #1280, opened 2022-08-21
Make targets for a virtualenv

handrews:
It's common to make a virtualenv inside of your cloned repo
to ensure that you have the right enviornment for that repository,
and .venv is a common convention for that virtualenv.

This adds targets to create and clean a .venv virtual env using
the checked-in requirements.txt file.  It also adds the .venv
to the .gitignore, and updates the README.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 XML2RFC ?= xml2rfc
+VENV ?= .venv
 
 OUT = \
 	jsonschema-core.html jsonschema-core.txt \
 	jsonschema-validation.html jsonschema-validation.txt \
 	relative-json-pointer.html relative-json-pointer.txt
-
 
 all: $(OUT)
 
@@ -24,6 +24,16 @@ json-schema.tar.gz: $(OUT)
 	(cd json-schema && make)
 	tar -czf json-schema.tar.gz --exclude '.*' json-schema
 	rm -rf json-schema
+
+venv: $(VENV)
+
+$(VENV):
+	python -m venv .venv
+	. .venv/bin/activate && python -m pip install --upgrade pip setuptools wheel
+	. .venv/bin/activate && python -m pip install -r requirements.txt
+
+venv-clean:
+	rm -rf .venv
 
 clean:
 	rm -f $(OUT) json-schema.tar.gz
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ OUT = \
 	jsonschema-validation.html jsonschema-validation.txt \
 	relative-json-pointer.html relative-json-pointer.txt
 
-all: $(OUT)
+all: $(VENV) $(OUT)
 
 %.txt: %.xml
 	$(XML2RFC) --text $< -o $@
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -25,6 +25,16 @@ json-schema.tar.gz: $(OUT)
 	tar -czf json-schema.tar.gz --exclude '.*' json-schema
 	rm -rf json-schema
 
+venv: $(VENV)
+
+$(VENV):
+	python -m venv .venv
+	. .venv/bin/activate && python -m pip install --upgrade pip setuptools wheel
```
webknjaz, line 32: What are you trying to achieve with this? Upgrading setuptools and wheel shouldn't be necessary.

Makefile
```diff
@@ -25,6 +25,16 @@ json-schema.tar.gz: $(OUT)
...
+	. .venv/bin/activate && python -m pip install -r requirements.txt
```
webknjaz, line 33: No need to use the activation script in a non-interactive mode. You can just run things from virtualenvs directly.
```suggestion
	.venv/bin/python -m pip install -r requirements.txt
```

jviotti: Related to my previous comment: `$@/bin/python -m pip install -r requirements.txt` is even better

README.md
```diff
@@ -35,7 +35,15 @@ Labels are assigned based on [Sensible Github Labels](https://github.com/Releque
     * links.json - JSON Hyper-Schema's Link Description Object meta-schema
     * hyper-schema-output.json - The recommended output format for JSON Hyper-Schema links
 
-Install "xml2rfc" using "pip" (https://pypi.org/project/xml2rfc/) and type "make" at a shell to build the .txt and .html spec files:
+The Makefile can create the necessary Python virtual environment for you:
+
+```sh
+make venv
```
jviotti, line 41: If you make `requirements.txt` a dependency of `.venv`, you can then make `.venv` a dependency of the default target. If so, users would only need to run `make` instead of `make venv` followed by `make`.

Given that and @webknjaz suggestion about the activation script, this could become just `make` as it was originally documented

Makefile
handrews: @webknjaz copy-paste error from some notes for a project where I needed to uses these, thanks for catching!

handrews:
Sorry for the delay on this, and thanks for the valuable input @webknjaz and @jviotti .  My understanding of `make` is extremely basic.

I made a "clean" target that gets rid of the venv as well, and a "spec-clean" target that does not.  Does that seem correct, or should "clean" just remove the built specs, and then we have a "clean-all" or "clean-venv" or something that either removes everything or just removes the venv?

webknjaz:
FWIW many Python projects just use tox or nox and not Makefile which is not really native to the Python ecosystem.

Also, you may want to make use of `git clean`, it's a powerful tool.

handrews:
@webknjaz for now I just want something that works at all.  I don't have time to delve into additional tools, although if someone wants to pick that up they're more than welcome to.

Makefile
```diff
@@ -25,7 +25,15 @@ json-schema.tar.gz: $(OUT)
 	tar -czf json-schema.tar.gz --exclude '.*' json-schema
 	rm -rf json-schema
 
-clean:
+$(VENV): requirements.txt
+	python -m venv $@
+	$@/bin/python -m pip install --upgrade pip
+	$@/bin/python -m pip install -r requirements.txt
```
jviotti, line 31: You probably want to refer to `requirements.txt` as `$<`

README.md
jviotti: @handrews I think this comment still applies. The `venv` is now part of the `all` target, so users can just run `make`

jviotti (approved):

gregsdennis (approved):

PR #1285, opened 2022-09-08
Update output schema to describe new structure

### LINKED ISSUE #949: Some inconsistencies in output schema
https://github.com/json-schema-org/json-schema-spec/blob/master/output/schema.json
https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.10

1. "basic", "detailed" and "verbose" are defined exactly the same, which means, if output is valid for one of them, it is valid for all of them. This means that, unless the output format is flag, the "oneOf" validation would always fail.

2. 
-- Output schema requires "errors" key in case "valid" is false;
-- "errors" is recursively defined as array of "outputUnit"
-- "valid" is required
This means the only way to terminate recursion in case of "valid": false is to include "errors": [], which is misleading.
On the other hand, the output examples sometimes contain "error" key and not "errors". Although not strictly necessary, it'd be good to have "error" defined in the schema.

gregsdennis:
Updates the output schema file to match the new structure as described by the specification.  Follow-up to #1249.

I've tested this against the example output in the spec.

Resolves #949

### DIFF
--- a/output/schema.json
+++ b/output/schema.json
@@ -6,18 +6,17 @@
   "anyOf": [
     { "$ref": "#/$defs/flag" },
     { "$ref": "#/$defs/basic" },
-    { "$ref": "#/$defs/detailed" },
-    { "$ref": "#/$defs/verbose" }
+    { "$ref": "#/$defs/hierarchical" }
   ],
   "$defs": {
     "outputUnit":{
       "properties": {
         "valid": { "type": "boolean" },
-        "keywordLocation": {
+        "evaluationPath": {
           "type": "string",
           "format": "json-pointer"
         },
-        "absoluteKeywordLocation": {
+        "schemaLocation": {
           "type": "string",
           "format": "uri"
         },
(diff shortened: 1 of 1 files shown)


output/schema.json
```diff
@@ -6,75 +6,56 @@
...
         "valid": { "type": "boolean" },
-        "keywordLocation": {
+        "evaluationPath": {
           "type": "string",
           "format": "json-pointer"
         },
-        "absoluteKeywordLocation": {
+        "schemaLocation": {
           "type": "string",
           "format": "uri"
```
karenetheridge, line 21: Shouldn't this be iri now?

output/schema.json
```diff
@@ -89,8 +70,15 @@
...
-    "detailed": { "$ref": "#/$defs/outputUnit" },
-    "verbose": { "$ref": "#/$defs/outputUnit" }
+    "basic": {
+      "properties": {
+        "valid": { "type": "boolean" },
+        "nested": {
+          "$ref": "#/$defs/outputUnitArray"
+        }
+      },
+      "required": [ "valid", "nested" ]
```
karenetheridge, line 80: Why does the 'basic' format contain 'nested'? My understanding was that basic is still a flat structure.

output/schema.json
karenetheridge: ah I see you're already on it, re the other PR about uri->iri changes.

output/schema.json
gregsdennis: Basic is an object with only `valid` and `nested ` at the root.  The rest is normal output units in the `nested` array.

```json
{
  "valid": true,
  "nested": [ <flat list> ]
}
```

This provides the overall validation result at the root and a flat list.  It's the same as 2020-12, except that the location properties aren't required at the root (because they're always just `/` and not necessary).

output/schema.json
gregsdennis: I'm going to push this forward as I can't wait on your PR any longer.  It needs to be rebased anyway.

handrews (approved):

karenetheridge:
Shouldn't this have gone into draft-next, not main?

gregsdennis:
@karenetheridge draft-next was merged some time ago.

PR #1286, opened 2022-09-09
xref format changed again - fix broken build

handrews:
Now there's a dot between W3C and the rest, and the underscores are back to hyphens.  Apparently.

Previously changed in PR #1268


handrews:
I'm merging this b/c broken build.  It definitely won't make it worse.

PR #1289, opened 2022-09-18
Clarify prefixItems wording about array length

### LINKED ISSUE #1174: prefixItems does not explicitly mention instances arrays shorter than the keyword value length
The [`prefixItems` keyword section](http://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.1.1) says this:

> Validation succeeds if each element of the instance validates against the schema at the same position, if any. This keyword does not constrain the length of the array. If the array is longer than this keyword's value, this keyword validates only the prefix of matching length.

Note that it explicitly mentions if the instance array is _longer_ but nothing about when it's shorter.  It _does_ say that no length constraints are imposed, but it feels odd that it would mention one without the other.

We do have a [test for this](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft2020-12/prefixItems.json#L22) as well.

handrews:
Fixes #1174

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2349,7 +2349,7 @@
                         </t>
                         <t>
                             Validation succeeds if each element of the instance validates
-                            against the schema at the same position, if any.  This keyword
+                            against the subschema at the same position, if any.  This keyword
                             does not constrain the length of the array.  Only array positions
                             present in both the keyword's value and the instance value are
                             affected by this keyword.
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -2350,9 +2350,9 @@
                         <t>
                             Validation succeeds if each element of the instance validates
                             against the schema at the same position, if any.  This keyword
```
gregsdennis, line 2352: While we're here, should we update this to say "subschema" instead of "schema?"

gregsdennis (approved):

jsonschema-core.xml
handrews: good catch, done!

jdesrosiers (approved):

PR #1290, opened 2022-09-18
Missed a "schema document" -> "schema resource"

### LINKED ISSUE #1129: clarification needed: bundled metaschemas
in https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.1.2:

> The "$vocabulary" keyword SHOULD be used in the root schema of any schema document intended for use as a meta-schema.  It MUST NOT appear in subschemas.

I'm not sure if the intent here is that metaschemas cannot be bundled which would (assuming that there are "$schema" and "$id" keywords alongside), result in "$vocabulary" appearing in the root of its particular resource but not in the root of the actual document itself -- or if this language preceded the formalization of bundling.

I think this usecase should be explicitly called out as either allowed or disallowed.

example:

```
{
  "$defs": {
    "metaschema": {
      "$id": "https://localhost:1234/mymetaschema",
      "$schema": "https://json-schema.org/draft/2020-12/schema",
      "$vocabulary": {
        ...
      }
    }
  },
  "$id": "https://localhost:1234/myschema",
  "$schema": "https://localhost:1234/mymetaschema",
  ...,
}
```

handrews:
This was just an oversight regarding where $vocabualry is legal. It should have always been "schema resource root."

Fixes #1129.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1273,10 +1273,10 @@
                     </t>
                     <t>
                         The "$vocabulary" keyword SHOULD be used in the root schema of any schema
-                        document intended for use as a meta-schema.  It MUST NOT appear in subschemas.
+                        resource intended for use as a meta-schema.  It MUST NOT appear in subschemas.
                     </t>
                     <t>
-                        The "$vocabulary" keyword MUST be ignored in schema documents that
+                        The "$vocabulary" keyword MUST be ignored in schema resources that
                         are not being processed as a meta-schema.  This allows validating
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

handrews:
Updated after realizing there was another occurrence in the next paragraph.

jdesrosiers (approved):

handrews:
I'm going to go ahead and merge this since it has 2 approvals and is about as non-controversial as it gets, so I can avoid merge conflicts in other PRs.

PR #1293, opened 2022-09-18
Explain why plain name fragments are as they are

### LINKED ISSUE #901: Suggestion to add explanation of logic behind character set choice for $anchor
This issue suggests that some sentences or a paragraph be added to the $anchor section in the core spec describing the logic behind the character set choice. It doesn't match the "fragment" production from URI, and it's not clear why because it functions similarly.

In my experience, it's often the case that explaining these things reduces the temptation to be more lenient with the spec.

See: [8.2.3. Defining location-independent identifiers with "$anchor"](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2.3)

handrews:
Because XML is/was the most likely need for cross-media-type fragment identifiers.

Fixes #901

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -534,8 +534,6 @@
                 "application/schema+json" document are specified
                 in the <xref target="anchor">"$anchor" keyword</xref> section.
             </t>
-            <t>
-            </t>
         </section>
 
         <section title="General Considerations">
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1458,8 +1458,11 @@
                         If present, the value of this keyword MUST be a string and MUST start with
                         a letter ([A-Za-z]) or underscore ("_"), followed by any number of letters,
                         digits ([0-9]), hyphens ("-"), underscores ("_"), and periods (".").
-                        This matches the US-ASCII part of XML's
-                        <xref target="xml-names">NCName production</xref>.
+                        Due to the once-common practice of providing resource representations in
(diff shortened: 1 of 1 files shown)


jdesrosiers:
Right now, the spec is quite bloated with things that don't necessarily belong in a specification. Mostly I think that's because we didn't really have anywhere else to put things, but that's not the case anymore. This strikes me as maybe something that should be a blog post or maybe an ADR.

handrews:
@jdesrosiers I could see it as an ADR.  I don't think it's actually interesting enough for a blog post.  I've been thinking of writing ADRs for some decisions that people still ask about just to document them.  Although sometimes trying to make it fit the ADR format when I can't even remember what the alternatives were is a bit annoying.

handrews:
@jdesrosiers I considered making this an ADR, but it doesn't fit the format well because we kind of inherited this and assembled our own justifications for keeping it.  There wasn't really a clear decision point.

I also noticed that it's weird that the Fragment Identifiers section doesn't actually define the plain name syntax, which instead appears in the section of `$anchor` and `$dynamicAnchor`, where its source is less clear.

So I've removed the editorial explanation about XML and APIs, moved the syntax up to the Fragment Identifiers section, and added one extra line so that the reason for using NCName is clear.  Given the bewildering constellation of XML specifications, the additional link is, I think, useful as an informative reference.  But I'm open to counter-arguments.

jdesrosiers (approved):

PR #1295, opened 2022-09-19
Normative language for "$vocabulary"

### LINKED ISSUE #1294: Clarify semantics of omitting vocabularies from "$vocabulary"
Per the [§8.1.2](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-the-vocabulary-keyword) of the core spec:

> The "$vocabulary" keyword is used in meta-schemas to identify the vocabularies available for use in schemas described by that meta-schema. It is also used to indicate whether each vocabulary is required or optional, in the sense that an implementation MUST understand the required vocabularies in order to successfully process the schema. Together, this information forms a dialect. Any vocabulary that is understood by the implementation MUST be processed in a manner consistent with the semantic definitions contained within the vocabulary.

The intent of "identify the vocabularies available for use" is that schemas not identified as available for use are not available for use.  We have do have a [test for this scenario](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/9cb7da92f9647afad1abf39edddd1548d90214ff/tests/draft2020-12/vocabulary.json#L23-L27).

Supporting points for this view include:

* There wouldn't be a need for optional vocabularies if omitting a vocabulary still allowed you to use it
* There wouldn't be a need for the introduction of [§8](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-the-json-schema-core-vocabu) to discuss handling of both omitting the core vocabulary and including it as optional, with the recommendation to raise an error in such cases, if omitting it had no effect.

An exhaustive discussion of further historical details, complete with citations from old issue and PR discussions, can be found in json-schema-org/JSON-Schema-Test-Suite#574

We just need to add normative language around the omission aspect.  A related clarification (also discussed in the linked test suite issue) is that non-vocabulary keywords of the same name as vocabulary keywords are not considered the same keyword, any more than identically named keywords in different vocabularies are.  They can be distinguished by whether inclusion in or omission from `$vocabulary` has an effect on them.

handrews:
This reworks the language around "$vocabulary" to capture the original intent in normative language.  In particular, it explicitly addresses vocabularies omitted from "$vocabulary".  As noted in #1294, this is based on extensive recent discussion and does not add any actual new requirements.  It does use RFC 2119 language in places where it had not previously been used, though.

The new text emphasizes the concepts of required or optional vocabularies more than the true and false boolean values that implement them, and adds some explanation of the utility of optional vocabularies.

It also adds a clarification around identically-named vocabulary and non-vocabulary keywords, including that the standard keywords MUST be considered vocabulary keywords and subject to "$vocabulary" control even if the implementation does not support custom vocabularies at all.

Fixes #1294 .  Also includes the two-word #1290 change as it just read strangely without it given the reworked section headers.  Either way it will be a merge conflict anyway so I'll sort that out as needed.


jsonschema-core.xml
```diff
@@ -1256,32 +1252,79 @@
...
+                        </t>
+                        <t>
+                            Implementations that do not support a vocabulary required by a schema
+                            MUST refuse to process that schema.
+                        </t>
+                        <t>
+                            Implementations that do not support a vocabulary that is optionally used
+                            by a schema SHOULD proceed with processing the schema.  The keywords will
+                            be considered to be unrecognized keywords as addressed by section
+                            <xref target="unrecognized" format="counter"></xref>.  Note that since
```
gregsdennis, line 1288: This is fine for now, but may change with the outcome of [Proposal: Disallow non-vocabulary keywords](https://github.com/orgs/json-schema-org/discussions/241).

jsonschema-core.xml
```diff
@@ -1256,32 +1252,79 @@
...
+                            the same behavior when used optionally whether the implementation
+                            supports them or not.  This allows annotation-only vocabularies to
+                            be supported without custom code, even in implementations that do
+                            not support providing custom code for extension vocabularies.
+                        </t>
+                    </section>
+                    <section title="Vocabularies are schema resource-scoped">
+                        <t>
+                            The "$vocabulary" keyword SHOULD be used in the root schema of any schema
+                            resource intended for use as a meta-schema.  It MUST NOT appear in subschemas.
```
gregsdennis, line 1300:
```suggestion
                            resource intended for use as a meta-schema.  It MUST NOT appear in subschemas
                            that are not themselves schema resource roots.
```

jsonschema-core.xml
handrews: yes, definitely, and I'll be happy to further update it.  I did not want to assume the outcome of a complex topic.

gregsdennis: No worries.  Just wanted to have a link and state that this is still open discussion.

handrews:
The force-push just now was a rebase with no changes.

Issue #1288, opened 2022-09-14
Clarify the handling of "contentSchema"


handrews:
`contentSchema` is an annotation.  It is never applied as part of normal evaluation (although see #1287), and shouldn't be processed at schema load time like a subschema of an applicator or location (`$defs`) would be.  In other words, you shouldn't be able to target it with a `$ref` even though it looks like a schema.  It's just data.

1. We should clarify that bit about it being data, not a schema you can `$ref`
2. We should note that since the schema location of the keyword is the schema-as-data's retrieval URI, if you fail to set `$id` then your schema-from-data will have that retrieval URI as its base URI, which means after the fragment is chopped off it is likely to have URI collisions with the parent schema.  If you then try to feed it right back into the implementation to use in the same session, this will probably blow up in your face.  We probably don't need to forbid it, but we should be very clear about why it is a Bad Idea™.
3. If you don't include `$schema`, you might not have any idea how to process it.  In a large ecosystem, this might come from a different schema resource than you started from.  In a sufficiently complex system, the code asking for the validation (possibly over a network API) may not have direct access to that schema resource in order to check its `$schema`.  We could either warn about this or we could include the dialect URI in the annotation data.  Note that we need the dialect and not the specific vocabulary of the `contentSchema` keyword, because even if it matches the version of the core vocabulary, that doesn't tell us what other vocabularies are assumed.  Presumably if we added dialect info to the annotation info, we'd also need to add it to the output format somewhere (which probably deserves its own issue).

jdesrosiers:
> `contentSchema` is an annotation. It is never applied as part of normal evaluation (although see https://github.com/json-schema-org/json-schema-spec/issues/1287), and shouldn't be processed at schema load time like a subschema of an applicator or location (`$defs`) would be.

Why wouldn't you treat it like a normal schema? I don't see why being an annotation or not getting evaluated means it can't be treated as a schema. The problems you bring up only apply if you don't treat it as a schema. It seems much simpler to just treat it like a schema and everything just works like normal.

handrews:
@jdesrosiers `contentSchema` is specifically ***not*** an applicator or schema location keyword (meaning a keyword like `$defs` that primarily serves as a reference target).  Its value is returned as an annotation, at which point is has been removed from the context that allows "treat it like a normal schema" to work.

jdesrosiers:
> `contentSchema` is specifically **_not_** an applicator or schema location keyword

[...]

PR #1296, opened 2022-09-19
Disallow even optional "content*" processing

### LINKED ISSUE #1287: Proposal: remove optional automatic processing of "content*" keywords
In [§8.2 Implementation Requirements](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#section-8.2) (for the `content*` keywords), we have:

> Implementations MAY offer the ability to decode, parse, and/or validate the string contents automatically. However, it MUST NOT perform these operations by default, and MUST provide the validation result of each string-encoded document separately from the enclosing document. This process SHOULD be equivalent to fully evaluating the instance against the original schema, followed by using the annotations to decode, parse, and/or validate each string-encoded document. _&lt;CREF&gt;For now, the exact mechanism of performing and returning parsed data and/or validation results from such an automatic decoding, parsing, and validating feature is left unspecified. Should such a feature prove popular, it may be specified more thoroughly in a future draft. &lt;/CREF&gt;_
>
> See also the [Security Considerations](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#security) ([Section 10](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#security)) sections for possible vulnerabilities introduced by automatically processing the instance string according to these keywords.

This is a lot of potential complexity and security risk for a minimal increase in convenience:

* it creates an entirely new code path, where additional activity takes place between the main validation and returning the result
* it requires a different approach to results in order to separately indicate the results of the (potentially many) sets of `content*` keywords
* those additional results include not just any result from `contentSchema`, but the output of decoding and/or parsing based on `contentEncoding` and/or `contentMediaType`
* it involves a mandatory runtime configuration option, which is extra work and reduces the predictability of evaluation behavior compared to schema author intent
* it is a giant security hole (JavaScript, anyone?)
* it creates another extension point in the JSON Schema architecture in order to handle more media types (and maybe encodings- they're not said to be extensible but people will ignore that), but this one does not add anything to the capabilities of JSON Schema at all (other than needless complexity)

If we remove this, it just means that people need to look at the annotations, hand the instance value off to the right decoder and/or parser, and (if `contentSchema` is present), call the JSON Schema implementation again, providing the schema and the decoded/parsed instance value.

All of the information needed to do that is present in the annotations, and there's nothing JSON Schema-specific about it (calling back into the implementation doesn't count- since this came from a call into a JSON Schema implementation in the first place, clearly the application can call it again just fine).

Let's simplify our architecture by jettisoning this potentially complex and insecure, and entirely unnecessary, feature.  People who want to automatically do this can trivially implement it as a separate library (and we can note that in the spec, maybe in a CREF).

handrews:
This feature added substantial complexity and security exposure for essentially no benefit. The functionality could be trivially implemented as a library on top of a JSON Schema implementation that supports annotation collection.

Fixes #1287.  Intentionally avoids topics still under discussion in #1288.

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -950,7 +950,7 @@
 
                 <t>
                     If the instance value is a string, this property defines that the string
-                    SHOULD be interpreted as encoded binary data and, and applications wishing
+                    SHOULD be interpreted as encoded binary data and applications wishing
                     to decode it SHOULD do so using the encoding named by this property.
                 </t>
 
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -936,33 +936,13 @@
                 <t>
                     Due to security and performance concerns, as well as the open-ended nature of
                     possible content types, implementations MUST NOT automatically decode, parse,
-                    and/or validate the string contents by default.  This additionally supports
-                    the use case of embedded documents intended for processing by a different
-                    consumer than that which processed the containing document.
+                    and/or validate the string contents.  Applications are expected to use these
+                    annotations to invoke the appropriate libraries (including JSON Schema for
+                    any further schema-based validation) separately.
                 </t>
                 <t>
                     All keywords in this section apply only to strings, and have no
                     effect on other data types.
-                </t>
-                <t>
-                    Implementations MAY offer the ability to decode, parse, and/or validate
-                    the string contents automatically.  However, it MUST NOT perform these
-                    operations by default, and MUST provide the validation result of each
-                    string-encoded document separately from the enclosing document.  This
-                    process SHOULD be equivalent to fully evaluating the instance against
-                    the original schema, followed by using the annotations to decode, parse,
-                    and/or validate each string-encoded document.
-                    <cref>
-                        For now, the exact mechanism of performing and returning parsed
-                        data and/or validation results from such an automatic decoding, parsing,
-                        and validating feature is left unspecified.  Should such a feature
-                        prove popular, it may be specified more thoroughly in a future draft.
-                    </cref>
-                </t>
-                <t>
-                    See also the <xref target="security">Security Considerations</xref>
-                    sections for possible vulnerabilities introduced by automatically
-                    processing the instance string according to these keywords.
                 </t>
             </section>
 
@@ -970,8 +950,8 @@
 
                 <t>
                     If the instance value is a string, this property defines that the string
-                    SHOULD be interpreted as encoded binary data and decoded using the encoding
-                    named by this property.
+                    SHOULD be interpreted as encoded binary data and, and applications wishing
+                    to decode it SHOULD do so using the encoding named by this property.
                 </t>
 
                 <t>
@@ -1023,7 +1003,8 @@
                 </t>
                 <t>
                     This keyword MAY be used with any media type that can be mapped into
-                    JSON Schema's data model.
+                    JSON Schema's data model.  Specifying such mappings is outside of the
+                    scope of this specification.
                 </t>
                 <t>
                     The value of this property MUST be a valid JSON schema. It SHOULD be ignored if
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -937,8 +937,7 @@
                     Due to security and performance concerns, as well as the open-ended nature of
                     possible content types, implementations MUST NOT automatically decode, parse,
                     and/or validate the string contents.  Applications are expected to use these
-                    annotations to invoke the appropriate libraries (including JSON Schema for
-                    any further schema-based validation) separately.
+                    annotations to invoke the appropriate libraries separately.
                 </t>
                 <t>
                     All keywords in this section apply only to strings, and have no


jsonschema-validation.xml
```diff
@@ -936,42 +936,22 @@
...
-                </t>
             </section>
 
             <section title="contentEncoding">
 
                 <t>
                     If the instance value is a string, this property defines that the string
-                    SHOULD be interpreted as encoded binary data and decoded using the encoding
-                    named by this property.
+                    SHOULD be interpreted as encoded binary data and, and applications wishing
```
jdesrosiers, line 953:
```suggestion
                    SHOULD be interpreted as encoded binary data and applications wishing
```

jsonschema-validation.xml
```diff
@@ -936,42 +936,22 @@
                 <t>
                     Due to security and performance concerns, as well as the open-ended nature of
                     possible content types, implementations MUST NOT automatically decode, parse,
-                    and/or validate the string contents by default.  This additionally supports
-                    the use case of embedded documents intended for processing by a different
-                    consumer than that which processed the containing document.
+                    and/or validate the string contents.  Applications are expected to use these
+                    annotations to invoke the appropriate libraries (including JSON Schema for
+                    any further schema-based validation) separately.
```
jdesrosiers, line 941: I don't understand what this parenthetical is saying.

handrews: It's referring to treating the JSON Schema implementation as the "application" when you get a `contentSchema` and (after decoding and parsing whatever was in the string according to `contentEncoding` and/or `contentMediaType`) passing it back in to validate the string contents.

Let me know if you have any ideas on better wording, or if you think I should just drop the parenthetical aside.

jdesrosiers: I see. I think it can be dropped.

handrews: Cool- done!

jdesrosiers (approved):

karenetheridge:
I'm against this change; I've found it useful in OpenAPI validation. I understand that in some situations it might be a security risk, so it's reasonable to have its support be optional (as it already is), and to default to NOT performing validation. I've been thinking of splitting it into two vocabularies: Annotation and Validation. Like with `format`, it can be enabled when desired, and left disabled most of the time.

handrews:
@karenetheridge this is not about `content*` having validation behavior (which it does not).  This is about the complexity of shoving an entire second (but totally indpendent) validation process into an interface that doesn't allow it.

You could still create a validating content vocabulary after this change.

handrews:
@karenetheridge also, the "automatic feed-back-to-validate" process can trivially be implemented as a library on top of any JSON Schema implementation that supports annotation collection and any standard output beyond "flag".  This PR is about taking something that is better handled as a separate library out of the spec.  It just doesn't need to be here, and there is no other reason to add this much complexity to the basic interface of JSON Schema.

handrews:
@karenetheridge since the `content*` keywords in the `https://json-schema.org/draft/next/content` vocabulary are strictly annotations, there are two ways to validate string contents while remaining in compliance with the spec:

1.  take the output of the _completed_ initial evaluation, and for each instance location to which `content*` annotations are attached
    * decode the string according to `contentEncoding` (the spec gives no guidance on handling any errors)
    * parse the string according to `contentType` (the spec gives no guidance on handling any errors)
    * evaluate the parsed string against `contentSchema` as a new evaluation process
    * recurse if any `content*` annotations are present in this new output
    * store the output to return alongside but separately from the initial evaluation and any other secondary evaluations
2.  create a `content-assertion` extension vocabulary analogous the the `format-assertion` vocabulary, which would perform the encoding/parsing/additional validation as part of the regular (single) evaluation process

Option 1 can be done standalone or shoved into the main JSON Schema implementation, creating a completely different set of output requirements from everything else.  _Or_ it can be done as a separate library, without burdening the JSON Schema implementation's output interface that works for literally everything else in JSON Schema.

There is no functional difference whatsoever between doing it within the implementation or as a separate library.  Please let me know if you do see a difference, because I would like to understand that.  I am proposing this change because fundamentally, it's not a change.  It just moves non-essential functionality to a library, which could even have its own little specification to ensure interoperability of this area.  That would be an improvement over the current situation.

Option 2 is not impacted by this change at all.  Supporting a `contentSchema` assertion-applicator is tricky as it is not clear how to represent instance locations within a string, so for now any such feature would not be interoperable, but we could figure that out if we wanted to make it interoperable.

karenetheridge:
I don't understand your references to output requirements in Option 1. What does output have to do with anything here?

The issue I have with moving this functionality out of the spec is, if we only talk about annotations in the spec, we lose the information needed to implement these keywords as validations consistently.

But, more importantly, if we "..take the output of the completed initial evaluation..", then we can't use the results of evaluating the content\* keyword(s) and use them as inputs to applicator keywords one level up in the document. So I see Option 2 as much more preferrable.

> it is not clear how to represent instance locations within a string..  but we could figure that out if we wanted to make it interoperable

The string is decoded (using the specified media type) into something that matches the JSON document model, so I presumed (when implementing this) that the instance location just carries on using json pointers inside the decoded data structure. That's under-specified though, so it would be good to clarify that (assuming the use of validation stays in the spec at all).

handrews:
> But, more importantly, if we "..take the output of the completed initial evaluation..", then we can't use the results of evaluating the content* keyword(s) and use them as inputs to applicator keywords one level up in the document

The spec has _never_ allowed this since `contentSchema` was introduced in 2019-09.  It's expressly forbidden:

>    Implementations MAY offer the ability to decode, parse, and/or
   validate the string contents automatically.  However, it MUST NOT
   perform these operations by default, and MUST provide the validation
   result of each string-encoded document separately from the enclosing
   document.

You're currently not allowed to use that result _within_ the validation of the enclosing document.  It MUST be produced as separate output results, which puts `content*` assertion behavior into a totally unique processed divorced from normal implementation.

This PR removes that uniqueness.  If anything, _it makes what you want to do easier_ by removing this constraint.

> So I see Option 2 as much more preferrable.

That's great!  This PR actually makes that easier by removing the part that more-or-less forbids that.

I don't hear any actual objection from you regarding what this PR does.  If anything, this sounds like support.  You're saying "no, don't merge this", but also "my preferred solution is the thing that is easier if we merge this."

> so I presumed (when implementing this) that the instance location just carries on using json pointers inside the decoded data structure. That's under-specified though, so it would be good to clarify that (assuming the use of validation stays in the spec at all).

You should file an issue for that, as it is not relevant to this PR.  It was underspecified before, and it will be the same level of under-specified afterwards.

karenetheridge:
> It was underspecified before, and it will be the same level of under-specified afterwards.

Well no, not really - if you take out the descriptions of how the content\* keywords would validate, then that leaves the reader to make up their own (inconsistent) rules. ..But on looking again, I see you're not removing those paragraphs.

> I don't hear any actual objection from you regarding what this PR does. If anything, this sounds like support. You're saying "no, don't merge this", but also "my preferred solution is the thing that is easier if we merge this."

I guess? I don't see any path forward to allowing in-place evaluation as assertions, other than creating a new vocabulary (akin to the format-annotation and format-validation vocabulary split in 2020-12), but I'd been pondering that myself and it seems reasonable.

So I guess my vote is "not-no" - I'm not terribly happy with the spec staying in the state after merging this, but the parts you are removing are parts that I definitely disagreed with. If that makes sense :)

handrews:
@karenetheridge 

>  the parts you are removing are parts that I definitely disagreed with. If that makes sense :)

Yes, exactly!  This PR is about removing a _bad_ solution to the problem you want to solve.  It takes no position on other possible solutions, and you are quite welcome to file proposals for other solutions.  I would actually suggest a slightly different approach to _validation_, using more specific keywords.  A major problem with these a validation keywords is that's they're so open-ended.  But if you specifically want to validate JSON-in-JSON,  or or parse HTML-in JSON, etc., you can write a more clear spec for that behavior.

But that's a discussion for another issue (or discussion).

I'll go ahead and merge this, then.  Two "yes"es and a "not-no" seem sufficient :-)

PR #1297, opened 2022-09-19
Note non-anchoring of regexes in the regex section.

### LINKED ISSUE #897: "patternProperties" should mention that it's not implicitly anchored
The `"patternProperties"` section should mention that it's not implicitly anchored. `"pattern"` mentions that it's not.

handrews:
Fixes #897

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -564,6 +564,11 @@
                     dialect described in <xref target="ecma262">ECMA-262, section 21.2.1</xref>.
                 </t>
                 <t>
+                    Unless otherwise specified by a keyword, regular expressions MUST NOT be
+                    considered to be implicitly anchored at either end.  All regular expression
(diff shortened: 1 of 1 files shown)


jdesrosiers:
It looks like the consensus in the linked issue was to mention it in the `patternProperties` keyword rather than as a general statement. It seems that if you're going to do something different, it should have some discussion.

I think having a general statement is a good idea. I don't care too much whether the keywords mention it or not, but I think they should do the same thing. Either mention anchoring or don't. Currently `pattern` does and `patternProperties` doesn't.

handrews:
There was support for both locations.  I'll update `patternProperties` as well, but it should be in the general section (for anyone making extension keywords), and no one objected to that in the issue (in fact that is what the issue title was changed to say).

handrews:
Now it 's in both places.

jdesrosiers (approved):

PR #1308, opened 2022-09-21
add pr-dependencies-check workflow

gregsdennis:
I use this on my `json-everything` repo (and I've used in it a few other places) to track when PRs are dependent upon each other.  For example, how #1285 is dependent upon #1085.

https://github.com/marketplace/actions/pr-dependency-check


handrews (approved):
Seems fine to me although I have no particular sense of how much we want to automate such things.  But it looks useful.

PR #1312, opened 2022-09-22
Reverse "contains" <-> "minContains"/"maxContains" dependency

### LINKED ISSUE #1161: "Keyword is present" meaning
In the current draft, some keywords behavior depends on the presence and annotation results of keywords from other vocabularies.

Example:
https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.4.4
> If "contains" is not present within the same schema object, then this keyword has no effect.
>
> An instance array is valid against "maxContains" in two ways, depending on the form of the annotation result of an adjacent "contains" keyword. The first way is if the annotation result is an array and the length of that array is less than or equal to the "maxContains" value. The second way is if the annotation result is a boolean "true" and the instance array length is less than or equal to the "maxContains" value.

If the Applicator vocabulary isn't used to process the current schema, but the "contains" property exists in the schema object, should an implementation consider it as presented? If yes, then we will have annotation produced by the unknown keyword "contains" with an arbitrary value. Such a case isn't considered in the description of the "maxContains" keyword.

### LINKED ISSUE #1311: Reverse direction of "contains" <-> "minContains" / "maxContains" dependency
Note:  Other folks (@jdesrosiers ? @karenetheridge ?) came up with this a while ago and apparently I objected to it (which I 100% believe but no longer recall the reason).

JSON Schema has an architectural principle that individual assertion keywords that fail cannot be un-failed.  While the assertion result can be manipulated at the schema object level (e.g. by `not` or `oneOf`, which pass when one or more subschemas fail), doing so at the individual keyword level violates the principles of a layered constraint system.

This means that a keyword can only produce an assertion result if it has all of the information necessary to determine the final result.  Only `contains` can meet this criteria out of these three keywords, and only if it knows the values of `minContains` and `maxContains`.

This means that `minContains` and `maxContains` cannot produce a failed assertion result.  They only exist to provide extra input to `contains`.

One subtlety is that since we describe interactions in terms of annotations (even though they do not need to be implemented that way), and because the absence of a keyword _cannot_ produce an annotation ([§7.3 ¶2](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-default-behaviors)), this means that the default `minContains` behavior of `1` must be specified as how `contains` behaves in the absence of `minContains`, not simply as the default value of `minContains`.

To make the behavior of this group of keywords clear, and because `minContains` and `maxContains` no longer have assertion-failing behavior, they should be moved to the applicator vocabulary with `contains`.

handrews:
Fixes #1311.  Fixes #1161 by eliminating the confusing situation that produced the concern.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2122,7 +2122,7 @@
                         </t>
                         <t>
                             "contains", whose behavior is affected by the presence and value of
-                            "minContains", in the Validation vocabulary
+                            "minContains"
                         </t>
                     </list>
                 </t>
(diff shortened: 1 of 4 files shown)


gregsdennis:
I think this wording works, but do we want to have `minContains` & `maxContains` produce an annotation that is consumed by `contains`?  I understand the "missing keywords can't produce annotations" thing, but that's covered by the current language of assuming a min of 1.  Having this annotative communication maintains consistency with how other keywords work.

handrews:
@gregsdennis this wording was intended to specify the behavior in terms of annotations, and then note an alternative (which we do for some other keywords as well).  What are you looking for that isn't there?  Do you want more explicit discussion of the consumption of annotations in the description of "contains"?

gregsdennis:
I was thinking that `minContains` and `maxContains` would actively produce an annotation that would be consumed by `contains`.  This follows the model that we have established with `additional*` and `unevaluated*`.

Currently the wording you have for min/max doesn't say anything about annotations.  I think this needs to be added.

Essentially, the dependency is on the annotation rather than on the other keyword directly.

handrews:
@gregsdennis 

> Currently the wording you have for min/max doesn't say anything about annotations. I think this needs to be added.

Lines 2530 and 2544 say:

> The value of this keyword is used as its annotation result.

gregsdennis:
Okay, yeah.... I missed that.

gregsdennis (approved):

handrews:
@gregsdennis no worries, the mentions are very brief!  It actually took me a moment to find them again.

jsonschema-core.xml
```diff
@@ -2517,23 +2517,65 @@
...
-                            An array instance is valid against "contains" if at least one of
-                            its elements is valid against the given schema,
-                            except when "minContains" is present and has a value of 0, in which
-                            case an array instance MUST be considered valid against the "contains" keyword,
-                            even if none of its elements is valid against the given schema.
+                            This keyword applies its subschema to array elements or object property values.
+                        </t>
+                        <t>
+                            An instance is valid against "contains" if the number of elements or property
+                            values that are valid against its subschema is with the inclusive range of
```
jdesrosiers, line 2562:
```suggestion
                            values that are valid against its subschema is within the inclusive range of
```

jdesrosiers (changes requested):
Just one small typo, but otherwise looks good.

jdesrosiers (approved):

handrews:
The force-push just now was to rebase and (more importantly) credit @karenetheridge for first coming up with the idea in the commit log.

PR #1318, opened 2022-09-24
Remove annotation-applicator interaction section

### LINKED ISSUE #1313: Remove language about applicators aggregating annotation results
This section:

> [7.7.1.3. ](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-7.7.1.3)[Annotations and Applicators](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-annotations-and-applicators)
In addition to possibly defining annotation results of their own, applicator keywords aggregate the annotations collected in their subschema(s) or referenced schema(s).

should be removed or reworked.  The "aggregation" language came from earlier ideas where annotations would define semantics for combining multiple values attached to the same instance location.  For example, [`readOnly` and `writeOnly` would be ORed.](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-10.3)  A few draft-07 implementations [actually did support this to some degree](https://github.com/hotkit/json-schema/blob/6f56e6fec149519796fdfc30f17384bb0ff9a466/include/f5/json/validator.hpp#L36).

We later decided that JSON Schema should not attempt to do this sort of thing, and would just defer all annotation usage logic to applications.

But because of that history, I had a mindset of applicators actively propagating annotations.  This contributed to the [confusion](https://github.com/json-schema-org/community/discussions/57) around the `properties`->`additionalProperties` and similar annotation communication paths, as it made it more ambiguous as to where the decision to produce or remove annotations occurred.

Ultimately, the consensus settled on keywords being able to produce annotations for use within schema objects, and [only being dropped in response to failed validation at the schema object level](https://github.com/json-schema-org/json-schema-spec/pull/1154#pullrequestreview-893513660).

This means that _applicators **do not** play a role in managing annotations other than their own,_ and therefore §7.7.1.3 can be dropped entirely.  There are no special interactions between annotations and applicators as classes of keywords.

handrews:
Fixes #1313.  The idea of annotations and applicators interacting is outdated. Annotation deletion/dropping occurs at the schema object level (clarified in recent drafts), and annotation values have not defined aggregation semantics within JSON Schema evaluation since draft-07.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -655,8 +655,12 @@
                 Typically, applicator keywords are processed until a schema object with no
                 applicators (and therefore no subschemas) is reached.  The appropriate
                 location in the instance is evaluated against the assertion and
-                annotation keywords in the schema object, and their results are gathered
-                into the parent schema according to the rules of the applicator.
+                annotation keywords in the schema object.  The interactions of those
+                keyword results to produce the schema object results are governed by
+                section <xref target="annot-assert" format="counter"></xref>, while the
+                relationship of subschema results to the results of the applicator
+                keyword that applied them is described by section
+                <xref target="applicators" format="counter"></xref>.
             </t>
             <t>
                 Evaluation of a parent schema object can complete once all of its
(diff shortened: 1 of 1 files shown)


handrews:
I found more somewhat confusing / outdated language and removed or updated it.

Please note that while several of these sections could benefit from aggressive editing, I am assuming that will happen as we move to the new process.  The purpose of this PR is to ensure that the starting point for that work is correct (because I forgot to fix this before and don't trust that I'll remember it later).

gregsdennis (approved):

PR #1325, opened 2022-10-12
Fix a lot of build warnings (only 7 lines of changes)

handrews:
This gets rid of all of the build warnings except the ones for the lines being too long, which aren't worth trying to fix as we're changing formats anyway.

But these are simple fixes to remove everything else. As of RFC 7749 (xml2rfc v2 vocabulary), the RFC 2629 DTD is no longer needed.  The submissionType was also added then.  It is supposed to default to "IETF" but apparently newer versions of xml2rfc no longer do this.

I don't know why xml2rfc did not like the no-breaking space in the xref, but it's happier with it gone.


Relequestual (approved):
Seems fine by me. Works fine after rebase too.

Issue #1319, opened 2022-09-26
Hierarchical output structure needs clarification around which nodes are kept


gregsdennis:
Follow-up to #939

Currently, the specification says that all nodes are to be kept for the `hierarchical` structure.  This needs to be clarified to state more precisely which nodes are intended.

This structure was meant to be a replacement for the 2019-09/2020-12 `verbose` structure.  Therefore _all_ nodes, including those which do not impact validation would be retained.

However, I realize this makes for a significantly larger structure in many cases.  As such, I would like to propose that we default to only including those output units which affect the final outcome, but allowing for implementations to include all of them upon request (or configuration)

I'm open to opposing arguments and alternatives.

handrews:
@gregsdennis I've always been a little unclear about how `verbose` works.  I recently noticed this (I'm sure I saw it before, but I'd forgotten):

> This includes sub-schema validation results that would otherwise be removed (e.g. annotations for failed validations, successful validations inside a `not` keyword, etc.).

The spec mandates that annotations for failed validation scopes are deleted, so I don't understand what's happening here.  On the other hand, I can see the _value_ of such annotations being preserved for debugging, but technically doing so would require modifying the spec (unless those annotations are re-categorized as errors somehow).

gregsdennis:
I think its primary use is debugging.  It can be helpful knowing why, e.g., `unevaluatedProperties` processed `foo`, which you can find out by following annotations (another reason I output even the "internal" annotations).

But I think that use case is probably less common, and we should default to a more common use case, which is people want what affects final outcome.

handrews:
> another reason I output even the "internal" annotations

yes, I think most people do output them, if they support it at all.  And I definitely see the debugging use.

Are you including annotations from failed subschemas, though?  That is the part that I'm a bit confused about.  Because the spec requires deleting them, but I can see them being useful for debugging as well so I wasn't sure what happened with `verbose` in that case.

gregsdennis:
I output them where they were produced, yes, but it should be apparent from the validation results that they don't "flow down" (or up, depending on your tree orientation) or factor into the outcome.

handrews:
> I output them where they were produced, yes, but it should be apparent from the validation results that they don't "flow down" (or up, depending on your tree orientation) or factor into the outcome.

I don't really know what this means.  Does this mean that you save them for output but not for use in parent schemas?  Does this mean that sometimes there are both annotations and errors in the output unit?

[...]

PR #1329, opened 2022-10-18
add droppedAnnotations

gregsdennis:
Partially addresses #1319.

This adds `droppedAnnotations` to the output unit as an optional field.  While they can be useful for debugging, these annotations are usually just noise.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2915,8 +2915,9 @@ https://example.com/schemas/common#/$defs/allOf/1
                         validation result of the containing subschema was successful.
                     </t>
                     <t>
-                        Implementations that wish to include these annotations are encouraged
-                        to provide users with a configuration option to exclude them.
+                        Implementations that wish to provide these annotations MUST NOT provide
+                        them as their default behavior.  These annotations MUST only be included
+                        by explicitly configuring the implementation to do so.
                     </t>
                     <t>
                         The value for this property MUST be an object where the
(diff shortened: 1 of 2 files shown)


jsonschema-core.xml
```diff
@@ -2903,6 +2906,32 @@ https://example.com/schemas/common#/$defs/allOf/1
...
+                    <t>
+                        Any annotations produced and subsequently dropped by the evaluation
+                        due to an unsuccessful validation result of the containing subschema.
+                        This property MAY be included if the validation result of the containing
+                        subschema was unsuccessful.  It MUST NOT be included if the local
+                        validation result of the containing subschema was successful.
+                    </t>
+                    <t>
+                        Implementations that wish to include these annotations are encouraged
+                        to provide users with a configuration option to exclude them.
```
handrews, line 2919: Do we want this to be opt-in or opt-out?  It's potentially expensive, and because of `oneOf`, etc. it can happen on successful runs as well.

We don't need to resolve this to merge this PR, I'm just raising a philosophical point about what our default behavior ought to be, for both this and https://github.com/orgs/json-schema-org/discussions/236

handrews (approved):
I raised a concern but it's not a blocker.

jsonschema-core.xml
gregsdennis: Yeah, I would expect this to be opt-in.  I'll add default behavior language.

gregsdennis: Updated.

PR #1331, opened 2022-10-24
Update code of conduct in CONTRIBUTING.md

handrews:
I copied this file over to the referencing spec and was informed it was out of date, so here's a backport of the update.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ As an informative document, bugs in the meta-schema may be fixed in place to ali
 
 ## Conduct
 
-All official channels including the mailing list, GitHub organization, and Freenode channel, follow the IETF Guidelines for Conduct as specified in [RFC7154](https://tools.ietf.org/html/rfc7154).
+All official channels including the mailing list, GitHub organization, Slack server, and IRC channels, follow our [Code of Conduct](https://github.com/json-schema-org/.github/blob/main/CODE_OF_CONDUCT.md).
 
 
 ## Financial contributions


CONTRIBUTING.md
```diff
@@ -46,7 +46,7 @@ As an informative document, bugs in the meta-schema may be fixed in place to ali
 
 ## Conduct
 
-All official channels including the mailing list, GitHub organization, and Freenode channel, follow the IETF Guidelines for Conduct as specified in [RFC7154](https://tools.ietf.org/html/rfc7154).
+All official channels including the mailing list, GitHub organization, Slack server, and Freenode channel, follow our [Code of Conduct](https://github.com/json-schema-org/.github/blob/main/CODE_OF_CONDUCT.md).
```
Julian, line 49: Do we still have a Freenode channel? I assume we don't have an IRC channel at all at this point and if we did it's probably on Libera (but I haven't checked in, maybe I should)

Julian: There's a couple of users in the Libera channel looks like, including @karenetheridge. (I don't have access to Freenode anymore nor do I think anyone else does, they password reset everyone during the hostile takeover or something right?)

TL;DR maybe it's easy to avoid all my comments by changing this to say IRC channels and remove the reference to Freenode.

Other than that, lgtm.

handrews: @Julian thanks, I changed it to "IRC channels".

karenetheridge (approved):
If there's a "resources" section in our docs somewhere, it would be good to explicitly mention the libera irc channel (although I'd want more people connected to it, ideally :) )

handrews:
@karenetheridge I think I'll leave it to those who understand the state of the IRC stuff to update that.  For now I'd like to just get the CoC link fixed.  I'd be happy to quickly review and approve an IRC PR afterwards, though.

gregsdennis:
Merging as it's been approved for three weeks with no further activity.

PR #1345, opened 2022-10-27
Update `properties` and `patternProperties` annotation language

### LINKED ISSUE #1310: Annotation specification of `properties` could be improved
I had a realization today about what the text in [section 10.3.2.1](https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.3.2.1) says about the annotation result of `properties`

> The annotation result of this keyword is the set of instance property names matched by this keyword.

## How I interpreted this

I originally took this to mean that instance properties that were listed in `properties` but were not matched merely weren't included in the annotation output.  So in this schema and instance

```json
{
  "properties": {
    "foo": true,
    "bar": false
  },
  "additionalProperties": true
}

{ "foo": 1, "bar": 1 }
```

`properties` would produce `[ "foo" ]` because `foo` would always pass, and `bar` would never pass.  The impact of this would be that `bar` would then also be evaluated by `additionalProperties`, which would pass.

This never really made sense to me because `/properties/bar` would fail validation overall anyway, which means it doesn't matter that `bar` passes `additionalProperties`.

## My realization

I now see that this requirement was stated this way to address the annotation result of the case that _not all properties were present in the instance_ rather than the case where one of the properties failed validation.

This means that in the schema above and the instance `{ "foo": 1, "baz": 1 }`, `properties` would produce `[ "foo" ]` because `bar` isn't present in the instance and `baz` isn't listed in `properties`.

So that means that `properties` produces an annotation of the intersection of the properties it declares and the properties the instance contains, regardless of whether the property's value passes the associated subschema.

In the second case, it generally doesn't matter whether `additionalProperties` is depends on the annotation because the schema's going to fail anyway.

## Why does this matter?

It matters in error output.  To illustration this, let's change `additionalProperties` to `false`:

```json
{
  "properties": {
    "foo": true,
    "bar": false
  },
  "additionalProperties": false
}

{ "foo": 1, "bar": 1 }
```

In my original interpretation, `additionalProperties` evaluates `bar` and produces an error.

In my new interpretation, it doesn't evaluate `bar` and so no error is produced.

## What am I asking?

Can we improve this text to more clearly specify which of these cases (or a third case, even) is intended?  Personally, I like the "intersection" language I used above.

I think it was the "matched by this keyword" that threw me off.  I took "matched" to mean "successfully validated."

(I think this has come up somwhere in a conversation between @karenetheridge and me.)

gregsdennis:
Resolves #1310 

Improves the annotation definition language for `properties` and `patternProperties`.

`properties` - replaced "match" language to be more explicit
`patternProperties` - clarified "match" language

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2404,7 +2404,7 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names which are also present in this keyword.
+                            property names which are also present under this keyword.
                             This annotation affects the behavior of "additionalProperties" (in
                             this vocabulary) and "unevaluatedProperties" in the Unevaluated vocabulary.
                         </t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2404,7 +2404,7 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names matched by this keyword.
+                            property names which are present in both this keyword and the instance.
                             This annotation affects the behavior of "additionalProperties" (in
                             this vocabulary) and "unevaluatedProperties" in the Unevaluated vocabulary.
                         </t>
@@ -2430,7 +2430,7 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names matched by this keyword.
+                            property names matched by at least one key in this keyword.
                             This annotation affects the behavior of "additionalProperties" (in this
                             vocabulary) and "unevaluatedProperties" (in the Unevaluated vocabulary).
                         </t>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -2430,7 +2430,7 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names matched by at least one key in this keyword.
+                            property names matched by at least one property under this keyword.
                             This annotation affects the behavior of "additionalProperties" (in this
                             vocabulary) and "unevaluatedProperties" (in the Unevaluated vocabulary).
                         </t>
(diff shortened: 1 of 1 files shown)


gregsdennis:
@handrews made a [comment](https://github.com/json-schema-org/json-schema-spec/issues/1310#issuecomment-1255454255) that the introduction of "A Vocabulary for Unevaluated Locations" will need to be updated, but after reading it, I'm not sure what was meant.  I think the changes present so far in this PR are sufficient to satisfy the need I expressed in the issue.

handrews (approved):
I'm pretty sure that whatever I was talking about with the unevaluated vocabulary was not directly about this specific wording change.

jsonschema-core.xml
```diff
@@ -2430,7 +2430,7 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names matched by this keyword.
+                            property names matched by at least one key in this keyword.
```
karenetheridge, line 2433: Do we say "key" elsewhere? I think we generally say "property", as the thingy underneath a JSON-object.
Also "in" -> "under" as above.

```suggestion
                            property names matched by at least one property under this keyword.
```

jsonschema-core.xml
```diff
@@ -2404,7 +2404,7 @@
                         </t>
                         <t>
                             The annotation result of this keyword is the set of instance
-                            property names matched by this keyword.
+                            property names which are also present in this keyword.
```
karenetheridge, line 2407: I think "under" more fits language we use elsewhere, rather than "in"?

```suggestion
                            property names which are also present under this keyword.
```

karenetheridge (changes requested):

gregsdennis:
@karenetheridge I've incorporated your updates.  Can you re-review, please?

karenetheridge (approved):

Issue #1333, opened 2022-10-26
Write ADR describing the decision to move to the new SDLC process


gregsdennis:
Following the ADR to leave IETF, we needed a new process.  We have landed on @jdesrosiers' [Proposal: New Spec Development Lifecycle Model](https://github.com/orgs/json-schema-org/discussions/234) or SDLC.

We need another ADR that summarizes this discussion and explains the decision to use this model and its consequences.

PR #1346, opened 2022-11-04
Add ADR for choosing a new SDLC

jdesrosiers:
Resolves: https://github.com/json-schema-org/json-schema-spec/issues/1333

Here's the ADR for documenting the reasons for the paradigm shift in how we develop the spec. I included the approach @awwright laid out recently, but I don't see that so much as an alternative as complementary. If we decide we want to work toward that vision, it would fit just fine into the process we have been discussing and have agreement on.

### DIFF
--- a/adr/2022-11-stable-spec.md
+++ b/adr/2022-11-stable-spec.md
@@ -25,7 +25,7 @@ There have been two proposals put forward. Both address the goal of a stable
 specification with the ability to evolve. The third option represents sticking
 with the status quo.
 
-### Option 1 - TC-39 inspired
+### Option 1 - TC-39 Inspired
 The spec would be converted from I-D XML to Markdown, but can otherwise be
 structured however we choose. A system would be put in place to allow us to flag
 the stability level of any feature in the spec. There would be only one version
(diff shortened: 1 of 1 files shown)


adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,180 @@
...
+The benefit of this approach is that it's compatible with the IETF process
+without imposing some of the constraints and perception issues that we had with
+IETF. We can pursue an RFC in the future if we choose to without significant
+changes or spec restructuring.
+
+With this proposal, releases are done as a new document that replaces the
+previous documents. Compared to the constantly evolving spec in Option 1, change
+from non-functional clarifications and bug fixes to adding and evolving new
+features takes much longer if you have to wait for the next release to make a
+change. This lengthens the feedback loop slowing spec development progress.
```
gregsdennis, line 152: This description of Option 2 is what I thought it was.  The paragraphs above doesn't describe a "document replacement" process.  Instead they merely focus on extensions.

gregsdennis (changes requested):

adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,180 @@
...
+has no compatibility guarantees between versions.
+
+## Decision Outcome
+The decision is to go with Option 1 while leaving discussion open for nearly all
+of Option 2 because it is mostly compatible with Option 1. Option 2 uses an
+immutable spec where each release replaces the last while the Option 1 uses a
+mutable spec. The outcome of having only one current version of the spec is
+achieved with either option, but the mutable spec allows us to remove some
+unnecessary roadblocks in our development processes and allows us to release a
+stable spec much sooner. Other than that, discussion for the rest of Option 2
```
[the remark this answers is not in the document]
jdesrosiers: I'm not seeing that. I think both options have similar outcomes for implementers. Both only require implementers to code for the current release and provide the same compatibility guarantees. But, I'm happy to add that point if you can elaborate more on your reasoning.

adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,180 @@
...
+Option 1 puts no constraint on the structure of the spec and restructuring is
+allowed at any time as long as it doesn't break compatibility requirements.
+Therefore, restructuring the spec as "Core Semantics" and "Standard Extensions"
+is compatible with Option 1. We can move forward with Option 1 now while leaving
+the restructuring discussion open.
+
+Option 2 defines a new extension mechanism and some new keywords. These features
+can be introduced under the stability model described in Option 1. Therefore, we
+can move forward with Option 1 while leaving the discussion open for these new
+features.
```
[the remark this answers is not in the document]
jdesrosiers: The extension mechanism is a small and almost irrelevant part of what is being decided here. I included this paragraph mostly to make it clear that the proposed extension mechanism and keywords were not being explicitly rejected by this decision. Maybe the proposed extension mechanism and keywords shouldn't be mentioned at all.

adr/2022-11-stable-spec.md
jdesrosiers: You're right, Option 2's "Considered Options" section is missing a lot of important details. I'll expand that section. However, Option 2 will still a focus on extensions because it proposes that extensions are the primary way for the spec to evolve.

adr/2022-11-stable-spec.md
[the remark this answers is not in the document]
gregsdennis: > Maybe the proposed extension mechanism and keywords shouldn't be mentioned at all.

Yeah, I think this should just stick to the process.

adr/2022-11-stable-spec.md
gregsdennis: > However, Option 2 will still a focus on extensions because it proposes that extensions are the primary way for the spec to evolve.

I think I see where you're going with this, but perhaps being a bit more explicit on how extensions are used to evolve the spec may shed light on the difference in process between the two options.

adr/2022-11-stable-spec.md
```diff
@@ -44,66 +44,64 @@ designated by the year they were released.
...
-essential for implementations to support. Changes to this spec must be
-compatible with previous releases with exceptions only in extreme cases.
-
-A registry could be maintained that maps keywords to their specified semantics.
-User extensions that aren't in the registry should use a URI for their keyword
-to avoid conflicts with other third-party extensions or with future standard
-extensions.
+extension mechanism. This spec should rarely change. New features would be added
+through additional specifications that define extensions to the "Core Semantics"
+spec.
```
gregsdennis, line 57: 👍

adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Pros and Cons of the Options
+The biggest benefit is shared between Option 1 and Option 2. Both approaches
+result in a stable spec. This will have benefits for both implementers and
+users. Because of the compatibility requirements, whenever you write a schema,
+you will never need to change it just to keep up with changes to JSON Schema.
+This is also better for implementers because they don't have to maintain
+separate code with different semantics in different versions. They just need to
+code for the current release and they will automatically have support for past
+releases.
```
Relequestual, line 100: I think it's worth noting that "past releases" means "from when the first "stable" release is created, not including previous "draft era" versions of JSON Schema".
It's unlikely, but possible, that someone reading this might conclude that they should expect all previous versions, including "draf era" verisons of JSON Schema, should be supported in a single implementation which claims "stable era" support.

adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,178 @@
...
+spec evolution and the mutability of the spec document.
+
+Having a mutable spec allows us to make clarifications and bug fixes immediately
+rather than having to wait months or years for the next release to go out. It
+also allows us to iterate faster on unstable features which would allow us to
+get them to a stable state much sooner. For example, we have changes to dynamic
+references that have been agreed upon and ready to go for over a year, but users
+can't benefit from the change until we can get the next full release published.
+With this model, the change could have been made available for over a year now
+and we would have a years worth of feedback on it's use. Having a mutable spec
```
Relequestual, line 113: Thinking/spitballing: In such a situation, how would a user be able to determine if an implementation supported the RIGHT version of an unstable feature?
Should we provide guidance that says something like unstable features should be referenced by implementations by the specific commit date of the spec file and tests (if any) they are intending to implement?

That got me thinking further...
If we specified that pre-stable keywords MUST be postfixed with the date of the commit, or stage number, or something, idk, then you make sure users and implementations can differenciate easily between pre-stable and stable versions of that keyword. Although I guess you'd expect to be able to do this with the "release" dialect.

And, I wonder if this is somewhat a moot point if we prevent non-dialect/vocabulary defined keywords from being used.

adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,178 @@
...
+also allows us to introduce new features without having to wait for a release.
+For example, the `propertyDependencies` keyword has also been waiting for months
+for a release. Users could have been benefiting from it for months and providing
+feedback.
+
+The downside of a mutable spec is that it can be more difficult for implementers
+and users to track when changes happen. We will need to be better at
+communicating changes in blog posts or equivalent.
```
Relequestual, line 121: I have some automation ideas on this.
Any such concept though would require us to track "latest modified" status of keywords in a structured data file, so automation can be run against such data.
This would add a burden to spec authors, unless unstable features/keywords are written in individual files, and merged into a single spec artifact at workflow time. In which case, changes would be easier to catch, but you still wouldn't know if it was a change requiring implementation change or not unless a human specified such.

adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,178 @@
...
+world feedback without committing to full compatibility requirements. This makes
+it much more likely that we don't get stuck with something that doesn't work out
+or could be done better.
+
+The stability model also makes it clear to users which features are stable and
+how likely a feature is to change in the future. Whether they prefer to stick
+with stable features or want to use a new keyword, users have the information
+they need to make that decision.
+
+The downside of the stability model is that it presents a very high barrier for
```
Relequestual, line 134: It's debatable if a high barrier is a downside. If stability is the goal, a high barrier can actually be a benefit.

adr/2022-11-stable-spec.md
```diff
@@ -0,0 +1,178 @@
...
+* https://github.com/jdesrosiers/json-schema-spec/blob/main/adr/2022-09-decouple-from-ietf.md -
+  The ADR for the decision to decouple from IETF
+* https://github.com/orgs/json-schema-org/discussions/234 - Proposal submitted
+  by @jdesrosiers for a process to replace the IETF based process we'd been
+  using.
+* https://github.com/orgs/json-schema-org/discussions/257 - @awwright's vision
+  for JSON Schema including how it can continue to evolve while still having a
+  stable core.
+* https://github.com/json-schema-org/community/discussions/119 - When we first
+  started talking about forward compatibility and a stable spec.
```
Relequestual, line 178:
```suggestion
  started talking about forward compatibility and a stable spec.
  * https://json-schema.org/blog/posts/future-of-json-schema - User friendly
    comments on decoupling from the IETF.
```

Relequestual (changes requested):
I've made a few comments.
Some are change suggestions, others are thoughts that come to mind while reading it which I wonder if they are worth making notes of (should you keep a list of notes on the SDLC some place. Feel free to create new discussions if a related discussion doesn't exist).

adr/2022-11-stable-spec.md
jdesrosiers: This is a bit out of scope for this PR, so let's have this discussion elsewhere.

adr/2022-11-stable-spec.md
jdesrosiers: This is a bit out of scope for this PR, but I think I'd argue that this communication is better done by a person than an automation. An automation can't explain why something changed.

adr/2022-11-stable-spec.md
jdesrosiers: I agree. I thought I had addressed that in previous paragraphs, but it seems I only implied it and should be more clear. The high barrier is definitely a benefit, but there's always a trade off. I was trying to address the trade off in this paragraph. I'll try to make it more clear.

adr/2022-11-stable-spec.md
gregsdennis: Automation could provide reminders for one of us to post on a change, though.

adr/2022-11-stable-spec.md
gregsdennis: To the point being made in the text, when I decided to create JsonSchema.Net as a replacement for Manatee.Json, there was a LOT of work to do, and the vast majority of it was re-implementing all of the keywords.  This high barrier was a serious factor in my consideration of making a new library.  (Obviously I decided to do it, but I did have to think about the workload I was taking upon myself.)

jdesrosiers: Right. Although it creates high barriers for spec changes, it actually lowers the barrier for implementers, which is probably worth mentioning as well.

handrews (approved):
I had apparently started a review when this was first posted but never published it 🤦 
The comments I had written have since been addressed.  There may be points here I'll want to clarify or debate as we document the actual process, but as documentation of the decision this looks good to me!

Relequestual (approved):

Issue #1334, opened 2022-10-26
Formally document the new SDLC


gregsdennis:
@jdesrosiers' [Proposal: New Spec Development Lifecycle Model](https://github.com/orgs/json-schema-org/discussions/234) (SDLC) needs formal documentation.

There may be a few things to work out, but the concept is mostly defined.

gregsdennis:
Closed by #1500

PR #1348, opened 2022-11-16
Add PROCESS.md documenting the new SDLC

jdesrosiers:
Resolves: https://github.com/json-schema-org/json-schema-spec/issues/1334

This is the formal documentation of the process we've discussed in https://github.com/orgs/json-schema-org/discussions/234. That discussion was focused on getting agreement on the general direction, while this makes some concrete decisions. Please do challenge any of the concrete decisions I've made here, but keep in mind that this is just a starting point, so it might be more productive address issues after merge as separate PRs/Issues so we don't end up with multiple threads of discussion in this PR.

Several more concrete decisions still need to be made that should be addressed in the spec (for example: dialect and vocabulary URIs). Those decisions are out of scope for this document.

I made some minor changes from what we had discussed to how `STAGE-2` is defined, so pay particular attention to that section when reviewing.

### DIFF
--- a/PROCESS.md
+++ b/PROCESS.md
@@ -1,0 +1,163 @@
+# JSON Schema Specification Development Process
+This document formally defines the process used to evolve the JSON Schema
+specification as of the first stable release in 2023. It applies to the core
+specification and the specifications for any dialects and vocabularies
+maintained by the JSON Schema Org. It does not necessarily apply to everything
+the JSON Schema Org maintains. For example, media type registration follows the
+IETF process. Certain components used within JSON Schema, such as Relative JSON
+Pointer, may also follow a separate process.
+
+## Canonical URLs
+There MUST be a stable canonical URL for referencing any specification that
+follows this process. If the spec is made available from any other URLs, they
+SHOULD redirect to the canonical URL. If the canonical URL is changed in the
+future, all previous canonical URLs MUST remain accessible as redirects to the
+current URL.
+
+## Mutable Spec
+There is only one version of the specification and it can be updated at any time
+as long as those changes don't violate the compatibility rules identified by the
+stability model. Bug fixes, clarifications, and other non-functional spec
+updates can be merged at any time even for stable features. A change-log blog
+post SHOULD be prepared quarterly to give visibility to any changes that were
+merged in the last quarter.
+
+## The Stability Model
+Every feature in the spec has a release status. It's either evolving, stable, or
+deprecated. Flags are used to show the status a feature is in. If a feature
+doesn't have a flag, it's considered stable. If it has a year flag (such as
+`2024`) that's the year the feature reached stability. The other flags indicate
+a feature that is not yet stable or is deprecated.
+
+### Champions
+Any feature that is evolving SHOULD have a champion. The champion is responsible
+for shepherding the feature through the process to into stable status. The
+champion doesn't need to do all the work, they're just responsible for making
+sure the work gets done. The champion role MAY be shared by more than one
+person. An evolving feature without a champion SHOULD be removed from the spec.
+
+### Proposals
+New features start out as proposals. Proposals are informal GitHub Issues or
+Discussions and can be submitted by anyone.
+
+All proposals MUST be backward compatible with the current stable specification.
+Features that could be problematic for forward compatibility are highly
+discouraged, but may be considered if the feature is optional and disabled by
+default.
+
+**Promotion Checklist:**
+[] Consensus among JSON Schema Org core contributors that the proposal is worth
+  pursuing
+[] The proposal has a champion
+[] A PR for an entry for in the `STAGE-0` registry has been approved
+
+### STAGE-0
+`STAGE-0` features are tracked in the `STAGE-0` Registry on the website. This
+registry lists all `STAGE-0` features. Each entry SHOULD include the following
+checklist as well as an informal description with at least one example and any
+relevant links.
+
+This stage requires at least two implementations to implement the feature. The
+purpose of these implementations is to identify any potential implementation
+issues that might lead to low adoption such as difficulty of implementation or
+performance implications. Therefore, these implementations don't need to be
+publicly available or have real world use at this point.
+
+**Promotion Checklist:**
+[] There is general consensus among the core contributors for adding the feature
+  to the spec
+[] Tests are available in the test suite
+[] Two implementations have implemented the feature and pass the tests suite
+[] A blog post has been published introducing the feature
+[] A PR for the spec has been approved that completely specifies the feature and
+  removes the feature from the `STAGE-0` registry.
+
+### STAGE-1
+`STAGE-1` features are included to the specification, but flagged as `STAGE-1`
+and are not subject to the compatibility requirements of "stable" features. They
+can be added at any time as long as it meets all criteria for `STAGE-1` status.
+
+At this stage, the feature should be seeing real world use. These features MAY
+change or be removed all together based on what we learn from real world usage,
+but anything more than small changes should be extremely rare. Any necessary
+feature experimentation and evolution should have be done in `STAGE-0`.
+
+Implementers are encouraged to implement `STAGE-1` features, but are not
+expected to maintain support for previous versions if they change. Users who
+choose to use these features should be prepared for the possibility of changes
+and be willing to update their schemas if a change does occur.
+
+A `STAGE-1` feature can be promoted to `STAGE-2` at any time. Ideally a
+feature should stay in `STAGE-1` for about one year, but may stay longer if we
+don't see it used enough in the wild.
+
+**Promotion Checklist:**
+[] There is general consensus that the feature has been proven to be a good
+  addition to JSON Schema and is unlikely to change.
+[] We see the feature being used successfully in the wild and not generating a
+  lot of community support issues that could indicate a problem.
+
+### STAGE-2
+`STAGE-2` features are in the last stages of becoming stable. Changes are
+unlikely, but backward incompatible changes are still allowed in extreme cases.
+Users can be reasonably sure that the feature won't change, but can't expect all
+implementations to support the feature yet.
+
+`STAGE-2` is a staging area for features that are ready to be declared stable,
+but are waiting for the next release or don't yet have enough implementations.
+When we see the vast majority of actively maintained implementations support
+this feature, it will be promoted to stable in the next release. Implementers
+SHOULD support `STAGE-2` features so they aren't out-of-date when the next
+release happens.
+
+Being in `STAGE-2` isn't a guarantee that a feature will be declared stable in
+the next release. If we don't see that the vast majority of actively maintained
+implementation have implemented the feature, it may stay in `STAGE-2` another
+year to allow implementations to catch up.
+
+`STAGE-2` features can only be promoted to stable as part of a release.
+
+**Promotion Checklist:**
+[] The vast majority of actively maintained implementations support the feature
+
+### Stable
+Any part of the specification that doesn't have stability flag or has a year
+stability flag such as `2024` is considered stable. This flag indicates the year
+a feature was declared stable. Any changes to stable features MUST be backward
+compatible and not introduce problems for forward compatibility. Stable features
+are never removed, but may be deprecated.
+
+### Deprecated
+Any feature in the specification that is flagged in the form `DEPRECATED-{YEAR}`
+is considered deprecated as of the year specified. Implementations SHOULD still
+support these features to maintain backward compatibility. Deprecated features
+will never be removed from the spec, but schema authors SHOULD migrate away from
+using them as implementations may begin dropping support for these features at
+some point.
+
+A feature can only be deprecated as part of a release.
+
+## Release Process
+Releases occur on Jan 1 every year. Each release is identified with the
+four-digit year of when the release occurred. Releases never introduce anything
+new. The only change should be updating `STAGE-2` flags to `{YEAR}` flags for
+the features that are promoted to stable that year, or adding adding deprecation
+flags for any features that are deprecated that year.
+
+Implementations that express support for a particular release MUST support all
+features that are stable as of that release.
+
+A snapshot of the spec will be taken of the stable parts of the spec and made
+available as a PDF on the website at.
+
+## Meta-Schemas
+Meta-schemas associated with the specification(s) maintained under this process
+are subject to the same compatibility rules as the specification. They can be
+updated at any time, but changes MUST be backward compatible and allow for
+forward compatibility with future schemas.
+
+Meta-schema URIs SHOULD not change once published, but if they do, the old URI
+MUST remain available and redirect to the new URI.
+
+Meta-schemas are considered non-normative. It's sometimes not possible or not
+practical for the meta-schema to describe every constraint in the specification.
--- a/PROCESS.md
+++ b/PROCESS.md
@@ -11,6 +11,10 @@ This process doesn't apply to third-party dialects, vocabularies, or other
 extensions. However, third-parties are welcome to use this as a basis for the
 process they use for their specifications.
 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in RFC 2119.
+
 ## Canonical URLs
 There MUST be a stable canonical URL for referencing any specification that
 follows this process. If the spec is made available from any other URLs, they
@@ -32,17 +36,17 @@ deprecated. Flags are used to show the status a feature is in. If it has a year
 flag (such as `2024`) that's the year the feature reached stability. The other
 flags indicate a feature that is not yet stable or is deprecated.
 
-If a feature doesn't have a flag, it's considered stable. A feature can only be
-unflagged if it was stable in the initial stable release in 2023. Any functional
-change to the spec after the initial stable release MUST have a flag indicating
-it's status.
+If a feature doesn't have a flag, that means it was part of the original stable
+release and MUST be considered stable. Any functional change to the spec after
+the initial stable release MUST have a flag indicating it's status.
 
 ### Champions
 Any feature that is evolving SHOULD have a champion. The champion is responsible
 for shepherding the feature through the process to into stable status. The
 champion doesn't need to do all the work, they're just responsible for making
 sure the work gets done. The champion role MAY be shared by more than one
-person. An evolving feature without a champion SHOULD be removed from the spec.
+person or transferred to someone else if necessary. An evolving feature without
+a champion SHOULD be removed from the spec.
 
 ### Proposals
 New features start out as proposals. Proposals are informal GitHub Issues or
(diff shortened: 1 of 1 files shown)


PROCESS.md
```diff
@@ -0,0 +1,163 @@
+# JSON Schema Specification Development Process
+This document formally defines the process used to evolve the JSON Schema
+specification as of the first stable release in 2023. It applies to the core
+specification and the specifications for any dialects and vocabularies
+maintained by the JSON Schema Org. It does not necessarily apply to everything
+the JSON Schema Org maintains. For example, media type registration follows the
```
gregsdennis, line 6: Should we also say that it doesn't apply to vocabularies (like mine) or other extensions (like JSON Schema form generation) defined outside of the org?

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+stability model. Bug fixes, clarifications, and other non-functional spec
+updates can be merged at any time even for stable features. A change-log blog
+post SHOULD be prepared quarterly to give visibility to any changes that were
+merged in the last quarter.
+
+## The Stability Model
+Every feature in the spec has a release status. It's either evolving, stable, or
+deprecated. Flags are used to show the status a feature is in. If a feature
+doesn't have a flag, it's considered stable. If it has a year flag (such as
+`2024`) that's the year the feature reached stability. The other flags indicate
```
gregsdennis, line 29: If we're considering flagless features as stable, is there another indicator as to when the feature reached stability?  Or does "doesn't have a flag" only pertain to the status flag?  Do we expect all stable features to have a year flag?  If one doesn't is that just an admin error that needs to be fixed?

Maybe it's easier to require that features have flags so that if they don't it's something to be fixed.  Having an allowance for not having a flag seems to raise a lot of questions.

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+All proposals MUST be backward compatible with the current stable specification.
+Features that could be problematic for forward compatibility are highly
+discouraged, but may be considered if the feature is optional and disabled by
+default.
+
+**Promotion Checklist:**
+[] Consensus among JSON Schema Org core contributors that the proposal is worth
+  pursuing
+[] The proposal has a champion
+[] A PR for an entry for in the `STAGE-0` registry has been approved
```
gregsdennis, line 52: Grammar (maybe lose the second "for"?)

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+[] There is general consensus among the core contributors for adding the feature
+  to the spec
+[] Tests are available in the test suite
+[] Two implementations have implemented the feature and pass the tests suite
+[] A blog post has been published introducing the feature
+[] A PR for the spec has been approved that completely specifies the feature and
+  removes the feature from the `STAGE-0` registry.
+
+### STAGE-1
+`STAGE-1` features are included to the specification, but flagged as `STAGE-1`
```
gregsdennis, line 76:
```suggestion
`STAGE-1` features are included in the specification, but flagged as `STAGE-1`
```

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+### Proposals
+New features start out as proposals. Proposals are informal GitHub Issues or
+Discussions and can be submitted by anyone.
+
+All proposals MUST be backward compatible with the current stable specification.
+Features that could be problematic for forward compatibility are highly
+discouraged, but may be considered if the feature is optional and disabled by
+default.
+
+**Promotion Checklist:**
```
gregsdennis, line 48: I assume these checklists are for promotion _out of_ this stage to the next one?

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+
+### STAGE-1
+`STAGE-1` features are included to the specification, but flagged as `STAGE-1`
+and are not subject to the compatibility requirements of "stable" features. They
+can be added at any time as long as it meets all criteria for `STAGE-1` status.
+
+At this stage, the feature should be seeing real world use. These features MAY
+change or be removed all together based on what we learn from real world usage,
+but anything more than small changes should be extremely rare. Any necessary
+feature experimentation and evolution should have be done in `STAGE-0`.
```
gregsdennis, line 83:
```suggestion
feature experimentation and evolution should have been done in `STAGE-0`.
```

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+[] A blog post has been published introducing the feature
+[] A PR for the spec has been approved that completely specifies the feature and
+  removes the feature from the `STAGE-0` registry.
+
+### STAGE-1
+`STAGE-1` features are included to the specification, but flagged as `STAGE-1`
+and are not subject to the compatibility requirements of "stable" features. They
+can be added at any time as long as it meets all criteria for `STAGE-1` status.
+
+At this stage, the feature should be seeing real world use. These features MAY
```
gregsdennis, line 80: I'm curious about how we will guage/measure this real world use.  Wouldn't we be reliant solely on volunteered feedback?

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+
+### STAGE-2
+`STAGE-2` features are in the last stages of becoming stable. Changes are
+unlikely, but backward incompatible changes are still allowed in extreme cases.
+Users can be reasonably sure that the feature won't change, but can't expect all
+implementations to support the feature yet.
+
+`STAGE-2` is a staging area for features that are ready to be declared stable,
+but are waiting for the next release or don't yet have enough implementations.
+When we see the vast majority of actively maintained implementations support
```
gregsdennis, line 108: > the vast majority

Do we need to define this more rigidly?

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+four-digit year of when the release occurred. Releases never introduce anything
+new. The only change should be updating `STAGE-2` flags to `{YEAR}` flags for
+the features that are promoted to stable that year, or adding adding deprecation
+flags for any features that are deprecated that year.
+
+Implementations that express support for a particular release MUST support all
+features that are stable as of that release.
+
+A snapshot of the spec will be taken of the stable parts of the spec and made
+available as a PDF on the website at.
```
gregsdennis, line 151:
```suggestion
available as a PDF on the website at <placeholder>.
```

... if you intend to fill the blank later.

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+## Mutable Spec
+There is only one version of the specification and it can be updated at any time
+as long as those changes don't violate the compatibility rules identified by the
+stability model. Bug fixes, clarifications, and other non-functional spec
+updates can be merged at any time even for stable features. A change-log blog
+post SHOULD be prepared quarterly to give visibility to any changes that were
+merged in the last quarter.
+
+## The Stability Model
+Every feature in the spec has a release status. It's either evolving, stable, or
```
gregsdennis, line 26: Do we want to avoid contractions like we do in specs or do we think less formal language is okay for this doc?

gregsdennis (changes requested):
We mention "JSON Schema Org core contributors" in several places.  Do we have/need a process to determine on- and offboarding of people into this role?

Also, a lot of my questions may be candidates for issues.  Feel free to indicate which ones you'd like to defer.

PROCESS.md
jdesrosiers: I thought I had covered that by explicitly stating what this process applies to (specs maintained by the JSON Schema Org), but technically it doesn't say that's the only thing it applies to. I'll edit to make that less ambiguous.

PROCESS.md
jdesrosiers: Are we sure using contractions in specs is discouraged? I don't think I've ever heard that. I think the language should be as natural as possible without being ambiguous. Using natural language wherever possible makes specs easier for readers to read and understand.

PROCESS.md
jdesrosiers: A part of the spec can only be unflagged if it's a stable feature that was introduced in the original stable release.

My thinking was that it would be too noisy to flag everything in the spec. It would be hard to see `2024` flags in a sea of `2023` flags. The first stable release wouldn't have any `2023` flags because it's unambiguous when those features were introduced since `2023` is the first release. Stable flags would start to be added for `2024`. That means that the number of flags in the spec should be fairly small and therefore stand out to readers.

I'll revise to clarify the purpose of unflagged features, but let's continue the discussion and get some more opinions if you still dislike the idea.

PROCESS.md
jdesrosiers: Correct. Any suggestions to make that more clear?

PROCESS.md
```diff
@@ -0,0 +1,163 @@
...
+A snapshot of the spec will be taken of the stable parts of the spec and made
+available as a PDF on the website at.
+
+## Meta-Schemas
+Meta-schemas associated with the specification(s) maintained under this process
+are subject to the same compatibility rules as the specification. They can be
+updated at any time, but changes MUST be backward compatible and allow for
+forward compatibility with future schemas.
+
+Meta-schema URIs SHOULD not change once published, but if they do, the old URI
```
Julian, line 159:
```suggestion
Meta-schema URIs SHOULD NOT change once published, but if they do, the old URI
```

PROCESS.md
jdesrosiers: Maybe? Assuming bowtie grows to include most implementations, this decision could be made based on hard data rather than gut feel or a time consuming audit of all known implementations.

PROCESS.md
jdesrosiers: Oops. Yeah, I had a placeholder originally and decided to remove it. I forgot to remove the "at" as well.

PROCESS.md
gregsdennis: Maybe we keep the "unflagged implies stable" idea, and instead of flagging the year for future features, we have a `newly-stable` flag that identifies features that are new for the year's release.  Then the next year, those `newly-stable` flags are cleared from those features so that the new `newly-stable` features stand out.

PROCESS.md
gregsdennis: Maybe explicitly say "Checklist for promotion to stage-X"

PROCESS.md
gregsdennis: I think in general, maybe.  But that may also be skewed toward more popular implementations that may or may not support a feature.

I like your answer below about including Bowtie and extracting metrics from there.  It shows support for a feature, but not necessarily usage.

PROCESS.md
gregsdennis: Your changes look fine.

PROCESS.md
gregsdennis: Just scanning the spec for `'s`s, it's predominantly possessives.  There are a couple `it's`s and `let's` (and even a `what's`), though.

I personally tend to steer clear of contractions in formal writing, but it doesn't appear that we as a group have been sufficiently consistent in our productions to say we have a preference.

PROCESS.md
gregsdennis: I don't think this is something that we have to resolve now.

gregsdennis (approved):
The changes that have been made are good.  I don't think the remaining items warrant holding this up.

PROCESS.md
jdesrosiers: Let's discuss it at the next OCWM

PROCESS.md
jdesrosiers: This is a tough one. I think it will always be largely a judgement call, but we should definitely be looking for more and better ways we can gather data we can use to make those calls.

Relequestual (changes requested):
Left some comments and questions.

Some parts feel under specified, which is fine for a draft, but I feel like there are some key issues to iron out.

Additionally, I feel like we should be stating...
"This policy document uses RFC2119 defined key words to indicate requirement levels" =]

Bulk of this looks like a great first start though. Thanks for all your work on this!

gregsdennis:
@Relequestual have a read through the comments that I made and the responses given. Maybe that will answer some of yours.

PROCESS.md
```diff
@@ -0,0 +1,171 @@
...
+## The Stability Model
+Every feature in the spec has a release status. It's either evolving, stable, or
+deprecated. Flags are used to show the status a feature is in. If it has a year
+flag (such as `2024`) that's the year the feature reached stability. The other
+flags indicate a feature that is not yet stable or is deprecated.
+
+If a feature doesn't have a flag, it's considered stable. A feature can only be
+unflagged if it was stable in the initial stable release in 2023. Any functional
+change to the spec after the initial stable release MUST have a flag indicating
+it's status.
```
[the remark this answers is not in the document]
jdesrosiers: You quoted the answer immediately before asking the question, so I'm not sure how to answer.

> A feature can only be unflagged if it was stable in the initial stable release in 2023.

I can expand on that if it's not clear enough, but I'm not sure what's not clear. Does this discussion with @gregsdennis help? https://github.com/json-schema-org/json-schema-spec/pull/1348#discussion_r1025595948

PROCESS.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Champions
+Any feature that is evolving SHOULD have a champion. The champion is responsible
+for shepherding the feature through the process to into stable status. The
+champion doesn't need to do all the work, they're just responsible for making
+sure the work gets done. The champion role MAY be shared by more than one
+person. An evolving feature without a champion SHOULD be removed from the spec.
```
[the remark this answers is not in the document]
jdesrosiers: I fully expect champions to disappear on occasion. In that case, a new champion would be named or the feature would be abandoned and removed. The "SHOULD" is to give us some leeway if for some reason we can't or don't want to assign a champion right away, but also don't want to remove the feature. It's not a likely scenario. I'm fine with calling it a MUST if we want to.

PROCESS.md
```diff
@@ -0,0 +1,171 @@
...
+choose to use these features should be prepared for the possibility of changes
+and be willing to update their schemas if a change does occur.
+
+A `STAGE-1` feature can be promoted to `STAGE-2` at any time. Ideally a
+feature should stay in `STAGE-1` for about one year, but may stay longer if we
+don't see it used enough in the wild.
+
+**Promotion to STAGE-2 Checklist:**
+[] There is general consensus that the feature has been proven to be a good
+  addition to JSON Schema and is unlikely to change.
```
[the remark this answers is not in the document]
jdesrosiers: Whoever the JSON Schema Org active contributors happen to be at the time. Maybe at some point that will be a specific and official group of people like a TSC, but for now it's squishy concept. For now I'll make the text more specific. We can update the wording when/if we have an official group we can name.

PROCESS.md
```diff
@@ -0,0 +1,171 @@
...
+[] We see the feature being used successfully in the wild and not generating a
```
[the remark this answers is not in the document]
jdesrosiers: See, https://github.com/json-schema-org/json-schema-spec/pull/1348#discussion_r1028497441

PROCESS.md
```diff
@@ -0,0 +1,171 @@
...
+
+Being in `STAGE-2` isn't a guarantee that a feature will be declared stable in
+the next release. If we don't see that the vast majority of actively maintained
+implementation have implemented the feature, it may stay in `STAGE-2` another
+year to allow implementations to catch up.
+
+`STAGE-2` features can only be promoted to stable as part of a release.
+
+**Promotion to Stable Checklist:**
+[] The vast majority of actively maintained implementations support the feature
```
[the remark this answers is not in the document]
jdesrosiers: > "vast majority"

see https://github.com/json-schema-org/json-schema-spec/pull/1348#discussion_r1024614138

> "actively maintained"

Yep. The more I think about it, the more I agree that this needs more definition. For example, an implementation may be active for bug fixes, but don't intent to add new features. Should that count? This probably needs discussion and probably doesn't need to be decided now as it will be a while before we have to make this determination.

> "support the feature"

The meaning of those words seem pretty clear to me. We could explicitly say it has to pass all the tests. Is there more nuance that I'm missing?

> I think we should investigate further sponsoring implementation work at some point

Sure. I think that speaks to increasing the pool of "actively maintained" implementations.

PROCESS.md
```diff
@@ -0,0 +1,171 @@
...
+features that are stable as of that release.
+
+A snapshot of the spec will be taken of the stable parts of the spec and made
+available as a PDF on the website.
+
+## Meta-Schemas
+Meta-schemas associated with the specification(s) maintained under this process
+are subject to the same compatibility rules as the specification. They can be
+updated at any time, but changes MUST be backward compatible and allow for
+forward compatibility with future schemas.
```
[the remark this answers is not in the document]
jdesrosiers: Just like the spec, there is one set of mutable meta-schemas that describe the current state of the specification. Because of backward and forward compatibility constraints, no matter what release of the spec the schema is written for, the latest meta-schemas is all you need.

> Are we going to maintain meta-schemas attached to individual release snapshots?

We could have snapshots for reference like with the spec, but there wouldn't be reason to actually use those snapshots for anything other than historical reference.

> Are we going to have a "current" meta-schema which contains all keywords, even stage-0?
> Are we going to have a "current" meta-schema which per stage? For example, all stage-1 and above or all stage-2 and above?

The meta-schemas should include anything that's in the spec. That includes `STAGE-1`, `STAGE-2`, and deprecated.

> If we have a "current" meta-schema, which includes all of the stages, how can we update meta-schemas yet maintain backwards compatiblity?

Meta-schema changes should be backward compatible because spec changes are backward compatible.

> This becomes a very specific problem if we disallow unknown keywords by default.
>
> What if we wanted to update the meta-schemas to use a new keyword? That wouldn't be backwards compatible.

I was thinking that the standard dialect meta-schema would allow additional keywords. If a keyword appears in the meta-schema but is not supported by the implementation, the implementation should raise an error. It's not something the meta-schema can know, so I consider it a runtime constraint the implementation needs to deal with, not the meta-schema.

However, meta-schemas could disallow additional keywords and it shouldn't hurt anything. If an implementation is using an older version of the meta-schema that doesn't have a newer keyword used by the schema, then the implementation also wouldn't have support for the keyword. The result is always that you get an error that the implementation doesn't know the keyword.

Either way, there shouldn't be a backward or forward compatibility issue and whether we choose to allow additional keywords in the meta-schema is not relevant to process, so we can discuss that when we get to writing the meta-schemas.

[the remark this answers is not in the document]
jdesrosiers: I expect there to be some discussion about meta-schema lifecycle that may end up as a bigger conversation. This was the kind of thing I thought might be best to get into after the initial merge. Like I said, I don't expect merging this document to mean it's "done". I expect it to be a first draft that we will iterate on.

PROCESS.md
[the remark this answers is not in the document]
Relequestual: I'd rather call it a MUST. As for remaining in the spec, I think that would need to be a case by case basis.

PROCESS.md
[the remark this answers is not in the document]
Relequestual: There are a number of assertions in your response which I don't agree with, but I don't think here's the right place to capture this discussion.

>I expect there to be some discussion about meta-schema lifecycle that may end up as a bigger conversation. This was the kind of thing I thought might be best to get into after the initial merge. Like I said, I don't expect merging this document to mean it's "done". I expect it to be a first draft that we will iterate on.

I think we need to discuss how we can break up the SDLC into more discreet chunks for further discussion.

PROCESS.md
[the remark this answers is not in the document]
jdesrosiers: I see. I'll try to reword to avoid that misunderstanding. Removing or changing flags isn't disallowed, but compatibility rules apply. So, you couldn't remove a stable flag (such as `2024`) because it would be a backward compatible change, but you could change a `STAGE-2` to a `2024` because `STAGE-2` is allowed backward-incompatible changes. Similarly, you couldn't remove a `STAGE-2` flag without replacing it with a stable flag because that would break backward compatibility.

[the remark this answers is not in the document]
jdesrosiers: Here's an example of the kind of thing I have in mind from the node.js docs https://nodejs.org/dist/latest-v19.x/docs/api/fs.html. They have stable, unstable, and deprecated features and flag them inline just like we're talking about here.

PROCESS.md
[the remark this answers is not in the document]
jdesrosiers: That's exactly why I proposed merging this as a work-in-progress and opening issues to discuss the parts that need more discussion. Holding this up until we decide every detail makes it more difficult to have those discussions.

Relequestual:
@jdesrosiers Can this be closed in favour of https://github.com/json-schema-org/json-schema-spec/pull/1500 ?

jdesrosiers:
Yes, it can

PR #1353, opened 2022-11-21
Put boundaries on lack of "$schema" behavior

### LINKED ISSUE #1315: Document intended existing constraints on "$schema" handling
This issue is _not_ intended to make fundamental changes, just address the overly-broad "implementation-defined" behavior by excluding obviously problematic behaviors.  Doing this does not preclude further changes to `$schema` and/or `$vocabulary` — I'm just filing it to ensure that if we do not otherwise change `$schema`, it will be more clear in the next release.

My intention was that, in the absence of `$schema`, or if `$schema` is not recognized and not available (since we already have clear guidance for how to handle and available `$schema` with and without a core vocabulary declared), implementations MUST do one of the following:

* Refuse to process the schema, e.g. raise an error rather than returning a pass or fail result
* Assume a specific documented dialect meta-schema URI/IRI, which can be for either a standard or custom meta-schema, but can't just be "the latest" without qualification as that could change
* Document the process by which the implementation determines its behavior, which MUST result in behavior consistent with one of however-many specific dialect meta-schema URIs/IRIs the implementation wants to consider, e.g. "if `$recursive*` or the array form of `items` is present, then the 2019-09 standard default dialect is assumed, otherwise the 2020-12 standard default dialect is assumed."

This would avoid the appearance that an implementation can proceed with completely arbitrary behavior, which was never the intent.

This issue is _not_ intended to address what else we may or may not want `$schema` to do, which is being discussed in numerous other places.  The above set of options is based on a survey of implementations and how they handle such things.

handrews:
Fixes #1315 

This may well change prior to the next release, but documents the intended range of options so as to avoid crashes or completely arbitrary behavior.  I believe it covers the observed behavior of implementations in the wild, so even though there is a new MUST I don't think it would cause a problem.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1228,7 +1228,15 @@
                         The "$schema" keyword SHOULD be used in the document root schema object,
                         and MAY be used in the root schema objects of embedded schema resources.
                         It MUST NOT appear in non-resource root schema objects.  If absent from
-                        the document root schema, the resulting behavior is implementation-defined.
+                        the document root schema, the resulting behavior is implementation-defined,
+                        but MUST fall within the following options:
+                        <ul>
+                            <li>Refuse to process the schema, as with unsupported required
+                                vocabularies</li>
+                            <li>Assume a specific, documented meta-schema</li>
+                            <li>Document the process by which it examines the schema and determines
+                                which of a specific set of meta-schemas to assume</li>
+                        </ul>
                     </t>
                     <t>
                         Values for this property are defined elsewhere in this and other documents,
@@ -3547,9 +3555,9 @@ https://example.com/schemas/common#/$defs/allOf/1
             <t>
                 Instances and schemas are both frequently written by untrusted third parties, to be
                 deployed on public Internet servers.
-                Validators should take care that the parsing and validating against schemas does not consume excessive
-                system resources.
-                Validators MUST NOT fall into an infinite loop.
+                Implementations should take care that the parsing and evaluating against schemas
+                does not consume excessive system resources.
+                Implementations MUST NOT fall into an infinite loop.
             </t>
             <t>
                 A malicious party could cause an implementation to repeatedly collect a copy


jsonschema-core.xml
```diff
@@ -1228,7 +1228,14 @@
                         The "$schema" keyword SHOULD be used in the document root schema object,
                         and MAY be used in the root schema objects of embedded schema resources.
                         It MUST NOT appear in non-resource root schema objects.  If absent from
-                        the document root schema, the resulting behavior is implementation-defined.
+                        the document root schema, the resulting behavior is implementation-defined,
+                        but MUST fall within the following options:
+                        <ul>
+                            <li>Refuse to process the schema, as with unsupported required vocabularies
+                            <li>Assume a specific, documented meta-schema
```
Julian, line 1235: I'll just point out that under this language, it is still valid behavior to proceed with validation and consider all instances *valid*, or all instances *invalid* (because I can document my specific documented meta-schema is `{}` or an equivalent of `{"not": {}}`).

I don't see that as an issue personally, I think there are reasonable reasons that may make sense, but will mention it explicitly since I know there was argument about those cases and whether they're desired behavior or not.

handrews: Yeah, the main issue is that we can't, in the core spec, restrict the options to just the standard meta-schemas.  There are perfectly good reasons to default to something non-standard (e.g. OpenAPI's dialect).  There are endless ways to write always-passing and always-failing schemas, so trying to forbid those feels like a bit of a lost cause.

Requiring that one document the meta-schema should at least highlight the possibility, and I suspect people would feel a bit silly documenting `{}` as their default meta-schema.  But at least it would be clear.

gregsdennis (approved):

jdesrosiers (approved):
I expect this to change a bit as we prepare for the next release, but this is good for now.

jsonschema-core.xml
```diff
@@ -3547,9 +3555,9 @@ https://example.com/schemas/common#/$defs/allOf/1
             <t>
                 Instances and schemas are both frequently written by untrusted third parties, to be
                 deployed on public Internet servers.
-                Validators should take care that the parsing and validating against schemas does not consume excessive
-                system resources.
-                Validators MUST NOT fall into an infinite loop.
+                Implementations should take care that the parsing and evaluating against schemas
```
gregsdennis, line 3558: Maybe lose "against"

gregsdennis (approved):

Issue #1335, opened 2022-10-26
Convert the spec to Markdown


gregsdennis:
As we are no longer publishing with IETF, we don't need to use their tools.  Going forward, we'd like to use markdown to author the spec instead of the current XML.

The first tasks are:

- [ ] decide markdown flavor and toolset
- [ ] convert the existing document

jdesrosiers:
I have an idea about how hopefully make this process a bit less painful that it sounds. I believe the IETF tooling has the ability to convert from XML style to Markdown style. It's an IETF specific markdown flavor, but it should be easier to convert from one Markdown flavor to another than it would be to convert from XML to Markdown.

gregsdennis:
We now have two approaches: #1357 and #1412.

I thought it would be wise to enumerate the document features we need/want.  Then we can decide which of these approaches best captures that.

Looking at the current (IETF) published spec in [HTML form](https://json-schema.org/draft/2020-12/json-schema-core.html#name-json-schema-documents), here's what's used:

- Headers (automatic numbering)
- Paragraph styles
  - Normal
  - Code blocks (no syntax highlighting)
- Inline comments (CREFs)
- Links (internal & external)
- Ordered & unordered (both plain and bulleted) lists
- Table of Contents (automatic)

On addition to the above, [JSON Path](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-14.html), which uses IETF Kramdown (as proposed in #1412), also has these:

- Text styles
  - italic
  - code (monospace)
- Paragraph styles
  - Code blocks (with captions, no syntax highlighting)
- Tables (with captions)
- Table of Contents (automatic)

In addition to the features our current spec uses, [OpenAPI](https://spec.openapis.org/oas/v3.1.0), which [appears](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md?plain=1) to use simple Markdown, also uses:

- Text styles
  - bold
  - italic
- Paragraph styles
  - Code blocks (with syntax highlighting)
- Tables (no captions)
- Table of Contents (written in the markdown, but rendered as side bar)

In addition to all of the above, do we also want (or could we make use of):

- Text styles
  - underline (not supported directly by markdown, but we can use `<u>` tags)
  - strikethrough?
- Paragraph styles
  - Block quotes
  - Comments? (see https://chirpy.cotes.page/posts/text-and-typography/#prompts for some examples of this idea)
- Images? (with captions?)

---

Lastly,

[...]

PR #1357, opened 2022-11-28
Convert the spec to Markdown

yakimun:
#1335

In this PR, I've tried to start working on the spec conversion from XML to Markdown (currently only the Core document).

I couldn't find a working solution quickly, so I wrote a small program that does most of the conversion. Things like references are a bit more complicated, so it was faster to convert them manually.

The Markdown flawor used here is GFM, but it could be changed quite easily.
GFM has tools for rendering to HTML, so it can be used to publish a specification to a website.

Not all elements from IETF XML can be presented as-is in GFM. So there are rules that I used in the conversion process:
| XML | GFM |
|------|-------|
| `<title>` | 1st-level header |
| `<abstract>`, `<note>` (from `<front>` block) | block with 2nd-level header |
| `<section>` | block with n-level header (where n is nesting level) |
| `<t>`, `<preamble>`, `<postamble>` | new paragraph |
| `<list>` | `-` or `1.` list |
| `<eref>` | plain URL (will be rendered as link) |
| `<xref>` | `[text](target)` |
| `<cref>` | footnote |
| `<artwork>` | code block |
| `<spanx>`, `<sourcecode>` | inline code block |

Also, all JSON code blocks are marked as `json`, so syntax highlighting is applied for them.
JSON was previously formatted inconsistently, so this has been fixed as well (JSON formatting rules can easily be changed in the future).

### Notes
* `<list style="hanging">` - don't see any option to represent it in GFM, so for now converting it as `- *hangText* text`
* `<cref>` - converted as [footnote](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes), but [quote](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-text) seems like a good option too.
* `<preamble>` and `<postamble>` - in IETF XML these elements are connected to `<artwork>`, but in MD they are just new paragraphs.
* `<spanx>`, `<sourcecode>` - I don't see any difference between them in the XML spec, so both elements represented as inline code blocks.

### Questions:
* Is GFM a good choice?
* Block before the title (containing "workgroup", "published", "expires", etc.), "Status of This Memo" and "Copyright Notice" blocks are not presented in the current version of MD spec. Should them be added?
* The links generated by `<xref>` element are rendered as `RFC 1111 [RFC1111]` and `Some header (Section 1.2.3)`. Is it necessary to follow these rendering rules in MD?

### Problems
* References (links) require more attention than in IETF XML (but a good editor makes this problem less significant).
* References (block) generation automated in IETF XML. In MD it's done manually.
* I have absolutely no idea how to validate this. The only option I see is to read the entire document very carefully.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -225,8 +225,7 @@ Extension keywords, meaning those defined outside of this document
 and its companions, are free to define other behaviors as well.
 
 A JSON Schema MAY contain properties which are not schema keywords or are not recognized as schema keywords.
-The behavior of such keywords is governed by section
-[6.5.1](#651-handling-of-unrecognized-or-unsupported-keywords).
+The behavior of such keywords is governed by [Section 6.5.1](#651-handling-of-unrecognized-or-unsupported-keywords).
 
 An empty schema is a JSON Schema with no properties.
 
@@ -297,8 +296,7 @@ with fragments are considered to be non-canonical.
 
 The root schema is the schema that comprises the entire JSON document
 in question. The root schema is always a schema resource, where the
-IRI is determined as described in section
-[9.1.1](#911-initial-base-iri).[^1]
+IRI is determined as described in [Section 9.1.1](#911-initial-base-iri).[^1]
 [^1]: Note that documents that embed schemas in another format will not
 have a root schema resource in this sense. Exactly how such usages
 fit with the JSON Schema document and resource concepts will be
@@ -320,8 +318,7 @@ and the schema titled "root" is the root schema.
 
 As with the root schema, a subschema is either an object or a boolean.
 
-As discussed in section
-[8.2.1](#821-the--id--keyword), a JSON Schema document
+As discussed in [Section 8.2.1](#821-the--id--keyword), a JSON Schema document
 can contain multiple JSON Schema resources. When used without qualification,
 the term "root schema" refers to the document's root schema. In some
 cases, resource root schemas are discussed. A resource's root schema
@@ -467,10 +464,9 @@ applicators (and therefore no subschemas) is reached. The appropriate
 location in the instance is evaluated against the assertion and
 annotation keywords in the schema object. The interactions of those
 keyword results to produce the schema object results are governed by
-section [7.7.1.2](#7712-annotations-and-assertions), while the
+[Section 7.7.1.2](#7712-annotations-and-assertions), while the
 relationship of subschema results to the results of the applicator
-keyword that applied them is described by section
-[7.5](#75-applicators).
+keyword that applied them is described by [Section 7.5](#75-applicators).
 
 Evaluation of a parent schema object can complete once all of its
 subschemas have been evaluated, although in some circumstances evaluation
@@ -607,8 +603,7 @@ to the assertion results of subschemas, but MUST NOT introduce new
 assertion conditions of their own.
 
 [Annotation](#77-annotations) results from subschemas
-are preserved in accordance with section
-[7.7.1](#771-collecting-annotations) so that applications
+are preserved in accordance with [Section 7.7.1](#771-collecting-annotations) so that applications
 can decide how to interpret multiple values. Applicator keywords
 do not play a direct role in this preservation.
 
@@ -847,8 +842,8 @@ Their purpose in the core vocabulary is to ensure that locations are
 available for certain purposes and will not be redefined by extension
 keywords.
 
-While these keywords do not directly affect results, as explained in section
-[9.4.2](#942-references-to-possible-non-schemas) unrecognized
+While these keywords do not directly affect results, as explained in
+[Section 9.4.2](#942-references-to-possible-non-schemas) unrecognized
 extension keywords that reserve locations for re-usable schemas may have
 undesirable interactions with references in certain circumstances.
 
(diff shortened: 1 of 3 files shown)


handrews:
This is a bit premature as there are a number of discussions going on right now about how the next version of the spec should be structured.  Those issues marked "SDLC" are not yet ready for PRs.

handrews:
@yakimun BTW a lot of the currently open issues are not ready for PR.  We welcome PRs from new contributors but you probably want to confirm that an issue is ready before working on one.  I know not all issues have a discussion about PR-readiness but that's usually because the longer-term contributors have some other context for deciding it's ready (that we should probably be better about documenting).

jdesrosiers:
@yakimun Thanks for doing this! I haven't reviewed the result in detail, but it looks like a good start and gets us over the biggest hurdle.

> there are a number of discussions going on right now about how the next version of the spec should be structured.

This shouldn't matter. The first step is just to convert what we have to markdown. Any restructuring that needs to be done should be done as a separate step.

The thing that I think makes this a bit premature is that we haven't yet talked about how we are going to make this transition. We'll probably want to to do a code freeze on the spec, then convert to markdown and delete XML, then continue work on the markdown.

Have there been updates to the spec since you did this conversion?

How long would it take to re-convert given the script you've developed? In other words, would it be reasonable for us to ask you to do the conversion again in a month or two if we the team decides not to code freeze and switch over just yet?

> Is GFM a good choice?

Yep. We don't know exactly what our needs will be for presenting on the website, so we might need to change at some point, but GFM is a good place to start.

> Block before the title (containing "workgroup", "published", "expires", etc.), "Status of This Memo" and "Copyright Notice" blocks are not presented in the current version of MD spec. Should them be added?

I think that's all just IETF stuff and it's ok to leave it out.

> The links generated by <xref> element are rendered as RFC 1111 [RFC1111] and Some header (Section 1.2.3). Is it necessary to follow these rendering rules in MD?

No. We can render it however we want.

> I have absolutely no idea how to validate this. The only option I see is to read the entire document very carefully.

Agreed. It's just going to need eyes on it. We're going to need to go through everything very closely for the next release, so I don't think it's terribly important right now to catch every broken link or visual issue that might be lurking.

handrews:
It should be easier to determine the restructuring and convert each piece in stages rather than do it all right now, which is a big change that is difficult to review.

jdesrosiers:
@handrews I strongly disagree. Let's discuss more a the OCWM.

yakimun:
@jdesrosiers 

> Have there been updates to the spec since you did this conversion?

I've used the last version (https://github.com/json-schema-org/json-schema-spec/blob/38bc78a0ba5c05e6f5c20cb93bd9f3e1c8a9ba0b/jsonschema-core.xml).

It took a couple of hours to manually fix the MD after automatic conversion. If the XML spec changes a lot in the future, then it's not a big problem to repeat the conversion process from the scratch. If the changes in the XML are minor, then I would just copy them to the MD.

> No. We can render it however we want.

So what do you think about removing the second part of the link? (`Some link text [SomeTarget]` -> `Some link text`). From my point of view, it makes everything cleaner.

jdesrosiers:
> it's not a big problem to repeat the conversion process from the scratch.

That's good to hear. We'll try to avoid having to do that, but it's good to know it's an option if we need it. It might be a little while before we have agreement to merge this, so don't be surprised if there's no activity here for a few weeks. It doesn't mean we've forgotten about it.

> So what do you think about removing the second part of the link?

I agree. Go ahead and remove it.

awwright:
Like @handrews said, there’s still many details to figure out before we can actually start editing the specification in Markdown. It's a bit like changing the side of the road that a whole country drives on. I have a couple of different solutions I'm evaluating myself, and we'll want to review these too. Can you share the script you’re using instead?

handrews:
Yes, I'm on board with changing over now-ish - things changed soon after that first comment of mine.  Although comparing a few script solutions as @awwright seems reasonable as long as we don't take too long on it.  If there's a script that's really well road-tested already that would mean less work verifying correctness.

The exact presentation of the new markdown is less important as we can change that easily in the future with less disruption than the xml->markdown change.

yakimun:
@awwright https://github.com/yakimun/json-schema-spec-converter

This solution can now automatically generate links, section numbers and other things. So, the output requires almost no manual fixes (autoformatting with any good tool is a good next step though).

The solution quickly gets messy, but that shouldn't be a problem for such a short-lived project.

Some tags conversion produces not perfect output in terms of formatting. For example, there are extra empty lines, and the lists with multiline elements are not padded nicely enough.
But it's easy to fix with any autoformatter. Also, it doesn't affect anything when spec is rendered.

yakimun:
I've added all specs converted to MD. Also, everything has been updated with changes from last commits. I hope this helps.

awwright:
I've been able to compile and run the [converter](https://github.com/yakimun/json-schema-spec-converter), so I'm going to see how many PRs we can merge in or close, before I run this.

Since I'll have to re-run the converter after these are merged in, I'll close this out, and continue discussion on #1335.

PR #1359, opened 2022-12-06
Plain name IRI fragments can use unicode

### LINKED ISSUE #1314: Expand plain name fragments to unicode / IRIs
Currently, our plain name fragments are restricted to the US-ASCII subset of the relevant XML plain name fragment syntax.  Now that we support IRIs should we expand this as well?  I believe the actual XML syntax we reference covers a wider set of unicode.  And perhaps we wouldn't need to spell out the syntax in our spec if we use the XML syntax exactly.

handrews:
Fixes #1314 as discussed in yesterday's community call.  This just brings plain name fragment support into parity with the rest of our IRI support, and continues to rely on XML's NCName rule for the syntax.  NCName always supported unicode, we just excluded all of it except for the US-ASCII characters.

I *think* I converted the W3C's grammar notation into ABNF properly, but if anyone is comfortable with ABNF (@awwright ?) a check would be appreciated.  The W3C's grammar involves [NCName, NCNameChar, and NCStartNameChar](https://www.w3.org/TR/2006/REC-xml-names11-20060816/#NT-NCName), which are just [Name, NameChar, and StartNameChar](https://www.w3.org/TR/xml11/#NT-NameStartChar) minus the `:` character.  I simplified the ABNF since we did not need the `:` variation at all.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -516,13 +516,34 @@
                 locally named schemas.
             </t>
             <t>
-                Plain name fragments MUST start with a letter ([A-Za-z]) or underscore ("_"),
-                followed by any number of letters, digits ([0-9]), hyphens ("-"),
-                underscores ("_"), and periods (".").  This matches the US-ASCII part of XML's
+                Plain name fragments MUST follow XML's 
                 <xref target="xml-names">NCName production</xref>, which allows for compatibility
                 with the recommended plain name <xref target="W3C.REC-xptr-framework-20030325">syntax</xref> for
-                XML-based media types.
+                XML-based media types.  For convenience, the NCName syntax is reproduced here in ABNF form, using
+                a minimal set of rules:
+                <cref>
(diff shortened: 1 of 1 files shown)


awwright:
@handrews Like this:

```abnf
NCNameStartChar = "_" / ALPHA
                / %xC0-D6 / %xD8-F6 / %xF8-2FF
                / %x370-37D / %x37F-1FFF
                / %x200C-200D / %x2070-218F
                / %x2C00-2FEF / %x3001-D7FF
                / %xF900-FDCF / %xFDF0-FFFD
                / %x10000-EFFFF

NCNameChar      = NCNameStartChar / "-" / "." / DIGIT
                / %xB7 / %x0300-036F / %x203F-2040

NCName          = NCNameStartChar *NCNameChar
```

However if we're not going to use ABNF for making a normative requirement, then maybe it'd be better to just give the regular expression. Or both.

gregsdennis:
Replaced by ☝️

PR #1360, opened 2022-12-07
Add all changes since last release to Change Log

handrews:
As promised in Monday's call, here is the change log i had researched covering all changes to date (plus PR #1359 which was agreed to in principle on the call as well).

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -4295,10 +4295,34 @@ https://example.com/schemas/common#/$defs/allOf/1
                 <list style="hanging">
                     <t hangText="draft-bhutton-json-schema-next">
                         <list style="symbols">
-                            <t>"contains" now applies to objects as well as arrays</t>
-                            <t>Use IRIs instead of URIs</t>
+                            <t>Use IRIs instead of URIs, including allowing unicode in plain-name fragments</t>
+                            <t>Clarify that detecting duplicate IRIs for different schemas SHOULD raise an error</t>
+                            <t>Consolidate and clarify the syntax and rationale for plain-name fragments</t>
+                            <t>"$id" MUST be an absolute-IRI, without any fragment, even an empty one</t>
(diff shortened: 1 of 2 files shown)


jdesrosiers (commented):
That's impressively thorough! I don't think clarifications need a change log entry. I'd just include things that effect behavior.

jsonschema-core.xml
```diff
@@ -4295,10 +4295,34 @@ https://example.com/schemas/common#/$defs/allOf/1
...
                     <t hangText="draft-bhutton-json-schema-next">
                         <list style="symbols">
-                            <t>"contains" now applies to objects as well as arrays</t>
-                            <t>Use IRIs instead of URIs</t>
+                            <t>Use IRIs instead of URIs, including allowing unicode in plain-name fragments</t>
+                            <t>Clarify that detecting duplicate IRIs for different schemas SHOULD raise an error</t>
+                            <t>Consolidate and clarify the syntax and rationale for plain-name fragments</t>
+                            <t>"$id" MUST be an absolute-IRI, without any fragment, even an empty one</t>
+                            <t>Note that an empty string "$id" results in duplicate IRIs for different schemas</t>
+                            <t>Define empty schemas as empty (no longer allowing unrecognized keywords</t>
```
Julian, line 4303:
```suggestion
                            <t>Define empty schemas as empty (no longer allowing unrecognized keywords)</t>
```

Julian:
> That's impressively thorough! I don't think clarifications need a change log entry. I'd just include things that effect behavior.

I agree though another riff on perhaps the same underlying suggestion is attempting to sort the list by some imprecise notion of "likelihood someone cares to read about it", which should be doable. E.g. `contains` changes are probably top of the list, and all the clarifying stuff likely falls near the bottom, if it isn't in fact cut out entirely.

gregsdennis:
Replaced by ☝️

PR #1361, opened 2022-12-07
Fix accidental ABNF omission for Relative JSON Pointer

### LINKED ISSUE #1175: Relative JSON Pointer specification inconsistency
First issue: text part of section [3. Syntax](https://datatracker.ietf.org/doc/html/draft-bhutton-relative-json-pointer-00#section-3) does not describe `[index-manipulation]` which is present in ABNF.

> A Relative JSON Pointer is a Unicode string in UTF-8 encoding (see
   RFC 8259, Section 8 [RFC8259]), comprising a non-negative integer,
   followed by either a '#' (%x23) character or a JSON Pointer (RFC 6901
   [RFC6901]).

```
relative-json-pointer =  non-negative-integer [index-manipulation] <json-pointer>
relative-json-pointer =/ non-negative-integer "#"
index-manipulation    =  ("+" / "-") non-negative-integer
non-negative-integer      =  %x30 / %x31-39 *( %x30-39 )
        ; "0", or digits without a leading "0"
```

Second issue: one of [examples](https://datatracker.ietf.org/doc/html/draft-bhutton-relative-json-pointer-00#section-5.1) (`0-1#`) does not comply with ABNF -- `#` can't be used together with `[index-manipulation]`.
```
                   For example, given the JSON document:

                         {
                            "foo": ["bar", "baz"],
                            "highly": {
                               "nested": {
                                  "objects": true
                               }
                            }
                         }

     Starting from the value "baz" (inside "foo"), the following JSON
               strings evaluate to the accompanying values:

                     "0"                         "baz"
                     "1/0"                       "bar"
                     "0-1"                       "bar"
                     "2/highly/nested/objects"   true
                     "0#"                        1
                     "0-1#"                      0
                     "1#"                        "foo"
```

Is ABNF the ultimate truth? Should I refer to it when implementing Relative JSON Pointers? Or does ABNF require fixes? In any case, I could make a pull request with fixes, if anyone can tell me what exactly needs to be fixed.

handrews:
Fixes #1175 (doing another RJP draft was discussed and agreed to [several months ago](https://github.com/orgs/json-schema-org/discussions/225)).

Taking the index of the result of an index manipulation was shown in the example and intended to work, but left out of the ABNF apparently by accident.

Also, rework the ABNF to keep it within the text RFC line width.

### DIFF
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -92,27 +92,19 @@
                 length or start with a '/' (%x2F).  Similarly, a JSON Pointer will
                 never be ambiguous with the '#'.
             </t>
-            <figure align="center">
-                <preamble>
-                    The ABNF syntax of a Relative JSON Pointer is:
-                </preamble>
-                <artwork>
-<![CDATA[
-   relative-json-pointer =  origin-specification <json-pointer>
-   relative-json-pointer =/ origin-specification "#"
-   origin-specification  =  non-negative-integer [index-manipulation]
-   index-manipulation    =  ("+" / "-") non-negative-integer
-   non-negative-integer  =  %x30 / %x31-39 *( %x30-39 )
-           ; "0", or digits without a leading "0"
-]]>
-                </artwork>
-                <postamble>
-                    where &lt;json-pointer&gt; follows the production defined in
-                    <xref target="RFC6901">RFC 6901, Section 3</xref> ("Syntax").
-                </postamble>
-            </figure>
             <t>
+                The ABNF syntax of a Relative JSON Pointer is:
             </t>
+            <sourcecode type="abnf9110"><![CDATA[
+  relative-json-pointer = origin-specification json-pointer
+                        / origin-specification "#"
+                        ; json-pointer from RFC 6901
+
+  origin-specification  = non-negative-integer [ index-manipulation ]
+  index-manipulation    = ( "+" / "-" ) non-negative-integer
+  non-negative-integer  = "0" / %x31-39 *DIGIT
+                        ; zero, or digits without a leading zero
+]]></sourcecode>
         </section>
 
         <section title="Evaluation">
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

awwright:
This is good, also note `*( %x30-39 )` is the same as `*DIGIT` if you want to incorporate that here too.

handrews:
@awwright I take it that `0` is written in the ABNF as `%x30` because we're describing a string rather than numbers?  `*DIGIT` is more readable but since we have to have the hex for the others maybe it makes more sense to leave it?  Hmm...

awwright:
I would propose this:

```xml
            <t>
                The ABNF syntax of a Relative JSON Pointer is:
            </t>
            <sourcecode type="abnf9110"><![CDATA[
  relative-json-pointer = origin-specification json-pointer
                        / origin-specification "#"
                        ; json-pointer from RFC 6901

  origin-specification  = non-negative-integer [ index-manipulation ]
  index-manipulation    = ( "+" / "-" ) non-negative-integer
  non-negative-integer  = "0" / %x31-39 *DIGIT
                        ; zero, or digits without a leading zero
]]></sourcecode>
```

In addition to the things I talked about above, this replaces the figure element with sourcecode; I changed to the `/` operator from `=/` which is usually reserved for extending a production in another specification; and I changed `json-pointer` to a literal rule name, from a prose-val which is usually reserved for describing non-regular rules.

handrews:
@awwright I've incorporated your suggestion, does this look correct?

handrews:
The 2023 year updates overlap with #1375 , which will most likely be updated first in which case I'll rebase this if it's not yet re-approved, or just fix the conflicts before merging if it has been re-approved.

gregsdennis:
Closing in favor of ☝️ which has some additional commits.

PR #1362, opened 2022-12-08
Remove some superfluous format="counter" attributes

awwright:
This shouldn't cause any changes except that "Section" or "Appendix" text is now included as part of the link.

@handrews I think you wrote all of these cases, can you double check?

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -352,8 +352,7 @@
                     </t>
                     <t>
                         A JSON Schema MAY contain properties which are not schema keywords or are not recognized as schema keywords.
-                        The behavior of such keywords is governed by section
-                        <xref target="unrecognized" format="counter"></xref>.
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -439,8 +438,7 @@
                     <t>
                         The root schema is the schema that comprises the entire JSON document
                         in question.  The root schema is always a schema resource, where the
-                        IRI is determined as described in section
-                        <xref target="initial-base" format="counter"></xref>.
+                        IRI is determined as described in section <xref target="initial-base"></xref>.
```
handrews, line 441: Left behind an extra "section" on this one.

handrews (commented):
Aside from the one stray extra "section", this looks good.  I copied it locally and rendered it and the HTML output looks fine to me with these changes.

handrews (approved):

PR #1369, opened 2022-12-16
Use Cases and Requirements document

awwright:
This is a document I've been working on for a little bit that I think would be very relevant to some of the discussion right now.

We're trying to figure out what parts of the specification are "stable," but (IMO) stability is something that you determine _with respect to a particular use case._ An example I pointed out earlier, if I publish a GPU-intensive program that stops being effective as a space heater because I improved the algorithm, too bad for you because it was never "stable" for home heating purposes, only cat detection purposes (or whatever).

This document would describe the use cases that we actually intend to support; when we make changes to the specification, we can point to this document and say "this feature in Use Cases would break if we made this change." Or we can make a change and point at this to say "yes, the change breaks cat detection, but JSON Schema was never intended to perform image analysis."

Here's a PDF rendering (note how it's not a "draft"): [jsonschema-use-cases.pdf](https://github.com/json-schema-org/json-schema-spec/files/10845363/jsonschema-use-cases.pdf)

While it currently specifies applications as "Use Cases," a "Requirements" section would be expanded to actually list the particular methods that implement each use case.

### DIFF
--- a/jsonschema-use-cases.xml
+++ b/jsonschema-use-cases.xml
@@ -1,0 +1,310 @@
+<?xml version="1.0" encoding="US-ASCII"?>
+<!DOCTYPE rfc [
+<!ENTITY RFC3986 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3986.xml">
+<!ENTITY RFC3987 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.3987.xml">
+<!ENTITY RFC5234 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.5234.xml">
+<!ENTITY RFC6906 SYSTEM "http://xml.resource.org/public/rfc/bibxml/reference.RFC.6906.xml">
+]>
+<?rfc toc="yes"?>
+<?rfc symrefs="yes"?>
+<?rfc compact="yes"?>
+<?rfc subcompact="no"?>
+<?rfc strict="no"?>
+<?rfc rfcedstyle="yes"?>
+<?rfc comments="yes"?>
+<?rfc inline="yes" ?>
+<rfc category="info" docName="draft-wright-jsonschema-use-cases-00" ipr="none" submissionType="independent">
+    <front>
+        <title abbrev="JSON Schema">JSON Schema: Use Cases and Requirements</title>
+
+        <author fullname="Austin Wright" initials="A" surname="Wright" role="editor">
+            <address>
+                <email>aaa@bzfx.net</email>
+            </address>
+        </author>
+
+        <date year="2023"/>
+        <workgroup>JSON Schema</workgroup>
+        <keyword>JSON</keyword>
+        <keyword>Schema</keyword>
+        <keyword>Hyper Schema</keyword>
+        <keyword>Hypermedia</keyword>
+
+        <abstract>
+            <t>
+                To foster development of JSON Schema, this document contains a list of use cases and requirements that may be used to inform its development and evolution.
+            </t>
+        </abstract>
+        <note title="Note to Readers">
+            <t>
+                The issues list for this document can be found at
+                <eref target="https://github.com/json-schema-org/json-schema-spec/issues"/>.
+            </t>
+            <t>
+                For additional information, see <eref target="https://json-schema.org/"/>.
+            </t>
+            <t>
+                To provide feedback, use this issue tracker, the communication methods listed on the
+                homepage, or email the document editors.
+            </t>
+        </note>
+    </front>
+
+    <middle>
+        <section title="Scope and Motivation">
+            <t>
+                JSON Schema is a JSON media type for defining the structure of JSON data. JSON Schema
+                is intended to define validation, documentation, hyperlink navigation, and interaction
+                control of JSON data.
+            </t>
+            <t>
+                This document elaborates in detail what this means, and the specific use cases that shall be supported.
+            </t>
+        </section>
+
+        <section title="Conventions and Terminology">
+            <t>
+                Objectives specify the class of problems that are in the scope of the specification.
+            </t>
+
+            <t>
+                Use Cases catalog a variety of personal objectives that users may have, due to various motivations and constraints, that the specification shall accommodate, but without prescribing a specific design or implementation.
+            </t>
+
+            <t>
+                Requirements list functional, non-functional, and quality requirements, the use cases they may be derived from/related to, and reference how each use case is implemented. Requirements are not detailed at this time but may be specified in the future.
+            </t>
+        </section>
+
+        <section title="Objectives">
+            <t>
+                JSON Schema shall be built to support the following objectives, supporting expansion into uses not currently described by any use case, but which fall within the objectives.
+            </t>
+
+            <section title="Validation">
+                <t>
+                    The first objective of JSON Schema is to describe sets of JSON documents; specifically, to notate a language of JSON documents using a machine-readable, set-builder notation.
+                    This covers the key use case of validating input using a validator, as well as numerous other tools that depend on describing to each other which kinds of JSON documents are and are not acceptable.
+                </t>
+            </section>
+
+            <section title="Annotation">
+                <t>
+                    The second objective of JSON Schema is to map an input JSON document to an arbitrary output described by the user.
+                    Annotations may be combined with validation (in the same schema) to specify the domain of inputs for which an output is defined.
+                    This covers the key use case of documenting the meaning of properties and values in JSON documents, and other uses where the input document is being interpreted in some fashion.
+                </t>
+            </section>
+
+            <section title="Internet">
+                <t>
+                    JSON is a technology standardized as a part of a larger ecosystem of Internet technology, and likewise, JSON Schema may also specify its role in this ecosystem; for example, use in HTTP, or the meaning of media type parameters.
+                </t>
+            </section>
+
+            <section title="Out of scope">
+                <t>
+                    Use cases or requirements that do not advance these objectives are likely out of scope, and better suited in separate work that references JSON Schema.
+                </t>
+                <t>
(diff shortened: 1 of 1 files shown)


Relequestual:
This is actually super useful, so thanks.
I think there are a few things missing...
- Code generation, specifically "models" or classes, which is a really common use case, and "types" for typed languages
- UI generation, mostly generating forms or other UI elements, which are pretty common. (Some impelementations have larger followings than we do... at least I think)
- Documentation generation - I've seen at least 3 articles talk about JSON Schema primerily for generating documentation. I don't see that covered in your doc

Given how frequently I see these things, I'd expect them to be explicitly called out.

---

You mention database constraints in section 3.7. Can you explain why this is in section 3 as opposed to a note in section 4, aka a use case for extensions.

Given this work, maybe you could propose a short bulleted list of what's "in scope" for our charter work?
👉 https://github.com/json-schema-org/community/discussions/286#discussioncomment-4391250

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,300 @@
...
+
+            <t>
+                Requirements list functional and non-functional or quality requirements, and the use cases they may be derived from/related to.
+            </t>
+        </section>
+
+        <section title="Use Cases">
+            <section title="Structural validation" anchor="uc-structural">
+                <t>
+                    The most basic function of validating JSON is the ability to express any subset of JSON that can be described by a deterministic context-free grammar, except for distinguishing between different encodings of the same value. Schemas that limit themselves to this functionality may be compiled to a deterministic pushdown automata.
```
gregsdennis, line 85: > Schemas that limit themselves to this functionality may be compiled to a deterministic pushdown automata.

While (potentially) true, is it necessary to mention this?  This seems like an analysis/theory topic, and doesn't add to the definition of "validation."

My point is saying this is that while "pushdown automata" may be a valid implementation approach, it's not universal.  Mentioning it here may only serve to invalidate implementations which have a more procedural or functional architecture.

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,300 @@
...
+            <section title="Semantic annotation" anchor="uc-annotations">
+                <t>
+                    There is a need to annotate values within a JSON document: for machine readability, and for documentation purposes.
+                    Given a document and directions for annotating it, you should should be able to:
+                </t>
+                <list>
+                    <t>document the meaning of a property,</t>
+                    <t>suggest a default value for new documents of a given type,</t>
+                    <t>fill in missing values (in objects or arrays) with values of equivalent meaning,</t>
+                    <t>generate a list of hyperlinks,</t>
```
gregsdennis, line 113: This seems very specific to hyperschema.

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,300 @@
...
+                    <t>or declare relationships between data.</t>
+                </list>
+                <t>
+                    A schema may be used to describe a machine-readable <xref target="RFC6906">profile</xref> of JSON document. Even when two schemas identify the same set of JSON documents; depending on the context, a given JSON document may be a profile of one but not the other.
```
gregsdennis, line 117: "profile" again seems very specific to hyperschema.

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,300 @@
...
+                </t>
+            </section>
+
+            <section title="Domain-specific language">
+                <t>
+                    Application authors may use an entirely self-contained inside an application that consumes JSON. By using a declarative language, the application requirements can be better optimized by the application.
```
gregsdennis, line 123: > Application authors may use an entirely self-contained ??? inside an application that consumes JSON.

Seems like a word is missing here?

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,300 @@
...
+                </t>
+            </section>
+
+            <section title="Embedded database constraints">
+                <t>
+                    A database that uses JSON may need a way to declare, enforce, or guarantee certain constraints on the JSON document being stored. The database may use JSON Schema as a way to annotate certain fields as having a special meaning for uniqueness or indexing purposes.
+                </t>
+            </section>
+
+            <section title="Partial validation" anchor="partial-validation">
```
gregsdennis, line 163: This section seems like it would be better covered as an "applicability" section of the spec.

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,300 @@
...
+                </t>
+                <t>
+                    Like in the <xref target="partial-validation">"Partial validation"</xref> use case, validators need to understand when the extension is not supported, and allow the application to determine if validation may continue without it. For example, if a Web API does not use XML, it would be unnecessary for its JSON validator to support keywords dealing with XML, so in this case, it's appropriate for its JSON validator to reject constraints dealing with XML altogether.
+                </t>
+            </section>
+        </section>
+
+        <section title="Use cases for extensions">
+            <t>
+                This section lists use cases that not required for interoperability in media type handling, but which should be defined as as standard extensions to maximize support and feature interoperability.
```
gregsdennis, line 206:
```suggestion
                This section lists use cases that are not required for interoperability in media type handling, but which should be defined as standard extensions to maximize support and feature interoperability.
```

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,300 @@
...
+
+        </section>
+
+        <section title="Requirements" anchor="requirements">
+            <t>
+                Requirement 1.
+            </t>
+            <section title="R1. " anchor="r1">
+                <t>
+                    Requirement 1.
```
gregsdennis, line 245: Missed description?

gregsdennis (changes requested):
Overall, I question the need for this document.  The use cases that @Relequestual mentioned exist _as a result of_ not explicitly listing use cases.  I fear that explicitly listing use cases will hinder our users' creativity as they will see that their need doesn't fit within the scope of the project and they'll go find another solution.

Furthermore, some of these are not areas that we have, in the past, officially supported.  By declaring them, you are changing our stance on that.  While I agree that we should start supporting some of these, I don't know that we have the expertise or personnel required to do so to the degree that declaring "officially supported use cases" would necessitate.

Personally, I think data validation and annotation should remain our focus, and we should isntead encourage others to devise new use cases and join our community in order to support and spread knowledge of their ideas.

jsonschema-use-cases.xml
awwright: This is a good suggestion, the way it's worded does nothing to explain why this is meaningful. I pushed an update that I think is an improvement.

Though, this isn't intended for the general public, but more as a mission statement for editors and authors to follow, and maybe for validator implementers to contribute to. So of any document anywhere, I'm least worried about jargon in this document.

jsonschema-use-cases.xml
awwright: That's exactly what it's discussing... but it makes sense here, the goal is to list features that "ought to be possible" and not necessarily "must be supported by everyone."

jsonschema-use-cases.xml
awwright: The word is "schema"... fixed

jsonschema-use-cases.xml
awwright: Yes, this whole "Requirements" section needs to be filled out. Or maybe removed.

awwright:
I updated the PDF rendering: [jsonschema-use-cases.pdf](https://github.com/json-schema-org/json-schema-spec/files/10845363/jsonschema-use-cases.pdf)

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,310 @@
...
+            </t>
+        </section>
+
+        <section title="Conventions and Terminology">
+            <t>
+                Objectives specify the class of problems that are in the scope of the specification.
+            </t>
+
+            <t>
+                Use Cases catalog a variety of personal objectives that users may have, due to various motivations and constraints, that the specification shall accommodate, but without proscribing a specific design or implementation.
```
gregsdennis, line 71:
```suggestion
                Use Cases catalog a variety of personal objectives that users may have, due to various motivations and constraints, that the specification shall accommodate, but without prescribing a specific design or implementation.
```

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,310 @@
...
+                <t>
+                    For example, a method of describing an API interface would exceed the scope of JSON Schema, although JSON Schema may be used as a part of such a description, such as when the API is using JSON and needs a way to describe these JSON documents.
+                </t>
+            </section>
+
+        </section>
+
+        <section title="Use Cases">
+            <t>
+                JSON Schema shall written to standardize these use cases, or shall accommodate implementations targeted at these uses.
```
gregsdennis, line 118:
```suggestion
                JSON Schema shall be written to standardize these use cases, or shall accommodate implementations targeted at these uses.
```

jsonschema-use-cases.xml
```diff
@@ -0,0 +1,310 @@
...
+                    <t>Some programming languages cannot distinguish between an ordered array of values and a key-value map; or an empty array is identical to an empty object.</t>
+                    <t>Other validators may have a limited amount of memory and cannot support assertions more complicated than a deterministic context-free grammar.</t>
+                </list>
+                <t>
+                    Users of these validators need a way to determine if the missing functionality is essential to correctly validating input, and if not, get a validation result that would be correct but-for the unimplemented functionality.
+                </t>
+                <t>
+                    <cref>
+                        This should work both ways; an application should be able to use a third-party schema and understand if the assertions go beyond what the environment supports; and an application should be able to publish a schema that is compatible with the subset of the value space that it supports. Consider if these are separate use cases.
```
gregsdennis, line 224: > Consider if these are separate use cases.

I'm not sure what the intent behind this sentence is.  Is it an exercise for the reader?

gregsdennis (approved):
Looks good.  Just some minor nits.

Issue #1365, opened 2022-12-14
Improved handling of unknown properties


awwright:
There are many applications where a schema is distributed for use by numerous, arbitrary validators. For example, an API describes to clients a range of values that are acceptable as input, or to describe exception cases that require special handling. In these applications, it’s important that all validators agree on the set of instances that a particular schema accepts and rejects. (It goes without saying by “validator” I mean “compliant validator.”)

However, JSON Schema has one quirk in particular that prevent this from being possible: Unknown keywords are treated if they didn’t exist. For a constraint-based system, this is a serious problem when a new keyword is introduced. If a schema author uses a newly standardized validation keyword, it’s likely for an important reason that should not be ignored—which is what older validators will do. While it should be possible to write keywords that are never ignored, this is not actually possible.

There’s two obvious ways to support this application; one is to introduce a scheme so that some properties in an object can be designated “must-understand” and require validators to implement this. The other is to report/handle keywords that are unknown, so the application can react appropriately.

The $vocabulary and $schema keywords offer partial solutions of the first class, but they are not complete. These keywords are a level of abstraction removed from the keywords the schema actually requires for correct processing; as a consequence, in the worst case, you can have two validators that cannot handle any schemas in common: you may have situations where two clients cannot be supported at the same time, even if their behavior is the same.

It seems that two schemas which have the same behavior, should also be supported by the same validators. So a general solution to this problem is likely to be a of the second class: Don’t _require_ unknown keywords be ignored; but define some option to handle them unless they’re known to not be validation keywords (i.e. they are known to never reject an instance).

That is, the specification should describe how to identify those keywords that are unknown; and describe special cases where unknown keywords should not be ignored, but instead can be reported, raise an exception, or (if the details are unimportant) simply reject. Most validators can suffice with a runtime option: "Unknown keyword behavior: accept/error/reject". (The behavior currently required of implementations would be the "accept" option.)

I carefully word it this way so that, at least for now, there is no change in how schemas are processed (by default); rather this accommodates a class of applications that can enable stricter handling when that's what the application requires, and the impact of setting “error” as the default behavior can be better analyzed. (Though this can be decided now too, if some need presents itself.)

---

This is the first in a series of changes to implement [JSON Schema Extensions](https://github.com/orgs/json-schema-org/discussions/257), including:

* Standard way to indicate annotation-only keywords
* Refine definition of keywords, arguments, validation
* Typed “format” validation keywords
[...]

PR #1370, opened 2022-12-17
Split Overview into the two specific use cases

awwright:
This is an alternative to #1244, and like that issue, this one should be incorporated prior to #1365. This PR makes a clearer distinction between validation uses and annotation uses that 1365 will draw on (it relies on classifying which use each keyword is being used for, since validation use is different from annotation use).

This PR moves vocabulary-related discussion deeper into the specification, and replaces Overview with a specific description of what a JSON Schema can do, especially in terms of its output.

Once its capabilities are described, the rest of the specification can describe how you write a schema that can do these things. 

This replaces the Overview section, which is a little bit redundant with the above Abstract, then Introduction, sections.

CC @handrews: Could you provide your take on if this touches Vocabularies too much?

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -125,57 +125,51 @@
             </t>
         </section>
 
-        <section title="Overview">
+        <!-- JSON Schema accomplishes two objectives, which each get their own section. -->
+        <section title="Validation">
             <t>
-                This document proposes a new media type "application/schema+json" to identify a JSON
-                Schema for describing JSON data.
-                It also proposes a further optional media type, "application/schema-instance+json",
-                to provide additional integration features.
-                JSON Schemas are themselves JSON documents.
-                This, and related specifications, define keywords allowing authors to describe JSON
-                data in several ways.
+                A JSON Schema document describes a validator (also known as a "recognizer" or "acceptor") which classifies a provided JSON document as "accepted" or "rejected."
+                It supports "structural validation" (context-free grammars), and certain more complicated conditions.
+                Validation follows JSON semantics, so two documents that are value-equal, but vary only by character escapes, property ordering, or whitespace, will validate with the same result.
             </t>
             <t>
-                JSON Schema uses keywords to assert constraints on JSON instances or annotate those
-                instances with additional information.  Additional keywords are used to apply
-                assertions and annotations to more complex JSON data structures, or based on
-                some sort of condition.
+                With respect to a given schema, an input document accepted by that schema is called an "instance."
+                A JSON Schema may be used to specify sets of JSON documents, by referring to the set of all possible instances of that schema.
             </t>
             <t>
-                To facilitate re-use, keywords can be organized into vocabularies. A vocabulary
-                consists of a list of keywords, together with their syntax and semantics.
-                A dialect is defined as a set of vocabularies and their required support
-                identified in a meta-schema.
+                A condition for accepting a document is called an "assertion".
+                Assertions impose constraints that instances must conform to.
+                Given a schema and an instance, the schema "accepts" an input whenever all the assertions are met,
+                and the schema "rejects" when any of the assertions fail.
+                Schemas without any assertions accept all JSON documents.
             </t>
             <t>
-                JSON Schema can be extended either by defining additional vocabularies,
-                or less formally by defining additional keywords outside of any vocabulary.
-                Unrecognized individual keywords simply have their values collected as annotations,
-                while the behavior with respect to an unrecognized vocabulary can be controlled
-                when declaring which vocabularies are in use.
+                Assertions are encoded into a JSON Schema using "keywords," described below.
+            </t>
+        </section>
+
+        <section title="Annotation">
+            <t>
+                A schema may also describe an "annotator," a way to read an instance and output a set of "annotations."
+                Annotations can be any output metadata about that instance.
             </t>
             <t>
-                This document defines a core vocabulary that MUST be supported by any
-                implementation, and cannot be disabled.  Its keywords are each prefixed
-                with a "$" character to emphasize their required nature.  This vocabulary
-                is essential to the functioning of the "application/schema+json" media
-                type, and is used to bootstrap the loading of other vocabularies.
+                For example, you can document the meaning of a property,
+                suggest a default value for new instances,
+                generate a list of hyperlinks from the instance,
+                or declare relationships between data.
+                Applications may make use of annotations to query for arbitrary information;
+                for example, to extract a list of names from a document with a known structure.
+                Annotations may also describe values within the instance in a standard way;
+                for example, extracting a common type of hyperlink from many different types of documents, using a different schema for type.
             </t>
             <t>
-                Additionally, this document defines a RECOMMENDED vocabulary of keywords
-                for applying subschemas conditionally, and for applying subschemas to
-                the contents of objects and arrays.  Either this vocabulary or one very
-                much like it is required to write schemas for non-trivial JSON instances,
-                whether those schemas are intended for assertion validation, annotation,
-                or both.  While not part of the required core vocabulary, for maximum
-                interoperability this additional vocabulary is included in this document
-                and its use is strongly encouraged.
-            </t>
-            <t>
-                Further vocabularies for purposes such as structural validation or
-                hypermedia annotation are defined in other documents.  These other
-                documents each define a dialect collecting the standard sets of
-                vocabularies needed to write schemas for that document's purpose.
+                Like assertions, the instructions for producing annotations are encoded in a schema using keywords.
+                Output is only defined over valid instances,
+                so annotations are not returned until the input has been validated.
+                However, not all valid input is meaningful or true to a given application.
+                That is, if you process an arbitrary instance with nonsense data,
+                the resulting annotations may not necessarily be true, even though the input is valid.
             </t>
         </section>
 
@@ -394,6 +388,42 @@
                     </t>
                 </section>
                 <section title="Schema Vocabularies">
+                    <t>
+                        To facilitate re-use, keywords can be organized into vocabularies. A vocabulary
+                        consists of a list of keywords, together with their syntax and semantics.
+                        A dialect is defined as a set of vocabularies and their required support
+                        identified in a meta-schema.
+                    </t>
+                    <t>
+                        JSON Schema can be extended either by defining additional vocabularies,
+                        or less formally by defining additional keywords outside of any vocabulary.
+                        Unrecognized individual keywords simply have their values collected as annotations,
+                        while the behavior with respect to an unrecognized vocabulary can be controlled
+                        when declaring which vocabularies are in use.
+                    </t>
+                    <t>
+                        This document defines a core vocabulary that MUST be supported by any
+                        implementation, and cannot be disabled.  Its keywords are each prefixed
+                        with a "$" character to emphasize their required nature.  This vocabulary
+                        is essential to the functioning of the "application/schema+json" media
+                        type, and is used to bootstrap the loading of other vocabularies.
+                    </t>
+                    <t>
+                        Additionally, this document defines a RECOMMENDED vocabulary of keywords
+                        for applying subschemas conditionally, and for applying subschemas to
+                        the contents of objects and arrays.  Either this vocabulary or one very
+                        much like it is required to write schemas for non-trivial JSON instances,
+                        whether those schemas are intended for assertion validation, annotation,
+                        or both.  While not part of the required core vocabulary, for maximum
+                        interoperability this additional vocabulary is included in this document
+                        and its use is strongly encouraged.
+                    </t>
+                    <t>
+                        Further vocabularies for purposes such as structural validation or
+                        hypermedia annotation are defined in other documents.  These other
+                        documents each define a dialect collecting the standard sets of
+                        vocabularies needed to write schemas for that document's purpose.
+                    </t>
                     <t>
                         A schema vocabulary, or simply a vocabulary, is a set of keywords,
                         their syntax, and their semantics.  A vocabulary is generally organized
@@ -1357,7 +1387,7 @@
                             specification and the companion Validation specification.
                         </t>
                     </section>
-                    <section title="Non-inheritability of vocabularies ">
+                    <section title="Non-inheritability of vocabularies">
                         <t>
                             Note that the processing restrictions on "$vocabulary" mean that
                             meta-schemas that reference other meta-schemas using "$ref" or


handrews:
@awwright I generally like the split.  I'd overall be more comfortable working out high-level organization and terminology in a few issues.  I find it hard to think about terminology changes in PRs like this because they have systemic implications.  Manageable-sized PRs make it hard to think about the systemic impact, and PRs that change terminology everywhere to show consistency are too large to work with in general.

Some of this also comes from recent experiences of having wording I've written questioned- the questions were understandable, but we were only able to work out what the spec really meant because there was a solid record of discussion in issues that we could research in addition to the text.

I've also had too many experiences (both as a writer and reader) where people thought they agreed on terminology, but actually had different meanings in mind.  Discussions in issues, where work can focus on the concepts and fit words to those concepts once they are thoroughly understood and agreed to, gives me much more confidence.  It would also help us avoid some of the terminology inconsistencies that have accumulated because the terminology was developed piecemeal instead of worked out as a system.

There's a balance between working out conceptual pieces and working out the whole system that I'm not sure how to manage, but in both cases I find working from concrete wording in PRs difficult.  It makes me nervous for reasons that have little to do with the PR and a lot to do with how things have ended up misunderstood in the past.

handrews:
As far as vocabularies, I see the point of moving things down.  I also feel like it's important to get the concept established early (although it doesn't need as much detail as is currently present in the overview).  This is another thing where working at an outline level (or slightly more detailed) would help a lot more than looking at text changes and movements.

awwright:
If this looks reasonable, I'd like to move this through next.

Then I'd like to focus mostly on resolving outstanding PRs to accommodate a conversion to Markdown, but I'd also like to squeeze in #1390—this would help organize the spec to accommodate writing #1365/[Discussion #329](https://github.com/orgs/json-schema-org/discussions/329).

jsonschema-core.xml
```diff
@@ -125,57 +125,51 @@
...
+        <section title="Validation">
             <t>
-                This document proposes a new media type "application/schema+json" to identify a JSON
-                Schema for describing JSON data.
-                It also proposes a further optional media type, "application/schema-instance+json",
-                to provide additional integration features.
-                JSON Schemas are themselves JSON documents.
-                This, and related specifications, define keywords allowing authors to describe JSON
-                data in several ways.
+                A JSON Schema document describes a validator (also known as a "recognizer" or "acceptor") which classifies a provided JSON document as "accepted" or "rejected."
```
gregsdennis, line 131: The "accept"/"reject" terminology is new.  I see you use it later in the PR as well, but it's not used throughout the document.

gregsdennis, line 131: Does the schema describe a validator?  I would expect people think of the "validator" as the implementation, not the document.

jsonschema-core.xml
```diff
@@ -125,57 +125,51 @@
...
-                Further vocabularies for purposes such as structural validation or
-                hypermedia annotation are defined in other documents.  These other
-                documents each define a dialect collecting the standard sets of
-                vocabularies needed to write schemas for that document's purpose.
+                Like assertions, the instructions for producing annotations are encoded in a schema using keywords.
+                Output is only defined over valid instances,
+                so annotations are not returned until the input has been validated.
+                However, not all valid input is meaningful or true to a given application.
+                That is, if you process an arbitrary instance with nonsense data,
+                the resulting annotations may not necessarily be true, even though the input is valid.
```
gregsdennis, line 172: The use of "true" here is odd.  What does it mean for an input to be "true" to an application?

gregsdennis (changes requested):
I'm still on the fence about needing to split validation and annotation use cases like this, although this does give a deeper dive into each.  I think these sections may need to be children of a "base use cases" section to ease the reader into it a bit more.

Extracting the vocab bit is fine with me.

jsonschema-core.xml
awwright: It's new to this spec, but it is used widely outside JSON Schema and may help new readers understand what is going on. I'm going to suggest we should use accept/reject more often (it greatly simplifies the phrasing of many sentences), but that'll be an issue for later.

gregsdennis: Can you remove that language from this PR and open an issue for that change, please?

I'm not opposed to it, but I think vernacular should be an agreed-upon change, not something that's just snuck in.

awwright: Yeah that makes sense... There's a sense in which these two uses are actually the same, the "validator implementation" is just a generic form of validator that is configurable. Like if I have a schema, then if the program is written or compiled to work only with that schema, or if it's generic and configured at runtime, makes no difference.

Is there a better name for "the program that tests an input against some specific schema"?

jsonschema-core.xml
```diff
@@ -125,57 +125,51 @@
...
+        </section>
+
+        <section title="Annotation">
             <t>
-                This document defines a core vocabulary that MUST be supported by any
-                implementation, and cannot be disabled.  Its keywords are each prefixed
-                with a "$" character to emphasize their required nature.  This vocabulary
-                is essential to the functioning of the "application/schema+json" media
-                type, and is used to bootstrap the loading of other vocabularies.
+                A schema may also describe an "annotator," a way to read an instance and output a set of "annotations."
```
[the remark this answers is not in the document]
awwright: Yeah, similar situation, I have a schema, and I want to use it to compile a program that takes a JSON input and returns an output format. It's not otherwise configurable, maybe this is an HTTP service. What do I call that program?

jsonschema-core.xml
awwright: Yeah, I struggled a bit with how to phrase this. I'm trying to explain the phenomenon of "garbage in garbage out" and that the assertions don't have to be 100% completely defined.

jsonschema-core.xml
```diff
@@ -125,57 +125,51 @@
...
             </t>
             <t>
-                To facilitate re-use, keywords can be organized into vocabularies. A vocabulary
-                consists of a list of keywords, together with their syntax and semantics.
-                A dialect is defined as a set of vocabularies and their required support
-                identified in a meta-schema.
+                A condition for accepting a document is called an "assertion".
+                Assertions impose constraints that instances must conform to.
+                Given a schema and an instance, the schema "accepts" an input whenever all the assertions are met,
+                and the schema "rejects" when any of the assertions fail.
```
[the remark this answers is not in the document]
awwright: The input JSON document, as was mentioned in 'the schema "accepts" an input whenever...'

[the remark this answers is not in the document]
gregsdennis: Yes, but grammatically, you need to repeat the object.

jsonschema-core.xml
gregsdennis: I don't think you understand. Colloquially, the "validator" _is_ the implementation, not the schema.  I think we need to stick with this.

Saying the schema itself is the validator will be confusing.  A validator evaluates JSON _against_ a schema.  The schema is no more than configuration.

awwright: Well my point is there's a certain segment who may see _our_ language as new, and "accepts" is the existing term they're familiar with. I think we should use a variety of language to introduce and define the concepts, and then we can use our choice of term for the rest of the document. Is there a problem with this line of thinking?

awwright: > Colloquially, the "validator" is the implementation, not the schema.

I believe I see the point you're making, but I'm adding, this is similar to how we discuss compilers and interpreters. You're pointing out a definition of "validator" that functions like an interpreter: there's a library that reads the schema (the source code), then uses this interpretation to validate JSON.

But you can also compile source code to a program, and run the program directly. In this paradigm, there is no interpreter (what is usually called the validator), but the compiled program is still a "validator" (a thing that performs validation). It just has no concept of a schema (any more than a compiled C program can parse C).

So with JSON Schema, the schema is not the validator (as such), but I think you can say it _describes_ a validator.

gregsdennis: I see where you're coming from, but never in my experience with this project have we used "validator" that way.  It has always been used to mean the implementation.

jsonschema-core.xml
gregsdennis: I think dropping "true" and sticking with "meaningful" is the right way here.

jsonschema-core.xml
gregsdennis: I don't have a problem with introducing them, but this PR doesn't seem the place for it.  I'd like to get the opinions of the other maintainers.

awwright: @gregsdennis I think the interpreter/compiler paradigm is a useful way to think about it because it anticipates a broader range of use cases... but I can understand how it sounds weird, do you have any suggestions?

I don't think this is actually incompatible with how we've been using the term, would additional explanation help?

gregsdennis: I think at this point we should get other opinions.

Julian: To me the right question is "what's the most important thing that someone first reading the spec should see"?

I don't think what's here previously was right (because I'd assert most people reading the spec don't know or care what a media type is -- maybe to a lesser extent if we were publishing an IETF draft but certainly if we aren't). But I don't see this as being a step forward personally.

Very superficially to start, I think calling the section "Overview" is a better choice than the change (as it directs someone "hey look over here if you want to read something short"). Having subheadings may still be useful, but yeah, strictly removing the title makes it harder to skim.

I'd probably also question the assertion that an "acceptor/rejector/recognizer" language is widely used everywhere but by us. Which circles are you referring to, and again, regardless of those circles, is that the first thing someone should encounter when trying to answer "what is JSON Schema"?

All in all I find the first few paragraphs here to be a step back, I'd personally be trying to deemphasize the notion of a media type but sticking quite close to what was originally here, i.e. something like:

>                 This document proposes a JSON-based format for schematizing and annotating JSON documents, known as JSON Schema.
                JSON Schemas are themselves JSON documents.
                This, and related specifications, define keywords allowing authors to describe JSON
                data in several ways.
                Two media types for schemas are also defined.
                etc.

More generally, I'd say I'm personally reasonably comfortable with the CS concepts you seem to be trying to introduce into the spec, but whilst I feel I sufficiently understand them, I don't see them as adding *understanding* to someone reading the spec. Yes some of these things have names, but [as Feynman said about explaining science](https://bigthink.com/personal-growth/how-to-use-the-feynman-technique-to-identify-pseudoscience/):

> Suppose a student would say, ‘I don’t think energy makes [a dog] move.’ Where does the discussion go from there? I finally figured out a way to test whether you have taught an idea or you have only taught a definition. Test it this way: You say, ‘Without using the new word which you have just learned, try to rephrase what you have just learned in your own language. Without using the word “energy,” tell me what you know now about the dog’s motion.’ You cannot. So you learned nothing about science. That may be all right. You may not want to learn something about science right away. You have to learn definitions. But for the very first lesson, is that not possibly destructive?

I am of course not accusing this of being pseudoscience. But I do think that a lot of the terms you seem keen on introducing, whilst *accurate*, don't necessarily add any understanding -- someone familiar with them already quite possibly recognizes some of these things without them being mentioned. And someone who isn't, doesn't gain anything from seeing the word "context-sensitive" in the first paragraph about JSON Schema, nor from trying to understand what you mean by saying schemas define sets of documents. It's not inaccurate clearly in both cases, but it's not helpful to say so, it doesn't add enough *meaning*. My 0.02.

I suppose I'm leaving this comment on the first bit of the PR even though it applies to a bunch of what's below as well -- to me, what's here in this whole PR does too much, and does it without being clear on specific motivation for a number of the changes. Moving some discussion of vocabularies to sit before the introduction of concrete vocabularies I might agree with. Rewording the intro paragraphs to make them *more* of an overview I probably would to, in a less drastic way. But on introducing some of the terminology, likely less so.

awwright: One of the quirks I'm trying to correct that the spec currently has an abstract, then an introduction, then an overview. This seems like too much setup. The middle should begin building concepts one on top of another.

In technical writing, we need to accommodate audiences from a variety of backgrounds. Terms should be introduced so they are understandable by the layman, at least so you get a feeling quickly that you're not in the target audience for the spec. e.g. If I am looking for a way to write computation in JSON, I can read the introduction and understand that's not what this does.

It should also define its relationships to other technical terminology, to accommodate the technical audiences. People familiar with finite state machines will read this and may wonder if what we are defining is different than an acceptor, [and it is not](https://en.wikipedia.org/wiki/Finite-state_machine#Acceptors). But the fact the introduction says "also known as an acceptor" should not throw off the layman.

Julian: > One of the quirks I'm trying to correct that the spec currently has an abstract, then an introduction, then an overview. This seems like too much setup. The middle should begin building concepts one on top of another

I didn't read all of the surrounding sections again, but this would indeed sound problematic to me and like the sort of thing I'd agree we should address.

The rest of what you mentioned though, I'd stick with what I said personally. I know what a finite state machine is, I still don't find the references you're adding helpful. Again, my opinion.

gregsdennis: I agree with Julian.  Even if people know what an FSM is, significantly fewer people have a real understanding of them or how a JSON Schema can be mapped into one.  I think these are more theory-oriented ideas that would be better suited for supplemental literature rather than being in the spec itself.

awwright: > I know what a finite state machine is, I still don't find the references you're adding helpful

> significantly fewer people have a real understanding of them or how a JSON Schema can be mapped into one

Ok, though my argument is that not every part of the intro has to be helpful to everyone; it has to be written so that the widest possible audience will understand what JSON Schema accomplishes for them.

The two biggest audiences, I think, will be application developers ("I want a DSL for checking JSON, instead of doing it in code") and formal grammars ("I know what ABNF and DTDs are, I want this for JSON").

I think you'll find that other similar technology uses technical terms much _more_ heavily than I'm suggesting we do.

I looked at the [introduction for ABNF](https://www.rfc-editor.org/rfc/rfc5234#section-1), which I found far too technical for most people to understand. It talks in technical terms that it's a formal syntax, but doesn't really describe why you'd want to use it at all, or use it over other languages.

[XML DTD also talks about formal grammars](https://www.w3.org/TR/xml11/#sec-prolog-dtd), validators, and uses the accepts/rejects terminology; but it too is somewhat technical and it's not immediately obvious to me who the target audience is.

So what I'm looking for is (1) is accommodating the audience familiar with formal grammars actually unimportant? (Since ABNF and DTDs both seem to be written exclusively for this audience, I would suggest no.)

And (2) if we should accommodate the formal grammar audience, is there a better way to write it so that it's more helpful for them, and less confusing to others?

awwright: @gregsdennis Well, what's your opinion? Specifically, if I write a program that's not an interperter (like most validators), but a compiler (i.e. the output is a single Javascript function that takes one argument, the instance)—what term should be used to refer to that _compiled_ function?

@Julian I'd like to hear your feedback too.

awwright: > not simply offering my own anecdote about my personal reading

Ok, I ask because saying "I don't find it helpful" is suggestive of a personal opinion without projecting what others will think; saying "I don't believe this will be helpful" is a general observation of the sort I'm looking for.

I'm going to have to think about what else to say, if it's not immediately obvious that formal grammars are related here, as that's the formal study of what JSON Schema is fundamentally doing.

>  I also don't see the "accepts/rejects" terminology in the section you linked. It uses "valid", as we already do.

XML does not use the term "validates" (in the third person singular) to refer to an outcome (and actually it doesn't use it in that form at all). It uses "validate" to describe a process, "accept"/"matches", and "reject" to describe outcomes of that process, and "valid" to describe documents that have been accepted by the process, but nothing like "validates successfully" as we do.

Julian: > Ok, I ask because saying "I don't find it helpful" is suggestive of a personal opinion without projecting what others will think; saying "I don't believe this will be helpful" is a general observation of the sort I'm looking for.

At the risk of quoting myself, the comment I left before that was quite clear on which I was intending, please don't ignore it:

> All in all I find the first few paragraphs here to be a step back

> I don't see them as adding understanding to someone reading the spec

> what's here in this whole PR does too much

Julian: I don't believe we need terminology for such a concept in the spec at all (and certainly not at this point in time). What we use today is fine, "implementation", which refers to the executable program capable of doing things with schemas.

gregsdennis: > the function that accepts the instance would be the "validator"

This I agree with, but it doesn't follow from this that the _schema_ is a validator.  The schema is still just "configuration" (if you want to call it that.  It still goes through a library/application, and you get an output.  It's just that your example also produces an intermediate output of an executable function that represents a specific schema.  The _system_ is inputting the JSON Schema (most likely as JSON or YAML text) and an instance and getting out whether the instance is valid according to that schema.  That "compile" step is an intermediate implementation detail that doesn't need to be covered in the spec.

The spec needs to concern itself with one thing:

- inputs: a schema and an instance
- output: validation results and/or annotations

Anything an implementation does to get from input to output is necessarily beyond the scope of the spec.

awwright: > but it doesn't follow from this that the schema is a validator

I see, this isn't what I intended to convey. By saying "the schema _describes_ a validator" I think that would disconnect the schema (the description) from the validator (the actual process). Is a different word is in order here, or some additional explanation ("the schema describes the behavior of a validator")?

gregsdennis:
I do want to address some of this with the upcoming release, but I fear this particular PR may be too far gone at this point and will need to be resubmitted.

I'm going to leave this open for now as a reminder to address it.

gregsdennis:
Resubmitting these changes in the markdown format as #1579. Closing this.

PR #1372, opened 2022-12-26
Upgrade to rfc2xml v3

awwright:
This updates the syntax to the v3 vocabulary which is what the Markdown toolchain generates. This should be pulled in before #1369, #1370.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -220,14 +220,14 @@
                         An instance has one of six primitive types, and a range of possible values
                         depending on the type:
 
-                        <list style="hanging">
-                            <t hangText="null:">A JSON "null" value</t>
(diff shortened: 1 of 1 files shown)


gregsdennis:
Just to be clear, are you saying that this should be pulled in first so that we can more easily see what the markdown changeover does?

awwright:
To compare a Markdown document to an XML document, the xml2rfc toolchain can generate XML from Markdown, but it generates v3 XML, with the current XML, lists and other things will mismatch. This is a long overdue upgrade anyways.

awwright:
This has been sitting for a considerable amount of time, it shouldn't change any of the output (except it fixes some whitespace output problems), and is easily fixed or reverted later on, so I'm going flag this for merge. May someone contribute an approval please?

Issue #1340, opened 2022-10-26
Update spec to disallow unknown keywords


gregsdennis:
Following @handrews' discussion on [disallowing non-vocab keywords](https://github.com/orgs/json-schema-org/discussions/241), we decided that this is a feature that we want in order to help guarantee forward compability.

The spec needs to be updated accordingly.

handrews:
We did?  I mean, I'm all for it!  I just didn't realize there was consensus.  There should probably be an ADR first as this is a pretty big deal, and there are some details to nail down.

gregsdennis:
As far as I could tell from the proposal, I felt that we all seemed to agree on doing it.

Others are, of course, encouraged to speak up.

handrews:
I can probably write it up as an ADR easily enough (the conversation wasn't that lengthy so I'll just do a bit of cut/paste/edit) and we can have an official PR for sign-off.  If we're all in agreement it should go quickly anyway.

jdesrosiers:
> There should probably be an ADR

The ADR I'm going to write will cover that we are committing to forward compatibility, but I don't think it will cover the specific changes (like this issue) that need to happen to make that possible. So, if you write an ADR specifically for this issue, I think there would be some overlap, but it could still make sense as a separate ADR.

handrews:
@jdesrosiers sounds good- I'm basically planning to just edit that discussion into a document.  I don't plan on addressing the larger forwards compatibility topic, it would just be assumed from the SDLC and could link to the ADR for that.

PR #1376, opened 2023-01-18
Strengthen requirements on optional vocabularies and unknown keywords

### LINKED ISSUE #1300: Strengthen requirements around optional vocabularies
Issue #1294 (implemented by PR #1295) includes a better explanation of optional vocabulary use cases, specifically supporting annotation-only vocabularies without needing any custom code.

Currently, implementations SHOULD process optional vocabularies that they do not understand, and SHOULD handle unrecognized keywords as annotations.

I don't remember why the first SHOULD exists - since it's already possible to disable annotation collection (if it's supported at all), and if annotations can't be collected then unrecognized keywords MUST be ignored, I think it makes sense to upgrade this to a MUST.  Any negative impact can be mitigated by disabling annotation collection.

The second SHOULD ought to be converted to a MUST collect as annotations when annotations are enabled, and a MUST ignore otherwise.  It was a SHOULD to avoid forcing collecting large values (e.g. subschemas of unknown applicators), which was a concern with non-vocabulary keywords.  However, a vocabulary with such keywords can be included as required to avoid collecting applicators as annotations, so there's now another way to avoid that problem that makes more sense.  If we keep non-vocabulary keywords, annotation collection can remain a SHOULD for those.

Finally, we should overall make it mandatory to support unrecognized optional vocabularies.  We want people to make these as they don't require code, so we want them to be widely supported.  Doing so is trivial (again, if annotations aren't supported then supporting unrecognized optional vocabularies means literally doing nothing).

handrews:
This fixes #1300, and has been posted to clear up confusion over the difference between it and #1294 / #1295.

_The explanation is in #1300, and if there is any controversy over this, or a desire to split the the two SHOULD-to-MUST changes to accept one and reject the other, then this draft PR should be closed unmerged and the discussion should take place in issue #1300.  If this happens, anyone should feel free to close this- there is no need to check with me first._

This strengthens two SHOULDs to MUSTS:

* Implementations MUST proceed with processing schemas using unrecognized optional vocabularies
* Implementations MUST collect unknown keywords as annotations if such collection is supported and not disabled

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -620,8 +620,9 @@
 
                 <section title="Handling of unrecognized or unsupported keywords" anchor="unrecognized">
                     <t>
-                        Implementations SHOULD treat keywords they do not recognize, or that
-                        they recognize but do not support, as annotations, where the value of
+                        If annotation collection is supported and not disabled, implementations
+                        MUST treat keywords they do not recognize, or that they recognize but
+                        do not support, as annotations, where the collected value of
                         the keyword is the value of the annotation.  Whether an implementation
                         collects these annotations or not, they MUST otherwise ignore the keywords.
                     </t>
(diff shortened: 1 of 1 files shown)


handrews:
See also https://github.com/orgs/json-schema-org/discussions/295#discussioncomment-4721326, #1340, and https://github.com/orgs/json-schema-org/discussions/241 for related ideas, some of which are compatible and others are incompatible.

awwright (approved):
This seems exactly correct to me.

gregsdennis:
Closing this as

1. We are discussing the viability of optional vocabularies in light of...
2. Unknown keywords are now disallowed.

As such, this PR is moot.

PR #1380, opened 2023-02-07
Add copyright notice as per required by the OpenJS Foundation

Relequestual:
Resolves https://github.com/json-schema-org/community/issues/321

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,3 +96,7 @@ Here are our top sponsors. You could be next! [[Become a sponsor](https://openco
 ## License
 
 The contents of this repository are [licensed under](./LICENSE) either the BSD 3-clause license *or* the Academic Free License v3.0.
+
+## Copyright
+
+Copyright Contributors to the JSON Schema project.


Julian:
The first line of the LICENSE file they don't want to match? Or is this supposed to update both?

Relequestual:
> The first line of the LICENSE file they don't want to match? Or is this supposed to update both?

The line item says add the copyright notice on GitHub, so that's what I've done.
If they want it somewhere specific, they should specify, but to my mind, the readme is the most prominent.

Relequestual:
Ah I see.
The one we have in the licence file does not match the [desired format](https://github.com/openjs-foundation/cross-project-council/blob/main/governance/IP_POLICY_GUIDANCE.md#2-copyright-notices).

I agree they should match. I think the one in this PR is preferable, and the one in the licence should be updated. From what I read, @Julian would need to approve any modification to that line in the licence file as he authored it.

I guess the sensible way forward is for this PR to update both, and require Julian to approve, in addition to normal requirements.

Relequestual:
Per https://github.com/openjs-foundation/cross-project-council/issues/1155#issuecomment-1691873835 I have emailed OpenJS Foundation legal. I'll update with any response.

Relequestual:
After speaking to a number of people, I'm going to close this PR.
Things that are copyrighted, are copyrighted by default. Adding it to the readme doesn't change that.
If it was added, updating it could be problematic.

PR #718, opened 2019-02-24
Add a "duration" format.

handrews:
Fixes #565

PR #1383, opened 2023-02-27
feat: add period format

pwmcintyre:
Similar to https://github.com/json-schema-org/json-schema-spec/pull/718 — this proposal completes the set by adding the `period` format — as specific in RFC3339 Appendix A

Snippet:
```
Periods:

   period-explicit   = iso-date-time "/" iso-date-time
   period-start      = iso-date-time "/" duration
   period-end        = duration "/" iso-date-time

   period            = period-explicit / period-start / period-end
```
— https://datatracker.ietf.org/doc/html/rfc3339#appendix-A

### DIFF
--- a/jsonschema-validation.xml
+++ b/jsonschema-validation.xml
@@ -643,14 +643,14 @@
 
             <section title="Defined Formats">
 
-                <section title="Dates, Times, and Duration">
+                <section title="Dates, Times, Duration, and Period">
                     <t>
                         These attributes apply to string instances.
                     </t>
                     <t>
                         Date and time format names are derived from
                         <xref target="RFC3339">RFC 3339, section 5.6</xref>.
-                        The duration format is from the ISO 8601 ABNF as given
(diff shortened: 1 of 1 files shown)


gregsdennis:
Thanks, but at this time, I don't think we're up for adding onto `format`.  It has a history of poor support and interoperability.  Even the tests for validating `format` are optional.

However, as an annotation keyword (its default mode), you can just put `format: period`, and most validators will just allow the value.  You'll need to actually do the validation yourself in app code.

Beyond that, I think your best bet is to work with the implementation you're using to see what support they have for custom formats.

awwright:
As a well-defined and useful standard, I think this makes sense to be in core. But like @gregsdennis mentioned, there has been a bit of a change happening in how we approach "format" keywords and extensibility in general, so I filed #1391 to address that aspect.

Is "period" discussed in any other repo? If so, I think that issue plus 1391 will suffice and we can close this.

PR #1387, opened 2023-03-07
Gregsdennis/explicit annotations

gregsdennis:
I don't think that I can rightly add the stuff about `@` keywords without first disallowing unknown keywords.  But that's a big job.

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -630,6 +630,10 @@
                         collected as annotations.
                     </t>
                     <t>
+                        Keywords which begin with the "@" symbol MUST NOT affect evaluation
+                        of a schema in any way other than annotation collection.
+                    </t>
+                    <t>
                         Consequently, the "@" symbol is reserved for this purpose, and
                         extension vocabularies which define keywords that start with "@"
                         MUST define them as annotation-only keywords.
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -638,9 +638,10 @@
 
                 <section title="Handling of unrecognized or unsupported keywords" anchor="unrecognized">
                     <t>
-                        Implementations MUST refuse to process schemas which contain
-                        keywords which they do not recognize or which they recognize
-                        but do not support.
+                        Implementations SHOULD treat keywords they do not recognize, or that
+                        they recognize but do not support, as annotations, where the value of
+                        the keyword is the value of the annotation.  Whether an implementation
+                        collects these annotations or not, they MUST otherwise ignore the keywords.
                     </t>
                 </section>
             </section>
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -626,17 +626,16 @@
 
                 <section title="Explicit annotation keywords" anchor="explicit-annotation">
                     <t>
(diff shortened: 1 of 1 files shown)


jsonschema-core.xml
```diff
@@ -624,6 +624,18 @@
...
 
+                <section title="Explicit annotation keywords" anchor="explicit-annotation">
+                    <t>
+                        The values of keywords which begin with the "@" symbol MUST be
+                        collected as annotations.
+                    </t>
+                    <t>
+                        Consequently, the "@" symbol is reserved for this purpose, and
+                        extension vocabularies which define keywords that start with "@"
+                        MUST define them as annotation-only keywords.
```
jdesrosiers, line 635: I don't think we should allow vocabularies to define `@` keywords. `@` should explicitly mean that it's an unknown keyword. It's at best confusing to allow vocabularies to define keywords that looks exactly like non-vocabulary keywords.

jdesrosiers (changes requested):

jdesrosiers:
> I don't think that I can rightly add the stuff about `@` keywords without first disallowing unknown keywords.

I don't think there's any conflict in doing this first. It's an unnecessary feature until unknown keywords are removed, but I don't think it creates any inconsistencies.

> But that's a big job.

I could take a stab at it.

jsonschema-core.xml
gregsdennis: I thought it might be nice for a vocab to be able to still define a collection of them.  I don't think it hurts anything.

jdesrosiers: Let's move this conversation back to the discussion. It seems we have more to work out.  https://github.com/orgs/json-schema-org/discussions/329#discussioncomment-5273278

jsonschema-core.xml
```diff
@@ -624,6 +624,18 @@
                     for registering and implementing such handlers is implementation-dependent.
                 </t>
 
+                <section title="Explicit annotation keywords" anchor="explicit-annotation">
+                    <t>
+                        The values of keywords which begin with the "@" symbol MUST be
+                        collected as annotations.
```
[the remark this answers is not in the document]
gregsdennis: Added with [c8c669c](https://github.com/json-schema-org/json-schema-spec/pull/1387/commits/c8c669c0f7f6848c631b69a9883e056127f9aa14)

awwright (dismissed):
This language is good but I have a different suggestion for how to organize the sections; I'm going to propose to rewrite the "JSON Schema Objects and Keywords" section, and add requirements on how to classify keywords there.

Also as I pointed out in https://github.com/json-schema-org/community/issues/356 I object to the `x-` prefix because that implies something specific about who is doing the naming of the keyword, rather than the behavior of the keyword as intended. But this is a minor bikeshedding issue.

jdesrosiers (approved):

gregsdennis:
> This language is good but I have a different suggestion for how to organize the sections; I'm going to propose to rewrite the "JSON Schema Objects and Keywords" section, and add requirements on how to classify keywords there. - @awwright 

Do you have an objection to adding this now and reorganizing in another PR?  Seems like rewriting entire sections is beyond the scope of this change.

PR #1388, opened 2023-03-07
reserve dollar keywords for core vocab

### LINKED ISSUE #1321: Forbid custom keywords that start with `$`
The Core keywords are easily identifiable by the fact that they start with `$`.

The specification [recommends](https://json-schema.org/draft/2020-12/json-schema-core.html#name-the-json-schema-core-vocabu) against defining custom keywords which also start with `$`, but this recommendation seems to be repeatedly ignored by implementations (`$data` seems to be a common violator).

> While the "$" prefix is not formally reserved for the Core vocabulary, it is RECOMMENDED that extension keywords (in vocabularies or otherwise) begin with a character other than "$" to avoid possible future collisions.

Since this recommendation has been in the spec for multiple iterations, maybe it's time to move this to a `MUST NOT` requirement.

gregsdennis:
Resolves #1321

I'm not sure how far we want to take this.  Should implementations check vocabulary extensions to ensure they don't define `$` keywords and throw an error if they do?

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -1157,9 +1157,8 @@
                 <eref target="https://json-schema.org/draft/next/meta/core"/>.
             </t>
             <t>
-                While the "$" prefix is not formally reserved for the Core vocabulary,
-                it is RECOMMENDED that extension keywords (in vocabularies or otherwise)
-                begin with a character other than "$" to avoid possible future collisions.
+                The "$" prefix is reserved for use by the Core vocabulary.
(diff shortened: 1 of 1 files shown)


jdesrosiers (approved):
> I'm not sure how far we want to take this. Should implementations check vocabulary extensions to ensure they don't define $ keywords and throw an error if they do?

I think it's sufficient for it to be up to implementations to decide if they want to enforce this restriction. I think the proposed wording is all that's necessary.

Relequestual (approved):

PR #1399, opened 2023-04-20
Specify a naming scheme for non-assertion keywords

awwright:
This is an alternative to #1387 and writes in #1340.

To implement annotation keywords, there's already a section a bit higher up (JSON Schema Objects and Keywords), which defines the usage of keywords, and so any requirements about naming schemes and what may or must not be assertions should probably go here.

Additionally it writes in "disallow unknown keywords" (at least partially with a SHOULD), because I found the purpose of annotation-only keywords becomes more evident.

Please closely review the language, even without "disallow unknown keywords" it may be slightly different than what #1387 is attempting to do. It provides a slightly different rationale for it, at least: While the practical way to discuss may be "an annotation keyword", the technical reason is better understood as a "non-assertion keyword".

### DIFF
--- a/jsonschema-core.xml
+++ b/jsonschema-core.xml
@@ -313,43 +313,70 @@
                 </t>
                 <section title="JSON Schema Objects and Keywords">
                     <t>
-                        Object properties that are applied to the instance are called keywords,
-                        or schema keywords.  Broadly speaking, keywords fall into one
-                        of five categories:
+                        Object properties that are applied to the instance are called keywords, or schema keywords.
+                        Two properties in the same schema object are referred to as adjacent keywords.
+                        As an authoring convenience, adjacent keywords can sometimes affect each other in a way that a property from a different schema (including sub-schemas) cannot.
+                    </t>
+                    <t>
+                        Keywords are classified into one or more of six categories:
                     </t>
                     <dl>
-                        <dt>identifiers</dt>
+                        <dt>Core keywords</dt>
                         <dd>
-                            control schema identification through setting a IRI
-                            for the schema and/or changing how the base IRI is determined
+                            Keywords that may configure how the schema is processed,
+                            and can modify the meaning of other keywords in the schema.
                         </dd>
-                        <dt>assertions</dt>
+                        <dt>Assertion keywords</dt>
                         <dd>
-                            produce a boolean result when applied to an instance
+                            Keywords that may decide an instance is invalid.
                         </dd>
-                        <dt>annotations</dt>
+                        <dt>Annotation keywords</dt>
                         <dd>
-                            attach information to an instance for application use
+                            Keywords that produce output generated from an instance.
                         </dd>
-                        <dt>applicators</dt>
+                        <dt>Applicator keywords</dt>
                         <dd>
-                            apply one or more subschemas to a particular location
-                            in the instance, and combine or modify their results
+                            A keyword that has one or more subschemas,
+                            that are applied against the instance,
+                            or some value within the instance,
+                            and logically aggregated together.
                         </dd>
-                        <dt>reserved locations</dt>
+                        <dt>Reserved keywords</dt>
                         <dd>
-                            do not directly affect results, but reserve a place
-                            for a specific purpose to ensure interoperability
+                            Keywords that are permitted in a schema,
+                            but that do not do anything by themselves,
+                            though their value may be read by or referred to by another keyword.
+                        </dd>
+                        <dt>Arguments</dt>
+                        <dd>
+                            Arguments are properties in a schema that are used by an adjacent keyword,
+                            and whose meaning may only be defined in the presence of that keyword.
                         </dd>
                     </dl>
                     <t>
-                        Keywords may fall into multiple categories, although applicators
-                        SHOULD only produce assertion results based on their subschemas'
-                        results.  They should not define additional constraints independent
-                        of their subschemas.
+                        Some properties may be classified into multiple categories.
+                        Some keywords are also arguments to other keywords,
+                        and applicator keywords are both assertions and annotations.
                     </t>
                     <t>
-                        Keywords which are properties within the same schema object are referred to as adjacent keywords.
+                        The categories that a keyword belongs to is defined per keyword.
+                        Additionally, there are naming conventions that impose limits or requirements on the categories that keywords belong to:
+                    </t>
+                    <ul>
+                        <li>
+                            A keyword that begins with "annotation-" is never a core keyword or assertion keyword, but may be an annotation or reserved keyword. These keywords MUST NOT affect a validation result.
+                            <cref>
+                                This naming is subject to change.
+                            </cref>
+                        </li>
+                        <li>
+                            A keyword that begins with "$" is a core keyword, unless defined otherwise.
+                            (For example, "$comment" behaves as a reserved keyword.)
+                        </li>
+                    </ul>
+                    <t>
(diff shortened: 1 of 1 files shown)


jdesrosiers:
You're introducing a new keyword classification system in this PR. I think that should be discussed separately and have it's own PR when there's consensus.

awwright:
JSON Schema already has a keyword classification system, so I think it's appropriate to modify it as needed to accommodate annotation/SVA-only keywords. I mentioned that I made the changes in that section to support this addition, I can (post-hoc) explain further.

If we're going to have a naming scheme for identifying annotation-only keywords, the "$" naming scheme should also be explained here too. So this expands the scope of the "identifiers" category into "Core keywords" in turn.

And if there's going to be an option or requirement to reject unknown keywords, then we must distinguish reserved keywords (which are meaningful by themselves) and arguments (which are only meaningful in the presence of an adjacent keyword that reads it). So, this separates "Reserved locations" into "Reserved keywords" and "Arguments".

These are changes that would not be very convincing by themselves, but I think make more sense when submitted together.

Finally, I'm submitting this as a _draft_ to solicit feedback and recognize that further changes may be necessary. At least, the "annotation-" prefix that's serving as a placeholder to bypass bikeshedding. And other changes, e.g. maybe there's someone who relies on distinguishing "identifiers" from other "core keywords"—I'm not aware of such a thing, but maybe someone wants to argue that.

jdesrosiers:
I'm totally fine with changing the classification system or even dropping it altogether (I never found it useful) if that's what people want to do. I just thought it was odd lumped in with the annotation changes.

jsonschema-core.xml
```diff
@@ -313,43 +313,70 @@
...
+                            </cref>
+                        </li>
+                        <li>
+                            A keyword that begins with "$" is a core keyword, unless defined otherwise.
+                            (For example, "$comment" behaves as a reserved keyword.)
+                        </li>
+                    </ul>
+                    <t>
+                        If a validator encounters a keyword whose assertion behavior is unknown (a keyword besides an "annotation-" keyword),
+                        then attempting to compute the validation result SHOULD raise an error.
```
jdesrosiers, line 379: Instead of "raise an error", would it be better to say the result is "indeterminate"? (Has the "indeterminate" concept made it into the spec yet? It might need to be defined.) I often use "raise an error" as shorthand, but not all languages have a concept of raising an error and even then, it's not required that they use that mechanism to report the indeterminate result. But, I'm probably being overly pedantic.

jdesrosiers (commented):
This is presented as an alternative to #1387, but I think some combination of the two would be best.

The other PR states that `x-` keywords MUST be collected as annotations. I think that's important to include.

The other PR states that `x-` keywords MUST NOT be defined in vocabularies. I think that's important to include.

This PR defines the behavior when an unknown keyword is encountered, which the other PR is missing.

jsonschema-core.xml
awwright: Yes, that's more specific too.

awwright:
That's a good observation, I was mostly concerned with describing the validation behavior, but you're right, more rigorous annotation behavior can be defined too.

jsonschema-core.xml
karenetheridge: What does "attempting to compute the validation result" mean? Does it mean "it is incorrect for the evaluator to compute a validation result from this keyword"?

jsonschema-core.xml
```diff
@@ -313,43 +313,70 @@
...
                         </dd>
                     </dl>
                     <t>
-                        Keywords may fall into multiple categories, although applicators
-                        SHOULD only produce assertion results based on their subschemas'
-                        results.  They should not define additional constraints independent
-                        of their subschemas.
+                        Some properties may be classified into multiple categories.
+                        Some keywords are also arguments to other keywords,
+                        and applicator keywords are both assertions and annotations.
```
gregsdennis, line 359: Applicator keywords _may_ be assertive and annotative, but not necessarily so.  `anyOf` isn't annotative.  Its subschemas produce annotations, but the keyword itself does not.

jsonschema-core.xml
```diff
@@ -313,43 +313,70 @@
...
+                    <ul>
+                        <li>
+                            A keyword that begins with "annotation-" is never a core keyword or assertion keyword, but may be an annotation or reserved keyword. These keywords MUST NOT affect a validation result.
+                            <cref>
+                                This naming is subject to change.
+                            </cref>
+                        </li>
+                        <li>
+                            A keyword that begins with "$" is a core keyword, unless defined otherwise.
+                            (For example, "$comment" behaves as a reserved keyword.)
```
gregsdennis, line 374: See https://github.com/json-schema-org/json-schema-spec/pull/1388

jsonschema-core.xml
```diff
@@ -313,43 +313,70 @@
...
+                            and logically aggregated together.
                         </dd>
-                        <dt>applicators</dt>
+                        <dt>Reserved keywords</dt>
                         <dd>
-                            apply one or more subschemas to a particular location
-                            in the instance, and combine or modify their results
+                            Keywords that are permitted in a schema,
+                            but that do not do anything by themselves,
+                            though their value may be read by or referred to by another keyword.
```
gregsdennis, line 348: Do you have an example of a "reserved" keyword?  What would the purpose of this be?

jsonschema-core.xml
```diff
@@ -313,43 +313,70 @@
...
                         </dd>
-                        <dt>reserved locations</dt>
+                        <dt>Arguments</dt>
                         <dd>
-                            do not directly affect results, but reserve a place
-                            for a specific purpose to ensure interoperability
+                            Arguments are properties in a schema that are used by an adjacent keyword,
+                            and whose meaning may only be defined in the presence of that keyword.
```
gregsdennis, line 353: I assume these are keywords like `minContains` which affect `contains` but do nothing else.

jsonschema-core.xml
```diff
@@ -313,43 +313,70 @@
...
+                    </t>
+                    <t>
+                        Keywords are classified into one or more of six categories:
                     </t>
                     <dl>
-                        <dt>identifiers</dt>
+                        <dt>Core keywords</dt>
+                        <dd>
+                            Keywords that may configure how the schema is processed,
+                            and can modify the meaning of other keywords in the schema.
```
gregsdennis, line 327: What kind of core keyword can modify the meaning of another keyword?

There's maybe `$vocabulary`, but this _defines_ what the keywords mean; it doesn't modify them.

jsonschema-core.xml
```diff
@@ -313,43 +313,70 @@
...
+                        and applicator keywords are both assertions and annotations.
                     </t>
                     <t>
-                        Keywords which are properties within the same schema object are referred to as adjacent keywords.
+                        The categories that a keyword belongs to is defined per keyword.
+                        Additionally, there are naming conventions that impose limits or requirements on the categories that keywords belong to:
+                    </t>
+                    <ul>
+                        <li>
+                            A keyword that begins with "annotation-" is never a core keyword or assertion keyword, but may be an annotation or reserved keyword. These keywords MUST NOT affect a validation result.
```
gregsdennis, line 367: I agree with @jdesrosiers that this needs to align with the `x-` decision that we've already made.

gregsdennis:
As #1387 has been merged, I'm closing this.  If further discussion is desired, let's do that in a new issue/discussion.

PR #1400, opened 2023-04-20
Updates for Relative JSON Pointer

gregsdennis:
Replaces #1361 

~Adds an optional index manipulation to the syntax.~

It looks like this was already in the text, but the ABNF was just missing it.  This PR adds the missing syntax into the ABNF.

### DIFF
--- a/relative-json-pointer.xml
+++ b/relative-json-pointer.xml
@@ -92,27 +92,19 @@
                 length or start with a '/' (%x2F).  Similarly, a JSON Pointer will
                 never be ambiguous with the '#'.
             </t>
-            <figure align="center">
-                <preamble>
-                    The ABNF syntax of a Relative JSON Pointer is:
-                </preamble>
-                <artwork>
-<![CDATA[
-   relative-json-pointer =  origin-specification <json-pointer>
-   relative-json-pointer =/ origin-specification "#"
-   origin-specification  =  non-negative-integer [index-manipulation]
-   index-manipulation    =  ("+" / "-") non-negative-integer
-   non-negative-integer  =  %x30 / %x31-39 *( %x30-39 )
-           ; "0", or digits without a leading "0"
-]]>
-                </artwork>
-                <postamble>
-                    where &lt;json-pointer&gt; follows the production defined in
-                    <xref target="RFC6901">RFC 6901, Section 3</xref> ("Syntax").
-                </postamble>
-            </figure>
             <t>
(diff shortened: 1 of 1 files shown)


relative-json-pointer.xml
```diff
@@ -83,35 +83,31 @@
...
-                <postamble>
-                    where &lt;json-pointer&gt; follows the production defined in
-                    <xref target="RFC6901">RFC 6901, Section 3</xref> ("Syntax").
-                </postamble>
-            </figure>
             <t>
+                The ABNF syntax of a Relative JSON Pointer is:
             </t>
+            <sourcecode type="abnf9110"><![CDATA[
+  relative-json-pointer = origin-specification json-pointer
```
Relequestual, line 101: According to the only online ABNF validator I could find, it's not quite valid here.
The error is that `json-pointer` is not defined.
Removing it from that line seems fix the issue.

I'm not super familiar with ABNF to know if this is a legitimate issue or not.

gregsdennis: Okay. I'll have a look later.  I didn't actually check it.

gregsdennis: Looks like it was originally in the postamble on line 109 before the change.  Not sure if it was removed on purpose, but I think it should be in there. I can confirm offline.

gregsdennis: It says on line 103

```
; json-pointer from RFC 6901
```

Not sure if that's sufficient.

gregsdennis: Offline sources say that ABNF comment is sufficient to define `json-pointer` for this spec.  I'm happy with it.

relative-json-pointer.xml
```diff
@@ -83,35 +83,31 @@
...
+                        / origin-specification "#"
```
eemeli, line 102:
```suggestion
  relative-json-pointer = origin-specification ( "#" / json-pointer )
```
It might be better to not duplicate the `origin-specification` as it's not necessary; this equivalent formulation lets the syntax stay LL1-parseable.

eemeli (approved):
I was asked to take a look. The ABNF looks like it matches the prose; I've left a small optimization suggestion inline.

relative-json-pointer.xml
gregsdennis: I can't seem to push to this branch.  I'll create a secondary PR for this.

Relequestual (approved):
Thank you for this work!

PR #1428, opened 2023-08-27
markdown errata

gregsdennis:
This is in prep for extracting the output sections into their own spec in order to support multiple outputs for different purposes (cc: @karenetheridge).  I expect this to take place over several PRs, but #1385 shows the gist of the end goal.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -3165,7 +3165,7 @@ This meta-schema combines several vocabularies for general use.
     {"$ref": "https://json-schema.org/draft/next/meta/core"},
     {"$ref": "https://json-schema.org/draft/next/meta/applicator"},
     {"$ref": "https://json-schema.org/draft/next/meta/validation"},
-    {"$ref": "https://example.com/meta/example-vocab",
+    {"$ref": "https://example.com/meta/example-vocab"},
   ],
   "patternProperties": {
     "^unevaluated": false
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -2079,10 +2079,10 @@ consumers will need to properly interpret validation results.
 ### Format
 
 JSON Schema output is defined using the JSON Schema data instance model as
-described in [Instance Data Model](#data-model). Implementations MAY deviate from this as supported
-by their specific languages and platforms, however it is RECOMMENDED that the
-output be convertible to the JSON format defined herein via serialization or
-other means.
+described in [Instance Data Model](#data-model). Implementations MAY deviate
+from this as supported by their specific languages and platforms, however it
+is RECOMMENDED that the output be convertible to the JSON format defined herein
+via serialization or other means.
 
 ### Output Formats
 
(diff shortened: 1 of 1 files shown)


jsonschema-core.md
```diff
@@ -2324,9 +2322,9 @@ allows their users to craft their own messages.
 
 The passing instance will produce the following annotations:
 
-- The keyword `title` evaluated at `` by following the path `` will produce
```
gregsdennis, line 2327: This rendered weird in VSCode, but I wasn't sure if it was a problem for the "publish" render engine.  VSCode rendered it with " by folloing the path" as monospace text instead of recognizing that `` is an empty string.  (Not really sure how do formatting on an empty string.)

My solution was to just reword it.

gregsdennis:
This may not be necessary.  ☝️ overwrites these changes.

jsonschema-core.md
```diff
@@ -2079,7 +2079,7 @@ consumers will need to properly interpret validation results.
 ### Format
 
 JSON Schema output is defined using the JSON Schema data instance model as
-described in section 4.2.1. Implementations MAY deviate from this as supported
+described in [Instance Data Model](#data-model). Implementations MAY deviate from this as supported
```
jdesrosiers, line 2082: The markdown linter that's setup will flag this line as too long. It might be a good idea to setup a Github Action to check the linting results sooner than later as we start getting into the changes. Or, we need to be disciplined about running the build locally (`npm run build`) when we make changes.

jdesrosiers (approved):
Thanks for finding these issues. I'm sure we'll find more as we go.

jsonschema-core.md
jdesrosiers: It looks like the remark build has the same problem.

jsonschema-core.md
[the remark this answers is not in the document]
gregsdennis: Do we have a way to link to another document?  I'm not particularly keen on the current method of referencing down to the reference section since it can't link directly to a section.  (I don't need this here, but I couldn't find a way to do it.)

gregsdennis: I actually prefer not maintaining line length.  I see the benefit in having new lines on new sentences, but not really in the middle of sentences.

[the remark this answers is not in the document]
jdesrosiers: You can link to another document using the normal markdown link syntax. Personally, I'm not a fan of the whole linking to the references section pattern. I was going to suggest we remove that and just link to whatever we need to link to.

jdesrosiers: I edit in vim where very long lines are difficult to work with and reflowing the text is simple. I strongly prefer maintaining line lengths, but I can deal with it if necessary.

By default the linter plugin is designed to enforce something called the "Markdown Style Guide". I don't know how widely adopted that style guide is, but I like the idea of adopting a common standard. It appears that the linter can be customized to some degree. We can probably remove the line length rule if we decided we want to, but I haven't looked into how easy or difficult that might be.

gregsdennis: I don't mind adjusting this for now, but I think we should have this bigger conversation.  I think there are practical reasons (not just my preferred editor) to use long lines and word wrapping.

Relequestual (approved):

PR #1429, opened 2023-08-28
Extract output to its own spec

### LINKED ISSUE #1320: Extract output to a separate specification
This proposal by @Relequestual (via @jdesrosiers) was originally documented in [this discussion](https://github.com/orgs/json-schema-org/discussions/63#discussioncomment-2816059).

Moving the output to a separate specification would be advantageous in several ways.

## Audience-specific

One of the realizations that came out of @karenetheridge's [output format proposal](https://github.com/orgs/json-schema-org/discussions/195) was that there is a difference in needs between output intended to be read by machines vs humans.

What's currently in the spec is intended for machine consumers.  Karen's proposal is intended for humans.

We could have separate specifications for each, and implementations could adopt whichever is needed for their intended consumers.

## Independent versioning

The current output is versioned along with the Core spec.  As such, there were no changes between 2019-09 and 2020-12.

Having the output external to the Core spec allows independent versioning.  With the changes coming in draft-next, there will a large burden on implementations to support the new output.  This is _on top of_ any changes to schema evaluation.  Such a burden can be a deterrent for implementations to support the new schema features.

Having an independent output spec would remove that burden by allowing implementations to support new schema features while still supporting an older output format, and they could separately update the output as they have availability.

## Why we maybe shouldn't

@handrews identified that there could be features in JSON Schema (such as annotations) that impose requirements on output.  He maintains that such a relationship requires that the output requirements remain in Core.

gregsdennis:
This is my proposal for extracting the output into its own specification.  I've left some definitions in Core so that satellite specs can reference those and use a common language, but I think this still gives sufficient freedom to those satellite specs and implementations.

Probably renders #1428 unnecessary as it just deletes those changes.

There are still changes I want to make to the output spec itself, but this will do to get started.

It also incorporates aspects of [annotation filtering](https://github.com/orgs/json-schema-org/discussions/236) and [output unit pruning](https://github.com/json-schema-org/json-schema-spec/issues/1319).

Resolves #1320


jdesrosiers (changes requested):
I haven't looked closely at the content. I assume it's basically just extracted from "core" and moved almost as is. My comments are all style issues to maintain consistency throughout the spec and with the other spec documents.

---

The name of the new spec is odd. Is the "machine" part on purpose? What's that supposed to mean?

---

Instead of using `jsonc` to add titles to code blocks, you should use the [remark-code-titles](https://github.com/json-schema-org/json-schema-spec#plugins) syntax I setup exactly for this situation.

```markdown
```jsonschema "Example Schema"
```

```markdown
```json "Failing instance"
```

I should also update that plugin to style `jsonc` the same as `json` in case there's a reason we do need `jsonc`.

---

When I converted the old spec, I went through all of our JSON and tried to normalize the code style. I'm sure I missed a lot of them and the changes I'm suggesting are probably things that I missed.

---

I noticed several places where you have comments. Comments would only show to authors, but these look like the audience is reader. Should these be footnotes (until we come up with a better alternative), or is this maybe a good use for a "note" block using the [remark-flexible-contanier](https://github.com/json-schema-org/json-schema-spec#plugins) plugin?

```markdown
::: note {title}
{content}
:::
```

---

The current code style used in the spec documents is to hard wrap long lines. I know that 's controversial and needs discussion, but until we decide otherwise, we should maintain consistency.

---

I too am old and learned to type in the time where two spaces after a period was the norm, but they've gone and [changed the rules](https://www.grammarly.com/blog/spaces-after-period/) on us since and all the cool kids are using one space now. The old spec had a mix of both styles. I normalized to one space when converting so everything was consistent. We should keep up that consistency.

gregsdennis:
> I assume it's basically just extracted from "core" and moved almost as is.

Yeah, mostly.  The definitions remain in Core, and I've added some optional filtering stuff at the bottom to address some discussions.

> The name of the new spec is odd. Is the "machine" part on purpose? What's that supposed to mean?

In discussions (primarily with Henry and @karenetheridge), it became apparent that this output format was intended to be consumed by other systems and applications.  That's part of why this is being extracted: it allows @karenetheridge (or anyone else) to also create an output format for human consumption.

> Instead of using `jsonc` to add titles to code blocks...

I use `jsonc` because the JSON has comments in it.  It's not for titles.

> I too am old and learned to type in the time where two spaces after a period was the norm, but they've gone and changed the rules on us since and all the cool kids are using one space now.

Maybe we should have some automatic linter.  I'm not sure this is something I have the energy to change.  Markdown doesn't care.

jsonschema-validation-output-machines.md
```diff
@@ -0,0 +1,587 @@
...
+        }
+      }
+    }
+  }
+}
+
+// failing instance
+{
+  "foo": {"foo-prop": "not 1", "other-prop": false},
+  "bar": {"bar-prop": 2}
```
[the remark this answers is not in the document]
gregsdennis: Spaces for `{}` but not for `[]`?  Seems inconsistent.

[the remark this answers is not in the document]
jdesrosiers: I agree that it's inconsistent, but my impressing is that that's the most common convention by a wide margin. I know that's just my experience and could be biased, so I asked ChatGPT to see what it would do. It used exactly the style I was expecting, two space indent, padded objects, and not padded arrays.

Screenshot from 2023-08-30 16-34-12

jdesrosiers:
> I use `jsonc` because the JSON has comments in it. It's not for titles.

The comments are used only to label the individual JSON values. Where you have one code block with three JSON values each labeled with comments, I'm suggesting that you make a separate code block for each value and use the comment as the title. I think that would look a lot nicer.

> Maybe we should have some automatic linter. I'm not sure this is something I have the energy to change. Markdown doesn't care.

We do have a linter, but it doesn't check for that one. But, I think you meant an auto-formatter. I hate those, but it would be possible to set something up to do that. You're right, the markdown to html conversion collapses it to one space, so no one would see it but us. But, I'd still prefer that we were consistent.

gregsdennis:
I might set up something locally to "unformat" the text so it's how I like it while I'm editing then reformat before I commit, kinda like how git manages line endings.

gregsdennis:
@jdesrosiers I've updated the opening comment to include links to the discussions for the additional things this does (filtering).

Relequestual (changes requested):
I found it difficult to review this PR, but broadly it looks OK.

This PR however does not just extract output to its own spec, but also changes requirements, removing SHOULD in terms of strongly recommending using at least "flag" output.

I didn't comment on https://github.com/json-schema-org/json-schema-spec/issues/1320, but I feel now like I should have done.

In my mind, I imagined that a "human readable" version of validation output, would be computable from the (now dubbed) "machine readable" version of validation output. This, to me, was one of the main drivers to support and push for it.

My thought process was, implementations would then only need to worry about the machine readable version, and then (they or developers) could use external/additional libraries to provide human readable feedback. Not only would this allow for customization, but also for additional languages, or different algorithms to assess such output structures to find the helpful human readable error.

I looked in the associated [Issue](https://github.com/json-schema-org/json-schema-spec/issues/1320) and [Discussion](https://github.com/orgs/json-schema-org/discussions/195), but couldn't see anyone suggesting removing the recomendation. (Please direct me to a specific comment if I've missed it.)

I would find it acceptable if, having put back such a recomendation, a clause like the following was added: "Implementations MAY choose not to support any standard output format."

gregsdennis:
> removing SHOULD in terms of strongly recommending using at least "flag" output.

Output _as a whole_ was a recommendation.  The flag format was part of that.

Like vocabularies, output as a feature was added as a loose requirement so that we could iterate on it.  This change is merely one of those iterations.

Aside from extracting it, the output formats themselves have changed significantly from 2020-12.  The requirements are already broken.  Moving output into its own spec doesn't change anything.

jsonschema-core.md
```diff
@@ -2071,71 +2071,44 @@ Omitting this keyword has the same assertion behavior as an empty schema.
...
-The absolute, dereferenced location of the schema object that produced the
-output unit. The value MUST be expressed using the canonical IRI of the relevant
-schema resource plus a JSON Pointer fragment that indicates the schema object
-that produced the output. It MUST NOT include by-reference applicators such as
-`$ref` or `$dynamicRef`.[^14]
+The schema location is the absolute, dereferenced location of the schema object
+that produced a result. The value MUST be expressed using the canonical IRI of
+the relevant schema resource plus a JSON Pointer fragment that indicates the
+schema object that produced the output. It MUST NOT include by-reference
+applicators such as `$ref` or `$dynamicRef`.[^14]
```
jdesrosiers, line 2111: Saying that it MUST NOT include `$ref` is, I think, a little more restrictive than is intended. It makes sense for the pointer to include a `$ref`, but only as the terminating value. Since 2019-09, `$ref` is a normal keyword and it make sense for schema location to point to that keyword as much as any other.

jsonschema-core.md
```diff
@@ -2071,71 +2071,44 @@ Omitting this keyword has the same assertion behavior as an empty schema.
...
 
 [^14]: Note that "absolute" here is in the sense of "absolute filesystem path"
 (meaning the complete location) rather than the "absolute-IRI" terminology from
```
jdesrosiers, line 2114: I know this isn't a think that's actually changing in this PR, but I found this footnote confusing. What does filesystem have to do with this?

jsonschema-core.md
```diff
@@ -2071,71 +2071,44 @@ Omitting this keyword has the same assertion behavior as an empty schema.
...
-#### Schema Location
+### Schema Location
 
-The absolute, dereferenced location of the schema object that produced the
-output unit. The value MUST be expressed using the canonical IRI of the relevant
-schema resource plus a JSON Pointer fragment that indicates the schema object
-that produced the output. It MUST NOT include by-reference applicators such as
-`$ref` or `$dynamicRef`.[^14]
+The schema location is the absolute, dereferenced location of the schema object
+that produced a result. The value MUST be expressed using the canonical IRI of
```
jdesrosiers, line 2108: I can imagine a realistic exception to this "MUST". If the schema is used as a file, it's canonical IRI would include a full filesystem path. An implementation that doesn't want to expose full file paths for security reasons might choose to express the schema location as a relative-reference IRI instead of a full non-relative IRI.

jsonschema-core.md
```diff
@@ -2146,540 +2119,48 @@ a fragment in order to identify the specific schema object.
...
-```
+### Instance Location
 
-Because no errors or annotations are returned with this format, it is
-RECOMMENDED that implementations use short-circuiting logic to return failure or
-success as soon as the outcome can be determined. For example, if an `anyOf`
-keyword contains five subschemas, and the second one passes, there is no need to
-check the other three. The logic can simply return with success.
+The instance location is the location of the JSON value within the root instance
+being validated. The value MUST be expressed as a JSON Pointer.
```
jdesrosiers, line 2127: I have a use case in mind related to a future hyper-schema-ish implementation where instances would have IRIs and output results may include more than one instance. It would be nice if the instance location was defined in a way that could support such a case. If this was defined as an IRI instead of just a JSON Pointer, current applications could use a fragment-only IRI fragment and it would leave open the possibility of future uses that might want to use the a full IRI.

I could introduce a new field type when I eventually have the need, but that would mean it wouldn't be compatible with existing tooling, which would be a bummer.

Another option is to introduce a new field later that is just the absolute IRI of the instance and that would work in conjunction with this keyword. That would probably at least mostly work, but the naming would be weird. A better name might be `instancePointer`. Then if we introduce a new field to be used for the instance IRI, the naming is more consistent with `schemaLocation` which is also an IRI.

If you want to discuss this separately, that's fine, but I wanted to mention once more that this being just a pointer is potentially limiting for novel applications of JSON Schema.

jsonschema-core.md
```diff
@@ -2146,540 +2119,48 @@ a fragment in order to identify the specific schema object.
...
-all referenced schemas were inlined in place of their associated by-reference
-keywords.
+### Annotations
 
-All output units are included in this format.
+Many keywords are defined to produce annotations, whether intended for
+inter-keyword communication (e.g. between `properties` and
+`unevaluatedProperties`) or for application consumption (e.g. `title` or
+`readOnly`). Annotation values may be of any type and are defined by the
+keywords that produced them.
```
jdesrosiers, line 2146: It might be useful to have a link in here somewhere to the section that defines annotations.

jsonschema-core.md
```diff
@@ -2146,540 +2119,48 @@ a fragment in order to identify the specific schema object.
...
+A dropped annotation is any annotation produced and subsequently dropped by the
+evaluation due to an unsuccessful validation result of the containing subschema.
+This information MAY be included if the validation result of the containing
+subschema was unsuccessful. It MUST NOT be included if the local validation
+result of the containing subschema was successful.
 
-#### Output validation schemas
+Implementations that wish to provide dropped annotations MUST NOT provide them
+as their default behavior. Dropped annotations MUST only be included when the
+implementations is explicitly configured to do so.
```
jdesrosiers, line 2160: IMO, the spec is overstepping here. I think it's perfectly fine to allow an implementation to make this choice themselves.

jdesrosiers (commented):
I like the idea of defining the core concepts in the spec and having the actual formats described elsewhere. I didn't realize that's what you were doing here, but I like it. It's a similar concept to [Structured Fields](https://datatracker.ietf.org/doc/html/rfc8941) that's used as essentially a framework for defining new HTTP headers.

jsonschema-core.md
gregsdennis: There's not a schema at a `$ref`, only a string (URI).

```json
{
  "type": "array",
  "items": { "$ref": "foo.json" }
}
```

The schema location indicates the schema object.  There are only two schema objects here:

- the root
- the `items` subschema

If an evaluation result comes from _inside_ the ref, then the schema location is `foo.json` or some subschema contained by it.

Thus, "schema location" never needs to carry a reference keyword.

jsonschema-core.md
gregsdennis: This note was added as a clarification.

Ref: https://github.com/json-schema-org/json-schema-spec/commit/4a1035137f90ffff425bea257f48858dcd0d301c  (see Henry's commit note)

jsonschema-core.md
gregsdennis: The point of this being a fully dereferenced IRI is so that the consumer can identify the exact (sub)schema immediately without having to perform any dereferencing operations.  That can't be done with a relative IRI.

However, given that the spec is now only defining the term and not requiring the value's inclusion, I think it's reasonable that an output spec could allow returning relative IRIs, but it would need to call it something else (e.g. `relativeSchemaLocation`).

jsonschema-core.md
gregsdennis: > and output results may include more than one instance.

This isn't covered by JSON Schema anywhere.  Throughout the spec, an evaluation takes place between a single (root) schema and a single instance.  A secondary instance means a secondary evaluation.

Secondly, while an instance may _have_ an identifier, the specification does not mention or make use of them.

It doesn't make sense for output to allow for multiple instances or instance identifiers when they're not mentioned elsewhere in the spec.  (And, no, I'm not interested in adding that.  Such things are extension behaviors.)

jsonschema-core.md
gregsdennis: I see your point, but I still want to emphasize that this data point is generally intended for debugging purposes, and making it available in a production scenario will likely lead to confusion.

gregsdennis: (For a separate conversaion, but worth mentioning here.)

Moreover, if we drop annotations as an driving mechanism for evaluation, many of the keywords which now produce annotations won't need to.  For example, if `additionalProperties` just looks at the adjacent `properties` keyword's keys to know what remains to evaluate, then `properties` doesn't need to emit its annotation.

That said, I'd probably still track that data internally and treat it as a sort of "internal annotation" in order to avoid duplicate effort like having to re-analyze the `patternProperties` keys against an instance's keys.  But that doesn't mean that it needs to be exposed.

I still have to do some research here to find out what annotations are being used.

jsonschema-core.md
gregsdennis: I think I prefer to keep the MUST here and possibly add a note to the security considerations.  (We should utilize that section more, actually.)

Relequestual:
> > removing SHOULD in terms of strongly recommending using at least "flag" output.
> 
> Output _as a whole_ was a recommendation. The flag format was part of that.
> 
> Like vocabularies, output as a feature was added as a loose requirement so that we could iterate on it. This change is merely one of those iterations.
> 
> Aside from extracting it, the output formats themselves have changed significantly from 2020-12. The requirements are already broken. Moving output into its own spec doesn't change anything.

I guess I'm just a little hesitent to make no recomendation for output, but I think I understand the reasoning here.

I'd be less concerned if we aim to create a test suite for output, making it clear that we would like implementations to make use of the output spec.

As far as I can tell, we haven't seen much use of the output spec, and I'd like that to chage. You can't create downstream tools if stream isn't flowing.

jsonschema-core.md
jdesrosiers: I was thinking that "schema location" meant the location _in_ the schema rather than the location _of_ a schema, but reading more carefully, I see I had that wrong. However, if this section is intended to be a set of universal concepts output format authors can use when creating custom output formats, it seems a bit limiting to not provide a concept that allows them to point to a keyword. I won't hold up this PR for that. It can be worked out later if necessary.

jsonschema-core.md
jdesrosiers: I think if we changed the word from "absolute" to "non-relative" we can avoid to confusion and skip the footnote that IMO is even more confusing.

jsonschema-core.md
```diff
@@ -2071,70 +2071,44 @@ Omitting this keyword has the same assertion behavior as an empty schema.
...
 
-#### Schema Location
+### Schema Location
 
-The absolute, dereferenced location of the schema object that produced the
-output unit. The value MUST be expressed using the canonical IRI of the relevant
-schema resource plus a JSON Pointer fragment that indicates the schema object
-that produced the output. It MUST NOT include by-reference applicators such as
-`$ref` or `$dynamicRef`.[^14]
+The schema location is the absolute, dereferenced location of the schema object
```
jdesrosiers, line 2107: It occurred to me that I don't think the word "dereferenced" makes sense here. "Dereferenced" usually means getting a representation from a URI, but this value is a URI, not schema. Maybe the word "normalized" was intended?

jsonschema-core.md
jdesrosiers: > This isn't covered by JSON Schema anywhere.

In our charter we say we want to enable and support the use of JSON Schema in ways that it wasn't specifically designed for. I see this as falling under that umbrella. We can't predict every way JSON Schema might be used, but when we have someone (me) saying they have an idea for novel way to apply JSON Schema, we shouldn't be knowingly throwing up roadblocks for them. We should be trying to clear the roadblocks.

> an evaluation takes place between a single (root) schema and a single instance

Hyper-schema adds links to plain JSON via a schema. The variation of hyper-schema I have in mind adds references (`$ref`) to plain JSON via a schema. So, just like you have a single (root) schema, you still have a single (root) instance. Just like you can end up in a different schema through a reference, you can end up in a different JSON document through a reference.

> Such things are extension behaviors.

Of course it's an extension, but there's no reason an implementation that supports this extension shouldn't be able to use a standard output format and tools that make use of that format. The output for evaluations using this extension can be this same as a standard evaluation. There's no difference other than that there could be more than one instance involved.

---

Do whatever you want. I think it's closed-minded, but I'm tired of every little thing turning into a fight with you and this one isn't worth it.

jsonschema-core.md
gregsdennis: Yeah, the concepts currently listed are the ones that are currently used by the (now split) output spec.  I don't mind additional things being added, but I think I'd want to see them being used first.  You know, see the term in another output spec, and then import it into this spec to standardize it so that other output specs can use the same term.

jsonschema-core.md
gregsdennis: What's the distinction that you see between those terms?

jsonschema-core.md
gregsdennis: > In our charter we say we want to enable and support the use of JSON Schema in ways that it wasn't specifically designed for.

I hear you, but the spec is still written with validation and annotation of a single JSON instance as its primary function.  The terms defined herein need to reflect that.

As mentioned [above](https://github.com/json-schema-org/json-schema-spec/pull/1429#discussion_r1322173829), you're welcome to define additional terms that make sense in the context of evaluating multiple instances.

jsonschema-core.md
gregsdennis: Given our recent conversations on URIs and the word "location," I'm open to renaming this.  However, I'd prefer we discuss that elsewhere and leave this for now.

I think "dereferenced" is correct because it indicates that any references have been resolved and this is the most direct URI that leads to the (sub)schema (the canonical URI).

Relequestual: I think strictly, this is the correct use of "dereferenced", as it relates to URIs as opposed to a `$ref`. People use it in error in relation to a `$ref` when they mean "transclusion". What they really mean is "de-referencing" as in taking out references".
This is my understanding when reading RFC 3986. Please correct me if I'm wrong.

jsonschema-core.md
Relequestual: In terms of the highlighted content, I feel like SHOULD NOT and SHOULD would be enough here. There may be implementations that are specifically designed to do that by default, for example.

@gregsdennis to your comment, I don't think there's any harm in saying explicitly that this sort of behaviour is intended to aid debugging in non-production environments. Of course, that could mean for developer tooling, such a thing is wanted by default.

jsonschema-core.md
jdesrosiers: They mean exactly the same thing, it just avoids the confusion caused by the common misunderstanding of what the term "absolute-IRI" means.

This value is an IRI (must contain a scheme and may have a fragment), so when we use the term "absolute", the reader may assume that that means it's an absolute-IRI (no fragment allowed), which is not what we meant. As best I can figure out, this footnote is trying to clarify that the use of "absolute" does not mean "absolute-IRI" and you should not assume that a fragment is not allowed.

I've taken to adding the term "non-relative" when talking about IRIs (must have a scheme and may have a fragment). When people say "absolute IRI", that's generally what they mean, but that's not technically correct and that leads to confusion. Using the term "non-relative" is explicit, clear, and doesn't have common misunderstandings that we need to clarify with a footnote.

jsonschema-core.md
jdesrosiers: > This is my understanding when reading RFC 3986. Please correct me if I'm wrong or you have a different understanding.

From RFC 3986 ...

> Given a URI, a system may attempt to perform a variety of operations on the resource, as might be characterized by words such as "access", "update", "replace", or "find attributes".  Such operations are defined by the protocols that make use of URIs, not by this specification.  However, we do use a few general terms for describing common operations on URIs.  URI "resolution" is the process of determining an access mechanism and the appropriate parameters necessary to dereference a URI; this resolution may require several iterations.  **_To use that access mechanism to perform an action on the URI's resource is to "dereference" the URI._**

That's how I understand the term and I don't see how that applies in this situation.

> I think "dereferenced" is correct because it indicates that any references have been resolved and this is the most direct URI that leads to the (sub)schema (the canonical URI).

Sorry, I'm not following what you mean. What would this value look like if it wasn't "dereferenced"?

gregsdennis: > What would this value look like if it wasn't "dereferenced"?

I'm saying that it shouldn't look like the evaluation path, which explicitly preserves any reference keywords that were followed.

It should be the schema's normalized canonical IRI _plus_ a JSON Pointer fragment (which might be empty) to the specific subschema location.

So it's both dereferenced _and_ normalized, but not "absolute" in the RFC 3986 sense of an absolute URI.  I'm happy to use both terms...

jsonschema-core.md
gregsdennis: I think https://github.com/json-schema-org/json-schema-spec/pull/1429#discussion_r1323584899 may resolve this conversation as well.

jsonschema-core.md
jdesrosiers: Got it. Now I understand what it's trying to say. It just means that it should be the canonical URI for the schema. Since the intended meaning of both "absolute" and "dereferenced" is unclear and saying it's the canonical URI says everything it needs to say, I suggest dropping both of those words.

gregsdennis:
I think all comments have been addressed either through replies or changes.

@Relequestual @jdesrosiers please re-review.

Relequestual (approved):

jdesrosiers (approved):
The last round of changes look great. However, in one last attempt to try to convince you to to use code titles rather than jsonc for the output examples, here's some screenshots illustrating the difference. I think the one with code titles is much nicer. This was exactly the kind of thing we introduced the code title plugin for.

With jsonc
Screenshot from 2023-09-13 12-12-34

---

With code titles
Screenshot from 2023-09-13 12-12-55

gregsdennis:
I'm going ahead with merging this.  @karenetheridge I'd still like your thoughts, but feel free to express those as a follow-up issue/PR or here if you like.

PR #1430, opened 2023-08-28
Plain name IRI fragments can use unicode (markdown version)

gregsdennis:
Replaces #1359.  Translated to markdown.

See that PR for discussion.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -341,9 +341,7 @@ Plain name fragments MUST follow XML's [`NCName` production](#xml-names), which
 allows for compatibility with the recommended plain name
 [syntax](#w3crec-xptr-framework-20030325) for XML-based media types.  For
 convenience, the `NCName` syntax is reproduced here in ABNF form, using
-a minimal set of rules: [^20]
-
-[^20] Note that the previous syntax for plain name fragment is a subset of this syntax, which used only the first line of each of the `NCNameStartChar` and `NCNameChar` rules.
+a minimal set of rules:
 
 ```abnf
 NCName          = NCNameStartChar *NCNameChar
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -3301,7 +3301,7 @@ to the document.
 
 ### draft-bhutton-json-schema-next
 - `contains` now applies to objects as well as arrays
-- Use IRIs instead of URIs
+- Use IRIs instead of URIs, including allowing unicode in plain-name fragments
 - Clarify that detecting duplicate IRIs for different schemas SHOULD raise an error
 - Consolidate and clarify the syntax and rationale for plain-name fragments
 - "$id" MUST be an absolute-IRI, without any fragment, even an empty one
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -339,7 +339,7 @@ named schemas.
 
 Plain name fragments MUST follow XML's [`NCName` production](#xml-names), which
 allows for compatibility with the recommended plain name
-[syntax](#W3C.REC-xptr-framework-20030325) for XML-based media types.  For
+[syntax](#w3crec-xptr-framework-20030325) for XML-based media types.  For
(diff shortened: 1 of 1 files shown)


jsonschema-core.md
```diff
@@ -337,11 +337,26 @@ identifiers](#w3cwd-fragid-best-practices-20121025), plain name fragment
...
 named schemas.
 
-Plain name fragments MUST start with a letter ([A-Za-z]) or underscore ("\_"),
-followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores
-("\_"), and periods ("."). This matches the US-ASCII part of XML's [NCName
-production](#xml-names), which allows for compatibility with the recommended
-plain name [syntax](#w3crec-xptr-framework-20030325) for XML-based media types.
+Plain name fragments MUST follow XML's [NCName production](#xml-names), which
+allows for compatibility with the recommended plain name
+[syntax](#W3C.REC-xptr-framework-20030325) for XML-based media types.  For
```
gregsdennis, line 342: I'm not sure how to handle this reference.  The text from the original PR was:

```xml
<xref target="W3C.REC-xptr-framework-20030325">syntax</xref>
```

jsonschema-core.md
```diff
@@ -337,11 +337,26 @@ identifiers](#w3cwd-fragid-best-practices-20121025), plain name fragment
...
-("\_"), and periods ("."). This matches the US-ASCII part of XML's [NCName
-production](#xml-names), which allows for compatibility with the recommended
-plain name [syntax](#w3crec-xptr-framework-20030325) for XML-based media types.
+Plain name fragments MUST follow XML's [`NCName` production](#xml-names), which
+allows for compatibility with the recommended plain name
+[syntax](#W3C.REC-xptr-framework-20030325) for XML-based media types.  For
+convenience, the `NCName` syntax is reproduced here in ABNF form, using
+a minimal set of rules: [^20]
+
+[^20] Note that the previous syntax for plain name fragment is a subset of this syntax, which used only the first line of each of the `NCNameStartChar` and `NCNameChar` rules.
```
gregsdennis, line 346: I'd also like to explore a non-numbered solution to these notes.  I had to search the document to find all of the ones we had to know that "20" was the next one.  Now they're out of order.

jsonschema-core.md
jdesrosiers:
```suggestion
[syntax](#w3crec-xptr-framework-20030325) for XML-based media types.  For
```

jsonschema-core.md
jdesrosiers: A ["note" callout](https://jdesrosiers.github.io/md-spec-demo/spec.html#section-5) would probably be a good solution for cases like this.

```markdown
::: note Note
Note that the previous syntax for plain name fragment is a subset of this syntax, which used only the first line of each of the `NCNameStartChar` and `NCNameChar` rules.
:::
```

gregsdennis: Is that already supported?  If not, maybe we make that transition in another PR.

jdesrosiers: Yes, it's supported (although inexpertly styled by me). Anything in the [demo page](https://jdesrosiers.github.io/md-spec-demo/spec.html) is supported. You should also be able to build locally and see for yourself what happens. Let me know if you have issues building. It should be easy.

jdesrosiers: Looking at this again, I'm not entirely sure a "note" callout is the right thing. I'm not sure who the audience of this message is. A "note" is for calling the reader's attention to something. This seems to be something of a historical side note. I'm not sure what relevance it has to anyone.

gregsdennis: Yeah, this seems to be for the editor's benefit.

gregsdennis: Externally confirmed that this can be removed.

Relequestual (approved):

Issue #1078, opened 2021-02-27
Suggestion: String JSON Schemas as shortcut for $ref


moander:
A schema is either an object or a [boolean](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.4.3.2). What if there was a third option (string) that is a shortcut for `$ref`?

So instead of models like this
```json
{
  "$id": "https://example.com/schemas/my-model.json",
  "properties": {
    "foo": {
      "$ref": "shared-type1.json"
    },
    "bar": {
      "$ref": "shared-type2.json"
    },
    "jau":{
      "type": "string"
    },
    "hei":{
      "type": "string"
    }
  }
}
```

You get
```json
{
  "$id": "https://example.com/schemas/my-model.json",
  "properties": {
    "foo": "shared-type1.json",
    "bar": "shared-type2.json",
    "jau": "#string",
    "hei": "#string"
  },
}
```

gregsdennis:
Thanks for the suggestion!  I think it's a really interesting idea, and it shouldn't be very hard to implement.

Relequestual:
This would have complex consequences.

For example, what are the annotation results now?
How does such a change impact the additionalProperties, patternProperties, and unevaluatedProperties chain?

I'm generally against making these complex things more complex and potentially more confusing.

moander:
> This would have complex consequences.

My thought was that the complexity and consequences of having such shortcuts already is taken care of by implementors because of the boolean schema. Can the same rules and arguments that lead to the boolean schema be used to support this feature or will it have different impact on things like additionalProperties and so on?

gregsdennis:
So the boolean schemas don't have any annotations or provide any assertions.  As a result, they have no impact on these other keywords, which rely upon those annotation and assertions.

However, an implicit `$ref` _would_ have the results of the references schemas, so it must be considered how those results are incorporated into the mix.

Another point is output.  Currently there are two location indicators in the output format, `schemaLocation` which shows the relative path to the keyword being processed and `absoluteSchemaLocation` which gives the direct, fixed path.  Having an implicit `$ref` would affect the relative path (`schemaLocation`) because `$ref` is currently part of the path to indicate that a jump from one location to another was made.

Relequestual:
I don't want to open this can of worms right now, and we have other issues we DO need to solve.
[...]

PR #1432, opened 2023-08-29
add implicit reference string schemas

gregsdennis:
This is just a concept PR to get an idea of what #1078 might look like.

I'm not too particular on the wording of the new core spec section, but it's interesting to see the effect on the meta-schemas.  It definitely eliminates the cruft of having to wrap every reference in `{ "$ref": ... }`.

There _is_ an [argument](https://github.com/json-schema-org/json-schema-spec/issues/1078#issuecomment-1019409047) to be made about readability, but I'm sure it's just a matter of getting used to seeing it (like boolean schemas was).

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -190,7 +190,7 @@ always be given the media type `application/schema+json` rather than
 defined to offer a superset of the fragment identifier syntax and semantics
 provided by `application/schema-instance+json`.
 
-A JSON Schema MUST be an object or a boolean.
+A JSON Schema MUST be an object, a boolean, or a string.
 
 #### JSON Schema Objects and Keywords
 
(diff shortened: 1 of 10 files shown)


jsonschema-core.md
```diff
@@ -239,6 +239,28 @@ While the empty schema object is unambiguous, there are many possible
...
+"https://example.com/schema"
+```
+
+is equivalent to
+
+```json
+{
+  "$ref": "https://example.com/schema"
+}
+```
```
moander, line 259: Maybe a little more descriptive example?

```json
{
  "properties": {
    "foo": "https://example.com/schema"
  }
}
```

is equivalent to

```json
{
  "properties": {
    "foo": {
      "$ref": "https://example.com/schema"
    }
  }
}
```

gregsdennis: Yeah, I was wondering about that.  Feedback is good.

jsonschema-core.md
```diff
@@ -244,17 +244,27 @@ intent is clear to both human readers and implementations.
...
+  }
+}
 ```
 
 is equivalent to
 
 ```json
 {
-  "$ref": "https://example.com/schema"
+  "type": "array",
```
moander, line 262: `object` as on line 251?

gregsdennis: Ah, yes. I was going to do something different, then didn't.

gregsdennis:
I need to rework this as a proposal.

PR #1434, opened 2023-09-03
require `$schema` in schemas

### LINKED ISSUE #1420: Schema Update: Require dialect to be declared and understood
Update the spec to require that implementations require that a dialect is declared. Implementations MUST refuse to evaluate schemas without a known dialect.

In order of precedence, a dialect can be determined in the following ways.
1. The `$schema` keyword
2. The `schema` media type parameter (if the schema was retrieved over HTTP)
3. The user provides a default through an implementation's API in some way. (optional)

Implementations MUST NOT guess the dialect based on the contents of the schema and MUST NOT use a default not specified by the user. If a dialect is declared, but the implementation doesn't understand it, the implementation MUST refuse to evaluate the schema. It MUST NOT attempt to evaluate the schema using a known dialect.

gregsdennis:
Resolves #1420

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -862,7 +862,9 @@ The meta-schema serves two purposes:
 Declaring the vocabularies in use: The `$vocabulary` keyword, when it appears in
 a meta-schema, declares which vocabularies are available to be used in schemas
 that refer to that meta-schema. Vocabularies define keyword semantics, as well
-as their general syntax.
+as their general syntax. By combining various vocabularies, distinct
+sets of keywords can be made available for use in a schema. This collection of
+vocabularies defines a dialect.
 
 Describing valid schema syntax: A schema MUST successfully validate against its
 meta-schema, which constrains the syntax of the available keywords. The syntax
@@ -879,6 +881,30 @@ vocabulary's keywords.
 
 Meta-schema authoring is an advanced usage of JSON Schema, so the design of
 meta-schema features emphasizes flexibility over simplicity.
+
+### Dialect Determination
+
+When the schema evaluation process begins, the first task is to determine the
+dialect used by the schema. To do this, implementations MUST determine the
+dialect using the following prioritized steps.
+
+1. The `$schema` keyword - Implementations MUST process the schema according to
+   the dialect it declares.
+2. `application/schema+json` media type with a `schema` parameter -
+   Implementations which support media type parameter inputs MUST process the
+   schema according to the dialect the parameter declares. A media type will
+   generally only be available if the schema has been retrieved and only applies
+   to the document root.
+3. Parent dialect - An embedded schema resource which does not itself contain a
+   `$schema` keyword MUST be processed using the same dialect as the schema
+   which contains it.
+4. User configuration - Implementations MAY provide means for the user to
+   configure the dialect under which a schema should be processed.
+
+(Note that steps 2 and 3 are mutually exclusive.)
+
+If the dialect is not specified through one of these methods, the implementation
+MUST refuse to process the schema, as with unsupported required vocabularies.
 
 #### The `$schema` Keyword {#keyword-schema}
 
@@ -899,22 +925,6 @@ type `application/schema+json`.
 The `$schema` keyword SHOULD be used in the document root schema object, and MAY
 be used in the root schema objects of embedded schema resources. When the
 keyword appears in non-resource root schema object, the behavior is undefined.
-
-If present in the resource root schema, an implementation MUST process the
-schema in accordance with the associated dialect.
-
-If absent from the resource root schema, and no parent dialect is defined, the
-schema MUST be processed with the following priorities:
-
-1. If the implementation accepts media type parameter inputs, and an
-   `application/schema+json` media type with a `schema` parameter is available,
-   then the schema MUST be processed in accordance with the dialect indicated by
-   the `schema` parameter.
-2. An implementation MAY offer a mechanism for the user to explicitly set the
-   dialect.
-
-If the dialect is not specified through one of these methods, the implementation
-MUST refuse to process the schema, as with unsupported required vocabularies.
 
 Values for this property are defined elsewhere in this and other documents, and
 by other parties.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -882,7 +882,7 @@ vocabulary's keywords.
 Meta-schema authoring is an advanced usage of JSON Schema, so the design of
 meta-schema features emphasizes flexibility over simplicity.
 
-### Dialect Determination
+#### Dialect Determination
 
 When the schema evaluation process begins, the first task is to determine the
 dialect used by the schema. To do this, implementations MUST determine the
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -884,7 +884,7 @@ meta-schema features emphasizes flexibility over simplicity.
 
 #### Dialect Determination
 
-When the evaluation encounters a new schema resource (i.e. the lexical scope
+When evaluation encounters a new schema resource (i.e. the lexical scope
 changes), the first task is to determine the dialect used by the schema. To do
 this, implementations MUST determine the dialect using the following prioritized
 steps.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -885,8 +885,8 @@ meta-schema features emphasizes flexibility over simplicity.
 #### Dialect Determination
 
 When evaluation encounters a new schema resource (i.e. the lexical scope
-changes), the first task is to determine the dialect used by the schema. To do
-this, implementations MUST determine the dialect using the following prioritized
+changes), the first task is to determine the dialect used by the schema.
+Implementations MUST determine the dialect using the following prioritized
 steps.
 
 1. The `$schema` keyword - Implementations MUST process the schema according to
@@ -898,7 +898,9 @@ steps.
    source and only applies to the document root.
 3. Parent dialect - An embedded schema resource which does not itself contain a
    `$schema` keyword MUST be processed using the same dialect as the schema
-   which contains it.
+   which contains it. If the schema is embedded in a non-schema document, the
+   semantics for determining the dialect MAY be determined by any specification
+   which applies to that document.
 4. User configuration - Implementations MAY provide means for the user to
    configure the dialect under which a schema should be processed.
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -884,7 +884,10 @@ meta-schema features emphasizes flexibility over simplicity.
 
 The `$schema` keyword is both used as a JSON Schema dialect identifier and as
 the identifier of a resource which is itself a JSON Schema, which describes the
-set of valid schemas written for this particular dialect.
+set of valid schemas written for this particular dialect. The identified dialect
+applies to the resource in which it is declared as well as any embedded schema
+resources, unless such a resource itself declares a different dialect by
+including the `$schema` keyword with a different value.
 
 The value of this keyword MUST be an [IRI](#rfc3987) (containing a scheme) and
 this IRI MUST be normalized. The current schema MUST be valid against the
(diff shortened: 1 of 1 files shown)


jdesrosiers (changes requested):
Glad to see these this change happening!

Something that's missing is dialect inheritance in embedded schemas. An embedded schema that doesn't declare `$schema` inherits the dialect of the schema it's embedded in (how ever the parent schema's dialect was determined, not necessarily `$schema`).

jsonschema-core.md
```diff
@@ -895,14 +895,22 @@ type `application/schema+json`.
...
 be used in the root schema objects of embedded schema resources. It MUST NOT
-appear in non-resource root schema objects. If absent from the document root
-schema, the resulting behavior is implementation-defined, but MUST fall within
-the following options:
-
-- Refuse to process the schema, as with unsupported required vocabularies
-- Assume a specific, documented meta-schema
-- Document the process by which it examines the schema and determines which of a
-  specific set of meta-schemas to assume
+appear in non-resource root schema objects.
```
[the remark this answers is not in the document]
gregsdennis: I think it would be a good idea to explicitly state that it's undefined.  I'm not sure that completely removing it is right.

jsonschema-core.md
```diff
@@ -895,14 +895,22 @@ type `application/schema+json`.
...
+
+If present in the document root schema, an implementation MUST process the
```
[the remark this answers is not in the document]
gregsdennis: I used "document" because that's what was used above.  I can change it.

jdesrosiers (commented):
I feel like this could be organized a bit better. There are four rules for determining the dialect, but they're not all in the same place and not presented consistently. The first rule is in a different place than the other three, Those three are in the same place, but the third and fourth are presented differently (as a list) from the second.

I think this could be more clear if the rules for determining the dialect were presented consistently and in one place in a way implementers could follow like an algorithm. I think the third and fourth rules are presented in this way, but it would be nice if the first and second were included as well.

Also, it just occurred to me that this is all defined in the section about the `$schema` keyword. Since the `$schema` keyword is only one one part of dialect determination, it seems like it would be better to talk about dialect determination in it's own section that references to `$schema` keyword as one way the dialect could be determined.

jdesrosiers:
The fourth rule is an embedded schema inheriting from its parent context. I think it belongs after the media-type parameter and before the user provided default.

I had forgotten about that one when I wrote that issue, but it's definitely part of the process. If you'd rather address that in a separate PR, I'm ok with that.

gregsdennis:
I disagree then.  And this needs to be discussed in that issue, not in this PR.

Relequestual (approved):
Can't wait to see the tests for this get merged! =D

jsonschema-core.md
```diff
@@ -880,29 +882,50 @@ vocabulary's keywords.
...
+1. The `$schema` keyword - Implementations MUST process the schema according to
+   the dialect it declares.
+2. `application/schema+json` media type with a `schema` parameter -
+   Implementations which support media type parameter inputs MUST process the
+   schema according to the dialect the parameter declares. A media type will
+   generally only be available if the schema has been retrieved from an external
+   source and only applies to the document root.
+3. Parent dialect - An embedded schema resource which does not itself contain a
+   `$schema` keyword MUST be processed using the same dialect as the schema
+   which contains it.
```
jdesrosiers, line 901: I think this concept also applies to schemas embedded in other types of documents like OpenAPI. I'm not sure how this might be reworded to include that case, or if it should be considered a different case, or if that case should be addressed elsewhere entirely.

jsonschema-core.md
```diff
@@ -862,7 +862,9 @@ The meta-schema serves two purposes:
 Declaring the vocabularies in use: The `$vocabulary` keyword, when it appears in
 a meta-schema, declares which vocabularies are available to be used in schemas
 that refer to that meta-schema. Vocabularies define keyword semantics, as well
-as their general syntax.
+as their general syntax. By combining various vocabularies, distinct
+sets of keywords can be made available for use in a schema. This collection of
+vocabularies defines a dialect.
 
 Describing valid schema syntax: A schema MUST successfully validate against its
 meta-schema, which constrains the syntax of the available keywords. The syntax
```
jdesrosiers, line 870: I noticed that a schema must validate against it's meta-schema in two places. The other place is in the `$schema` section on line 920. I don't have an opinion on which place to say this, but it doesn't need to be both places.

This isn't technically related to this change, but it might be a good time to clean that up since we're making changes in this area anyway. I'm also ok with dealing with this separately.

jdesrosiers (approved):
This looks great! I left a couple notes, but either could be address separately if we want.

jdesrosiers:
> Can't wait to see the tests for this get merged! =D

I'm not sure much, if any, of this testable with our current setup.

jsonschema-core.md
gregsdennis: I think I'd drop it from 920 since this is a more general "must validate against the meta-schema regardless of how you get the meta-schema" whereas 920 is specifically targeted to `$schema`.

jsonschema-core.md
```diff
@@ -880,29 +882,50 @@ vocabulary's keywords.
...
+this, implementations MUST determine the dialect using the following prioritized
+steps.
+
+1. The `$schema` keyword - Implementations MUST process the schema according to
+   the dialect it declares.
+2. `application/schema+json` media type with a `schema` parameter -
+   Implementations which support media type parameter inputs MUST process the
+   schema according to the dialect the parameter declares. A media type will
+   generally only be available if the schema has been retrieved from an external
+   source and only applies to the document root.
```
karenetheridge, line 898: I think we can go stronger here and say that if the schema is fetched over HTTP, it MUST respect the media type parameter if it is present, and add a reference to the section that defines the name and syntax of this header. This may be implied by the existing wording, but we can spell it out more explicitly.

jsonschema-core.md
karenetheridge: We can say something like:

If the schema is embedded in a larger document (such as OpenAPI), the semantics for determining the dialect may be specified by that document's specification. Unless explicitly dictated by that document's specification, the presence of a `$schema` keyword shall always override the dialect imposed by the containing document.

jsonschema-core.md
```diff
@@ -880,29 +882,50 @@ vocabulary's keywords.
...
 the identifier of a resource which is itself a JSON Schema, which describes the
-set of valid schemas written for this particular dialect.
+set of valid schemas written for this particular dialect. The identified dialect
+applies to the resource in which it is declared as well as any embedded schema
+resources, unless such a resource itself declares a different dialect by
+including the `$schema` keyword with a different value.
 
-The value of this keyword MUST be a [IRI](#rfc3987) (containing a scheme) and
+The value of this keyword MUST be an [IRI](#rfc3987) (containing a scheme) and
 this IRI MUST be normalized. The current schema MUST be valid against the
```
karenetheridge, line 920:
```suggestion
this IRI MUST be normalized. The current schema MUST be valid when evaluated against the
```

karenetheridge (commented):
I don't know if it's been discussed elsewhere but there is still a problem with allowing `$schema` keywords to appear in non-root locations in the document -- because we also require that the schema must validate against its metaschema, and this requirement can be violated if a contained sub-schema uses a different dialect than the document root that conflicts in some way.

Are we expecting that implementations need to have a special evaluation mode when validating schemas, and somehow switch to a different metaschema when $schema keywords are encountered? We don't spell that out in the spec in any way, and instead assume that when a schema is being evaluated against its metaschema, we treat it just like any other data instance.

jsonschema-core.md
gregsdennis: I like that @karenetheridge.  I'll try to work that in.

jsonschema-core.md
gregsdennis: Is this not what it's already saying?  895 contains the "MUST" requirement.  I don't want to call out HTTP; I'd prefer to stay more general and say that if the media type parameter is available in general.  It could come from anywhere, e.g. the OS or app.

jsonschema-core.md
Relequestual: I agree. Currently, the OpenAPI spec allows you to define a default dialect.

@karenetheridge What's the use case you see for allowing a containing document specification to override `$schema` for it's contained schemas?

Relequestual:
> Are we expecting that implementations need to have a special evaluation mode when validating schemas, and somehow switch to a different metaschema when $schema keywords are encountered?

Yup.

> We don't spell that out in the spec in any way, and instead assume that when a schema is being evaluated against its metaschema, we treat it just like any other data instance.

First, we have to look at the definition of a Compound Schema Document...

> A Compound Schema Document is defined as a JSON document (sometimes called a "bundled" schema) which has multiple embedded JSON Schema Resources bundled into the same document to ease transportation. - [Core 9.3](https://json-schema.org/draft/2020-12/json-schema-core.html#section-9.3)

I would argue that this defines any OpenAPI definition as a compound schema document.
While the examples in the apendix section is a JSON Schema, it doesn't have to be.

Then, the validation section...

> Given that a Compound Schema Document may have embedded resources which identify as using different dialects, these documents SHOULD NOT be validated by applying a meta-schema to the Compound Schema Document as an instance. It is RECOMMENDED that an alternate validation process be provided in order to validate Schema Documents. Each Schema Resource SHOULD be separately validated against its associated meta-schema. - [Core 9.3.3](https://json-schema.org/draft/2020-12/json-schema-core.html#section-9.3.3)

Given this, we might want to tighten up the language of meta-schema useage for validation to make the target a "schema resource" to make this clearer? I don't know. Open to thoughts.

Relequestual:
> I don't think the comment was about OpenAPI, but rather any time a $schema is used in a subschema resource, which is allowed currently.

> If the schema is embedded in a larger document (such as OpenAPI)... - One of @karenetheridge [comments](https://github.com/json-schema-org/json-schema-spec/pull/1434#discussion_r1320628013).

However, I agree, it applies more broadly.

> Still, let's work that out separately from this PR, please.

I agree. As a reference, we discussed this at quite some length already: https://github.com/json-schema-org/json-schema-spec/issues/936

jsonschema-core.md
karenetheridge: I guess I was confused what "Implementations which support media type parameter inputs" meant. Does that cover "if you used HTTP, you better check the Content-Type header and do what it says"?

jsonschema-core.md
karenetheridge: No, the other way around - the openapi spec has language about following `jsonSchemaDialect` to set the dialect for all contained schemas, but if one of those schemas has a `$schema` keyword in it, that overrides the jsonSchemaDialect. (Unless OpenAPI wanted to add a rule saying that `$schema` is no good here.)

jsonschema-core.md
gregsdennis: I think it covers any scenario, including http, where you might receive a media type.  So, yeah.

jsonschema-core.md
gregsdennis: Right, so that falls in line with the precedence order here.  `$schema` first, then media type (if it exists), then parent document, then config.

I think Ben was wondering about the edge case where a document wanted to override `$schema`.  I'm not sure we can define requirements for that.

karenetheridge (approved):

karenetheridge:
> I agree. As a reference, we discussed this at quite some length already: https://github.com/json-schema-org/json-schema-spec/issues/936

That issue is talking about how to _evaluate with a schema_ when there are embedded `$schema` keywords that change the dialect midway through. The problem I raised is how to _evaluate that schema as a data instance_ against its metaschema -- because the metaschema you're evaluating with is changing midway through the evaluation, but the normal evaluation process doesn't know anything about switching schemas depending on what it finds in the data.

gregsdennis:
@karenetheridge I get what you're saying. I'll create an issue for it for us to discuss.

Issue #1231, opened 2022-05-19
Security considerations should mention treating URIs as URLs (from $ref and $schema)


Relequestual:
Related https://github.com/json-schema-org/community/discussions/176

karenetheridge:
We mention earlier on that it is NOT required for an implementation to support loading resources from the web (https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.2.3), but this may not be prominent enough as a warning; we could do with repeating it in the security section. (And possibly generalize to loading from anywhere, including disk.)

awwright:
Yeah, specifically mention the risks of loading resources controlled by third parties, as well as performance and privacy concerns. e.g. validation that relies on network operations can potentially be blocked by DoS attacks.

Network operations should be limited to hypermedia APIs and similar applications where this risk already exists and is built into the architecture. We sort of mention this in the appropriate sections, but it can be duplicated in security concerns.

PR #1435, opened 2023-09-04
Gregsdennis/locatable iri adjustments

gregsdennis:
Resolves https://github.com/orgs/json-schema-org/discussions/460

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1058,10 +1058,7 @@ IRI.
 
 Note that this IRI is an identifier and not necessarily a network locator. In
 the case of a network-addressable URL, a schema need not be downloadable from
-its canonical IRI. However, schema authors SHOULD refrain from creating schemas
-that use a locatable `$id` value unless the schema will be available at the
-indicated location. Otherwise, an IRI which does not convey location semantics
-SHOULD be used.
+its canonical IRI.
 
 If present, the value for this keyword MUST be a string, and MUST represent a
 valid [IRI-reference](#rfc3987). This IRI-reference SHOULD be normalized, and
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1277,16 +1277,18 @@ The use of IRIs to identify remote schemas does not necessarily mean anything is
 downloaded, but instead JSON Schema implementations SHOULD understand ahead of
 time which schemas they will be using, and the IRIs that identify them.
 
-When schemas are downloaded, for example by a generic user-agent that does not
-know until runtime which schemas to download, see [Usage for
-Hypermedia](#hypermedia).
-
 Implementations SHOULD be able to associate arbitrary IRIs with an arbitrary
 schema and/or automatically associate a schema's `$id`-given IRI, depending on
 the trust that the validator has in the schema. Such IRIs and schemas can be
 supplied to an implementation prior to processing instances, or may be noted
 within a schema document as it is processed, producing associations as shown in
 {{idexamples}}.
+
+Implementations MAY provide functionality to automatically fetch schemas based
+on location semantics expressed by the URI, however such functionality MUST be
+disabled by default to prefer offline operation. When schemas are downloaded,
+for example by a generic user-agent that does not know until runtime which
+schemas to download, see {{hypermedia}}.
 
 #### Detecting a Meta-Schema
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1058,7 +1058,10 @@ IRI.
 
 Note that this IRI is an identifier and not necessarily a network locator. In
 the case of a network-addressable URL, a schema need not be downloadable from
-its canonical IRI.
+its canonical IRI. However, schema authors SHOULD refrain from creating schemas
+that use a locatable `$id` value unless the schema will be available at the
+indicated location. Otherwise, an IRI which does not convey location semantics
+SHOULD be used.
 
 If present, the value for this keyword MUST be a string, and MUST represent a
 valid [IRI-reference](#rfc3987). This IRI-reference SHOULD be normalized, and
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1282,7 +1282,7 @@ within a schema document as it is processed, producing associations as shown in
 {{idexamples}}.
 
 Implementations MAY provide functionality to automatically fetch schemas based
-on location semantics expressed by the IRI, however such functionality MUST be
+on location semantics expressed by the IRI, however such functionality SHOULD be
 disabled by default to prefer offline operation. When schemas are downloaded,
 for example by a generic user-agent that does not know until runtime which
 schemas to download, see {{hypermedia}}.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1144,7 +1144,7 @@ provides the ability to validate recursive structures through self-reference.
 The resolved IRI produced by these keywords is not necessarily a network
 locator, only an identifier. A schema need not be downloadable from the address
 if it is a network-addressable URL. Implementations which can access the network
-MUST default to operating offline.
+SHOULD default to operating offline.
 
 ##### Direct References with `$ref` {#ref}
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1282,7 +1282,7 @@ within a schema document as it is processed, producing associations as shown in
 {{idexamples}}.
 
 Implementations MAY provide functionality to automatically fetch schemas based
-on location semantics expressed by the URI, however such functionality MUST be
+on location semantics expressed by the URI, however such functionality SHOULD be
 disabled by default to prefer offline operation. When schemas are downloaded,
 for example by a generic user-agent that does not know until runtime which
 schemas to download, see {{hypermedia}}.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1146,9 +1146,8 @@ provides the ability to validate recursive structures through self-reference.
 
 The resolved IRI produced by these keywords is not necessarily a network
 locator, only an identifier. A schema need not be downloadable from the address
-if it is a network-addressable URL, and implementations SHOULD NOT assume they
-should perform a network operation when they encounter a network-addressable
-IRI.
+if it is a network-addressable URL. Implementations which can access the network
+SHOULD default to operating offline.
 
 ##### Direct References with `$ref` {#ref}
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1282,7 +1282,7 @@ within a schema document as it is processed, producing associations as shown in
 {{idexamples}}.
 
 Implementations MAY provide functionality to automatically fetch schemas based
-on location semantics expressed by the URI, however such functionality MUST be
+on location semantics expressed by the IRI, however such functionality MUST be
 disabled by default to prefer offline operation. When schemas are downloaded,
 for example by a generic user-agent that does not know until runtime which
 schemas to download, see {{hypermedia}}.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1144,7 +1144,7 @@ provides the ability to validate recursive structures through self-reference.
 The resolved IRI produced by these keywords is not necessarily a network
 locator, only an identifier. A schema need not be downloadable from the address
 if it is a network-addressable URL. Implementations which can access the network
-SHOULD default to operating offline.
+MUST default to operating offline.
 
 ##### Direct References with `$ref` {#ref}
 
@@ -1282,7 +1282,7 @@ within a schema document as it is processed, producing associations as shown in
 {{idexamples}}.
 
 Implementations MAY provide functionality to automatically fetch schemas based
-on location semantics expressed by the URI, however such functionality SHOULD be
+on location semantics expressed by the URI, however such functionality MUST be
 disabled by default to prefer offline operation. When schemas are downloaded,
 for example by a generic user-agent that does not know until runtime which
 schemas to download, see {{hypermedia}}.


jsonschema-core.md
```diff
@@ -1045,7 +1045,10 @@ IRI.
 
 Note that this IRI is an identifier and not necessarily a network locator. In
 the case of a network-addressable URL, a schema need not be downloadable from
-its canonical IRI.
+its canonical IRI. However, schema authors SHOULD refrain from creating schemas
+that use a locatable `$id` value unless the schema will be available at the
+indicated location. Otherwise, an IRI which does not convey location semantics
+SHOULD be used.
```
gregsdennis, line 1051: I'm still iffy on including this.  We don't have much in the way for recommendations for schema authors mainly because we really can't control what they do.

jsonschema-core.md
```diff
@@ -1262,17 +1264,19 @@ The use of IRIs to identify remote schemas does not necessarily mean anything is
...
 the trust that the validator has in the schema. Such IRIs and schemas can be
 supplied to an implementation prior to processing instances, or may be noted
 within a schema document as it is processed, producing associations as shown in
 {{idexamples}}.
 
+Implementations MAY provide functionality to automatically fetch schemas based
+on location semantics expressed by the URI, however such functionality MUST be
+disabled by default to prefer offline operation. When schemas are downloaded,
+for example by a generic user-agent that does not know until runtime which
+schemas to download, see {{hypermedia}}.
```
jdesrosiers, line 1278: > such functionality MUST be disabled by default to prefer offline operation

This would mean you can't write a hyper-schema client that functions at all in it's default configuration. We can't do that. It also contradicts line 1137 that says this is a "SHOULD". Saying it once on that line should be enough. I also don't think strengthening this "SHOULD" to a "MUST" accurately reflects the opinions expressed in the discussion.

Here's an alternative that I think avoids the useless-hyper-schema-client problem and reflects the opinions expressed in the discussion.

```suggestion
Some types of implementations, such as generic user-agents (see {{hypermedia}}), don't know what schemas they will need until runtime. Those implementations MAY automatically fetch schemas based on location semantics expressed by the URI.
```

We aren't explicitly giving permission to fetch schemas to any and all types of implementations, just ones where that's the only mode in which the implementation could possibly work. This doesn't apply to typical validator implementations that can know schemas ahead of time. The recommended behavior of those types of implementations is given on line 1137.

jdesrosiers (changes requested):
The two main changes here are addressing the same kind of thing, but in different sections far away from each other. I think it makes sense to address all of this in the same place. I think the "Loading Schemas" section is the best place. The "Schema References" section can link to the other section if we think that's necessary.

jsonschema-core.md
Relequestual: I'm not convinced we should include this in the spec either.
I think this is the only section which strays into best practice though.

> I also don't think this recommendation makes sense given the direction these changes and the discussions behind them have gone.

I think we have (speaking personally) intrenched to almost always use https, that using a non-locating semantics URI scheme seems strange. But I do think it's something we should strive to wards suggesting as best practice. It would avoid confusion, as you've said. But I also think we shouldn't allow things to be required that way. (Won't say any more on the topic here.)

I think we can/should track potential best practice recomendations as Issues in the Community repo, and maybe migrating them to the new website repo when appropriate.

jsonschema-core.md
Relequestual: I think the proposed suggestion above is fine.

gregsdennis: He's right that it conflicts with 1137, though.  I'm happy to make the MUST a SHOULD, but I think I prefer the wording that's there.

jsonschema-core.md
```diff
@@ -2081,7 +2082,7 @@ consumers will need to properly interpret validation results.
 JSON Schema output is defined using the JSON Schema data instance model as
 described in section 4.2.1. Implementations MAY deviate from this as supported
 by their specific languages and platforms, however it is RECOMMENDED that the
-output be convertible to the JSON format defined herein via serialization or
+output be convertib le to the JSON format defined herein via serialization or
```
karenetheridge, line 2085:
```suggestion
output be convertible to the JSON format defined herein via serialization or
```

gregsdennis: Nice catch, yeah.  The fix for that came in when I rebased just now.

jsonschema-core.md
karenetheridge: How does it conflict?  This line just says that the config option can exist; it doesn't say that it's allowed to default to on.

I would prefer to say that the config option to download, if it exists at all, MUST default to off. Otherwise we have security considerations that must be documented more visibly.

gregsdennis: From "Schema References" section

> Implementations which can access the network SHOULD default to operating offline.

Then this text said

> ... such functionality MUST be disabled by default to prefer offline operation.

gregsdennis: Regarding the redudancy, I'm happy leaving it as it's pertinent to both cases, and I don't feel that it's necessary to reference an entire section when all we need from it is a single sentence.  I think the duplication is fine in this case.

gregsdennis: > Otherwise we have security considerations that must be documented more visibly.

We are also looking at (separately) augmenting the security considerations section to address this.

karenetheridge: Yes, duplication in a single line is ok as long as it's consistent.
I'm still strongly in favour of "MUST default to offline" though.

gregsdennis: I'm inclined to agree that MUST is best here.

> This would mean you can't write a hyper-schema client that functions at all in it's default configuration. - @jdesrosiers 

I wouldn't think a hyper-schema client is strictly a JSON Schema implementation.  I would say that it _uses_ a JSON Schema implementation, even if that implementation is built in.  Since that's the case, the client would need to configure the schema implementation to access the network.

These requirements really only apply to JSON Schema implementations that are exposed as such.  If an implementation is built into some other functionality (e.g. a hyper-schema client) specifically to serve that functionality, then it makes sense for the implementation to take exception to things that hinder that functionality.

Thus, if your hyper-schema client needs its JSON Schema implementation to access the internet by default, go for it.  It's not violating the spec for doing so because no one is using it _as_ a JSON Schema implementation.

On the other hand, if you expose it _as_ a JSON Schema implementation, then when it's operating in that capacity, it needs to follow these requirements.

gregsdennis:
Also, digging this up from the abyss of forgetfulness: https://github.com/json-schema-org/json-schema-spec/issues/1231

jsonschema-core.md
jdesrosiers: I see what you're saying, but I think your definition of what constitutes an "implementation" is more narrow than what our charter indicates. Validators are not the only type of implementation this spec needs to support. We intend to enable and support other types of implementations that are not validators such as code generators, form generators, documentation generators, hyper-schema clients, and more that haven't been thought of yet.

The way I see it, the "core" spec is a common set of rules that applies to all types of JSON Schema implementations including non-validator implementations like hyper-schema. The "validation" spec is specific to validator implementations. There's currently a little cross over in that the applicator vocabulary is defined in "core", but that was done specifically because it was thought that any type of implementation would need applicators.

So, I can't agree that a hyper-schema client is not a JSON Schema implementation. Although it uses a validation implementation it is also in itself independently a non-validator JSON Schema implementation.

jdesrosiers: > Regarding the redudancy, I'm happy leaving it as it's pertinent to both cases, and I don't feel that it's necessary to reference an entire section when all we need from it is a single sentence. I think the duplication is fine in this case.

I don't see that these are two distinct cases. I think we're just saying the same thing in two different places. It doesn't need to be said twice.

jdesrosiers: > I think I prefer the wording that's there.

If you don't like the way I phrased it, that's fine, but it really needs to somehow acknowledge that the requirement is nonsense in some applications and that the requirement doesn't apply in those cases.

jdesrosiers:
> restore requirement to MUST

This is not what we discussed and I don't think there's strong support for strengthening this requirement. The discussion was about whether or not to remove the requirement. We didn't discuss strengthening it. My understanding is that Juan and I are in favor of this being entirely an implementation decision based on the unique needs of their use-case. Julian is a proponent of implementers having the freedom to make the choice, but thinks it should be discouraged. Greg and Karen and for strengthening the requirement. I'm not clear where others fall, but so far there's definitely not a clear consensus for strengthening this requirement.

jsonschema-core.md
gregsdennis: > I think your definition of what constitutes an "implementation" is more narrow than what our charter indicates.

https://github.com/json-schema-org/community/pull/325/files#diff-91d16dd057bd39f6431e90a1d78e9224eafafcedc611eba0878d77e6fda86601R10

> JSON Schema aims to enable the confident and reliable use of the JSON data format. It does this primarily by providing specification documents which define a declarative language that allows annotation and validation of JSON documents.

The (proposed) charter clearly states that the primary purposes are "annotation and validation."  Hypermedia and other concerns are explicitly listed as secondary.

> We intend to enable and support other types of implementations that are not validators such as code generators, form generators, documentation generators, hyper-schema clients, and more that haven't been thought of yet.

But not at the cost of compromising the integrity of the primary concerns.  Being able to reach out to the internets or access the file system is not a requirement of annotation or validation.  The current text _recognizes_ the need by permitting implementations to have an online mode, but I maintain that the default behavior must be offline.

Hyperschema, codegen, and all of those other uses are _uses_ of JSON Schema.  They are not targets.  I will not weaken the spec to accommodate secondary and tertiary uses for the spec.  Strictly speaking, none of these are JSON Schema implementations.  They _use_ JSON Schema.  As such, JSON Schema only needs to be defined to serve its primary purposes, and users of JSON Schema (including secondary specifications) need to work within those bounds.

gregsdennis:
@Relequestual I think we need a tie breaker from our leader.

Do you prefer MUST or SHOULD default to offline operation?

jsonschema-core.md
jdesrosiers: We should all agree what constitutes an "implementation". Let's discuss that topic at the next OCWM.

gregsdennis:
I should add that the current [spec](https://json-schema.org/draft/2020-12/json-schema-core.html#name-dereferencing) (2020-12) does have a note in the context of dereferencing:

> In the case of an evolving API described by Hyper-Schema, it is expected that new schemas will be added to the system dynamically, so placing an absolute requirement of pre-loading schema documents is not feasible.

I think this supports Jason's case for a SHOULD.

gregsdennis:
@jdesrosiers what are your thoughts on replacing that note with

> Implementations which have the ability to access externally defined resources MUST do so per the requirements outlined in {{loading}}.

It seems to fit in with this PR.

(then of course, we still need to resolve the SHOULD/MUST question for the loading section)

jdesrosiers:
> what are your thoughts on replacing that note with

If `{{loading}}` says something equivalent to that note, then that would be great. Currently that's not the case. If you like the words used in that note better than the ones I used in my suggestion, then by all means use those words instead, but I do think it needs to be said.

gregsdennis:
`{{loading}}` is the section containing second change in this PR.

jdesrosiers:
> `{{loading}}` is the section containing second change in this PR.

I know. Like I said, that section doesn't say anything equivalent to the note. The only way I'd be in favor of removing that note and pointing to that section is if we say something equivalent to the note in that section.

jsonschema-core.md
gregsdennis: I think for the purposes of this, we're on the same page on the definition of an implementation.  I think MUST is more appropriate, but I'm okay with this being a SHOULD.  Technically the hyper-schema bits can annotate an instance offline, but to do so outside of the context of a client (which definitely needs online access) doesn't really seem useful.

jdesrosiers: > Technically the hyper-schema bits can annotate an instance offline

Fetching a schema can't be completely decoupled from evaluation of the schema in the case of hyper-schema. The schema may include references to other schemas that also need to be fetched. You don't know about those additional schemas until you are evaluating the schema, so it's necessary for the fetch functionality to be integrated with the evaluation process.

jsonschema-core.md
```diff
@@ -1143,9 +1143,8 @@ provides the ability to validate recursive structures through self-reference.
 
 The resolved IRI produced by these keywords is not necessarily a network
 locator, only an identifier. A schema need not be downloadable from the address
-if it is a network-addressable URL, and implementations SHOULD NOT assume they
-should perform a network operation when they encounter a network-addressable
-IRI.
+if it is a network-addressable URL. Implementations which can access the network
+MUST default to operating offline.
```
jdesrosiers, line 1147: This needs to be a "SHOULD" to not contradict the "SHOULD" on 1285.

jdesrosiers (commented):
@gregsdennis, can you address my suggestion to not have this defined in two different sections? Do you think there is good reason for this separation?

> The two main changes here are addressing the same kind of thing, but in different sections far away from each other. I think it makes sense to address all of this in the same place. I think the "Loading Schemas" section is the best place. The "Schema References" section can link to the other section if we think that's necessary.

> > Regarding the redudancy, I'm happy leaving it as it's pertinent to both cases, and I don't feel that it's necessary to reference an entire section when all we need from it is a single sentence. I think the duplication is fine in this case.
>
> I don't see that these are two distinct cases. I think we're just saying the same thing in two different places. It doesn't need to be said twice.

gregsdennis:
> Do you think there is good reason for this separation?

The separation isn't the intent.  It's isolating the specific requirement.

We can really only reference an entire section, but what I specifically need is just "functionality SHOULD be disabled by default to prefer offline operation."  I don't think a lnik to another section reads well.  I'm happy to get a proposal.

jdesrosiers (approved):
>  I don't think a lnik to another section reads well. I'm happy to get a proposal.

I'll look into it another time. It's good enough for now and I don't want to hold this up any longer.

Issue #1444, opened 2023-09-20
Clean-up of stable features and removal of unstable features


gregsdennis:
In [What current features should be considered unstable?](https://github.com/orgs/json-schema-org/discussions/472#discussioncomment-6918674) we determined that most everything available in 2020-12 will be considered stable, with some things having a few changes.

This issue will catalog the things we need to do to resolve that discussion.

- [x] [[PR `$id`](https://github.com/json-schema-org/json-schema-spec/pull/1537)] Scrutinize `$schema` and `$id` to ensure that everything is right because we can't change it if it's wrong (@gregsdennis)
- [ ] [[Issue]](https://github.com/json-schema-org/json-schema-spec/issues/1140) [[PR]](https://github.com/json-schema-org/json-schema-spec/pull/1555) `$dynamic*` (@gregsdennis)
  - Remove initial resolution requirement from `$dynamicRef`
  - Remove `$anchor` behavior from `$dynamicAnchor`
  - Create clear separation between the two referencing systems (`$ref` vs `$dynamicRef`)
- [x] [[PR]](https://github.com/json-schema-org/json-schema-spec/pull/1506) Extract `propertyDependencies` (see also #1443) (@gregsdennis & @jdesrosiers)
- [x] [[PR]](https://github.com/json-schema-org/json-schema-spec/pull/1429) Extract output (@gregsdennis)
- [x] [[PR]](https://github.com/json-schema-org/json-schema-spec/pull/1445) Add language that makes it clear ~using annotations~ any mechanism used to describe keyword interaction is for illustration purposes and not a prescriptive implementation architecture (@gregsdennis)
- [x] [[PR]](https://github.com/json-schema-org/json-schema-spec/pull/1451) ~Update interaction description for `if`/`then`/`else` to use annotation mechanism~ (@gregsdennis)
- [x] ~Update interaction description for `contains`/`minContains`/`maxContains` to use annotation mechanism (requires #1358 first)~
- [x] [[PR](https://github.com/json-schema-org/json-schema-spec/pull/1541)] Ensure interaction description for `unevaluated*` (which currently uses annotation mechanism) is clear (@gregsdennis)
- [x] [[PR](https://github.com/json-schema-org/json-schema-spec/pull/1541)] Remove annotation dependency descriptions from keywords where it's not required ~(basically everything except `unevaluated*`)~ (@gregsdennis)
[...]

PR #1445, opened 2023-09-20
add section about keyword interaction mechanics being illustrative

gregsdennis:
Holding this PR based on the outcome of https://github.com/orgs/json-schema-org/discussions/491

Related to #1444

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -533,7 +533,7 @@ Unless otherwise specified, keywords act independently.
 
 This document uses annotations as a mechanism to describe the interactions
 between keywords, however this is for illustrative purposes and not
-prescriptive. Implementations SHOULD use whatever mechanisms makes sense given
+prescriptive. Implementations MAY use whatever mechanisms makes sense given
 their architecture and language needs in order to achieve the specified
 behaviors. Implementations are not required to collect annotations in order to
 manage keyword interaction.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -524,8 +524,8 @@ to inform the user of the evaluation path that produced the error or annotation.
 ### Keyword Interactions
 
 Keywords MAY modify their behavior based on the presence or absence of another
-keyword in the same [schema object](#schema-document). Such keywords MUST NOT
-result in a circular dependency.
+keyword or its value in the same [schema object](#schema-document). Such
+keywords MUST NOT result in a circular dependency.
 
 Unless otherwise specified, keywords act independently.
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -533,7 +533,7 @@ Supplementary specifications are encouraged to specify any dependencies as part
 of the dependent keyword (i.e. the keyword whose behavior is modified).
 
 Within this document, keyword dependencies are expressed using one of the
-following mechanics:
+following mechanisms:
 
 - Static dependencies, in which the dependency relies on the presence or
   contents of another keyword.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -523,11 +523,23 @@ to inform the user of the evaluation path that produced the error or annotation.
 
 ### Keyword Interactions
 
-Keyword behavior MAY be defined in terms of the annotation results of
-[subschemas](#root) and/or adjacent keywords (keywords within the same schema
-object) and their subschemas. Such keywords MUST NOT result in a circular
-dependency. Keywords MAY modify their behavior based on the presence or absence
-of another keyword in the same [schema object](#schema-document).
+Keywords MAY modify their behavior based on the presence or absence of another
+keyword in the same [schema object](#schema-document). Such keywords MUST NOT
+result in a circular dependency.
+
+Unless otherwise specified, keywords act independently.
+
+#### Annotations as an Interaction Mechanism
+
+This document uses annotations as a mechanism to describe the interactions
+between keywords, however this is for illustrative purposes and not
+prescriptive. Implementations SHOULD use whatever mechanisms makes sense given
+their architecture and language needs in order to achieve the specified
+behaviors. Implementations are not required to collect annotations in order to
+manage keyword interaction.
+
+Supplementary specifications are also encouraged to use annotations to describe
+any interactions of the keywords they define.
 
 ### Default Behaviors {#default-behaviors}
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -529,6 +529,8 @@ MUST NOT result in a circular dependency.
 
 Unless otherwise specified, keywords act independently.
 
+Supplementary specifications are encouraged to specify any interactive behavior as part of the keyword whose behavior is modified.
+
 #### Annotations as an Interaction Mechanism
 
 This document uses annotations as a mechanism to describe the interactions
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -542,7 +542,7 @@ following mechanisms:
   annotations produced by the keyword or the validation result of its
   subschema(s).
 
-These mechanics are used merely to describe dependencies; they are for
+These mechanisms are used merely to describe dependencies; they are for
 illustrative purposes and not prescriptive. Implementations MAY use whatever
 mechanisms make sense given the needs of their architecture and language in
 order to achieve the specified behaviors.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -523,25 +523,29 @@ to inform the user of the evaluation path that produced the error or annotation.
 
 ### Keyword Interactions
 
-Keywords MAY modify their behavior based on the presence, absence, or value of
-another keyword in the same [schema object](#schema-document). Such keywords
-MUST NOT result in a circular dependency.
-
 Unless otherwise specified, keywords act independently.
 
-Supplementary specifications are encouraged to specify any interactive behavior as part of the keyword whose behavior is modified.
+Keywords MAY modify their behavior based on the presence, absence, or value of
+another keyword in the same schema object. Such keywords MUST NOT result in a
+circular dependency.
 
-#### Annotations as an Interaction Mechanism
+Supplementary specifications are encouraged to specify any dependencies as part
+of the dependent keyword (i.e. the keyword whose behavior is modified).
 
-This document uses annotations as a mechanism to describe the interactions
-between keywords, however this is for illustrative purposes and not
-prescriptive. Implementations MAY use whatever mechanisms makes sense given
-their architecture and language needs in order to achieve the specified
-behaviors. Implementations are not required to collect annotations in order to
-manage keyword interaction.
+Within this document, keyword dependencies are expressed using one of the
+following mechanics:
 
-Supplementary specifications are also encouraged to use annotations to describe
-any interactions of the keywords they define.
+- Static dependencies, in which the dependency relies on the presence or
+  contents of another keyword.
+- Dynamic dependencies, in which the dependency relies on the evaluation of
+  another keyword against an instance. This dependency may be on either the
+  annotations produced by the keyword or the validation result of its
+  subschema(s).
+
+These mechanics are used merely to describe dependencies; they are for
+illustrative purposes and not prescriptive. Implementations MAY use whatever
+mechanisms makes sense given the needs of their architecture and language in
+order to achieve the specified behaviors.
 
 ### Default Behaviors {#default-behaviors}
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -523,9 +523,9 @@ to inform the user of the evaluation path that produced the error or annotation.
 
 ### Keyword Interactions
 
-Keywords MAY modify their behavior based on the presence or absence of another
-keyword or its value in the same [schema object](#schema-document). Such
-keywords MUST NOT result in a circular dependency.
+Keywords MAY modify their behavior based on the presence, absence, or value of
+another keyword in the same [schema object](#schema-document). Such keywords
+MUST NOT result in a circular dependency.
 
 Unless otherwise specified, keywords act independently.
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -544,7 +544,7 @@ following mechanics:
 
 These mechanics are used merely to describe dependencies; they are for
 illustrative purposes and not prescriptive. Implementations MAY use whatever
-mechanisms makes sense given the needs of their architecture and language in
+mechanisms make sense given the needs of their architecture and language in
 order to achieve the specified behaviors.
 
 ### Default Behaviors {#default-behaviors}


jsonschema-core.md
```diff
@@ -523,11 +523,23 @@ to inform the user of the evaluation path that produced the error or annotation.
 
 ### Keyword Interactions
 
-Keyword behavior MAY be defined in terms of the annotation results of
-[subschemas](#root) and/or adjacent keywords (keywords within the same schema
-object) and their subschemas. Such keywords MUST NOT result in a circular
-dependency. Keywords MAY modify their behavior based on the presence or absence
-of another keyword in the same [schema object](#schema-document).
+Keywords MAY modify their behavior based on the presence or absence of another
```
gregsdennis, line 526: Do we need to be explicit that keywords need to be defined so that this interaction is one-way?  E.g. the way it's defined now, `minContains` can modify the validation result of `contains`.  (I have an item in the linked issue to address this.)

jsonschema-core.md
```diff
@@ -523,11 +523,23 @@ to inform the user of the evaluation path that produced the error or annotation.
...
+keyword in the same [schema object](#schema-document). Such keywords MUST NOT
+result in a circular dependency.
+
+Unless otherwise specified, keywords act independently.
+
+#### Annotations as an Interaction Mechanism
+
+This document uses annotations as a mechanism to describe the interactions
+between keywords, however this is for illustrative purposes and not
+prescriptive. Implementations SHOULD use whatever mechanisms makes sense given
```
jdesrosiers, line 536: I'm not sure that using "SHOULD" makes sense here because it's telling you a specific thing that you should be doing. Maybe this makes more sense as "should". I think "MAY" would also work.

jsonschema-core.md
jdesrosiers: Isn't that what the "circular dependency" clause says? I might be missing the point.

jdesrosiers (approved):
lgtm. I just left one minor suggestion.

jsonschema-core.md
gregsdennis: No, by "one-way" I mean that a keyword can't be used to explicitly modify another keyword's validation result.  The logic has to be in the dependent keyword.

jsonschema-core.md
gregsdennis: I'll go with MAY.

jsonschema-core.md
jdesrosiers: I'm still not following the distinction. As originally defined, the result of `contains` depends on the annotations of `minContains` and the result of `minContains` depends on the annotations of `contains`. That sounds like the circular dependency that's mentioned. I'm not seeing what else might need to be said.

jdesrosiers, line 526: It's not just the "presence" of another keyword, it's also the "value" of that keyword that can modify behavior.

gregsdennis: I feel that may be implicit or understood.  I'm not sure why a keyword would be sensitive to the presence of another keyword but not its value.

jdesrosiers: I think it's implicit enough for informal discussion, but not for a specification. I think the spec should be explicit.

jdesrosiers: I think I see what you're saying. It's confusing to me because the direction of operation depends on your view point. `additionalProperties` changes it's result based on the value of `properties`, but if you look at it from a different perspective, `properties` changes the result of `additionalProperties`. That is, if I add a property to `properties`, the result of `additionalProperties` changes. `contains`/`minContains` is the same. We want people to specify that `contains` depends on `minContains`, not that `minContains` changes the result of `contains` even though both perspectives are technically correct.

I think different directions of operation are just different perspectives of the same behavior. I don't think that direction is the reason `contains`/`minContains` was problematic. It was problematic because it can circular dependencies.

Assuming I'm not still missing something, this sounds like guidance for spec authors and although useful, probably doesn't belong in the spec.

gregsdennis: Okay...  The different perspectives thing is technically correct, I guess, but I'm asking if it's important that keywords need to be specified so that the dependent keyword has the variable behavior.

If `foo` depends on `bar` in some way, the specification shouldn't say that `bar` changes the outcome of `foo`; it should say that `foo` can change its own behavior depending on `bar`.

jdesrosiers: > I'm asking if it's important that keywords need to be specified so that the dependent keyword has the variable behavior.

I think we should definitely stick to defining keywords that way, but that's more of spec author style guide requirement. It doesn't constraint implementations in any way, so I don't think it needs to be in the spec.

gregsdennis: > that's more of spec author style guide requirement

I see no reason why a spec can't contain requirements for extension specs.  I'm [doing that](https://github.com/json-schema-org/json-schema-spec/pull/1429/files#diff-5f8683f3cbc29d6dcf36b2664eda6d2efbd18a7074546d1c9105a60e239e5f51R2085-R2087) with the output PR.

In this section, Core is describing how keyword interaction works _in general_.  I think it's a good idea to set some guidelines for people who want to define their own keywords.

jsonschema-core.md
```diff
@@ -523,11 +523,23 @@ to inform the user of the evaluation path that produced the error or annotation.
 
 ### Keyword Interactions
 
-Keyword behavior MAY be defined in terms of the annotation results of
-[subschemas](#root) and/or adjacent keywords (keywords within the same schema
-object) and their subschemas. Such keywords MUST NOT result in a circular
-dependency. Keywords MAY modify their behavior based on the presence or absence
-of another keyword in the same [schema object](#schema-document).
+Keywords MAY modify their behavior based on the presence or absence of another
+keyword or its value in the same [schema object](#schema-document). Such
```
jdesrosiers, line 527: I'm finding this wording awkward. I suggest something more like this ...

```suggestion
Keywords MAY modify their behavior based on the presence, absence, or value of 
another keyword in the same [schema object](#schema-document). Such
```

jsonschema-core.md
jdesrosiers: I don't feel too strongly about this one, but I think it would be good to get some other opinions. My opinion is that this is an example of putting stuff in the spec that doesn't necessarily belong there because we currently have nowhere else to put that kind of content.

jsonschema-core.md
gregsdennis: Yeah, it's awkward, but I think it's more correct.  It's:

- the presence or absence of the keyword
- the keyword's value

I feel they're separate concepts and so need to be listed separately.

jdesrosiers: I disagree. Listing them together makes sense to me. 

* the presence of another keyword
* the absence of another keyword
* the value of another keyword

These things all seem to fit and make sense together.

jsonschema-core.md
Relequestual: > > I'm asking if it's important that keywords need to be specified so that the dependent keyword has the variable behavior.
> 
> I think we should definitely stick to defining keywords that way, but that's more of spec author style guide requirement. It doesn't constraint implementations in any way, so I don't think it needs to be in the spec.


I agree we should be defining keywords that way, and I feel that it should be in the specifciation too. The specification [defines how vocabularies are defined](https://json-schema.org/draft/2020-12/json-schema-core.html#name-schema-vocabularies). Therefore, it sets up the foundations for the JSON Schema ecosystem, not just how implementations should function. You may suggest that such content should be elsewhere, and the specification should only contain the requirements of an implementaiton, however some prose and context is helpful and relevant, which is why it's here. It doesn't make any requirements (We could if we wanted to), but some directions like this do belong in the spec IMHO.

jsonschema-core.md
Relequestual: For me personally, I feel like these are all attributes of a keyword, so a single comma separated list feels more correct and easier to read.
But, I don't see that it matters a great deal, and isn't a blocker for merging this PR.

jsonschema-core.md
gregsdennis: Relevant text is now at line 541.  Leaving it in as a soft recommendation for supplementary specs.

jsonschema-core.md
```diff
@@ -523,11 +523,23 @@ to inform the user of the evaluation path that produced the error or annotation.
...
+#### Annotations as an Interaction Mechanism
+
+This document uses annotations as a mechanism to describe the interactions
+between keywords, however this is for illustrative purposes and not
+prescriptive. Implementations MAY use whatever mechanisms makes sense given
+their architecture and language needs in order to achieve the specified
+behaviors. Implementations are not required to collect annotations in order to
+manage keyword interaction.
+
+Supplementary specifications are also encouraged to use annotations to describe
```
Julian, line 541: I would delete this line personally. I don't know that I believe that it's anything other than a stylistic choice to define keywords in this way, and I'm not really convinced it's clearer to do so in general -- worse so given we ourselves have had unexpected issues from having done so (c.f. https://github.com/orgs/json-schema-org/discussions/57)

I'm going to LGTM the PR anyhow, but yeah, I would delete this personally.

Julian (approved):

jsonschema-core.md
gregsdennis: I agree, that's why I didn't use SHOULD or make it any kind of hard requirement.  I think it's good for related specs to be aligned in their language.  It's just a gentle nudge.

jdesrosiers: > I would delete this line personally.

I would too. I too don't think describing behaviors using annotations is more clear, but I'm more concerned that this line (as well as describing keyword behaviors in terms of annotations in general) will (indirectly) encourage people to implement things in a specific way. I personally don't think encouraging stateful evaluation through annotations is a good choice because it can lead to less efficient implementations.

So, while this PR isn't exactly what I would want, I think it's a significant step in the right direction and I'll take what I can get.

gregsdennis:
@jdesrosiers @Julian

Based on my ramblings [here](https://github.com/orgs/json-schema-org/discussions/491#discussioncomment-7191252), what do you think of this (before I add it as a commit)?

> ### Keyword Interactions
> 
> Keywords MAY modify their behavior based on the presence, absence, or value of another keyword in the same schema object. Such keywords MUST NOT result in a circular dependency.
> 
> Unless otherwise specified, keywords act independently.
> 
> Supplementary specifications are encouraged to specify any dependencies as part of the dependent keyword (i.e. the keyword whose behavior is modified).
> 
> Within this document, keyword dependencies are expressed using one of the following mechanics:
> 
> - Static dependencies, in which the dependency relies on the presence or contents of another keyword.
> - Dynamic dependencies, in which the dependency relies on the evaluation of another keyword against an instance. This dependency may be on either the annotations produced by the keyword or the validation result of its subschema.
> 
> These mechanics are used merely to describe dependencies; they are for illustrative purposes and not prescriptive. Implementations MAY use whatever mechanisms makes sense given the needs of their architecture and language in order to achieve the specified behaviors.
> 
> #### ~Annotations as an Interaction Mechanism~

Julian:
That looks pretty good to me! Though the words "in the same schema object" at the beginning make me double take a bit (maybe those words were there before -- it seems so) -- `unevaluatedProperties` changes its behavior not just based on keywords in the same schema object but also whenever `properties` and friends appear potentially in subschema resources does it not?

But the rest certainly seems better than before to me, I have no suggestions for further improvements there. (Though I have to reread your other comment on the discussion)

gregsdennis:
> unevaluatedProperties changes its behavior not just based on keywords in the same schema object but also whenever properties and friends appear potentially in subschema resources does it not?

I think it's argued that "subschemas" is included in "values."

gregsdennis:
If we like this, then I need to create a follow-up PR that removes annotation language from all of the static dependency keywords (that's gonna be a lot).

Also #1451 will need to be closed and those sections possibly re-evaluated.  I'm not sure I like the way it's set up, and I think it could be improved.  We'll see.

jdesrosiers:
I think this is fantastic. I have one suggestion, but I'll save it for when you actually push the change. I'm 100% on board with the direction.

jsonschema-core.md
```diff
@@ -523,11 +523,29 @@ to inform the user of the evaluation path that produced the error or annotation.
...
+
+Keywords MAY modify their behavior based on the presence, absence, or value of
+another keyword in the same schema object. Such keywords MUST NOT result in a
+circular dependency.
+
+Supplementary specifications are encouraged to specify any dependencies as part
+of the dependent keyword (i.e. the keyword whose behavior is modified).
+
+Within this document, keyword dependencies are expressed using one of the
+following mechanics:
```
jdesrosiers, line 536: The word "machanics" feels a awkward to me, but I can't say why. I feel like "mechanisms" makes more sense.

```suggestion
following mechanisms:
```

jsonschema-core.md
```diff
@@ -523,11 +523,29 @@ to inform the user of the evaluation path that produced the error or annotation.
...
+  contents of another keyword.
+- Dynamic dependencies, in which the dependency relies on the evaluation of
+  another keyword against an instance. This dependency may be on either the
+  annotations produced by the keyword or the validation result of its
+  subschema(s).
+
+These mechanics are used merely to describe dependencies; they are for
+illustrative purposes and not prescriptive. Implementations MAY use whatever
+mechanisms make sense given the needs of their architecture and language in
+order to achieve the specified behaviors.
```
jdesrosiers, line 548: I'm not sure this paragraph is necessary anymore. These are more classifications of behavioral properties than they are mechanisms. They're certainly not implementation details. It seems redundant to state that implementations can implement them however they want because we haven't said anything that constrains how they would implement them. We've only defined their external behaviors.

jdesrosiers (commented):
Left a couple small notes. In addition to that, I think it would be good to also include a recommendation to strongly prefer static dependencies when designing keywords with interactions.

jsonschema-core.md
gregsdennis: I don't think it hurts to be explicit here.

The original concern about annotations being the description mechanism was that it would lead implementors into thinking they had to use annotations to drive these dependencies.  I don't think the other changes address that concern, but this paragraph does so explicitly.

gregsdennis:
> I think it would be good to also include a recommendation to strongly prefer static dependencies when designing keywords with interactions.

I'm iffy about this being in the spec.  I think it would be good as implementor guidance on the website.

This is a bit different from recommending that dependencies are defined as part of the dependent keyword, where the spec is defining more-or-less defining that one keyword can't actively modify other keyword's behavior; rather the second keyword modifies its own behavior based on the first.

jsonschema-core.md
jdesrosiers: If we keep it, it needs to be made more clear what it's trying to say because I can't make sense of it. The way I understand it, the things you're calling "mechanisms" now are not mechanisms in the same way that annotation are a mechanism. If they were, everything would make sense. Before, we were telling implementations _how_ to get the right outcome using annotations. Now, we're going to tell implementations what the outcome should be and let implementations figure out how to get to that outcome on their own. Since the new way already isn't telling implementations _how_ to get the outcome, it doesn't make sense to me that we're still telling them they can get to that outcome however they need to. We're no longer giving them a "how", so telling them that the "how" is just illustrative doesn't make sense.

I'm sure there's some piece of this that you understand differently that's leading us to different places.

jdesrosiers:
> > I think it would be good to also include a recommendation to strongly prefer static dependencies when designing keywords with interactions.
>
> I'm iffy about this being in the spec. I think it would be good as (advanced) user guidance on the website.

Well, I don't disagree with that, but I do think it's inconsistent with our recent decisions to include guidance in other places in the spec including this section.

> This is a bit different from recommending that dependencies are defined as part of the dependent keyword, where the spec is more-or-less defining that one keyword can't actively modify other keyword's behavior; rather the second keyword modifies its own behavior based on the first.

I disagree. We established that the dependency directionality is a matter of perspective and the recommendation is merely an authoring style choice. Warning people off of dynamic keywords that are likely to be more difficult to implement and slower to evaluate than static keywords seems like more valuable guidance to include.

jsonschema-core.md
gregsdennis: What's unclear about it?  We invent some mechanics to describe behavior and then say they're only descriptive and implementations can do what they want as long as the behavior is met.

> We're no longer giving them a "how"

We _are_ giving them a "how."  It's just the "how" isn't always annotations.  Now it can also be "this keyword looks at that keyword."

gregsdennis:
> Warning people off of dynamic keywords that are likely to be more difficult to implement and slower to evaluate than static keywords seems like more valuable guidance to include.

I disagree.  Warning people off dynamic keywords limits the keywords that people propose.  Dynamic keywords aren't inherently more difficult to implement or slower to evaluate.  They may be in your implementation (and perhaps others), but that's not universal.

jsonschema-core.md
Relequestual: > The way I understand it, the things you're calling "mechanisms" now are not mechanisms in the same way that annotation are a mechanism.

I disagree. "Dynamic dependencies" even explicitly includes annotations as one of the sub-parts that may be used for determinatioin.

In computer science, a mechanisum is any function that produces an output. We define two classes of mechanisms, but that doesn't negate the fact that there are mechanisms.

Besides, if you take `unevaluatedProperties` as an example, it still defines using annotations as the mechanism for determining its assertion result.

> Before, we were telling implementations how to get the right outcome using annotations.

Aren't we still doing that?

I'm really struggling to reach any other conclusion. Can you help me understand how you reached yours @jdesrosiers?

Relequestual (approved):

jsonschema-core.md
jdesrosiers: > We _are_ giving them a "how." It's just the "how" isn't always annotations. Now it can also be "this keyword looks at that keyword."

This is the part I'm not following. "This keyword looks at that keyword" doesn't say what mechanism is used to do the looking. That statement describes the behavior, not the mechanism by which the behavior is achieved. It's like saying "Greg goes to the store". The mechanism by which he gets to the store is implicitly up to him. That statement doesn't describe the mechanism he should use to get to the store. He might drive, walk, ride a bike, take a bus, etc.

We're saying that there are two categories of keyword dependencies. One category is where the keyword depends on the value of the dependent keyword. The other category is where where the keyword depends on the value of the dependent keyword as well as an instance. Those are statements of dependency. I don't see how they can be interpreted as mechanisms (or imply a mechanism) for how that dependency is implemented.

> > Before, we were telling implementations how to get the right outcome using annotations.
>
> Aren't we still doing that?

No, the [plan](https://github.com/json-schema-org/json-schema-spec/pull/1445#issuecomment-1747716533) is to undo that except possibly in the case of `unevaluated*` whose definition is dependent on annotations.

> if you take `unevaluatedProperties` as an example, it still defines using annotations as the mechanism for determining its assertion result.

Right, but that's a different context. This section defines two "mechanisms": static keyword dependencies and dynamic keyword dependencies. Those are the "mechanisms" that are the subject of this passage. Annotations are no longer the subject. Is the point of that passage to say that keywords may describe their behavior in terms of some mechanism, but if they do implementations don't need to use that mechanism? If so, that makes perfect sense. That needs to be said either in this section or by the keyword definition using a mechanism to define behavior. But, that's not what this passage currently says.

> I'm really struggling to reach any other conclusion. Can you help me understand how you reached yours @jdesrosiers?

I too am trying very hard to understand this from another perspective. I specifically need help understanding how a statement of dependencies can be interpreted as a mechanism. I've laid out how I have reached by conclusions in this and the previous comments in this thread. I don't know how else to explain it. The only explanation for this misunderstanding is that there is some underlying assumption that we're making differently and not realizing. We are all coming to logical conclusions based on our assumption, but end up with different conclusions because our assumptions are different. I don't know how to tease out that fundamental difference of understanding.

Ultimately, I don't want this to become another big argument. I'm simply trying to understand what that passage is trying to say. The way I understand the passage, it's meaningless, so I'm not concerned about it doing harm. I just want to understand it's intent or remove it if it doesn't make sense.

jdesrosiers:
> Dynamic keywords aren't inherently more difficult to implement or slower to evaluate.

I disagree. I wasn't making a personal statement here. I believe that dynamic keywords are in general more difficult to implement and slower to evaluate. I certainly make `unevaluated*` harder due to my architectural choices, but I'm that's not what if talking about.

Dynamic keywords inherently require either maintaining state and managing order of evaluation or duplicating work. Both of those options carry complexity and performance costs. Implementing state management and managing order of operations adds complexity to your architecture and there is a performance cost associated with maintaining state that I've found to be significant. Duplicating validation can incur complexity in a different way and has a performance impact because evaluations are repeated. I acknowledge that it may be relatively easy to implement `unevaluated*` if you already have state management and order of evaluation as part of your architecture, but you still pay the cost of the added architectural complexity and performance to have those things.

From an anecdotal perspective, I think the current state of the ecosystem shows that implementers find dynamic dependencies more difficult to implement as they are the least likely to be implemented in implementations that claim partial support of 2019-09 and 2020-12 and more likely to be [complained](https://github.com/orgs/json-schema-org/discussions/491#discussioncomment-7210533) about by implementers trying to support the later versions. I would say the only thing in JSON Schema that gives implementers as much trouble as dynamic keywords is `$id`.

None of this is particularly relevant, but I wanted to respond to the accusation that my position is based on self-interest. The real issue here is whether or not we put guidance in the spec. Generally, I think no. I would hope that when we do, it's limited to times where we the guidance is universally agreed on. That's not the case here, but it's also not been the case for other guidance we recently decided should be in the spec, so it seems reasonable to consider if we're being consistent.

jsonschema-core.md
gregsdennis: > Is the point of that passage to say that keywords may describe their behavior in terms of some mechanism, but if they do implementations don't need to use that mechanism? If so, that makes perfect sense. That needs to be said either in this section or by the keyword definition using a mechanism to define behavior. But, that's not what this passage currently says. - @jdesrosiers 

I'm not sure what this passage says to you, but what you described is exactly its purpose: implementations can do what they want so long as the behavior is achieved.

jsonschema-core.md
```diff
@@ -523,11 +523,29 @@ to inform the user of the evaluation path that produced the error or annotation.
...
-Keyword behavior MAY be defined in terms of the annotation results of
-[subschemas](#root) and/or adjacent keywords (keywords within the same schema
-object) and their subschemas. Such keywords MUST NOT result in a circular
-dependency. Keywords MAY modify their behavior based on the presence or absence
-of another keyword in the same [schema object](#schema-document).
+Unless otherwise specified, keywords act independently.
+
+Keywords MAY modify their behavior based on the presence, absence, or value of
+another keyword in the same schema object. Such keywords MUST NOT result in a
+circular dependency.
```
notEthan, line 530: It seems unexpected to have a MUST NOT here about this. It's not wrong, but pretty much all the other MUSTs are either what implementations must do, or what schemas must conform to. This is something another vocabulary or keyword _specification_ must do, so only relevant for a minuscule fraction of readers. Not that it shouldn't be here, but maybe some clarification of the context, what this MUST applies to, would help.

gregsdennis: This was already present, I just flipped the sentences because it made more sense to me and felt less redundant this way.

gregsdennis: > Not that it shouldn't be here, but maybe some clarification of the context, what this MUST applies to, would help.

I have a bigger issue with the flow of the document, and I intend to make some bigger changes to the organization.  I think this is better saved for when I tackle that.

notEthan: Sounds good, no objection to that.

jsonschema-core.md
gregsdennis: (I did update "mechanics" to "mechanisms" to be consistent with the rest of the section.)

gregsdennis:
Build is running on an out of date JP spec.  The `main` branch has been updated.

PR #1449, opened 2023-09-24
add proposal document template

### LINKED ISSUE #1443: Should we have a supplementary spec template for proposals?
Since we've decided that new feature proposals will be defined in external documents, should we have a template document?

gregsdennis:
Resolves #1443 

I started working on the `propertyDependencies` extraction and the template kinda just worked itself out.  It turns out leaving questions is easier that actually filling out the sections.

### DIFF
--- a/proposals/proposal-template.md
+++ b/proposals/proposal-template.md
@@ -19,7 +19,7 @@ This document proposes a change to the ??? specification by adding ???.
 
 <!--
 Status is free text: just something to explain where the proposal is in
-its journey.  This status does not align with the SLDC feature life cycle.
+its journey.  This status does not align with the SDLC feature life cycle.
 -->
 
 ## Note to Readers
@@ -60,25 +60,26 @@ also apply to this document.
 
 ## Change Details
 
-### Target
-
 <!--
-Where does this change go?
+This is where the specification changes are defined. This must be precise as
+these changes will be made verbatim.
 
 For example
 
-    This proposal will add the {{propertyDependencies}} section contained herein as
-    a subsection of JSON Schema Core, section 10.2.2 "Keywords for Applying
-    Subschemas Conditionally."
-
+1. The following section will be added to the JSON Schema Core specification as
+   a subsection of "Keywords for Applying Subschemas Conditionally".
+    > ### {New section name}
+    >
+    > {Feature description}
+2. The following subschema will be added to the Applicator Vocabulary schema,
+   `https://json-schema.org/<version>/<release>/meta/applicator`, at
+   `/properties/{keyword}`:
+    ```jsonc
+    {
+      // keyword schema
+    }
+    ```
 -->
-
-This proposal will add the {{???}} section contained herein as
-a subsection of ???.
-
-### ???
-
-<!-- What is the text that will appear in the specification? -->
 
 ## [Appendix] Change Log
 
--- a/proposals/proposal-template.md
+++ b/proposals/proposal-template.md
@@ -15,15 +15,12 @@ For example
 
 This document proposes a change to the ??? specification by adding ???.
 
-## Status
+## Current Status
 
-**Current Status**: PROPOSAL
-
-<!-- TODO: We should have a short standard blurb outlining the stages involved in a
-feature making its way to stable status. -->
-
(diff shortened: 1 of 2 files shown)


gregsdennis:
See ☝️ for an example of this template in use.

jdesrosiers (approved):

gregsdennis:
We're having some debate on the purpose of this document over on #1450.  Holding this until that resolves.

Relequestual:
Based on the discussion and activity at #1450, I feel this PR needs a suplementary document which explains the purpose of the proposal document (or documents if that turns out to be the case).

I would think such discussion needs to be had at #1443 and agreed on before making further changes here.

I'm happy to facilitate these discussions, either in text form or via calls. @gregsdennis @jdesrosiers let me know which you would prefer. (Anyone else is also welcome to be involved, but Greg and Jason are the primary individuals driving this right now.)

proposals/proposal-template.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Status
+
+**Current Status**: PROPOSAL
+
+<!-- TODO: We should have a short standard blurb outlining the stages involved in a
+feature making its way to stable status. -->
+
+<!-- TODO: Link to a document that describes the proposal => stable process in
+detail. -->
```
gregsdennis, line 26: Starting to think this can go away completely.  The document _is_ a proposal.

Maybe we want to keep it around for posterity, but I think we could use file paths to handle that.  Maybe something like moving the files to a `completed/<release>/` folder and renaming the file to `<filename>-rejected.md` / `<filename>-accepted.md`.

Personally, I'm happy with just deleting the file once we don't need it anymore, whether it was rejected or accepted.

proposals/proposal-template.md
```diff
@@ -0,0 +1,98 @@
...
+
+<!-- What problem exists that needs solving? -->
+
+### Solution
+
+<!-- What is the solution?  Include examples of use. -->
+
+### Alternatives
+
+<!-- What other options have been considered? (summary, not detailed) -->
```
gregsdennis, line 62: Per https://github.com/json-schema-org/json-schema-spec/issues/1443#issuecomment-2099427543, maybe this should go.  This would be important for an ADR, but it's not necessarily something that the public needs to see.

proposals/proposal-template.md
jdesrosiers: I think it would be a good idea to use the status field to communicate the maturity of the proposal. For example, something like,

1. Development -- The proposal is in the early stages of being considered
2. Release Candidate -- No further changes are expected, but we want to see as much implementation and usage to flush out any remaining issues
3. Awaiting Release -- The proposal has been excepted and will be included in the next release. There will be no more changes.
4. Released
5. Rejected

We don't even need to have a fixed list of statuses we're limited to. I think what's important is that it communicate something useful.

As for whether to keep around completed proposals, I think it's best to keep them around. Technically, they're never gone because they're in the git history, but people aren't going to go digging through our git history to find a proposal they're watching. I'm imagining someone trying to find the status of a proposal they're interested in and being confused because they can't find it or any explanation why it disappeared. For the same reason, I think it's best not to move them to a "rejected" or "completed" folder because that would break any bookmarks people have to those proposals. They can maybe be removed after some generous period of time, but not right away.

proposals/proposal-template.md
jdesrosiers: I'm still not entirely sure where you want to draw the line between the two documents, but I agree that a discussion of the alternatives isn't necessary in this document except maybe as an appendix.

proposals/proposal-template.md
gregsdennis: I'm happy keeping this as an open text field.  I'll update so that's clear.

gregsdennis:
@jdesrosiers I've also updated the process document to provide direction around creating an ADR.

proposals/proposal-template.md
gregsdennis: Please see my https://github.com/json-schema-org/json-schema-spec/issues/1443#issuecomment-2108834298

jdesrosiers (approved):

PR #1451, opened 2023-09-26
specify if, then, and else together using annotations as interaction mechanism

gregsdennis:
Holding this PR based on the outcome of https://github.com/orgs/json-schema-org/discussions/491

Depends on #1445 
Relates to #1444

I kinda went a bit extra on this.  It seemed more appropriate (because they really work as a single unit) to specify all three together in one section rather than to keep them separate.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1714,62 +1714,35 @@ against the schema defined by this keyword.
 
 #### Keywords for Applying Subschemas Conditionally {#conditional}
 
-Three of these keywords work together to implement conditional application of a
-subschema based on the outcome of another subschema. The fourth is a shortcut
-for a specific conditional case.
+These keywords contain subschemas which are conditionally applied based on the
+instance.
 
-`if`, `then`, and `else` MUST NOT interact with each other across subschema
-boundaries. In other words, an `if` in one branch of an `allOf` MUST NOT have an
-impact on a `then` or `else` in another branch.
+##### `if`, `then`, and `else`
 
-There is no default behavior for `if`, `then`, or `else` when they are not
-present. In particular, they MUST NOT be treated as if present with an empty
-schema, and when `if` is not present, both `then` and `else` MUST be entirely
-ignored.
+These three keywords work together to implement conditional application of a
+subschema based on the outcome of another subschema.
 
-##### `if`
+The value for each of these keywords value MUST be a valid JSON Schema.
 
-This keyword's value MUST be a valid JSON Schema.
+The `if` keyword produces an annotation which is the boolean validation result
+of its subschema against the instance. The validation outcome of the `if`
+keyword's subschema MUST NOT directly effect the overall validation result.
 
-This validation outcome of this keyword's subschema has no direct effect on the
-overall validation result. Rather, it controls which of the `then` or `else`
-keywords are evaluated.
+The `then` and `else` keywords each consume the annotation from an adjacent `if`
+keyword and conditionally apply their subschemas to the instance as follows:
 
-Instances that successfully validate against this keyword's subschema MUST also
-be valid against the subschema value of the `then` keyword, if present.
+- When the annotation from the `if` keyword is `true`, the `then` keyword's
+  subschema MUST be applied to the instance and the `else` keyword's subschema
+  MUST be ignored.
+- When the annotation from the `if` keyword is `false`, the `else` keyword's
+  subschema MUST be applied to the instance and the `then` keyword's subschema
+  MUST be ignored.
+- When there is no adjacent `if` keyword, no annotation can be produced, so both
+  the `then` and `else` keywords' subschemas MUST be ignored.
 
-Instances that fail to validate against this keyword's subschema MUST also be
-valid against the subschema value of the `else` keyword, if present.
-
-If [annotations](#annotations) are being collected, they are collected from this
-keyword's subschema in the usual way, including when the keyword is present
-without either `then` or `else`.
-
-##### `then`
-
-This keyword's value MUST be a valid JSON Schema.
-
-When `if` is present, and the instance successfully validates against its
-subschema, then validation succeeds against this keyword if the instance also
-successfully validates against this keyword's subschema.
-
-This keyword has no effect when `if` is absent, or when the instance fails to
-validate against its subschema. Implementations MUST NOT evaluate the instance
-against this keyword, for either validation or annotation collection purposes,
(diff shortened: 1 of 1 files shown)


jsonschema-core.md
```diff
@@ -1714,62 +1714,35 @@ against the schema defined by this keyword.
 
 #### Keywords for Applying Subschemas Conditionally {#conditional}
 
-Three of these keywords work together to implement conditional application of a
-subschema based on the outcome of another subschema. The fourth is a shortcut
-for a specific conditional case.
+These keywords contain subschemas which are conditionally applied based on the
```
Julian, line 1717: I suppose this is a perfect example to me of https://github.com/json-schema-org/json-schema-spec/pull/1445#discussion_r1343191641 (I didn't look at this PR first, promise :)

But to me, this makes things way *less* clear than what was here I think.

Just my opinion of course.

jsonschema-core.md
```diff
@@ -1714,62 +1714,35 @@ against the schema defined by this keyword.
...
-valid against the subschema value of the `else` keyword, if present.
-
-If [annotations](#annotations) are being collected, they are collected from this
-keyword's subschema in the usual way, including when the keyword is present
-without either `then` or `else`.
-
-##### `then`
+The `if` keyword produces an annotation which is the boolean validation result
+of its subschema against the instance. The validation outcome of the `if`
+keyword's subschema MUST NOT directly effect the overall validation result.
```
Julian, line 1729:
```suggestion
keyword's subschema MUST NOT directly affect the overall validation result.
```

gregsdennis: Odd.  I don't usually get that wrong.

jsonschema-core.md
gregsdennis: Is this a comment on this line or the concept of the PR as a whole?

Julian: PR as a whole, sorry (for being unclear).

gregsdennis: The purpose of this change is actually to make the interaction of `if` with `then` and `else` more consistent.

All other keyword interactions (except `contains`/`minContains`, which I'm planning a PR for) are _described_ using annotations as the communication mechanism.  Implementations are required to actually do it that way, but it serves as a good description.  That's what https://github.com/json-schema-org/json-schema-spec/pull/1445 is about.

As they're currently defined, `then` and `else` depend on the _evaluation result_ of `if`.  Depending on an evaluation result of a sibling keyword doesn't exist anywhere else in the spec, and there's not a defined communication mechanism to describe that behavior.  But there is such a mechanism using annotations.  So redefining `then` and `else` to depend on the annotation result of `if` (which _is_ the evaluation result) makes sense.

Julian: To me -- again just me -- it is undoubtedly going to be clear to *current* readers of the spec what the intended behavior is, regardless of formality. It's the way any programmer really thinks about these things -- annotations are (I think incontrovertibly) not how anyone already knows if/then/else to work.

I do hear you on consistency -- my personal preference would be to go in the *other* direction and reduce the number of other keywords defined using annotations. Again though, it's a stylistic thing, I find things generally clearer when they're not mentioned, but it's obviously easier case by case to look at something and compare.

gregsdennis: > my personal preference would be to go in the other direction and reduce the number of other keywords defined using annotations.

I'm open to this, but we'd need to either:

- define that keywords can depend on the evaluation result of other keywords
- describe this interaction in some other way

Currently, only "presence, absence, and value" dependencies are defined (which means I need to actually add "annotation result" to that list in #1445).

If we decide to backtrack on the annotations-as-communication-mechanism thing, I'd want to be sure that all keywords can be described without them.  The difficult ones are `unevaluated*`.

I'm going to hold these PRs and start a discussion.

jsonschema-core.md
```diff
@@ -1714,62 +1714,35 @@ against the schema defined by this keyword.
...
-valid against the subschema value of the `else` keyword, if present.
-
-If [annotations](#annotations) are being collected, they are collected from this
-keyword's subschema in the usual way, including when the keyword is present
-without either `then` or `else`.
-
-##### `then`
+The `if` keyword produces an annotation which is the boolean validation result
+of its subschema against the instance. The validation outcome of the `if`
+keyword's subschema MUST NOT directly affect the overall validation result.
```
Relequestual, line 1729: Reading this, remembering [these tests](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/unevaluatedProperties.json#L449-L615) which make sure annotation collection happens through `if` for the purposes of determining the validation result of `unevaulatedProperties`, I think there's an opportunity to improve this phrasing.

I realise the phrasing is almost identical to what we had, however what was intedned is that not that it "MUST NOT directly affect the overall validation result", but rather that it "MUST NOT be used as an assertion.".

I don't know if that's less or more confusing, in truth.

When I started writing this suggestion, I thought it was obvious, but turned out to be less so. But I still thought it could be improved.

If anyone thinks it doesn't make it clearer, please just resolve the comment.

gregsdennis: > make sure annotation collection happens through if for the purposes of determining the validation result of unevaulatedProperties

That's why I said "directly affect."  What you've mentioned is an indirect effect.

Relequestual (approved):
I also checked if it impacted [these tests](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/unevaluatedProperties.json#L449-L615), and it does not.

jsonschema-core.md
Relequestual: Right. I think I have trouble passing "directly" the way it's intended. Maybe I need more coffee.

jdesrosiers (commented):
I'm not a fan of defining these keywords together. I think it was better defined as three separate sections with each keyword address its own concerns. I think the way `then` and `else` were defined were fine, but `if` shouldn't mention `then` and `else`. `if` is not dependent on `then` or `else`, so shouldn't mention them.

As for defining this interaction using annotations, I agree with Julian in that I don't think it's necessary and I'd rather it didn't (same for any keyword interactions that don't necessarily need be defined using annotations), but I'm reluctantly ok with defining it that way as long as we don't require implementations literally use the annotation method.

gregsdennis:
Closing this PR as the direction has changed.  These keywords may still need to be reworked, but using annotations isn't the way.

PR #1452, opened 2023-09-27
Revert "`contains` applies to objects"

### LINKED ISSUE #1358: Reconsider allowing "contains" to apply to objects
During yesterday's meeting, I brought up that in the time since we decided to allow `"contains"` to apply to objects, we've shifted to more of an emphasis on stability.

`"contains"` was added in draft-06, and while the details of its interactions with `"minContains"` and `"maxContains"` (added in 2019-09) have changed, its assertion behavior on its own has not.  Allowing these three keywords to apply to objects is a breaking change.

We could instead revert them to apply only to arrays and introduce keywords named something like `"containsPropertyValue"`, `"minContainsPropertyValue"`, `"maxContainsPropertyValue"` alongside them.  I'll let others sort out the ideal names.  `"containsProperty"` might be fine but also feels a bit like it is talking about property names rather than values.  Another option could be `"hasValue"` or `"hasPropertyValue"` or something of that sort.  All of the name options are less aesthetically appealing than having `"contains"` work on objects, but they'd preserve more stability.

I don't have an incredibly strong opinion on this, but it seemed to get some traction in the call so I thought I'd file it.

gregsdennis:
Resolves #1358
Relates to #1444

I think I got everything, but maybe another pair of eyes could help.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -2072,25 +2072,25 @@ The value of `unevaluatedProperties` MUST be a valid JSON Schema.
 
 The behavior of this keyword depends on the annotation results of adjacent
 keywords that apply to the instance location being validated. Specifically, the
-annotations from `properties`, `patternProperties`, `contains`, and
-`additionalProperties`, which can come from those keywords when they are
-adjacent to the `unevaluatedProperties` keyword. Those four annotations, as well
-as `unevaluatedProperties`, can also result from any and all adjacent [in-place
+annotations from `properties`, `patternProperties`, and `additionalProperties`,
+which can come from those keywords when they are adjacent to the
+`unevaluatedProperties` keyword. Those four annotations, as well as
+`unevaluatedProperties`, can also result from any and all adjacent [in-place
 applicator](#in-place) keywords. This includes but is not limited to the
 in-place applicators defined in this document.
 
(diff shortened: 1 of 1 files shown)


jsonschema-core.md
```diff
@@ -1970,24 +1970,23 @@ The minimum number of occurrences is provided by the `minContains` keyword
 within the same schema object as `contains`. If `minContains` is absent, the
 minimum number of occurrences MUST be 1.
 
-Implementations MAY implement the dependency on `minContians` and `maxContains`
+Implementations MAY implement the dependency on `minContains` and `maxContains`
 by inspecting their values rather than reading annotations produced by those
 keywords.
```
jdesrosiers, line 1975: Do we need to say this? We've already stated elsewhere that annotations are just an explanatory device and implementations aren't required to use annotations for inter-keyword communication.

jdesrosiers (changes requested):

jsonschema-core.md
gregsdennis: That's to be addressed in a follow-up PR.  I'm just removing objects here.

jdesrosiers (approved):

gregsdennis:
Can't seem to re-run the min open time job.  Going to close and reopen to see if that triggers it.

PR #1457, opened 2023-10-06
move output spec to `/output` (cleanup of root)

gregsdennis:
Figured it'd be better in it's own folder than with the primary specs.  The folder already existed and housed the schema file.


gregsdennis:
Closing and reopening to re-run the jobs

jviotti (approved):

PR #1462, opened 2023-11-01
remove `$vocabulary` from vocab meta-schemas (2020-12)

gregsdennis:
Relates to [Discussion #511](https://github.com/orgs/json-schema-org/discussions/511).


gregsdennis:
Interesting... this branch is from before we had workflows, so no checks.

jdesrosiers (approved):

PR #1463, opened 2023-11-03
Update the link to the JSON Schema website.

Julian:
Refs: json-schema-org/website#201


Julian:
(I'm merging these to get them off my PR tab in GitHub which they've been sitting on, under the assumption that they don't require review and are trivially correct. The only reason they were left open in the first place is the slightly silly merge block which doesn't allow merging without doing an admin force.)

PR #1467, opened 2023-11-03
Rename metaschema's `$defs/simpleTypes` to singular

notEthan:
This is somewhat trivial but I think this is more correct. The string is a simple type, not multiple types.

### DIFF
--- a/meta/validation.json
+++ b/meta/validation.json
@@ -11,10 +11,10 @@
     "properties": {
         "type": {
             "anyOf": [
-                { "$ref": "#/$defs/simpleTypes" },
+                { "$ref": "#/$defs/simpleType" },
                 {
                     "type": "array",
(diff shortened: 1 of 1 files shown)


gregsdennis:
I think it's correct as is.  It's common to name an enumeration type as the plural, effectively stating, "There are the available _X_."  In this case, this definition is listing the available types.

notEthan:
All the other $defs names refer to the contents of the instance, not what's in the schema. I have only ever seen conventions of naming $defs names per the instance described.
But, I thought this would be trivial/noncontroversial, not particularly wanting to debate it, so if that is not accurate, I am not bothered if you close this (or I can).

gregsdennis:
I'm happy to get other opinions, but plural makes most sense to me.

jdesrosiers:
Closing due to inactivity.

PR #1479, opened 2024-01-19
update RJP 2023 to 2024

gregsdennis:
Just updating dates to fix the build.


jdesrosiers (approved):
I think this is one of those times we can skip the minimum review period.

Issue #1480, opened 2024-01-20
Omitted `prefixItems` vs empty


notEthan:
The spec says of `prefixItems`:

> The value of "prefixItems" MUST be a non-empty array of valid JSON Schemas.

Then ends with:

> Omitting this keyword has the same assertion behavior as an empty array.

Specifying that an omitted `prefixItems` behaves the same as a value that it must not be looks wrong. Comparing to `allOf`/`anyOf`/`oneOf` which also must be non-empty arrays of schemas, they omit the sentence about omitting the keyword. 

I'm not sure why any of these must be non-empty - sure, when empty they end up either just allowing everything (`prefixItems`, `allOf`) or nothing (`anyOf`, `oneOf`), but plenty of other keywords allow similar degenerate cases, e.g. `minItems` may be 0 allowing everything and `enum` may be empty allowing nothing. That's just an observation, not a battle I am itching to take on, and I'm sure it's all been discussed before.

Related: https://github.com/json-schema-org/json-schema-spec/issues/975 / https://github.com/json-schema-org/json-schema-spec/pull/1001

notEthan:
I suppose this is straightforward enough I'll fix & PR

PR #1481, opened 2024-01-20
Fix `prefixItems` keyword omitted behavior

notEthan:
Fixes https://github.com/json-schema-org/json-schema-spec/issues/1480 to omit the sentence about omitting the keyword, as `allOf` etc do.
Also fixes a small markdown error while I'm touching that section.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1832,7 +1832,7 @@ results.
 
 ##### `prefixItems`
 
-The value of "prefixItems` MUST be a non-empty array of valid JSON Schemas.
+The value of `prefixItems` MUST be a non-empty array of valid JSON Schemas.
 
 Validation succeeds if each element of the instance validates against the
 subschema at the same position, if any. This keyword does not constrain the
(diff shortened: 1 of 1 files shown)


jsonschema-core.md
```diff
@@ -1844,8 +1844,6 @@ this keyword applied a subschema. The value MAY be a boolean true if a subschema
 was applied to every index of the instance, such as is produced by the `items`
 keyword. This annotation affects the behavior of `items` and `unevaluatedItems`.
 
-Omitting this keyword has the same assertion behavior as an empty array.
```
gregsdennis, line 1847: We still need to specify what the behavior of omitting the keyword is.  (Either that or remove all other omission clauses; probably not.)

notEthan: Compare to [`allOf`](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#name-allof) and friends, just removing it is consistent with how they are. Plenty of other keywords don't include a comparable sentence.

karenetheridge (approved):

jdesrosiers (approved):
Thanks for catching this and for writing up the fix!

PR #1483, opened 2024-02-02
Evaluation path example fix to use pointer rather than URI fragment syntax

notEthan:
per the "Evaluation path" section stating "The value MUST be expressed as a JSON Pointer"


gregsdennis (approved):
This section will likely be reworked, but for now this change looks right to me.

gregsdennis:
Closing and opening to rerun jobs.

Issue #1065, opened 2021-01-15
Add keyword id to output unit


jdesrosiers:
A schema using dialect-a can `$ref` a schema using dialect-b but, there is nothing in the output format that indicates the dialect that was used to determine each output unit result. This can be important for keywords whose semantics have changed over time, but their name hasn't changed. The ones that come to mind immediately are `maximum`/`exclusiveMaximum`, `minimum`/`exclusiveMinimum`, `format`, and `links`.

We could potentially determine the dialect for an output unit by using the `absouteKeywordLocation` to lookup the schema and inspect it's `$schema` property, but that has two problems. First, `absoluteKeywordLocation` is not required. Second, if the output is passed to a third party for processing, it might not have access to the original schema to look up it's dialect.

The simplest approach would be to add a field such as `dialect` to the output unit. However, It would be more useful to have a unique identifier for each keyword. That way, when we introduce a new draft, if a keyword doesn't change, it's ID doesn't have to change. That would make it easier for tools to do things with annotations where they don't care what dialect they are as long as they have the same semantics (such as `title` and `description`).

gregsdennis:
We should probably properly define a way to identify a dialect first.  From [2020-12](https://github.com/json-schema-org/json-schema-spec/blob/6c5b6729ce7482072404b09c02533291e9c528c7/jsonschema-core.xml#L144-L145):

> A dialect is defined as a set of vocabularies and their required support identified in a meta-schema.

It has no specific identifier, but perhaps the meta-schema `$id` (specified by `$schema` in the schema) could suffice since this effectively describes this set of vocabularies.

> First, `absoluteKeywordLocation` is not required. - @jdesrosiers

The reason it's not required is because the relative location is sufficient when the navigated path contains no references.

That said, maybe it's sufficient to include the `$schema` value and/or `$vocabulary` value in the cases where there is ambiguity because

- a dereference takes you to a new dialect (different meta-schema),
- a keyword contains multiple semantic definitions due to the specific set of vocabularies include (this also brings up the namespacing concept we had discussed somewhere else).

jdesrosiers:
I think identifying a dialect is sufficiently defined.

> The "$schema" keyword is both used as a JSON Schema dialect identifier and
> as the identifier of a resource which is itself a JSON Schema, which describes the
> set of valid schemas written for this particular dialect.

[...]

PR #1487, opened 2024-02-08
Update output to make the root object an overview with results contained inside

gregsdennis:
This makes progress toward some of what is discussed in https://github.com/json-schema-org/json-schema-spec/issues/1065.

Basically, we need a place to include metadata about an evaluation.  This PR wraps the existing output in a root node that can contains any such metadata.

(As a side effect, it creates a purpose for the root object in the case of the "list" format, which could have just been an array before.)


gregsdennis:
Re-reading through the issue, it seems that `dialect` should be part of the output unit itself rather than the summary.  Still... I think the summary root node is a good addition.

jdesrosiers (commented):
> Re-reading through the issue, it seems that `dialect` should be part of the output unit itself rather than the summary. Still... I think the summary root node is a good addition.

I agree that `dialect` should be part of the output unit. I'm not sure putting it in the summary makes sense. Validation can include references to other schema resources that use a different dialect. So, `dialect` in the summary is only telling you the dialect of the root schema. Maybe that's ok, but it might be misleading if the evaluation includes multiple dialects.

gregsdennis:
I'll remove dialect from this PR and do another one for that then.

I know we had a discussion somewhere on having some more "global" details in the root node, though...

output/schema.json
```diff
@@ -3,11 +3,25 @@
...
-  "anyOf": [
-    { "$ref": "#/$defs/flag" },
-    { "$ref": "#/$defs/list" },
-    { "$ref": "#/$defs/hierarchical" }
-  ],
+  "properties": {
+    "dialect": {
+      "type": "string",
+      "format": "iri"
+    },
```
jdesrosiers, line 10:
```suggestion
```

jsonschema-validation-output-machines.md
```diff
@@ -219,13 +216,14 @@ The passing instance will produce the following annotations:
...
-In the simplest case, merely the boolean result for the `valid` valid property
-needs to be fulfilled. For this format, all other information is explicitly
-omitted.
+In the simplest case, only the overview needs to be fulfilled. For the "flag"
+format, all other information is explicitly omitted.
 
 ```json "Flag Results"
 {
-  "valid": false
+  "dialect": "https://json-schema.org/draft/next/schema",
```
jdesrosiers, line 224:
```suggestion
```

output/schema.json
```diff
@@ -3,11 +3,25 @@
...
+    "valid": { "type": "boolean" },
+    "details": {
+      "anyOf": [
+        { "$ref": "#/$defs/list" },
+        { "$ref": "#/$defs/hierarchical" }
+      ]
+    }
+  },
+  "requires": [ "dialect", "baseUri", "valid" ],
```
jdesrosiers, line 23:
```suggestion
  "requires": [ "baseUri", "valid" ],
```

jdesrosiers (approved):
You missed a couple places removing `dialect`, but otherwise, this looks good to me.

gregsdennis:
Superceded by #1542

PR #1494, opened 2024-03-10
Remove the funding file to use the org wide settings.

benjagm:
As of now the button to sponsor on Github is not working in this repo because if this file. We have org wide version in the .github repo and if we remove this we'll have the consistent setup applied accross.


gregsdennis (approved):
I'm happy to merge early as this doesn't affect the spec.

PR #1505, opened 2024-05-08
Add ADR for extracting unstable features

gregsdennis:
As [mentioned](https://github.com/json-schema-org/json-schema-spec/issues/1443#issuecomment-2099429276), this ADR documents the decision to extract unstable keywords prior to stable spec release.

This is to support [this proposal](https://github.com/json-schema-org/json-schema-spec/issues/1443#issuecomment-2099427543) of including an ADR alongside proposal documents.

PR #1510, opened 2024-05-22
Extract vocabularies from the specs

gregsdennis:
Relates to https://github.com/orgs/json-schema-org/discussions/724 and https://github.com/json-schema-org/json-schema-spec/pull/1505.

Depends on #1518
Depends on #1512
(will likely need rebase after those merge)

The Core spec was a lot, but it was fairly straightforward.

This PR will make a best effort to include vocabularies as they are in 2020-12 as a start.  Some redesign is inevitible, but this this is a good starting point.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -226,7 +226,7 @@ never produce annotation results.
 
 These boolean schemas exist to clarify schema author intent and facilitate
 schema processing optimizations. They behave identically to the following schema
-objects (where `not` is defined in [later this document](#not)).
+objects (where `not` is defined [later in this document](#not)).
 
 - `true`: Always passes validation, as if the empty schema `{}`
 - `false`: Always fails validation, as if the schema `{ "not": {} }`
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -29,12 +29,11 @@ and interaction control of JSON data.
 
 This specification defines JSON Schema core terminology and mechanisms,
 including pointing to another JSON Schema by reference, dereferencing a JSON
-Schema reference, specifying the dialect being used, specifying a dialect's
-vocabulary requirements, and defining terms.
+Schema reference, specifying the dialect being used, and defining terms.
 
-Other specifications define the vocabularies that perform assertions about
-validation, linking, annotation, navigation, and interaction as well as output
-formats.
+Other specifications define keywords that perform assertions about validation,
+linking, annotation, navigation, interaction, as well as other related concepts
+such as output formats.
 
 ## Conventions and Terminology
 
@@ -69,23 +68,22 @@ JSON Schema can be extended either by defining additional vocabularies, or less
 formally by defining additional keywords outside of any vocabulary. Unrecognized
 individual keywords are not supported.
 
-This document defines a core vocabulary that MUST be supported by any
-implementation, and cannot be disabled. Its keywords are each prefixed with a
-"$" character to emphasize their required nature. This vocabulary is essential
-to the functioning of the `application/schema+json` media type, and is used to
-bootstrap the loading of other vocabularies.
+This document defines a set of core keywords that MUST be supported by any
+implementation, and cannot be disabled. These keywords are each prefixed with a
+"$" character to emphasize their required nature. These keywords are essential
+to the functioning of the `application/schema+json` media type.
 
-Additionally, this document defines a RECOMMENDED vocabulary of keywords for
+Additionally, this document defines a RECOMMENDED set of keywords for
 applying subschemas conditionally, and for applying subschemas to the contents
-of objects and arrays. Either this vocabulary or one very much like it is
+of objects and arrays. These keywords, or a set very much like them, are
 required to write schemas for non-trivial JSON instances, whether those schemas
 are intended for assertion validation, annotation, or both. While not part of
-the required core vocabulary, for maximum interoperability this additional
-vocabulary is included in this document and its use is strongly encouraged.
+the required core set, for maximum interoperability this additional
+set is included in this document and its use is strongly encouraged.
 
-Further vocabularies for purposes such as structural validation or hypermedia
+Further keywords for purposes such as structural validation or hypermedia
 annotation are defined in other documents. These other documents each define a
-dialect collecting the standard sets of vocabularies needed to write schemas for
+dialect collecting the standard sets of keywords needed to write schemas for
 that document's purpose.
 
 ## Definitions
@@ -132,20 +130,20 @@ depending on the type:
 
 Whitespace and formatting concerns, including different lexical representations
 of numbers that are equal within the data model, are thus outside the scope of
-JSON Schema. JSON Schema [vocabularies](#vocabulary) that wish to work with such
-differences in lexical representations SHOULD define keywords to precisely
-interpret formatted strings within the data model rather than relying on having
-the original JSON representation Unicode characters available.
+JSON Schema. Extensions to JSON Schema that wish to work with such differences
+in lexical representations SHOULD define keywords to precisely interpret
+formatted strings within the data model rather than relying on having the
+original JSON representation Unicode characters available.
 
 Since an object cannot have two properties with the same key, behavior for a
 JSON document that tries to define two properties with the same key in a single
 object is undefined.
 
-Note that JSON Schema vocabularies are free to define their own extended type
+Note that JSON Schema extensions are free to define their own extended type
 system. This should not be confused with the core data model types defined here.
-As an example, "integer" is a reasonable type for a vocabulary to define as a
-value for a keyword, but the data model makes no distinction between integers
-and other numbers.
+As an example, "integer" is a reasonable type to define as a value for a
+keyword, but the data model makes no distinction between integers and other
+numbers.
 
 #### Instance Equality
 
@@ -175,7 +173,7 @@ where an instance may be outside any of the six JSON data types.
 In this case, annotations still apply; but most validation keywords will not be
 useful, as they will always pass or always fail.
 
-A custom vocabulary may define support for a superset of the core data model.
+An extension may define support for a superset of the core data model.
 The schema itself may only be expressible in this superset; for example, to make
 use of the `const` keyword.
 
@@ -228,8 +226,7 @@ never produce annotation results.
 
 These boolean schemas exist to clarify schema author intent and facilitate
 schema processing optimizations. They behave identically to the following schema
-objects (where `not` is part of the subschema application vocabulary defined in
-this document).
+objects (where `not` is defined in [later this document](#not)).
 
 - `true`: Always passes validation, as if the empty schema `{}`
 - `false`: Always fails validation, as if the schema `{ "not": {} }`
@@ -238,37 +235,11 @@ While the empty schema object is unambiguous, there are many possible
 equivalents to the `false` schema. Using the boolean values ensures that the
 intent is clear to both human readers and implementations.
 
-#### Schema Vocabularies
-
-A schema vocabulary, or simply a vocabulary, is a set of keywords, their syntax,
-and their semantics. A vocabulary is generally organized around a particular
-purpose. Different uses of JSON Schema, such as validation, hypermedia, or user
-interface generation, will involve different sets of vocabularies.
-
-Vocabularies are the primary unit of re-use in JSON Schema, as schema authors
-can indicate what vocabularies are required or optional in order to process the
-schema. Since vocabularies are identified by IRIs in the meta-schema, generic
-implementations can load extensions to support previously unknown vocabularies.
-While keywords can be supported outside of any vocabulary, there is no analogous
-mechanism to indicate individual keyword usage.
-
-A schema vocabulary can be defined by anything from an informal description to a
-standards proposal, depending on the audience and interoperability expectations.
-In particular, in order to facilitate vocabulary use within non-public
-organizations, a vocabulary specification need not be published outside of its
-scope of use.
-
 #### Meta-Schemas
 
 A schema that itself describes a schema is called a meta-schema. Meta-schemas
-are used to validate JSON Schemas and specify which vocabularies they are using.
-
-Typically, a meta-schema will specify a set of vocabularies, and validate
-schemas that conform to the syntax of those vocabularies. However, meta-schemas
-and vocabularies are separate in order to allow meta-schemas to validate schema
-conformance more strictly or more loosely than the vocabularies' specifications
-call for. Meta-schemas may also describe and validate additional keywords that
-are not part of a formal vocabulary.
+are used to validate JSON Schemas and specify the set of keywords those schemas
+are using.
 
 #### Root Schema and Subschemas and Resources {#root}
 
@@ -415,14 +386,13 @@ neither at the beginning nor at the end. This means, for instance, the pattern
 
 ### Extending JSON Schema {#extending}
 
-Additional schema keywords and schema vocabularies MAY be defined by any entity.
-Save for explicit agreement, schema authors SHALL NOT expect these additional
-keywords and vocabularies to be supported by implementations that do not
-explicitly document such support.
+Additional schema keywords MAY be defined by any entity. Save for explicit
+agreement, schema authors SHALL NOT expect these additional keywords to be
+supported by implementations that do not explicitly document such support.
 
 Implementations MAY provide the ability to register or load handlers for
-vocabularies that they do not support directly. The exact mechanism for
-registering and implementing such handlers is implementation-dependent.
+keywords that they do not support directly. The exact mechanism for registering
+and implementing such handlers is implementation-dependent.
 
 #### Implicit annotation keywords {#implicit-annotations}
 
@@ -434,7 +404,7 @@ Implicit annotation keywords MUST NOT affect evaluation of a schema in
 any way other than annotation collection.
 
 Consequently, the "x-" prefix is reserved for this purpose, and extension
-vocabularies MUST NOT define any keywords which begin with this prefix.
+keywords MUST NOT begin with this prefix.
 
 #### Handling of unrecognized or unsupported keywords {#unrecognized}
 
@@ -563,19 +533,19 @@ in this document.
 
 Note that when no such alternate approach is possible for a keyword,
 implementations that do not support annotation collections will not be able to
-support those keywords or vocabularies that contain them.
+support those keywords.
 
 ### Identifiers
 
 Identifiers define IRIs for a schema, or affect how such IRIs are resolved in
-[references](#referenced), or both. The Core vocabulary defined in this document
-defines several identifying keywords, most notably `$id`.
+[references](#referenced), or both. This document defines several identifying
+keywords, most notably `$id`.
 
 Canonical schema IRIs MUST NOT change while processing an instance, but keywords
 that affect IRI-reference resolution MAY have behavior that is only fully
 determined at runtime.
 
-While custom identifier keywords are possible, vocabulary designers should take
+While custom identifier keywords are possible, extension designers should take
 care not to disrupt the functioning of core keywords. For example, the
 `$dynamicAnchor` keyword in this specification limits its IRI resolution effects
 to the matching `$dynamicRef` keyword, leaving the behavior of `$ref`
@@ -641,16 +611,16 @@ Most assertions only constrain values within a certain primitive type. When the
 type of the instance is not of the type targeted by the keyword, the instance is
 considered to conform to the assertion.
 
-For example, the `maxLength` keyword from the companion [validation
-vocabulary](#json-schema-validation): will only restrict certain strings (that
+For example, the `maxLength` keyword will only restrict certain strings (that
 are too long) from being valid. If the instance is a number, boolean, null,
 array, or object, then it is valid against this assertion.
 
 This behavior allows keywords to be used more easily with instances that can be
-of multiple primitive types. The companion validation vocabulary also includes a
-`type` keyword which can independently restrict the instance to one or more
-primitive types. This allows for a concise expression of use cases such as a
-function that might return either a string of a certain length or a null value:
+of multiple primitive types. The companion Validation specification also
+includes a `type` keyword which can independently restrict the instance to one
+or more primitive types. This allows for a concise expression of use cases such
+as a function that might return either a string of a certain length or a null
+value:
 
 ```jsonschema
 {
@@ -824,14 +794,19 @@ assertions.
 A fourth category of keywords simply reserve a location to hold re-usable
 components or data of interest to schema authors that is not suitable for
 re-use. These keywords do not affect validation or annotation results. Their
-purpose in the core vocabulary is to ensure that locations are available for
-certain purposes and will not be redefined by extension keywords.
+purpose is to ensure that locations are available for certain purposes and will
+not be redefined by extension keywords.
+
+While these keywords do not directly affect results, as explained in
+{{non-schemas}} unrecognized extension keywords that reserve locations for
+re-usable schemas may have undesirable interactions with references in certain
+circumstances.
 
 ### Loading Instance Data
 
-While none of the vocabularies defined as part of this or the associated
-documents define a keyword which may target and/or load instance data, it is
-possible that other vocabularies may wish to do so.
+While none of the keywords defined as part of this or the associated
+documents define a keyword which target and/or load instance data, it is
+possible that extensions may wish to do so.
 
 Keywords MAY be defined to use JSON Pointers or Relative JSON Pointers to
 examine parts of an instance outside the current evaluation location.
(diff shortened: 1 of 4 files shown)


jdesrosiers (commented):
I've only skimmed a little so far, but one thing that stood out is that we don't necessarily need to remove the word or the concept of vocabularies generally as collection of keywords. We only need to remove the mechanism (`$vocabulary`) and anything that refers to the behavior of that mechanism. The spec used the term "vocabulary" before the vocabulary system was introduced and I think it would be fine if we leave a few places that use the term in the same way. It could help make this a smaller change.

gregsdennis:
Yes, the word was present, but it was used very loosely.  Later we added it with a very specific (and vaguely different) meaning but didn't address the existing usages.  (See https://json-schema.org/draft-07/json-schema-validation.)

I figure it's better for now to just remove it completely.  I don't think that the text suffers.

jsonschema-validation.md
```diff
@@ -295,7 +284,7 @@ the name of a property in the instance.
 
 Omitting this keyword has the same behavior as an empty object.
 
-## Vocabularies for Semantic Content With `format` {#format}
```
gregsdennis, line 298: There are considerable impactful changes for `format`.  Specifically, schema authors will no longer have the ability to specify that `format` should be validated.  Tool configuration is the only way to validate `format` when we remove vocabularies.

See [discussion](https://github.com/orgs/json-schema-org/discussions/724#discussioncomment-9527084).

jdesrosiers:
> I figure it's better for now to just remove it completely. I don't think that the text suffers.

That's fine with me. I didn't think the text suffered, I was just trying to make this a smaller job :smile:.

gregsdennis:
Reminder (to me): open a new issue to discuss the future of the keyword and link all of the issues I've been linking ☝️

jsonschema-validation.md
gregsdennis: (Might not be an issue if #1520 goes through.)

jsonschema-core.md
```diff
@@ -816,74 +778,41 @@ circumstances.
...
 
-The current IRI for the Core vocabulary is:
-`https://json-schema.org/draft/next/vocab/core`.
+### Meta-Schemas
 
-The current IRI for the corresponding meta-schema is:
-`https://json-schema.org/draft/next/meta/core`.
+Meta-schemas are used to inform an implementation how to interpret a schema.
+Every schema has a meta-schema, which can be explicitly declared using the
+`$schema` keyword.
```
jdesrosiers, line 808: I think this section was effectively the introduction to and definition of the term "dialect". That ends up getting lost in the changes. Do we want to do something to include that defines "dialect"?

jdesrosiers (commented):
This is big, so I'm going to review in several parts. This first part is for `jsonschema-core.md`.

I like the wording changes. I especially like that works whether the vocabulary system is a thing or not which will make it easier to merge back in at some point if it ever reaches a stable state.

jsonschema-validation.md
```diff
@@ -320,115 +309,57 @@ can be used alongside the `type` keyword with a value of "integer", or could be
...
+identified by a format instance.
 
 #### Custom format attributes
 
 Implementations MAY support custom format attributes. Save for agreement between
 parties, schema authors SHALL NOT expect a peer implementation to support such
-custom format attributes. An implementation MUST NOT fail to collect unknown
-formats as annotations. When the Format-Assertion vocabulary is specified,
-implementations MUST fail upon encountering unknown formats.
+custom format attributes.
```
jdesrosiers, line 357: I think that given the changes in this section, we would always expect implementations to "support" custom format attributes, we just wouldn't expect them to validate to anything other than `true`. Since partial and even no validation is allowed for any format attribute, always returning `true` is a valid way to "support" that attribute.

I think the point of this sentence is to say that you can't expect it to assert the way you expect, but that ends up being the case for any format attribute including the ones defined in the spec. So, I'm not sure what we should do with this.

jsonschema-validation.md
```diff
@@ -320,115 +309,57 @@ can be used alongside the `type` keyword with a value of "integer", or could be
...
 
-Vocabularies do not support specifically declaring different value sets for
-keywords.  Due to this limitation, and the historically uneven implementation of
-this keyword, it is RECOMMENDED to define additional keywords in a custom
-vocabulary rather than additional format attributes if interoperability is
-desired.
+An implementation MUST NOT fail to collect unknown formats as annotations.
```
jdesrosiers, line 359: We've been qualifying this with "(if the implementation supports annotation collection)" everywhere. Should we be doing that here as well?

jsonschema-validation.md
```diff
@@ -320,115 +309,57 @@ can be used alongside the `type` keyword with a value of "integer", or could be
...
+
+When configured for assertion behavior for `format`, implementations MUST fail
+upon encountering unknown formats.
```
jdesrosiers, line 362: We say that even when assertion behavior is enabled, implementations might still just return `true`, so why must an implementation fail when encountering an unknown format rather than returning `true`? Or should they fail if they don't have a best-effort implementation for the format whether it's known or not?

I think this requirement made sense applied to the `format-assertion` vocabulary, but doesn't really work in this context.

jsonschema-validation.md
```diff
@@ -969,40 +883,6 @@ draft-bhutton-json-schema-01, June 2022,
 Hoehrmann, B., "Scripting Media Types", RFC 4329, DOI 10.17487/RFC4329, April
 2006, <<https://www.rfc-editor.org/info/rfc4329>>.
 
-## [Appendix] Keywords Moved from Validation to Core
```
jdesrosiers, line 972: I'm not sure removing this section is related to extracting vocabularies, but I fully agree with removing it.

jsonschema-core.md
```diff
@@ -816,74 +778,41 @@ circumstances.
...
+schema or meta-schema, including those split across multiple documents, or exist
+to reserve keywords for purposes that require guaranteed interoperability.
 
-The behavior of a false value for this vocabulary (and only this vocabulary) is
-undefined, as is the behavior when `$vocabulary` is present but the Core
-vocabulary is not included. However, it is RECOMMENDED that implementations
-detect these cases and raise an error when they occur. It is not meaningful to
-declare that a meta-schema optionally uses Core.
+Support for these keywords MUST be considered mandatory at all times in order to
+bootstrap the processing of further keywords.
```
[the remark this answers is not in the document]
gregsdennis: I see the `$` keywords as "directive" keywords, meaning they tell you how to process and navigate the schema.  In that way, they are sort of "bootstrapping", but I'm open to massaging it a bit.

jsonschema-validation.md
gregsdennis: The point might be moot with https://github.com/json-schema-org/json-schema-spec/issues/1520.

For the purpose here, though, I want to move away from specifying annotation or assertion behavior.  Maybe it's enough to say that for unknown formats, an implementation must collect the value as an annotation and provide a passing validation result.

jsonschema-validation.md
gregsdennis: Possibly.  I think this is a candidate for "things that need cleaning up later".

jsonschema-validation.md
gregsdennis: The answer to the question is: which is more acceptable, a false negative or a false positive?  I landed on the side of false negative being more acceptable.

jsonschema-validation.md
gregsdennis: Hm... I'm not sure.  It probably got caught up from one of my more aggressive change experiments I have locally.

Still, I agree it can be removed.  I'm not fussed about removing it if you're not.

jsonschema-core.md
gregsdennis: I'm not sure.  I wonder if that's something that needs to be re-introduced with the proposal.

jsonschema-validation.md
jdesrosiers: How does #1520 makes this irrelevant? Isn't that just about what the default behavior is? We still have multiple behavior that we have to define in consistent way.

> Maybe it's enough to say that for unknown formats, an implementation must collect the value as an annotation and provide a passing validation result.

That would work.

jsonschema-validation.md
jdesrosiers: I wasn't advocating for any particular behavior. I was just pointing out that the text is contradictory.

But, my opinion on the behavior is that in assert mode it should either:

- Allow partial/no validation and unknown formats return true; OR
- Require full validation and refuse to process unknown formats

Allowing no validation and refusing to process unknown formats doesn't make sense.

gregsdennis: Do we just remove assertion completely, then, and let implementations figure out what that mode means?

jsonschema-validation.md
gregsdennis: Maybe [this](https://github.com/json-schema-org/json-schema-spec/pull/1510#discussion_r1648195159) is the solution here, too?

proposals/vocabularies.md
```diff
@@ -0,0 +1,267 @@
...
+highlights that the structure of `$vocabulary` needs to be reconsidered.
+
+#### Machine Readability
+
+The vocabulary URI is an opaque value. There is no data that an implementation
+can reference to identify the keywords defined by the vocabulary. The vocabulary
+schema _implies_ this, but scanning a `properties` keyword isn't very reliable.
+Moreover, such a system cannot provide metadata about the keywords. As such, the
+user must explicitly ensure that the implementation recognizes and supports the
+vocabulary, which isn't much of an improvement over the current state.
```
jdesrosiers, line 212: That statement doesn't make sense to me. I don't see how the vocabulary description document has any effect for users other than reducing the work they need to do when adding support for their vocabulary in the library they're using.

jdesrosiers (commented):
Here's the last part of my review including the adr and proposal.

jsonschema-core.md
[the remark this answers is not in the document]
jdesrosiers: I see what you're trying to say. Maybe something like this,

```suggestion
Support for these keywords MUST be considered mandatory at all times as they are
necessary to navigate and process any schema.
```

jsonschema-core.md
jdesrosiers: I think the concept of a dialect should stay. The concept isn't dependent on the vocabulary system. I explain further in my latest review.

jsonschema-validation.md
jdesrosiers: As long as we define something before we release, I'm fine with leaving this somewhat undefined for now.

proposals/vocabularies-adr.md
```diff
@@ -0,0 +1,74 @@
...
+- Language-agnostic
+- Ease of use
+- Ease of implementation
+
+## Considered Options
+
+### Current design as included in 2019-09 and 2020-12.
+
+Vocabularies are external documents that describe how new keywords function.
+They can be in a specification style, or a blog post, or some other format.
```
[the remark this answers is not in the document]
gregsdennis: That's fair.  I'll massage this.

proposals/vocabularies.md
```diff
@@ -0,0 +1,267 @@
...
+  // ...
+}
+```
+
+A dialect is the set of vocabularies listed by a meta-schema. It is ephemeral
+and carries no identifier.
+
+_**NOTE** It is possible for two meta-schemas, which would have different `$id`
+values, to share a common dialect if they both declare the same set of
+vocabularies._
```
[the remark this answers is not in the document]
gregsdennis: I think dialects are more of a duck-typing scenario.

I think this is splitting hairs.  I'll need to do an analysis on the impact of each approach.  I expect they're not very dissimilar.

proposals/vocabularies.md
```diff
@@ -0,0 +1,267 @@
...
+    "https://example.org/vocab/vocab1": true,
+    "https://example.org/vocab/vocab2": true,
+    "https://example.org/vocab/vocab3": false
+  },
+  // ...
+}
+```
+
+A dialect is the set of vocabularies listed by a meta-schema. It is ephemeral
+and carries no identifier.
```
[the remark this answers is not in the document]
gregsdennis: https://github.com/json-schema-org/json-schema-spec/pull/1510/files#r1648263398 is definitely informing your comment here.

proposals/vocabularies.md
```diff
@@ -0,0 +1,267 @@
...
+  "unknown" and handled accordingly. These vocabularies are also known as
+  "optional" vocabularies.
+
+Typically, but not required, a schema will accompany the vocabulary description
+document. This _vocabulary schema_ should carry an `$id` value which is distinct
+from the vocabulary URI. The purpose of the vocabulary schema is to provide
+syntactic validation for the the vocabulary's keywords' values for when the
+schema is being validated by a meta-schema that includes the vocabulary. (A
+vocabulary schema is not itself a meta-schema since it does not validate entire
+schemas.) To facilitate this extra validation, when a vocabulary schema is
```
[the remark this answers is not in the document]
gregsdennis: He didn't update the spec or anything.  This actually came out of reviewing @jviotti's book and trying to rework my vocab schemas.  I wrote about it [here](https://blog.json-everything.net/posts/updating-vocabs/), and we pulled out `$vocabulary` from them in #1460, #1461, and #1462.

They're not meta-schemas because they don't themselves describe full schemas; they are used by meta-schemas as components.

proposals/vocabularies.md
gregsdennis: There was always an intent to build in some kind of machine-readable vocbaulary meta-data document.  What stalled that is that we couldn't identify what that meta-data should be.

I'm merely mentioning that the current solution doesn't address that intent, and whatever we come up with should.  I think it's fine to keep in here for now.

The last sentence is just stating that the current state of having to check if a vocab is supported is no different than having to check if explicit keywords are supported.  Having a machine-readable description document could automate some of that in some cases and ease the burden of the user.

proposals/vocabularies-adr.md
```diff
@@ -0,0 +1,74 @@
...
+
+An implementation declares support for a particular vocabulary via
+implementation of its keywords and documentation.
```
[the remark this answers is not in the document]
gregsdennis: > Also, you don't implement documentation, you implement the keywords in accordance with the vocabulary documentation.

... "via (implementation of its keywords) && documentation"

proposals/vocabularies.md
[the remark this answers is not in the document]
jdesrosiers: Maybe. And certainly it's influenced by the architecture of my implementation to some degree, but it also comes from the spec [Section 8.1.1](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#rfc.section.8.1.1)

> The "$schema" keyword is both used as a JSON Schema dialect identifier and as the identifier of a resource which is itself a JSON Schema, which describes the set of valid schemas written for this particular dialect.

proposals/vocabularies.md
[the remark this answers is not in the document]
jdesrosiers: I don't think we're quite talking about the same thing. I entirely agree that vocabulary meta-schemas shouldn't include the `$vocabulary` keyword.

> They're not meta-schemas because they don't themselves describe full schemas they are used by meta-schemas as components.

This distinction is what doesn't sit right with me. The way I see it, a component of a meta-schema is still a meta-schema. [Section 8.1.2.2](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#rfc.section.8.1.2.2) is what I thought you were referring to here. Is that correct?

It does seem like the wording there is not considering schemas referenced by a meta-schema a meta-schema, but that's never how I've understood the word or how we've defined the word. I've always used the terms dialect meta-schema and vocabulary meta-schema. The `$vocabulary` keyword is only meaningful when the meta-schema is used as a dialect meta-schema.

proposals/vocabularies.md
jdesrosiers: I'm not disputing that vocab description document wasn't always the intention or that it doesn't have value. It's very valuable and I'm 100% on board with having one. I see it enabling all kinds of things. It's just the last sentence that doesn't make sense to me. I don't see how a vocabulary description document can influence the need for a user to check if an implementation supports a vocabulary or not.

proposals/vocabularies-adr.md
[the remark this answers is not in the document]
jdesrosiers: :laughing: It makes so much more sense now. Still, that sentence could be reworked so it's harder to misunderstand.

proposals/vocabularies.md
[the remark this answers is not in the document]
gregsdennis: No, I wasn't referring to that section.

I'm looking at [Core 4.3.4](https://json-schema.org/draft/2020-12/json-schema-core#section-4.3.4) which _defines_ "meta-schema":

> A schema that itself describes a schema is called a meta-schema.

A vocabulary schema doesn't describe a schema, therefore it's not a meta-schema.

You wouldn't use the Meta-data vocab schema as a meta-schema; you reference it from a meta-schema.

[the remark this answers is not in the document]
jdesrosiers: Interesting. We're using the same definition, but interpreting it differently. The way I see it, a vocabulary schema is validating a schema. It validates the syntax of keywords in a schema. I don't think the definition implies that a schema is only a meta-schema if it describes the _entire_ dialect used by the schema.

gregsdennis:
I've rebased this on top of #1512.  Starting to address comments now.

jsonschema-core.md
gregsdennis: How do you want to define "dialect"?  Is it merely the collection of keywords?  Open to suggestions.

jsonschema-validation.md
gregsdennis: I think I want to leave this for #1520 to figure out.  For now, I just want to pull out the vocabulary language.  We'll definitely address it before release, though.

proposals/vocabularies-adr.md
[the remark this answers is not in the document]
gregsdennis: The vocabs that I've created (and the OpenAPI one) are definitely in the specification style, but I remember discussing it with Henry, and he mentioned that it didn't always have to be.  It just had to be defined somewhere people who wanted to implement it had access to.  That could be a blog post or an internal wiki or whatever.

I think that allows for wider participation in defining new keywords than writing a spec-like document would because most people wouldn't bother with the precision.

proposals/vocabularies.md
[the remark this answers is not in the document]
gregsdennis: It _is_ validating keyword usage inside of a schema, but it is not validating a schema.  It's a small distinction, but I think it's worth noting.

Calling the vocabulary schema a "meta-schema" is what led me to create my first vocab schemas _as_ meta-schemas, which multiple people recognized as the wrong way to do it.  I didn't understand why until Juan started to dig into it for his book.  I realized my mistake and started calling it a "vocab schema" from then on.  If _I_ got confused by this, others doubtlessly will (or have already) as well.  I just want the distinction to be apparent that this isn't used the in the same way that a full meta-schema would be.

A meta-schema is something that you'd reference with `$schema`.  You wouldn't put a vocabulary schema ID in `$schema`; its purpose is different.

proposals/vocabularies.md
gregsdennis: The idea was that an implementation could potentially "read in" a vocabulary and support it dynamically.  I think in practice this would be limited to vocabs which only defined annotation-only keywords, but that was the thought.

proposals/vocabularies.md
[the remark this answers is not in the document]
mwadams: I had the same realisation while implementing vocab support - which leads to another question. Do we need to publish a canonical vocab metaschema schema?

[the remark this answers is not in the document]
gregsdennis: > Do we need to publish a canonical vocab metaschema schema?

I'm not sure what you're leading toward.  Are you questioning the current practice of defining a vocab schema?  Or are you saying that we should have a vocab schema that _can_ function as a meta-schema?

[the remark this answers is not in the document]
jdesrosiers: > If _I_ got confused by this, others doubtlessly will (or have already) as well.

I 100% agree that we need to explain this better and using more descriptive terms is a big part of doing that well. But, the way you're currently expressing this is confusing to me because this doesn't mesh with my (and I'm sure others) understanding of the term "meta-schema". That's why I'd prefer to address this by introducing new terms that are more specific. Earlier I mention that I use the terms "dialect meta-schema" and "vocabulary meta-schema", but since the term "meta-schema" appears to be inconsistently understood, perhaps "dialect schema" and "vocabulary schema" is better.

proposals/vocabularies.md
jdesrosiers: I agree with that completely, but that seems like it's a different thing than what that sentence is saying. Maybe I may have misunderstood what you meant with the word "ensure". I read that to mean "check the documentation" or "run a tests to see if it works as expecte". But, if you meant something more like "write the code to configure the vocabulary", then the sentence makes sense to me.

proposals/vocabularies.md
```diff
@@ -0,0 +1,265 @@
...
+Having some sort of "vocabulary definition" file could alleviate this.
+
+One reason for _not_ having such a file is that, at least for functional
+keywords, the user generally needs to provide custom code to the implementation
+to process the keywords, thus performing that same explicit configuration
+anyway. (Such information cannot be gleaned from a vocabulary specification. For
+example, an implementation can't know what to do with a hypothetical `minDate`
+keyword.)
+
+Several ideas have been offeree for this sort of document:
```
jdesrosiers, line 215:
```suggestion
Several ideas have been offered for this sort of document:
```

proposals/vocabularies-adr.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Language-agnostic
+- Ease of use
+- Ease of implementation
+
+## Considered Options
+
+### Current design as included in 2019-09 and 2020-12.
+
+Vocabularies are collections of keywords and are defined by vocabulary document.
```
jdesrosiers, line 30:
```suggestion
Vocabularies are collections of keywords and are defined by a vocabulary document.
```

gregsdennis: We need to keep the plurality consistent.  Do you have a preference between "vocabularies are..." and "a vocabulary is..."?

proposals/vocabularies.md
gregsdennis: I read that to mean "check the documentation" or "run a tests to see if it works as expecte[d]". But, if you meant something more like "write the code to configure the vocabulary", then the sentence makes sense to me.

It's encompassing of both, really.  Definitely the former, but also up to and including the latter.

proposals/vocabularies-adr.md
jdesrosiers: Apparently I read carefully enough to notice that there was an mistake, but not carefully enough to make the right suggestion :man_facepalming:.

No strong preference from me, but I'd probably choose the singular.

gregsdennis: Updated to singular form

gregsdennis:
@jdesrosiers would you please re-review this? I've made changes according to our conversations.

Also, would you be satisfied with the proposal doc in its current state as a starting point?  I fully expect further issues and PRs to be created to edit the proposal.

gregsdennis:
_[Just some housekeeping]_

Issues to be opened and addressed in follow-up PRs:

- [define "dialect" and how vocabs relate to it](https://github.com/json-schema-org/json-schema-spec/pull/1510#discussion_r1646571299) (also [this comment](https://github.com/json-schema-org/json-schema-spec/pull/1510#discussion_r1648263398))
- [unknown format handling](https://github.com/json-schema-org/json-schema-spec/pull/1510#discussion_r1646589555)
- [defining meta-schema vs "vocab schema" vs "[whatever] schema" more clearly](https://github.com/json-schema-org/json-schema-spec/pull/1510#discussion_r1648268000)

proposals/vocabularies.md
```diff
@@ -0,0 +1,265 @@
...
+A schema that declares a meta-schema (via `$schema`) which contains
+`$vocabulary` is declaring that only those keywords defined by the included
+vocabularies are to be processed when evaluating the schema. All other keywords
+are to be considered "unknown" and handled accordingly.
+
+The boolean values in `$vocabulary` signify implementation requirements for each
+vocabulary.
+
+- A `true` value indicates that the implementation must recognize the vocabulary
+  and be able to process each of the keywords defined it. If an implementation
```
jdesrosiers, line 115:
```suggestion
  and be able to process each of the keywords defined in it. If an implementation
```

jdesrosiers (approved):
I'm still not happy with some parts. I feel like this is a change in the meaning of "meta-schema" and I'm not sure I'm comfortable with that. But, I'm ok with merging this as is as an iteration. The proposal will need to evolve or die anyway. This is not the final version in any case.

gregsdennis:
> I feel like this is a change in the meaning of "meta-schema" and I'm not sure I'm comfortable with that.

Regarding this, I feel we've been misusing "meta-schema" by allowing it to refer to schemas that describe vocabs, and I want us to return to using it to refer to schemas that describes schemas.

It's not a change in definition; it's a return to correct usage.

> But, I'm ok with merging this as is as an iteration.

Definitely, this PR is focused on getting vocabs out of the spec so that we can start iterating on the spec again.  This is the last piece that we agreed to remove.

The proposal is definitely unfinished.  The approach I took was to just make it as close to the 2020-12 behavior as I could.  I recognize that it won't work with the next spec, but we need a starting point.  The proposal will continue to evolve and will likely not look much like its current state when it's done.  Maybe it'll just be wholesale replaced by something better.

jsonschema-validation.md
```diff
@@ -320,115 +309,57 @@ can be used alongside the `type` keyword with a value of "integer", or could be
...
-Format-Assertion vocabulary since implementations are not required to provide
-full validation support when the Format-Assertion vocabulary is not specified.
-
-When the implementation is configured for assertion behavior, it:
+Implementing support for `format` as an assertion is OPTIONAL. Implementations
+which choose to support assertion behavior:
 
+- MUST still collect the keyword's value as an annotation (if the implementation
+  supports annotation collection),
+- MUST provide a configuration option to enable assertion behavior, defaulting to
```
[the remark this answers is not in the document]
gregsdennis: No.  In order to ensure interoperability, we need all implementations to have the same out-of-the-box behavior.  Implementations can do whatever, but they need non-spec behavior hidden behind configuration so that the user must explicitly opt in.  This applies _in general_.

This particular text will be revisited with the resolution of #1520.

jsonschema-validation.md
```diff
@@ -320,115 +309,57 @@ can be used alongside the `type` keyword with a value of "integer", or could be
...
-formats as annotations. When the Format-Assertion vocabulary is specified,
-implementations MUST fail upon encountering unknown formats.
+custom format attributes.
 
-Vocabularies do not support specifically declaring different value sets for
-keywords.  Due to this limitation, and the historically uneven implementation of
-this keyword, it is RECOMMENDED to define additional keywords in a custom
-vocabulary rather than additional format attributes if interoperability is
-desired.
+An implementation MUST NOT fail to collect unknown formats as annotations.
```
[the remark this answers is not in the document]
gregsdennis: Yeah, probably.  This was the text in previous versions.  I just reverted it to that.  Can be revisited.

proposals/vocabularies.md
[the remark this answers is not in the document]
jdesrosiers: > I would not have made the mistake if we'd had a "vocab metaschema" schema that clearly defined what was allowed in a vocab metaschema. It would have been apparent that we were talking about a very similar, but different entity.

What would be the difference between these two meta-schemas? Everything that's allowed in a vocab schema is allowed in a dialect schema. Everything that's allowed in a dialect schema is allowed in a vocab schema. The only difference I can see would be their identifier.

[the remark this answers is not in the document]
gregsdennis: I agree with @jdesrosiers on this:  syntactically they're just schemas.

The only difference is intent.  A meta-schema is intended to describe/validate a schema.  It can _use_ vocab schemas (by reference) to accomplish that task.  A vocab schema on its own does not describe a schema; it's merely a building block in a meta-schema.  This is the distinction I'm making.  (One _could_ create a single-vocab meta-schema.)

I don't think having a separate meta-schema for vocab schemas is needed since anything you do in a general schema should technically be allowed in a vocab schema.  I'm not sure how _useful_ all of those features would be, but I don't see a reason to forbid any features.

PR #1512, opened 2024-05-23
Remove support for unrecognized keywords

### LINKED ISSUE #1340: Update spec to disallow unknown keywords

### LINKED ISSUE #1365: Improved handling of unknown properties

gregsdennis:
We discussed this a lot, but never actually made the changes.

Resolves #1340
Resolves #1365

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -438,8 +438,8 @@ vocabularies MUST NOT define any keywords which begin with this prefix.
 
 #### Handling of unrecognized or unsupported keywords {#unrecognized}
 
-Implementations MUST refuse to process schemas which contain keywords they do
-not recognize, or that they recognize but do not support.
+Implementations MUST refuse to evaluate schemas which contain keywords which
+they do not know how to process or explicitly choose not to process.
 
 ## Keyword Behaviors
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -67,9 +67,7 @@ identified in a meta-schema.
 
 JSON Schema can be extended either by defining additional vocabularies, or less
 formally by defining additional keywords outside of any vocabulary. Unrecognized
-individual keywords simply have their values collected as annotations, while the
-behavior with respect to an unrecognized vocabulary can be controlled when
-declaring which vocabularies are in use.
+individual keywords are not supported.
 
 This document defines a core vocabulary that MUST be supported by any
 implementation, and cannot be disabled. Its keywords are each prefixed with a
@@ -216,8 +214,8 @@ adjacent keywords.
 Extension keywords, meaning those defined outside of this document and its
 companions, are free to define other behaviors as well.
 
-A JSON Schema MAY contain properties which are not schema keywords or are not
-recognized as schema keywords. The behavior of such keywords is governed by
+A JSON Schema MUST NOT contain properties which are not schema keywords or are
+not recognized as schema keywords. The behavior of such keywords is governed by
 {{unrecognized}}.
 
 An empty schema is a JSON Schema with no properties.
@@ -440,10 +438,8 @@ vocabularies MUST NOT define any keywords which begin with this prefix.
 
 #### Handling of unrecognized or unsupported keywords {#unrecognized}
 
-Implementations SHOULD treat keywords they do not recognize, or that they
-recognize but do not support, as annotations, where the value of the keyword is
-the value of the annotation. Whether an implementation collects these
-annotations or not, they MUST otherwise ignore the keywords.
+Implementations MUST refuse to process schemas which contain keywords they do
+not recognize, or that they recognize but do not support.
 
 ## Keyword Behaviors
 
@@ -831,11 +827,6 @@ re-use. These keywords do not affect validation or annotation results. Their
 purpose in the core vocabulary is to ensure that locations are available for
 certain purposes and will not be redefined by extension keywords.
 
-While these keywords do not directly affect results, as explained in
-{{non-schemas}} unrecognized extension keywords that reserve locations for
-re-usable schemas may have undesirable interactions with references in certain
-circumstances.
-
 ### Loading Instance Data
 
 While none of the vocabularies defined as part of this or the associated
@@ -1008,13 +999,7 @@ refuse to process that schema.
 
 Implementations that do not support a vocabulary that is optionally used by a
 schema SHOULD proceed with processing the schema. The keywords will be
-considered to be unrecognized keywords as addressed by {{unrecognized}}. Note
-that since the recommended behavior for such keywords is to collect them as
-annotations, vocabularies consisting only of annotations will have the same
-behavior when used optionally whether the implementation supports them or not.
-This allows annotation-only vocabularies to be supported without custom code,
-even in implementations that do not support providing custom code for extension
-vocabularies.
+considered to be unrecognized keywords as addressed by {{unrecognized}}.
 
 ##### Vocabularies are schema resource-scoped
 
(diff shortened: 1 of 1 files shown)


jsonschema-core.md
```diff
@@ -986,13 +977,7 @@ refuse to process that schema.
...
 Implementations that do not support a vocabulary that is optionally used by a
 schema SHOULD proceed with processing the schema. The keywords will be
-considered to be unrecognized keywords as addressed by {{unrecognized}}. Note
-that since the recommended behavior for such keywords is to collect them as
-annotations, vocabularies consisting only of annotations will have the same
-behavior when used optionally whether the implementation supports them or not.
-This allows annotation-only vocabularies to be supported without custom code,
-even in implementations that do not support providing custom code for extension
-vocabularies.
+considered to be unrecognized keywords as addressed by {{unrecognized}}.
```
Relequestual, line 980: This section is now contradictory.

> SHOULD proceed with processing the schema.

But considered unrecognized keywords...

> mplementations MUST refuse to process schemas which contain keywords they do not recognize, or that they recognize but do not support.

Are we really wanting to say that optinal support of a vocabulary is not something we want to support anymore?

Relequestual (changes requested):
Left a comment. Currently contradictory as far as I can tell.

Otherwise, looks good.

jsonschema-core.md
gregsdennis: Yes. This is a contradiction, which I've brought up in a discussion elsewhere (will provide the link shortly).

jsonschema-core.md
```diff
@@ -436,10 +434,8 @@ vocabularies MUST NOT define any keywords which begin with this prefix.
 
 #### Handling of unrecognized or unsupported keywords {#unrecognized}
 
-Implementations SHOULD treat keywords they do not recognize, or that they
-recognize but do not support, as annotations, where the value of the keyword is
-the value of the annotation. Whether an implementation collects these
-annotations or not, they MUST otherwise ignore the keywords.
+Implementations MUST refuse to process schemas which contain keywords they do
+not recognize, or that they recognize but do not support.
```
jdesrosiers, line 438: The wording here _feels_ contradictory although I get that it isn't and why it isn't. I get that `x-` keywords are considered "recognized", but I'm not sure we're making that clear. Maybe we need to be more clear what is considered "recognized" or maybe we need to say it different way that's harder to misunderstand.

gregsdennis: @Relequestual had this same concern in [Slack](https://json-schema.slack.com/archives/CT7FF623C/p1717978845635409).  It's addressed in #1518.

jsonschema-core.md
```diff
@@ -67,9 +67,7 @@ identified in a meta-schema.
 
 JSON Schema can be extended either by defining additional vocabularies, or less
 formally by defining additional keywords outside of any vocabulary. Unrecognized
-individual keywords simply have their values collected as annotations, while the
-behavior with respect to an unrecognized vocabulary can be controlled when
-declaring which vocabularies are in use.
+individual keywords are not supported.
```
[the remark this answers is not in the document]
gregsdennis: > The `x-` convention is an extension mechanism...

I'm curious of your definition of an extension mechanism.  The `x-` behavior is defined by the spec, so in my eyes, these keywords are not extensions.  This is clarified by #1518.

(This paragraph is going to change in #1510 anyway, so I'm not too concerned about getting it 100% right here.)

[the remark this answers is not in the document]
jdesrosiers: I think we agree that the vocabulary system is an extension mechanism. It allows you to use custom keywords in your schemas. But, we decided that the vocabulary system was too cumbersome for users who just wanted to use simple annotation custom keywords, so we introduced `x-` as an alternative for users to define simple custom keywords without the overhead of the vocabulary system. Both are extension mechanisms for using custom keywords. One is complicated and powerful and the other is simple and highly constrained, but they're serving the same purpose.

`x-` is so constrained that I can see how you can view it as one thing defined by the spec, but I think there's more to it. `x-foo` and `x-bar` have the same annotation collection and validation behavior, but they aren't the same keyword. They have different semantics given to them by the schema author, not by the spec. I see the spec defined behavior of `x-` as constraints that `x-` extension keywords must adhere to, but it's not a complete definition. It take the same author to define the semantics for it to be a complete keyword.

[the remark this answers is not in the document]
gregsdennis: Okay.  Are you happy, then, to have `x-` keywords clarified as being "recognized" (using that word) in #1518?

jsonschema-core.md
jdesrosiers: I'm not convinced the rewording in #1518 really addresses this concern. Not enough anyway. I think that rewording here is probably a better approach. Something like,

> Implementations MUST refuse to evaluate schemas which contain keywords they do not know how to process.

I think that's clear, covers all the bases, and avoids confusion about whether `x-` counts as a recognized keyword. Implementations know how to process `x-` keywords, so it's clear this statement isn't referring to those keywords.

jsonschema-core.md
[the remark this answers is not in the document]
jdesrosiers: I think that would work, but I just posted in the other tread an alternative that I think would be better. It avoids the problem altogether.

jsonschema-core.md
gregsdennis: That also addresses implementations which choose to support custom keywords outside of vocabs.

gregsdennis: Updated with one tweak.

jdesrosiers (approved):
I still think that paragraph in the overview section should reference `x-` as an extension mechanism, but I'll let you decide if you want to make that change. Otherwise, this looks good to me.

gregsdennis:
I'm going to leave it out of the extensions section.

> `x-foo` and `x-bar` have the same annotation collection and validation behavior, but they aren't the same keyword. They have different semantics given to them by the schema author, not by the spec.

You're not wrong about this.  However, those extra semantics only apply outside of the context of evaluating the schema, regardless of how or for what purpose the evaluation occurs.

The specification is saying that `x-` keywords are annotations only, which means the only semantics they carry or behavior they exhibit _during evaluation of the schema_ is collection of their values as annotations.  Any other semantics that a schema author or application infers _must_ be carried out outside of the context of evaluation.

jsonschema-core.md
Relequestual: I would be happier if there was an Issue which identifies the inconsistency which is part of the stable release milestone. I'm fine if #1510 resolves that Issue.
I'm just hesitent to approve a PR with a known contradiction **unless** there's a clear Issue to resolve it that likely won't be missed.

gregsdennis: There _is_ the discussion I linked.  It describes the problem in extreme detail, and I keep linking it in all the places that involves vocabs.

jviotti (approved):
This is my first time being involved in specs, so take my approval with a grain of salt, but at least all the proposed changes make perfect sense to me and it reads well. I think an implementer would definitely get it.

jsonschema-core.md
gregsdennis: The discussion isn't going to get lost, and it doesn't need to be addressed before the spec releases, or even before #1510 merges.  It doesn't need to be tracked on a project board or be associated with a milestone.

Relequestual: OK. I guess given #1510 is part of the milestone, I'm less concerned.

Relequestual (approved):

PR #1518, opened 2024-06-11
Improve "x-" keyword description

gregsdennis:
Per [this Slack thread](https://json-schema.slack.com/archives/CT7FF623C/p1717978845635409), we felt that the wording here doesn't do enough to describe "x-" keywords as "known" keywords.

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -424,11 +424,13 @@ Implementations MAY provide the ability to register or load handlers for
 vocabularies that they do not support directly. The exact mechanism for
 registering and implementing such handlers is implementation-dependent.
 
-#### Explicit annotation keywords {#explicit-annotations}
+#### Implicit annotation keywords {#implicit-annotations}
 
-The values of keywords which begin with "x-" MUST be collected as annotations.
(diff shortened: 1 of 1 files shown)


Relequestual (approved):
I was still unsure about the specific phrasing, but on reflection I think it does make sense and communicates the correct meaning and requirements.

jdesrosiers (commented):
Honestly, this still feels unclear to me. Readers have to make the connection that "implicitly defined" means that it's considered "recognized" by the implementation. I think it would be a lot easier on readers if we use the same terms in all the places we talk about this.

jdesrosiers (approved):
After discussion about avoiding the "recognized" ambiguity elsewhere, I'm no longer concerned about using that word in this section.

gregsdennis:
(Ref: https://github.com/json-schema-org/json-schema-spec/pull/1512#discussion_r1645256194)

gregsdennis:
Still going to pull these changes in since I think the wording is a bit cleaner, even if it doesn't solve the original ask.  That ask will be resolved as part o #1512.

jviotti (approved):

PR #1522, opened 2024-06-19
Fixes #1402: clarify `readOnly` applies to any value, not just properties

### LINKED ISSUE #1402: description of readOnly is ambiguous and refers to property when it should be instance
The [2020-12 validation spec](http://json-schema.org/draft/2020-12/json-schema-validation.html#name-readonly-and-writeonly) reads
> attempts by an application to modify the value of this property
This should be changed to
> attempts by an application to modify the value of this instance
since readOnly can apply to an object or array or primitive schema, not just properties on an object.

(I'd also like more clarity here when `readOnly: true` appears on a container instance - i.e. explicitly state whether this constraint does or does not extend to containers within the container. I believe it is local and only applies to the container)

DavidBiesack:
clarify that `readOnly` applies to any value, not just properties.
Re-format `true` and `false` code values

Resolves #1402

### DIFF
--- a/jsonschema-validation.md
+++ b/jsonschema-validation.md
@@ -1,4 +1,4 @@
-e# JSON Schema Validation: A Vocabulary for Structural Validation of JSON
+# JSON Schema Validation: A Vocabulary for Structural Validation of JSON
 
 ## Abstract
 
@@ -771,9 +771,10 @@ these keywords are applicable to a single sub-instance, the resulting behavior
 SHOULD be as for a `true` value if any occurrence specifies a `true` value, and
 SHOULD be as for a `false` value otherwise.
 
-If `readOnly` has a value of boolean `true`, it indicates that the value of the
-instance is managed exclusively by the owning authority, and attempts by an
-application to modify the value of an instance are expected to be ignored or
+If `readOnly` has a value of boolean `true`, it indicates that the
+corresponding value in the instance is
+managed exclusively by the owning authority, and attempts by an
+application to modify the value are expected to be ignored or
 rejected by that owning authority.
 
 An instance document that is marked as `readOnly` for the entire document MAY be
--- a/jsonschema-validation.md
+++ b/jsonschema-validation.md
@@ -1,4 +1,4 @@
-# JSON Schema Validation: A Vocabulary for Structural Validation of JSON
+e# JSON Schema Validation: A Vocabulary for Structural Validation of JSON
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


jsonschema-validation.md
```diff
@@ -1,4 +1,4 @@
-# JSON Schema Validation: A Vocabulary for Structural Validation of JSON
+e# JSON Schema Validation: A Vocabulary for Structural Validation of JSON
```
gregsdennis, line 1: accidental key press?

jdesrosiers, line 1:
```suggestion
# JSON Schema Validation: A Vocabulary for Structural Validation of JSON
```

jsonschema-validation.md
```diff
@@ -748,39 +748,39 @@ associated schema.
...
-SHOULD be as for a true value if any occurrence specifies a true value, and
-SHOULD be as for a false value otherwise.
+SHOULD be as for a `true` value if any occurrence specifies a `true` value, and
+SHOULD be as for a `false` value otherwise.
 
-If `readOnly` has a value of boolean true, it indicates that the value of the
+If `readOnly` has a value of boolean `true`, it indicates that the value of the
 instance is managed exclusively by the owning authority, and attempts by an
-application to modify the value of this property are expected to be ignored or
+application to modify the value of an instance are expected to be ignored or
```
jdesrosiers, line 776: When I read "an instance" my mind went to the "the entire JSON document", which isn't the intention. I think "instance" can be used to refer to a fragment of a JSON document, but we usually don't use it that way so this might be confusing.

What if we just leave it at "the value" and drop the "of" part entirely? Or we could say "that value within an instance"?

jdesrosiers (commented):
Thanks for doing this and for taking the time to do the additional cleanup.

jsonschema-validation.md
gregsdennis: or "the local instance"

DavidBiesack: how about : 
```
If `readOnly` has a value of boolean `true`, it indicates that the
corresponding value in the instance is
managed exclusively by the owning authority, and attempts by an
application to modify the value are expected to be ignored or
rejected by that owning authority.
```

gregsdennis: Got that from https://www.learnjsonschema.com/2020-12/meta-data/readonly/

gregsdennis (approved):

jdesrosiers (approved):

DavidBiesack:
is his ready to merge? I'm not sure how to reset the "PR Policy / Require Minimum Open Time (pull_request)" check

gregsdennis:
We can rerun it from our end, but you can also close then reopen the PR.

It'll fail if the PR hasn't been open for two weeks. It's a policy we've had in place for a long time.

PR #1525, opened 2024-06-20
update issue and pr templates

gregsdennis:
I'm not completely set on these.  Happy to change them if needed.  It's just something thrown together to get started.

Probably doesn't need the 2w minimum time.


jviotti (approved):
Looks decent. Mainly with these things, the best thing you can is to just try them out for real and see how it feels. Incrementally tweaking them if needed sounds more productive than discussing a lot here

Issue #1349, opened 2022-11-17
URI/IRI "Normalization" and Compatibility


watuwo:
The current core specification (2020-12) requires that
`"$schema"` (and some other) URIs "MUST be normalized".
While RFC 3986 makes suggestions for "normalization" steps,
it does not define a normal form for URIs (IRIs analogous).

"normalized"/"normal"/"normalization"/... (of URIs/IRIs) is
not well-defined. I propose that these terms are either
eliminated from the specification or defined.

My current personal preference: Eliminate the
"normalization" constraint.

---

A thought about the switch from URIs to IRIs:

An implementation could/should reject a schema if

- the resolution of `"$id"` yields an IRI that is not also
  a URI
- AND the schema's meta schema declares
  
  ```
  ...
  "$vocabulary": {
    ...
    "https://json-schema.org/draft/2020-12/vocab/core": true
    ...
  ```

Is this true now? Can this still occur in the future? Is
rejection still possible/desired? What about this:

> Meta-schemas that do not use "$vocabulary" MUST be
> considered to require the Core vocabulary as if its URI
> were present with a value of true.

handrews:
We specify normalization to improve the likelihood that a URI or IRI is recognized as a known resource.  Comparison and equivalence is the context in which normalization is defined in RFC 3986.

Normalization is a well-understood industry term, and while it is hard to be exact about it, JSON Schema should not try to step in on behalf of other specifications.  The one place where we can do this is to note that that for documents of media type `application/schema+json`, a URI with a trailing empty fragment can be normalized to a URI without a fragment (because that is under the control of the media type specification).

Otherwise, the point of the normalization requirement is to avoid requiring JSON Schema implementations to implement normalization themselves.  There is nothing in JSON Schema that requires or encourages an implementation to detect an insufficiently normalized URI and error on it.  Failing to normalize your `"$schema"` URI, for example, just means that an implementation might not recognize it and might therefore refuse to process your schema.

> the resolution of `"$id"` yields an IRI that is not also a URI

All IRIs can be [mapped to URIs](https://datatracker.ietf.org/doc/html/rfc3987#section-3.1).

watuwo:
I agree with everything you say about "normalization". The question is whether the phrase "MUST be normalized" expresses well what you wrote in your reply.

[...]

PR #1537, opened 2024-10-01
`$id` updates

gregsdennis:

Cleans up some language around `$id`.

- Related to #1444
- Related to #1349

### Summary

Just some clean-up and clarification that we've previously identified.

> Schema document roots SHOULD have an $id (maybe relax this? ...) - from https://github.com/json-schema-org/json-schema-spec/issues/1444#issuecomment-1749604538

I decided to leave this in for now.  We can discuss and change separately if needed.

No

### DIFF
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -895,8 +895,10 @@ by other parties.
 ### Base IRI, Anchors, and Dereferencing
 
 To differentiate between schemas in a vast ecosystem, schemas are identified by
-[IRI](#rfc3987), and can embed references to other schemas by specifying their
-IRI.
+[absolute IRIs](#rfc3987) (without fragments), and can embed references to other
+schemas by specifying their IRI. When comparing IRIs, implementations SHOULD
+interpret them using the normalization procedures defined in
+[RFC 3987](#rfc3987), section 5.3.
 
 Several keywords can accept a relative [IRI reference](#rfc3987), or a value
 used to construct a relative IRI reference. For these keywords, it is necessary
@@ -904,32 +906,23 @@ to establish a base IRI in order to resolve the reference.
 
 #### The `$id` Keyword {#id-keyword}
 
-The `$id` keyword identifies a schema resource with its [canonical](#rfc6596)
-IRI.
+The `$id` keyword identifies a schema resource. The value for this keyword MUST
+be a string, and MUST represent a valid [IRI reference](#rfc3987) (without a
+fragment).
+
+When the value of this keyword is resolved against the current base IRI, the
+resulting absolute IRI then serves as the identifier for the schema resource and
+as a base IRI for relative IRI references in keywords within that schema
+resource, in accordance with [RFC 3987 section 6.5](#rfc3987) and
+[RFC 3986 section 5.1.1](#rfc3986) regarding base IRIs embedded in content.
 
 Note that this IRI is an identifier and not necessarily a network locator. In
 the case of a network-addressable URL, a schema need not be downloadable from
 its canonical IRI.
 
-If present, the value for this keyword MUST be a string, and MUST represent a
-valid [IRI reference](#rfc3987). This IRI reference SHOULD be normalized per RFC
-3987, section 5.3, and MUST resolve to an [absolute IRI](#rfc3987) (without a
-fragment).
-
-The resulting absolute IRI serves as the base IRI for relative IRI references in
-keywords within the schema resource, in accordance with [RFC 3987 section
-6.5](#rfc3987) and [RFC 3986 section 5.1.1](#rfc3986) regarding base IRIs
-embedded in content.
-
-The presence of `$id` in a subschema indicates that the subschema constitutes a
-distinct schema resource within a single schema document. Furthermore, in
-accordance with [RFC 3987 section 6.5](#rfc3987) and [RFC 3986 section
-5.1.2](#rfc3986) regarding encapsulating entities, if an `$id` in a subschema is
-a relative IRI reference, the base IRI for resolving that reference is the IRI
-of the parent schema resource. Note that an `$id` consisting of an empty IRI or
-of the empty fragment only will result in the embedded resource having the same
-IRI as the encapsulating resource, which SHOULD be considered an error per
-{{duplicate-iris}}.
+Also note that an `$id` consisting of an empty IRI only will result in the
+embedded resource having the same IRI as the encapsulating resource, which
+SHOULD be considered an error per {{duplicate-iris}}.
 
 If no parent schema object explicitly identifies itself as a resource with
 `$id`, the base IRI is that of the entire document, as established by the steps
@@ -1386,6 +1379,17 @@ Note that single-level custom keywords with identical syntax and semantics to
 behave correctly under implementations that attempt to use any reference target
 as a schema. However, this behavior is implementation-specific and MUST NOT be
 relied upon for interoperability.
+
+A reference target under a keyword for which the value is known not to be a
+schema results in undefined behavior in order to avoid burdening implementations
+with the need to detect such targets.[^10]
+
+[^10]: These scenarios are analogous to fetching a schema over HTTP but
+receiving a response with a Content-Type other than `application/schema+json`.
+An implementation can certainly try to interpret it as a schema, but the origin
+server offered no guarantee that it actually is any such thing. Therefore,
+interpreting it as such has security implication and may produce unpredictable
+results.
 
 #### Failure to resolve references {#failed-refs}
 
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -896,9 +896,9 @@ by other parties.
 
 To differentiate between schemas in a vast ecosystem, schemas are identified by
 [absolute IRIs](#rfc3987) (without fragments) and can embed references to other
-schemas by specifying their respective IRIs. When comparing IRIs,
-implementations SHOULD first follow the IRI normalization procedures defined in
-[RFC 3987](#rfc3987), section 5.3.
+schemas by specifying their respective IRIs. When comparing IRIs for the
+purposes of resource identification, implementations SHOULD first follow the IRI
+normalization procedures defined in [RFC 3987](#rfc3987), section 5.3.
 
 Several keywords can accept a relative [IRI reference](#rfc3987), or a value
 used to construct a relative IRI reference. For these keywords, it is necessary
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -915,7 +915,7 @@ resulting absolute IRI then serves as the identifier for the schema resource and
 as a base IRI for relative IRI references in keywords within that schema
 resource and for embedded schema resources, in accordance with [RFC 3987 section
 6.5](#rfc3987) and [RFC 3986 section 5.1.1](#rfc3986) regarding base IRIs
-embedded in content.
+embedded in content and RFC 3986 section 5.1.2 regarding encapsulating entities.
 
 Note that this IRI is an identifier and not necessarily a network locator. In
 the case of a network-addressable URL, a schema need not be downloadable from
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -912,8 +912,9 @@ the case of a network-addressable URL, a schema need not be downloadable from
 its canonical IRI.
 
 If present, the value for this keyword MUST be a string, and MUST represent a
-valid [IRI reference](#rfc3987). This IRI reference SHOULD be normalized, and
-MUST resolve to an [absolute IRI](#rfc3987) (without a fragment).
+valid [IRI reference](#rfc3987). This IRI reference SHOULD be normalized per RFC
+3987, section 5.3, and MUST resolve to an [absolute IRI](#rfc3987) (without a
+fragment).
 
 The resulting absolute IRI serves as the base IRI for relative IRI references in
 keywords within the schema resource, in accordance with [RFC 3987 section
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -906,14 +906,14 @@ to establish a base IRI in order to resolve the reference.
 
 #### The `$id` Keyword {#id-keyword}
 
-The `$id` keyword identifies a schema resource. The value for this keyword MUST
-be a string, and MUST represent a valid [IRI reference](#rfc3987) without a
-fragment.
+An `$id` keyword in a schema or subschema identifies that schema or subschema as
+a distinct schema resource. The value for this keyword MUST be a string, and
+MUST represent a valid [IRI reference](#rfc3987) without a fragment.
 
 When the value of this keyword is resolved against the current base IRI, the
 resulting absolute IRI then serves as the identifier for the schema resource and
 as a base IRI for relative IRI references in keywords within that schema
-resource and for nested schema resources, in accordance with [RFC 3987 section
+resource and for embedded schema resources, in accordance with [RFC 3987 section
 6.5](#rfc3987) and [RFC 3986 section 5.1.1](#rfc3986) regarding base IRIs
 embedded in content.
 
@@ -1370,20 +1370,15 @@ recursive nesting like this; the behavior is undefined.
 #### References to Possible Non-Schemas {#non-schemas}
 
 Subschema objects (or booleans) are recognized by their use with known
-applicator keywords or with location-reserving keywords such as
-[`$defs`](#defs) that take one or more subschemas as a value. These keywords may
-be `$defs` and the standard applicators from this document or
-implementation-specific custom keywords.
+applicator keywords or with location-reserving keywords, such as
+[`$defs`](#defs), that take one or more subschemas as a value. These keywords
+include the standard applicators from this document or implementation-specific
+custom keywords.
 
-Note that single-level custom keywords with identical syntax and semantics to
-`$defs` do not allow for any intervening `$id` keywords, and therefore will
-behave correctly under implementations that attempt to use any reference target
-as a schema. However, this behavior is implementation-specific and MUST NOT be
-relied upon for interoperability.
-
-A reference target under a keyword for which the value is known not to be a
-schema results in undefined behavior in order to avoid burdening implementations
-with the need to detect such targets.[^10]
+A reference target under a keyword for which the value is not explicitly known
+to be a schema results in undefined behavior. Implementations MAY support
+references to these locations, however such behavior is not considered
+interoperable and should not be relied upon.[^10]
 
 [^10]: These scenarios are analogous to fetching a schema over HTTP but
 receiving a response with a Content-Type other than `application/schema+json`.
--- a/jsonschema-core.md
+++ b/jsonschema-core.md
@@ -1269,7 +1269,7 @@ desired to identify the object containing the `$ref` in the second
 Due to the potential break in functionality described above, the behavior for
 using JSON Pointer fragments that point to or cross a resource boundary is
 undefined.  Schema authors SHOULD NOT rely on such IRIs, as using them may
-reduce interoperability.
+reduce interoperability.[^8]
 
 [^8]: This is to avoid requiring implementations to keep track of a whole stack
 of possible base IRIs and JSON Pointer fragments for each, given that all but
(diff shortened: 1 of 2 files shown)


jviotti (approved):
Nice! Most of the changes are simple one-character edits, and the rest read very well to me

jsonschema-core.md
```diff
@@ -1271,10 +1266,10 @@ the `$id` of the embedded or referenced resource unless it is specifically
...
 (non-embedded) arrangement.
 
-An implementation MAY choose not to support addressing schema resource contents
-by IRIs using a base other than the resource's canonical IRI, plus a JSON
-Pointer fragment relative to that base. Therefore, schema authors SHOULD NOT
-rely on such IRIs, as using them may reduce interoperability.[^8]
+Due to the potential break in functionality described above, the behavior for
+using JSON Pointer fragments that point to or cross a resource boundary is
+undefined.  Schema authors SHOULD NOT rely on such IRIs, as using them may
+reduce interoperability.
```
notEthan, line 1272: It seems like it's not undefined - calling it undefined seems to me to contradict the above "an embedded schema resource and its subschemas can be identified by JSON Pointer fragments relative to either its own canonical IRI, _or relative to any containing resource's IRI_".

This also leaves `[^8]` with no reference to it.

And, minor, but two spaces after a period isn't consistent with the rest of the document.

jsonschema-core.md
```diff
@@ -1386,28 +1381,13 @@ applicator keywords or with location-reserving keywords such as
 be `$defs` and the standard applicators from this document or
 implementation-specific custom keywords.
 
-Multi-level structures of unknown keywords are capable of introducing nested
-subschemas, which would be subject to the processing rules for `$id`. Therefore,
-having a reference target in such an unrecognized structure cannot be reliably
-implemented, and the resulting behavior is undefined. Similarly, a reference
-target under a known keyword, for which the value is known not to be a schema,
-results in undefined behavior in order to avoid burdening implementations with
-the need to detect such targets.[^10]
```
notEthan, line 1395: This seems to me to be a useful thing to address a longstanding issue/ambiguity , why remove it? Addressing schemas in unknown locations should be called out as undefined behavior.

jsonschema-core.md
gregsdennis: It was discussed that we need to make pointer that cross boundaries an undefined behavior rather than allowing implementations to support it.  Of course, implementations can still support it if they want, but the spec should push schema authors a bit more forcefully toward using the proper `$id` instead.

I think the critical point here is "an embedded schema resource _and its subschemas_".  It's arguable that an embedded schema resource isn't strictly a subschema since it can be refactored out without any change in behavior.  If we need to clarify that, I'm happy to.

jsonschema-core.md
gregsdennis: Unknown keywords are no longer a thing, so calling out data structures under them no longer makes sense.

jsonschema-core.md
```diff
@@ -2230,32 +2210,22 @@ listed IRI in accordance with {{fragments}} and {{embedded}} above.
 `#/$defs/B`: canonical (and base) `IRI: https://example.com/other.json`
 - canonical resource IRI plus pointer fragment:
   `https://example.com/other.json#`
-- base IRI of enclosing (root.json) resource plus fragment:
-  `https://example.com/root.json#/$defs/B`
 
 `#/$defs/B/$defs/X`: base IRI: `https://example.com/other.json`
```
notEthan, line 2214: It's a bit weird that this refers to the location using `#/$defs/B/$defs/X` in IRI format but that's not actually an IRI to be used to reference it. I wonder if changing these to pointer format without being in an IRI fragment would be better to distinguish them from IRIs used to resolve references.

gregsdennis: I could be talked into using a plain pointer in this appendix for these, yeah.  Let's see what others think.

gregsdennis: For reference, I would reformat like this:

```markdown
Document location: `/$defs/B/$defs/X`
- base IRI: `https://example.com/other.json`
- canonical resource IRI plus plain fragment:
  `https://example.com/other.json#bar`
- canonical resource IRI plus pointer fragment:
  `https://example.com/other.json#/$defs/X`
```

There's not any IRI-encoding going on in these pointers, so the difference is minimal, but it could communicate the point that the item header pointers are just indicating the location in the above document as written rather than a location within the schema.

gregsdennis: I also considered doing a table here, but I wasn't sure whether we had set up the transpiler to support those.  We don't have any in the specs yet, either.  (Adventures in markdown!)

jsonschema-core.md
notEthan: Oh right, I'd forgotten that part, that's fine then.

jsonschema-core.md
notEthan: Right, I get making this undefined is the main point of this PR (I edited my comment a bit to clarify that, had it slightly wrong initially ... I'm not sure this _should_ be undefined but if that is the way it is to be, okay).

To me "an embedded schema resource ... can be identified by JSON Pointer fragments ... relative to any containing resource's IRI" reads as defined behavior and not consistent with making it undefined some paragraphs later, and needs some change. I see your point on ambiguity whether "subschema" includes embedded schemas but I don't see language present that actually calls embedded schemas subschemas.

gregsdennis: I read that paragraph as informative, not prescriptive.  It's setting up the reasoning behind the requirement.

> Embedded schemas _can_ be referenced this way, but because reasons it's a bad idea to actually do it, so we're gonna say probably don't.  We're also letting implementations decide whether they want to support it, no pressure either way.

jsonschema-core.md
gregsdennis: Done.  @notEthan please let me know if this is better.

jdesrosiers: > I also considered doing a table here, but I wasn't sure whether we had set up the transpiler to support those. We don't have any in the specs yet, either. (Adventures in markdown!)

The markdown compiler supports Github-flavored markdown, so you should have available to you all the [table syntax you have in Github](https://github.com/remarkjs/remark-gfm?tab=readme-ov-file#use). There are also [plugins to expand on the table features](https://github.com/wataru-chocola/remark-extended-table) we can add if we need it.

jsonschema-core.md
```diff
@@ -2265,6 +2235,29 @@ determines the canonical nature of the resulting full IRI.[^18]
 and direct you to read the CREF located in the [JSON Pointer fragments and
 embedded schema resources](#embedded) section for further comments.
 
+While the following IRIs do correctly indicate specific schemas, per the reasons
+outlined in {{embedded}}, they are to be avoided:
```
jdesrosiers, line 2245: I think this language needs to be a little stronger. As a reader, I might assume this statement means it's just poor style and not realize that it there is an interoperability concern.

```suggestion
While the following IRIs do correctly indicate specific schemas, per the reasons
outlined in {{embedded}}, they may not work in all implementations:
```

jdesrosiers (changes requested):
Thanks for tackling these issues!

jsonschema-core.md
gregsdennis: I think stating both is probably good.

```suggestion
While the following IRIs do correctly indicate specific schemas, per the reasons
outlined in {{embedded}}, they are to be avoided as they may not work in all implementations:
```

The "are to be avoided" is quite strong IMO, but I agree with adding _why_.

jsonschema-core.md
```diff
@@ -912,10 +912,11 @@ the case of a network-addressable URL, a schema need not be downloadable from
 its canonical IRI.
 
 If present, the value for this keyword MUST be a string, and MUST represent a
-valid [IRI-reference](#rfc3987). This IRI-reference SHOULD be normalized, and
-MUST resolve to an [absolute-IRI](#rfc3987) (without a fragment).
+valid [IRI reference](#rfc3987). This IRI reference SHOULD be normalized per RFC
+3987, section 5.3, and MUST resolve to an [absolute IRI](#rfc3987) (without a
+fragment).
```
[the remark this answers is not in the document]
gregsdennis: I'll have a think about this.  I don't disagree, but I'll need to figure out where to put such a requirement.  This section does warrant a note to authors that implementations will be performing this normalization; probably a link to wherever this ends up.

jsonschema-core.md
jdesrosiers: I'm unresolving this conversation because my comment wasn't addressed. I assume it was missed. The second part of this paragraph is still relevant. It deals with known keywords. That means the footnote is still relevant as well, but we don't necessarily need to keep it.

> Similarly, a reference target under a _known keyword_, for which the value is known not to be a schema, results in undefined behavior in order to avoid burdening implementations with the need to detect such targets.[^10]

gregsdennis: Yeah I missed it. I think there was a concurrency error between me resolving the thread and you adding your comment.

I'll add that bit about known keywords back in.

jsonschema-core.md
gregsdennis: I might play with tables later.

jsonschema-core.md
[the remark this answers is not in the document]
gregsdennis: Leaving it in place and changing to

> If present, the value for this keyword MUST be a string, and MUST represent a valid [IRI reference](#rfc3987). When processing this IRI reference, implementations SHOULD normalize it per RFC 3987, section 5.3, so that it resolves to an [absolute IRI](#rfc3987) (without a fragment).

seems to work.  It puts the requirement on the implementation, but it also lets the user (if they're even reading this) know what they can expect from the implementation.

I do fear, though, that this gives license to users to put IRIs _with_ fragments in `$id`.  (It's almost like the output needs the ability to report a warning as well as an error. "Hey, uh, there's a fragment on that IRI, guy. You sure you want to do that? Probably don't.")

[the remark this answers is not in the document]
jdesrosiers: I don't think leaving it in place works because `$id` isn't the only way a schema might be assigned an IRI. For example, implementations often also allow users to assign IRIs to schemas when they're registered. The normalization requirement should also apply to those user provided IRIs and therefore needs to be somewhere where it doesn't only apply to `$id`.

One place I think it could work is in "Base IRI, Anchors, and Dereferencing" above. Here's a suggestion.

> To differentiate between schemas in a vast ecosystem, schemas are identified by [absolute IRIs](#rfc3987) (without a fragment).
>
> Schemas can embed references to other schemas by specifying their IRI. When implementations dereference these references, they SHOULD use the normalization procedures defined in RFC 3987, section 5.3 when comparing URIs.
>
> Several keywords can accept an [IRI reference](#rfc3987). The schema's identifier serves as the base IRI for resolving relative references.

Then I'd remove everything about normalization and fragments from this paragraph. If we want, we could include an informal warning that fragments are ignored because schemas are identified by absolute IRIs, but it shouldn't technically be necessary.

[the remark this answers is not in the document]
gregsdennis: We _do_ need information on valid values for `$id`, though.  I feel that's what at least the first sentence does.  I'll play with it.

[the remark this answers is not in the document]
jdesrosiers: Right, I was only suggesting removing the part about normalization and fragments. The part about it being an IRI reference needs to stay.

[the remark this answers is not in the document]
gregsdennis: I've moved the normalization up to the previous section as suggested.  I had to play with the wording a bit.

gregsdennis:
@notEthan @jdesrosiers I've almost completely reworked the `$id` section.  There were some redundancies in it that didn't sit right with me.

I moved the syntactic definition of the keyword up to the top sentence next to its purpose, and I reworked how it talks about defining a resource, needing to be resolved against the current base IRI, and then providing a base IRI for child resources.

This relationship is necessarily recursive (the `$id` value needs a base IRI but then also serves as a base IRI), so it's quite difficult to state simply.

jsonschema-core.md
```diff
@@ -895,49 +895,44 @@ by other parties.
 ### Base IRI, Anchors, and Dereferencing
 
 To differentiate between schemas in a vast ecosystem, schemas are identified by
-[IRI](#rfc3987), and can embed references to other schemas by specifying their
-IRI.
+[absolute IRIs](#rfc3987) (without fragments) and can embed references to other
+schemas by specifying their respective IRIs. When comparing IRIs for the
```
jdesrosiers, line 899: This sentence doesn't sound right. I think there's something missing?

```suggestion
[absolute IRIs](#rfc3987) (without fragments) and can be used to embed references to other
schemas by specifying their respective IRIs. When comparing IRIs for the
```

jsonschema-core.md
```diff
@@ -895,49 +895,44 @@ by other parties.
...
+used to construct a relative IRI reference. For these keywords, it is necessary
 to establish a base IRI in order to resolve the reference.
 
 #### The `$id` Keyword {#id-keyword}
 
-The `$id` keyword identifies a schema resource with its [canonical](#rfc6596)
-IRI.
+The `$id` keyword identifies a schema resource. The value for this keyword MUST
+be a string, and MUST represent a valid [IRI reference](#rfc3987) (without a
+fragment).
```
jdesrosiers, line 911: I suggest removing the parens in this statement. Previously, we used this with "absolute-IRI". In that case, the definition of absolute IRI already disallows a fragments and the parenthetical is just a helpful reminder of that fact.

However, fragments _are_ normally allowed in IRI references, so disallowing fragments is an important part of the requirement, not an aside, which makes it confusing why it's in parens.

```suggestion
The `$id` keyword identifies a schema resource. The value for this keyword MUST
be a string, and MUST represent a valid [IRI reference](#rfc3987) without a
fragment.
```

jsonschema-core.md
```diff
@@ -1271,10 +1260,10 @@ the `$id` of the embedded or referenced resource unless it is specifically
 desired to identify the object containing the `$ref` in the second
 (non-embedded) arrangement.
 
-An implementation MAY choose not to support addressing schema resource contents
-by IRIs using a base other than the resource's canonical IRI, plus a JSON
-Pointer fragment relative to that base. Therefore, schema authors SHOULD NOT
-rely on such IRIs, as using them may reduce interoperability.[^8]
+Due to the potential break in functionality described above, the behavior for
+using JSON Pointer fragments that point to or cross a resource boundary is
+undefined.  Schema authors SHOULD NOT rely on such IRIs, as using them may
```
jdesrosiers, line 1265:
```suggestion
undefined. Schema authors SHOULD NOT rely on such IRIs, as using them may
```

jsonschema-core.md
```diff
@@ -895,49 +895,44 @@ by other parties.
...
-keywords within the schema resource, in accordance with [RFC 3987 section
-6.5](#rfc3987) and [RFC 3986 section 5.1.1](#rfc3986) regarding base IRIs
-embedded in content.
-
-The presence of `$id` in a subschema indicates that the subschema constitutes a
-distinct schema resource within a single schema document. Furthermore, in
-accordance with [RFC 3987 section 6.5](#rfc3987) and [RFC 3986 section
-5.1.2](#rfc3986) regarding encapsulating entities, if an `$id` in a subschema is
-a relative IRI-reference, the base IRI for resolving that reference is the IRI
-of the parent schema resource. Note that an `$id` consisting of an empty IRI or
```
jdesrosiers, line 928: This needs to stay. As much as I'd like to remove the concept of embedded schemas from the spec, I don't think that was your intention.

jsonschema-core.md
```diff
@@ -1386,13 +1375,15 @@ applicator keywords or with location-reserving keywords such as
...
-having a reference target in such an unrecognized structure cannot be reliably
-implemented, and the resulting behavior is undefined. Similarly, a reference
-target under a known keyword, for which the value is known not to be a schema,
-results in undefined behavior in order to avoid burdening implementations with
-the need to detect such targets.[^10]
+Note that single-level custom keywords with identical syntax and semantics to
+`$defs` do not allow for any intervening `$id` keywords, and therefore will
+behave correctly under implementations that attempt to use any reference target
+as a schema. However, this behavior is implementation-specific and MUST NOT be
+relied upon for interoperability.
```
jdesrosiers, line 1382: I find this paragraph confusing. The word "intervening" doesn't make sense in this context. I also found it awkward that it starts by saying it will behave correctly, then says it must not be relied upon. I think it would be more clear if it was the other way around.

I think this was originally about unknown keywords. Since unknown keywords aren't allowed anymore, it becomes about custom vocabulary defined keywords. Therefore, I think this is saying that you can't use `$id` in a schema in a custom keyword. I don't think that's what we want. `$id` should work in any known schema, even in custom keywords.

jdesrosiers (changes requested):
Sorry, for the delay in reviewing.

jsonschema-core.md
gregsdennis: I think "which can be used to" is probably what's intended.  The IRIs are doing the embedding, not the schemas.  Then the "by specifying their respective IRIs" probably just gets removed.

jsonschema-core.md
gregsdennis: Was thinking similarly.

jsonschema-core.md
jdesrosiers: Yes, that sounds better.

Does the "embed" part make sense? Should it just say "to reference schemas".

jsonschema-core.md
gregsdennis: This paragraph was just moved up.  It was previously at line 1404 (which you can see just below).

> I also found it awkward that it starts by saying it will behave correctly, then says it must not be relied upon.

The text is "will behave correctly under implementations that attempt to use any reference target as a schema" which is correct.

> I think this was originally about unknown keywords.

It's not just about unknown keywords.  It's also about keywords that can define structure but not subschemas, like `default` or `const` (or custom keywords with structured data).  The point of this paragraph is to say that `$id` in any place where a schema is not expected is to treated as data, not an identifier.

jsonschema-core.md
gregsdennis: What about this?

> To differentiate between schemas in a vast ecosystem, schemas are identified by [absolute IRIs](#rfc3987) (without fragments).  These identifiers are used to create references between schemas.  When comparing...

jsonschema-core.md
jdesrosiers: Yes, I noticed that it was moved. I'm not arguing about its correctness. I can't even say anything about its correctness at this point because I don't understand what it's trying to say.

> It's not just about unknown keywords. It's also about keywords that can define structure but not subschemas, like `default` or `const` (or custom keywords with structured data).

I don't see how that paragraph can be understood to be about that. It appears this only applies to custom keywords with identical semantics to `$defs`.

> The point of this paragraph is to say that `$id` in any place where a schema is not expected is to treated as data, not an identifier.

That's well stated and clear. I suggest replacing the paragraph with some variation of that sentence.

jsonschema-core.md
jdesrosiers: > The first sentence in this paragraph is just restating the first sentence of the section, "The $id keyword identifies a schema resource."

I see what you mean, but I don't think readers will make the connection that it's not just an identifier for the root schema, but also away to embed a schema resource somewhere inside the root schema. Those who do make the connection will be looking for something else to confirm that connection is correct and won't find it.

> The second sentence here is replaced by the new text on lines 913-918.

Those two things don't say the same thing. The first refers to base IRIs embedded in content and the second refers to encapsulating entities. Those are treated as two different cases in RFC 3986/7, so I think we should treat them as two different cases as well.

I think the spec is better with that paragraph included.

gregsdennis: Previous text:

> The presence of `$id` in a subschema indicates that the subschema constitutes a
distinct schema resource within a single schema document. ... if an `$id` in a subschema is a relative IRI-reference, the base IRI for resolving that reference is the IRI of the parent schema resource.

New text:

> The `$id` keyword identifies a schema resource. ... When the value of [`$id`] is resolved against the current base IRI, the resulting absolute IRI then serves as the identifier for the schema resource and as a base IRI for relative IRI references in keywords within that schema resource and for nested schema resources...

These texts are saying the same things, specifically:

- `$id` identifies and declares a schema resource
- the value of `$id` is resolved against whatever comes from its parent

Maybe we need to define "current base IRI" better, but I'm almost sure it's defined elsewhere.

As I mentioned [here](https://github.com/json-schema-org/json-schema-spec/pull/1537#issuecomment-2408016806), the resolution of `$id` is somewhat recursive, which makes it hard to explain.

I'll have another go at this, but what's here needs improvement.

jdesrosiers: > for nested schema **_resources_**

I missed that last word. I thought this was talking about subschemas, not embedded schema resources. I think using the term "embedded" rather than "nested" is important to be consistent with how we talk about this concept throughout the spec and in discussion. Using a different word primed my brain to think you were explicitly not talking about embedded schema resources.

So, I think the new text works if the RFC reference mentions both "base IRIs embedded in content" _and_ "encapsulating entities".

To summarize, my suggestions are now
* Make it more explicit that `$id` in a subschema creates an embedded schema resource
* Use "embedded" instead of "nested"
* Reference "encapsulated entities" as well as "base IRIs embedded in content".

If we can do those things, I'm comfortable with the removed paragraph staying removed.

jsonschema-core.md
```diff
@@ -894,50 +894,45 @@ by other parties.
 
 ### Base IRI, Anchors, and Dereferencing
 
-To differentiate between schemas in a vast ecosystem, schemas are identified by
-[IRI](#rfc3987), and can embed references to other schemas by specifying their
-IRI.
+To differentiate between schemas in a vast ecosystem, schema resources are
+identified by [absolute IRIs](#rfc3987) (without fragments).  These identifiers
+are used to created references between schema resources. When comparing IRIs for
```
jdesrosiers, line 899:
```suggestion
identified by [absolute IRIs](#rfc3987) (without fragments). These identifiers
are used to create references between schema resources. When comparing IRIs for
```

jsonschema-core.md
gregsdennis: > Reference "encapsulated entities" as well as "base IRIs embedded in content".

I'm not sure where you're meaning for this.

jsonschema-core.md
gregsdennis: Yeah, reading it again, I don't think it's necessary.  It looks like it's just trying to explain that you _can't_ have an `$id` between the keyword simply because of the structure of the keyword.  I'm not sure that needs explaining, so I'll just remove it.

I'm not sure if we need to keep the "requirement" about "MUST NOT be relied upon". It's definitely a requirement on schema authors, but it's a weird one at that.

The next paragraph actually covers exactly what I was saying:

> A reference target under a keyword for which the value is known not to be a schema results in undefined behavior in order to avoid burdening implementations with the need to detect such targets.

jsonschema-core.md
jdesrosiers: Looks like you figured out what I meant.

> in accordance with [RFC 3987 section
6.5](#rfc3987) and [RFC 3986 section 5.1.1](#rfc3986) regarding base IRIs
embedded in content.

:point_up: That was what I wanted.

jdesrosiers: For example,

> in accordance with [RFC 3987 section 6.5](#rfc3987) and [RFC 3986 section 5.1.1](#rfc3986) regarding base IRIs embedded in content ***and [RFC 3986 section 5.1.2](#rfc3986) regarding encapsulating entities.***

Section 5.1.1 (base IRIs embedded in content) covers base IRIs.
Section 5.1.2 (encapsulating entities) covers embedded schema resources.

Both need to be referenced.

jdesrosiers (commented):
Latest changes look great. I think, https://github.com/json-schema-org/json-schema-spec/pull/1537/files#r1819699325 is the only thing left.

gregsdennis:
@jdesrosiers this should be done now.

jdesrosiers (approved):
Thanks for working with me through the details. This is a huge improvement and I'm excited to see this merged. :tada:

PR #1542, opened 2024-10-12
add a root object to the output

gregsdennis:

feature

Should probably merge #1457 first, but it's not a hard dependency.  If anything, I can move the file in this PR and just close that one.

### Summary

This PR addresses some conversations that @jdesrosiers and I have had intermittently over the past few years.  His point was that the output needed a place to add more metadata about an evaluation.  So this is a first go at that.

It adds a root object that is defined separately from the output unit.  Previously, the root object was just another output unit, but having a distinct root object does make sense.

I've added dialect and schema identifiers to the root object and tweaked the wording throughout.


gregsdennis:
Looks like I tried this already ☝️.  Will review the comments there and update this.

gregsdennis:
Closing and reopening to rerun build

NoClueMike:
...

PR #1543, opened 2024-10-16
Add automation for linting the spec

jdesrosiers:
When I setup the build script for the markdown-based spec, I included a markdown linter and link checker. Unfortunately, that isn't being used during development and a lot of errors have gotten in. This PR adds automation to make sure schema authors are aware when they are introducing errors.

This also includes linting automation for the build scripts themselves.


proposals/propertyDependencies-adr.md
```diff
@@ -1,14 +1,16 @@
...
+- Status: proposed
+- Deciders: @gregsdennis, @jdesrosiers, @relequestual
+- Date: 2022-04-07
 
 Technical Story:
 
-- Issue discussing feature - https://github.com/json-schema-org/json-schema-spec/issues/1082
-- PR to add to the spec - https://github.com/json-schema-org/json-schema-spec/pull/1143
-- ADR to extract from the spec and use feature life cycle - https://github.com/json-schema-org/json-schema-spec/pull/1505
+- Issue discussing feature - <https://github.com/json-schema-org/json-schema-spec/issues/1082>
```
gregsdennis, line 9: Why the angle brackets?

gregsdennis, line 9: Why the angle brackets?

gregsdennis (approved):
Looks good to me.  Didn't check the JS stuff, but I trust you on that.

Just need to fix the build, and I posted a question.

gregsdennis:
Our `.gitignore` is getting pretty specific.  Would it be possible to output all of the files to a specific folder so that we can just ignore that folder?

proposals/propertyDependencies-adr.md
jdesrosiers: The angle brackets are markdown syntax for URLs. We aren't used to using them because Github-flavored markdown detects URLs automatically and turns them into links. The markdown style guide the linting plugin is based on [prefers the explicit syntax](https://cirosantilli.com/markdown-style-guide/#automatic-links-without-angle-brackets). I think the rationale makes sense. By forcing you to use angle brackets or backticks, it helps catch errors where a URL is converted to a link when it wasn't intended to be.

jdesrosiers:
I originally removed the ietf build from the ci, but then remembered that we still have a couple specs (RJP) that still use it, so I put it back.

gregsdennis (approved):
Thanks for these.  I think we should merge this ASAP.  We can discuss on [today's OCWM](https://github.com/json-schema-org/community/issues/803).

PR #1547, opened 2024-10-20
refer to core regex section from patternProperties and pattern

### LINKED ISSUE #1478: Pattern: clarification on whether leading and trailing delimiters are allowed or required
So the definition of pattern is:

> 6.3.3. pattern
> 
> The value of this keyword MUST be a string. This string SHOULD be a valid regular expression, according to the ECMA-262 regular expression dialect.
> A string instance is considered valid if the regular expression matches the instance successfully. Recall: regular expressions are not implicitly anchored.

Does that include the regex literal delimiter `/` which means that pattern values must look like `/pattern/flags` (ECMA-262 15.10 or 7.85) or does it just include the pattern (ECMA-262 15.10.1)?
The spec is unclear because it does not link to a specific section of ECMA-262
Can this be clarified? What is allowed? What is required?

gregsdennis:

Clarification

-  Closes #1478 

### Summary

Updates the `patternProperties` (Core) and `pattern` (Validation) sections to point back to the Core section about regex.

No


jdesrosiers:
The link checker fails for this because it references the built result. I'm not sure what the best fix will be for this. Maybe there's some config in [remark-validate-links](https://github.com/remarkjs/remark-validate-links) that we can use that will fix the problem. If not, we can have links point to `.md`, but transform them to `.html` as part of the build.

Sarrac3873 (approved):

Sarrac3873:
> 
> 
> 
> 
> 
> 
> 
> ### What kind of change does this PR introduce?
> 
> 
> Clarification
> 
> ### Issue & Discussion References
> 
> 
> -  Closes #1478 
> 
> ### Summary
> 
> 
> 
> Updates the `patternProperties` (Core) and `pattern` (Validation) sections to point back to the Core section about regex.
> 
> ### Does this PR introduce a breaking change?
> 
> 
> No
>

Sarrac3873 (approved):

specs/jsonschema-validation.md
```diff
@@ -210,7 +210,9 @@ Omitting this keyword has the same behavior as a value of 0.
 #### `pattern` {#pattern}
 
 The value of this keyword MUST be a string. This string SHOULD be a valid
-regular expression, according to the ECMA-262 regular expression dialect.
+regular expression as indicated in
+[JSON Schema Core, section 6.3](./jsonschema-core.md#regex).
+<!-- This link assume a publish path, which we don't know yet. -->
```
jdesrosiers, line 215: It's a relative path, so it shouldn't matter where it's published in the end as long as they are published in the same relative location.

jdesrosiers (approved):

PR #1548, opened 2024-10-21
remove redundant sections about keyword independence

### LINKED ISSUE #1218: Section "Keyword independence" is written twice
The section "Keyword independence" is a short little section that describes how keywords don't (usually) interact with each other, and that this is a design consideration of JSON Schema. It lists some of the exceptions to this rule.

However, somewhere along the lines, this section was duplicated:

10.1. [Keyword Independence](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.1)
11.1. [Keyword Independence](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.11.1)

The doppelgänger should be removed.

gregsdennis:

Clarification

-  Closes #1218 

### Summary

Removes 2 redundant sections about keywords being independent.  We now have a "Keyword Independence" section, and each keyword section explicitly states their relations to other keywords.

No

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -1468,19 +1468,6 @@ operators can contact the owner of a potentially misbehaving script.
 This section defines a set of keywords that enable schema combinations and
 composition.
 
-### Keyword Independence
-
-Schema keywords typically operate independently, without affecting each other's
-outcomes.
-
-For schema author convenience, there are some exceptions among these keywords:
-
-- `additionalProperties`, whose behavior is defined in terms of `properties` and
-  `patternProperties`
-- `items`, whose behavior is defined in terms of `prefixItems`
-- `contains`, whose behavior is affected by the presence and value of
-  `minContains`
-
 ### Keywords for Applying Subschemas in Place {#in-place}
 
 These keywords apply subschemas to the same location in the instance as the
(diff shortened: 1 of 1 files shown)


jdesrosiers (approved):
In case anyone else is trying to find where this was stated elsewhere, it's in the "Keyword Interactions" section line 487.

```
### Keyword Interactions

Unless otherwise specified, keywords act independently.

Keywords MAY modify their behavior based on the presence, absence, or value of
another keyword in the same schema object. Such keywords MUST NOT result in a
circular dependency.

Supplementary specifications are encouraged to specify any dependencies as part
of the dependent keyword (i.e. the keyword whose behavior is modified).

Within this document, keyword dependencies are expressed using one of the
following mechanisms:

- Static dependencies, in which the dependency relies on the presence or
  contents of another keyword.
- Dynamic dependencies, in which the dependency relies on the evaluation of
  another keyword against an instance. This dependency may be on either the
  annotations produced by the keyword or the validation result of its
  subschema(s).

These mechanisms are used merely to describe dependencies; they are for
illustrative purposes and not prescriptive. Implementations MAY use whatever
mechanisms make sense given the needs of their architecture and language in
order to achieve the specified behaviors.
```

So with this change, we are no longer being explicit about which keywords violation the keyword independence principle. I think that's fine.

PR #1553, opened 2024-10-28
make format validate by default

### LINKED ISSUE #1520: Proposal: Make `format` validate by default
There's a long and sticky history around `format`.

1. Going back all the way to [Draft 01](https://json-schema.org/draft-01/draft-zyp-json-schema-01.html), `format` has never required validation.
2. Whether to support `format` validation has always been the decision of the implementation.
3. The extent to which formats are validated has also been the decision of the implementation.

The result of all of this is that implementation support for validation has been spotty at best.  Despite the JSON Schema specs referencing very concretely defined formats (by referencing other specs), implementations that do support validation don't all support each format equally.  This has been the primary driving force behind _keeping_ `format` as an opt-in validation.

With [2019-09](https://json-schema.org/draft/2019-09/draft-handrews-json-schema-validation-02#rfc.section.7.2), we decided that it was time to give the option of `format` validation to the schema author.  They could enable validation by using a meta-schema which listed the Format Vocabulary with a `true` value, which meant, "`format` validation is required to process this schema."

In [2020-12](https://json-schema.org/draft/2020-12/json-schema-validation#section-7.2), we further refined this by offering two separate vocabularies, one that treats the keyword as an annotation and one that treats it as an assertion.  The [argument](https://github.com/json-schema-org/json-schema-spec/issues/1020) was that the behavior of a keyword shouldn't change based on whether the vocabulary was required or not.

However, the fact remains that our users consistently report (via questions in Slack, GitHub, and StackOverflow) that they expect `format` to validate.  (The most recent case I can think of was only last week, in [.Net's effort](https://github.com/dotnet/runtime/pull/103322) to build a short-term solution for schema generation from types.)

**This consistency in user expectations leads me to believe that we should officially make `format` an assertion keyword and strictly enforce it by moving the appropriate tests into the required section of the Test Suite.**

(Personally, I'm not passing all of the optional `format` tests, so I'll have to do some work to get there or document why they're not supported.)

gregsdennis:

Feature update

Closes #1520 
Depends on https://github.com/json-schema-org/TSC/issues/19

### Summary

Makes `format` an assertion to align with user expectations.

Quick notes:

- Implementations SHOULD support `format` with the defined values
- Implementations MAY support others, but only by explicit config
- Implementations MUST refuse to process a schema that contains an unsupported format
- Permissive allowances removed since every format is clearly defined by referencing a specification

Technically a breaking change, but aligns to user expectations. So... ????

### DIFF
--- a/adr/2024-11-2-assertion-format.md
+++ b/adr/2024-11-2-assertion-format.md
@@ -2,10 +2,13 @@
 
 * Status: proposed
   <!-- will update below to only those who participated in the vote -->
-* Deciders: @gregsdennis @jdesrosiers @jviotti @mwadams @karenetheridge @awwright @benjam @relequestual
+* Deciders: @gregsdennis @jdesrosiers @julian @jviotti @mwadams @karenetheridge @relequestual
 * Date: 2024-11-02
 * Technical Story: https://github.com/json-schema-org/json-schema-spec/issues/1520
 * Voting issue: https://github.com/json-schema-org/TSC/issues/19
+    For - @gregsdennis @jdesrosiers @jviotti @mwadams @karenetheridge
+    Neutral - @relequestual
+    Against - @julian
 
 ## Context and Problem Statement
 
@@ -25,8 +28,8 @@ However, the fact remains that our users consistently report (via questions in S
 
 Due to this consistency in user expectations, we have decided to:
 
-1. make format an assertion keyword and strictly,
-2. enforce it by moving the appropriate tests into the required section of the Test Suite.
+1. make format an assertion keyword, and
+2. strictly enforce it by moving the appropriate tests into the required section of the Test Suite and building them more completely.
 
 ## Decision Drivers
 
@@ -65,7 +68,8 @@ The TSC has decided via vote (see voting issue above) that we should change `for
 
 ### Negative Consequences <!-- optional -->
 
-* This is a breaking change, which means that we will likely have to re-educate our users.
+* This is a breaking change, which means that we will likely have to re-educate the users who correctly treat it as an annotation.
+* Older schemas which do not specify a version (`$schema`) may change their validation outcome.
 * The burden on implementations will be greater since format validation was previously optional.
 
 ## Links <!-- optional -->
--- a/adr/2024-11-2-assertion-format.md
+++ b/adr/2024-11-2-assertion-format.md
@@ -1,0 +1,74 @@
+# [short title of solved problem and solution]
+
+* Status: proposed
+  <!-- will update below to only those who participated in the vote -->
+* Deciders: @gregsdennis @jdesrosiers @jviotti @mwadams @karenetheridge @awwright @benjam @relequestual
+* Date: 2024-11-02
+* Technical Story: https://github.com/json-schema-org/json-schema-spec/issues/1520
+* Voting issue: https://github.com/json-schema-org/TSC/issues/19
+
+## Context and Problem Statement
+
+There's a long and sticky history around format.
+
+1. Going back all the way to Draft 01, format has never required validation.
+2. Whether to support format validation has always been the decision of the implementation.
+3. The extent to which formats are validated has also been the decision of the implementation.
+
+The result of all of this is that implementation support for validation has been spotty at best. Despite the JSON Schema specs referencing very concretely defined formats (by referencing other specs), implementations that do support validation don't all support each format equally. This has been the primary driving force behind keeping format as an opt-in validation.
+
+With 2019-09, we decided that it was time to give the option of format validation to the schema author. They could enable validation by using a meta-schema which listed the Format Vocabulary with a true value, which meant, "format validation is required to process this schema."
+
+In 2020-12, we further refined this by offering two separate vocabularies, one that treats the keyword as an annotation and one that treats it as an assertion. The argument was that the behavior of a keyword shouldn't change based on whether the vocabulary was required or not.
+
+However, the fact remains that our users consistently report (via questions in Slack, GitHub, and StackOverflow) that they expect format to validate. (The most recent case I can think of was only last week, in .Net's effort to build a short-term solution for schema generation from types.)
+
+Due to this consistency in user expectations have decided to:
+
+1. make format an assertion keyword and strictly,
+2. enforce it by moving the appropriate tests into the required section of the Test Suite.
+
+## Decision Drivers
+
+* User expectation
+* Current behavior
+* Historical context
+* Disparity of current implementation support vs the proposed requirements
+
+## Considered Options
+
+### `format` remains an annotation keyword by default
+
+This is the current state. The primary benefit is that we don't need to make a breaking change.
+
+The primary downside is that the current system of (1) configuring the tool or (2) incluing the `format-assertion` vocab[^1] is confusing for many and doesn't align with user expectations.
+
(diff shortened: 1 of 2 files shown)


jsonschema-validation.md
```diff
@@ -292,39 +292,46 @@ Structural validation alone may be insufficient to allow an application to
...
 
-The value of this keyword is called a format attribute. It MUST be a string. A
-format attribute can generally only validate a given set of instance types. If
-the type of the instance to validate is not in this set, validation for this
-format attribute and instance SHOULD succeed. All format attributes defined in
-this section apply to strings, but a format attribute can be specified to apply
-to any instance types defined in the data model defined in the [core JSON
-Schema.](#json-schema)[^1]
+The value of this keyword MUST be a string. While this keyword can validate any
+type, each distinct value will generally only validate a given set of instance
```
jviotti, line 300: If I'm reading this correctly, it means we let `format` apply to instances other than strings. If I recall correctly, the JSON Schema test suite very explicitly tests that `format` only applies to instances, which might be a change that could confuse various people and implementations.

gregsdennis: > If I'm reading this correctly, it means we let format apply to instances other than strings.

Yes, `format` can apply to instances of any type.  This isn't a new feature, however, the spec only defines formats that apply to strings.

jdesrosiers:
This being a breaking change and a potentially controversial one, I think there should be a TSC vote to formally document that we have consensus before this moves forward. It should also have an ADR to document the decision.

gregsdennis:
> It should also have an ADR to document the decision.

I'll add one.

> I think there should be a TSC vote to formally document that we have consensus before this moves forward

@benjagm can you organize this, please?

adr/2024-11-2-assertion-format.md
```diff
@@ -0,0 +1,74 @@
...
+
+With 2019-09, we decided that it was time to give the option of format validation to the schema author. They could enable validation by using a meta-schema which listed the Format Vocabulary with a true value, which meant, "format validation is required to process this schema."
+
+In 2020-12, we further refined this by offering two separate vocabularies, one that treats the keyword as an annotation and one that treats it as an assertion. The argument was that the behavior of a keyword shouldn't change based on whether the vocabulary was required or not.
+
+However, the fact remains that our users consistently report (via questions in Slack, GitHub, and StackOverflow) that they expect format to validate. (The most recent case I can think of was only last week, in .Net's effort to build a short-term solution for schema generation from types.)
+
+Due to this consistency in user expectations have decided to:
```
jdesrosiers, line 26:
```suggestion
Due to this consistency in user expectations, we have decided to:
```

adr/2024-11-2-assertion-format.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. make format an assertion keyword and strictly,
```
jdesrosiers, line 28: > and strictly

This sentence feels awkward to me. I'm not sure it's clear what strictly means in this context. I know what you mean, but it can be stated more clearly for the sake of outside readers and maybe future us.

specs/jsonschema-validation.md
```diff
@@ -293,75 +293,60 @@ Structural validation alone may be insufficient to allow an application to
...
-format attribute and instance SHOULD succeed. All format attributes defined in
-this section apply to strings, but a format attribute can be specified to apply
-to any instance types defined in the data model defined in the [core JSON
-Schema.](#json-schema)[^1]
+The value of this keyword MUST be a string. While this keyword can validate any
+type, each distinct value will generally only validate a given set of instance
+types. If the type of the instance to validate is not in this set, validation
+for this keyword SHOULD succeed. All format values defined in this section apply
+to strings, but a format value can be specified to apply to any instance types
+defined in the data model defined in the [core JSON Schema](#json-schema)[^1].
```
jdesrosiers, line 305: This sentence sounds incomplete. How about something like ...

```suggestion
defined in the data model defined in the [core JSON Schema](#json-schema) specification[^1].
```

(I realize this isn't something you changed, but it's something we can clean up while we're here.)

specs/jsonschema-validation.md
```diff
@@ -293,75 +293,60 @@ Structural validation alone may be insufficient to allow an application to
...
-  to annotation-only behavior
-- SHOULD provide an implementation-specific best effort validation for each
-  format attribute defined below;[^3]
-- MAY choose to implement validation of any or all format attributes as a no-op
-  by always producing a validation result of true;[^4]
-- SHOULD use a common parsing library for each format, or a well-known regular
-  expression;
+Implementations SHOULD provide assertion behavior for the format values defined
+by this document[^2] and MUST refuse to process any schema which contains an
+unsupported format value.
```
jdesrosiers, line 316: I feel like it's people might think it's unclear whether "unsupported" mean unsupported by the spec or unsupported by the implementation. I think a small change can make that less ambiguous.

```suggestion
Implementations SHOULD provide assertion behavior for the format values defined
by this document[^2] and MUST refuse to process any schema which contains an format value it doesn't support.
```

specs/jsonschema-validation.md
```diff
@@ -293,75 +293,60 @@ Structural validation alone may be insufficient to allow an application to
...
+The requirement for validation of format values in general is limited to
+syntactic checking; implementations SHOULD NOT attempt to send an email, connect
+to a URL, or otherwise check the existence of an entity identified by a format
+instance.
 
 #### Custom format attributes
 
 Implementations MAY support custom format attributes. Save for agreement between
 parties, schema authors SHALL NOT expect a peer implementation to support such
 custom format attributes.
```
jdesrosiers, line 348: We removed the part where it defines the term "format attributes". We shouldn't use it here.

```suggestion
#### Custom format values

Implementations MAY support custom format values. Save for agreement between
parties, schema authors SHALL NOT expect a peer implementation to support such
custom format values.
```

jdesrosiers (approved):
I have a few editorial suggestions, but otherwise, this looks great.

adr/2024-11-2-assertion-format.md
gregsdennis: I think this was a case where I was thinking of going somewhere, got distracted by something else, and didn't come back.  I have no idea what this was for now.  Will update.

specs/jsonschema-validation.md
gregsdennis: If you can edit the wrapping for that, I'll just commit it from here.

specs/jsonschema-validation.md
gregsdennis: Good catch.

Relequestual:
I was initially concerned with some of the wording and the implication for required or optional tests, but it was changed, and I'm now happy. Sharing after the merge as @gregsdennis asked on the TSC vote issue for me to do so, although that was a week ago (sorry). (Partly for my own records

Specifically https://github.com/json-schema-org/json-schema-spec/pull/1553/files#diff-a2feaccf762dd55c4d3dd625aa1b57c48df6f782b8088cd4210fc9b93f8aea17R314-R316

> Implementations SHOULD provide assertion behavior for the format values defined
by this document[^2] and MUST refuse to process any schema which contains a
format value it doesn't support.

I was concerned the "SHOULD" was too weak here, but the proceeding "MUST" gives it enough balance. Primarily, this avoids any false positives where a user expects assertion behavior and an implementation goes "*shrugs* I'll just allow it".

jjrauch32-lab (commented):
K

jjrauch32-lab (commented):
OK

PR #1560, opened 2024-11-12
clarify that `type` must be a non-empty array

### LINKED ISSUE #1404: Spec and meta-schema seem inconsistent w.r.t empty "type" array
The spec allows `"type": []`, but in the validation vocabuary there is a 'minItems: 1' for the "type" property. This seems inconsistent.

For details see this issue I filed earlier with the test-suite (but that was the wrong place): https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/673

gregsdennis:

clarification

-  Closes #1404

### Summary

Clarifies that `type` as an array must be non-empty.

Technically yes?  @awwright did point out that there is a niche use case as an alternative to `{not:{}}`, but the keyword certainly wasn't intended to be used this way.  Also, given that we present `{not:{}}` as the analog to `false`, it's extremely unlikely that anyone will be using it this way on purpose.

### DIFF
--- a/specs/jsonschema-validation.md
+++ b/specs/jsonschema-validation.md
@@ -108,7 +108,8 @@ an instance. These keywords are all assertions without any annotation behavior.
 #### `type`
 
 The value of this keyword MUST be either a string or an array. If it is an
-array, elements of the array MUST be strings and MUST be unique.
+array, it MUST be non-empty, and elements of the array MUST be strings and MUST
+be unique.
 
 String values MUST be one of the six primitive types ("null", "boolean",
 "object", "array", "number", or "string"), or "integer" which matches any number


jviotti (approved):
This is great. All of these edge cases and what is technically possible can be very confusing.

jdesrosiers (approved):
I'm a little reluctant to approve a breaking change without documented consensus and an ADR, but I think it's ok in this case because the support for `type` with an empty array was never reliable because the meta-schema has included `"minItems": 1` since at least draft-04 (I didn't bother looking further back). That means that implementations that use the meta-schema to validate schemas (probably most of them) have been rejecting `type` with an empty array.

PR #1561, opened 2024-11-13
Update deprecated Github action

jdesrosiers:
We got a notice that the version of `actions/upload-artifact` that we are using is going to be deprecated. This updates that action and other outdated actions to the latest version.


gregsdennis (approved):

gregsdennis:
Happy to let this go through early.

jviotti (approved):

PR #599, opened 2018-05-27
Update regex concerns for subschema application

handrews:
Move the main text into core, and just reference it from validation.

This somehow got left off the PR back when the applicators moved to core, and I only just now noticed.  Core now includes keywords that use regexes.

PR #1563, opened 2024-11-14
use more accurate regex vocabulary

gibson042:

Editorial

-  Related to #1470
- Related to #599
- Related to 314b5c3040c5af13f838165189df3e0585b35a26

### Summary

Referring to the difference between e.g. `.+` and `.+?` as "laziness" isn't wrong, but it is less accurate than "greediness".

No.


jdesrosiers (approved):

gregsdennis (approved):
Thanks for this!

PR #1568, opened 2025-01-05
Gregsdennis/external links

### LINKED ISSUE #1536: Link directly to targets
With the spec docs now in markdown and with us self-hosting the files, should we move to having links go directly to their targets rather than to an appendix (which is IETF style)?  Personally, I never liked linking to an appendix.

OpenAPI uses direct links, which you can see with the JSON Schema spec link in [Section 4.4](https://spec.openapis.org/oas/latest.html#data-types).

gregsdennis:

editorial

- Closes #1536

### Summary

Inline all of the links to external documents and remove the associated appendix.

- [ ] Still a couple open comments on the issue to address
- [ ] need to perform markdown cleanup

No


jdesrosiers (commented):
Reference style links might be a good choice for this. Especially if we end up making the same reference in multiple places.

```md
## Blah
Blah, blah, [JSON Pointer][RFC6901], blah.

### Again
Some other reference to [JSON Poiner][RFC6901] elsewhere in the spec.

<!-- Reference links can go in one place at the bottom -->
[RFC6901]: https://www.rfc-editor.org/info/rfc6901
```

gregsdennis:
Alright I think that's it for the primary changes of this PR.  The only thing left is figuring out cross-document links.

@jdesrosiers I'm not sure whether you want to support that as part of this PR.  Alternatives are we merge this in the broken state or it gets supported in another PR and this is rebased.  I'm happy with either, really, but the latter seems more proper.

jdesrosiers (commented):
Everything looks good so far.

> The only thing left is figuring out cross-document links.

I don't think there's an easy way around this problem. I'm going to take another look, but if I don't find anything in a relatively short amount of time, I think we should merge this. It really bothers me to commit something that's breaking the build, but it might be all we can do right now.

jdesrosiers:
I just remembered a solution I had in mind that I want to write down so I don't forget again.

I think the best approach is to write links to the `.md` source, but convert the link to `.html` as part of the build to HTML. In that case, the right thing would be to make the links point to `.md` before merging this and then add the link conversion plugin later.

jdesrosiers (approved):

jdesrosiers:
Rewriting the links was easy and works out great. Not only does it make the link checker work, but it also allows links to work both in Github, not just when built to HTML.

Issue #1552, opened 2024-10-28
✨ Proposal: `format` registry


gregsdennis:
### Describe the inspiration for your proposal

Open API already has defined several formats that are arguably useful outside of the context of their spec, but I don't think we should add them to our spec.

Additionally, we receive a lot of requests to add new formats to the spec.  A registry could be a good middle ground.

### Describe the proposal

We should support a registry for different formats.

We don't have any precedent for creating a registry, so I'm open to ideas on exactly how to do that.  I think a file in GH that ends up being push to the website that could then be referenced by the spec would be an okay idea.

### Additional context

There is a concern about adding additional burden for implementations around having to support all of the formats.  I'm open to ideas here as well.

Really just opening the floor for exploration.

jviotti:
I like the idea, but my main concern is that if most of these formats are optional, very few will actually implement them, rendering mostly unusable for any interoperable use case.

I would almost prefer a more limited set of formats defined by us and mandatory for everybody. I feel just _forcing_ everybody to do the same thing, even if imperfect, can be better than _suggesting_ everybody to do a lot of things.

mwadams:
Personally, I think a combination of the two.

I think we should enforce some "near-primitives" that makes interoperability hard if we don't (the date and time related formats specifically. They are ubiquitous and notoriously hard to interop without a solid spec)

Then there are the "arguables" that I think should be required because we use them in the metaschema - the URI/IRI family.

Then there are the things I think should be in a registry. (u)int32,64,128, uuid etc. etc.

jdesrosiers:
Open API already has a [format registry](https://spec.openapis.org/registry/format/). Ideally we should be involved with that registry, but I don't want to have a competing registry. In the past, they've expressed willingness to donate ownership of the registry to us if we want it. Co-ownership of the registry would be a good outcome as well.

gregsdennis:
The discussions have been to transfer that registry here to be in a centralized location closer to the source.  I should have been more explicit.

jdesrosiers:
Great. Then I'm 100% on board.

karenetheridge:
100% in favour.

We could include data on what implementations (of JSON Schema or OpenAPI, etc) are known to implement the formats, as well as information about their intended use and their origin, when known/relevant.

[...]

PR #1574, opened 2025-01-15
Gregsdennis/format registry

gregsdennis:

Feature

Related to #1552

### Summary

- Adds a `format` registry file and includes OpenAPI 3.1.1 formats.  In the issue, I mocked up the registry with an array at the root, but realized an object would be good since we really shouldn't have multiple definitions for the same format.  (I have other thoughts on this that I'll post in the issue and update here if necessary.)
- Adds text defining the registry and requirements for implementations to support them.  Having a link to a later section feels weird to me.  Maybe it's not?  Happy to reorganize if others think it's off.

No.  There is a new "SHOULD" requirement to support the registry formats.  I'd like this to be _slightly_ less required than the ones defined by JSON Schema, but "MAY" seems not required enough.  Open to entertaining ideas.


ralfhandl:
@gregsdennis What about the formats registered in https://spec.openapis.org/registry/format/index.html?
Will they be eventually registered here, too?

And what will be the process/hurdles for registering formats?

Goal would be to retire the OAS format registry and instead redirect here.

gregsdennis:
I didn't know that page existed.  I was just looking at your spec. 

Yeah, I expect those can be pulled in as well. The plan _was_ to transfer your registry to us.

Given the sheer volume of formats, I'm more inclined to use "MAY" for the registry formats, but also reserve those names so that they can only mean what's registered.  That way _if_ those names are supported, they MUST be these formats.

karenetheridge:
We'll want to support an array for hyperlinks that reference the specification, as I think we'll want to specify multiple versions of reference documents. For example, it's useful to note the earliest version of the OAS spec that added mention of a format, but there are also cases where the reference specification for a format has changed over time (e.g. for json-schema's hostname format, draft7 uses RFC1034, draft2019-09+ uses RFC1123 and that information shouldn't be hidden by simply linking to draft7 and draft2019-09. Perhaps we could say something like "RFC1034 via draft7, RFC1123 via draft2019-09 and draft2020-12", but I'm not sure how that could be visibly represented in a compact yet readable way.

The formats listed here so far are present in OpenAPI v3.1.1, but we should also start out by listing all the formats available in various drafts of JSON Schema itself.

Also note that the current OAS format registry is at https://spec.openapis.org/registry/format/index.html and we should include the information contained there if we agree to take on hosting of that registry, although I've spotted a few errors in that data so those will need to be fixed.

karenetheridge:
> I didn't know that page existed. I was just looking at your spec.

I'm confused, we talked about proposing a transfer of the registry several months ago here?  https://github.com/json-schema-org/json-schema-spec/issues/1552#issuecomment-2449071062

gregsdennis:
I've imported all of the formats.  It _is_ a direct import processed by GH Copilot.  Checking on the errors...

gregsdennis:
> I'm confused... @karenetheridge

The answer is in your remark: "several months ago" 😆

ralfhandl (approved):

gregsdennis:
@json-schema-org/tsc I think thisi is one that we can pull in and even publish before we release the spec.

There's still a fair amount of planning, design, and work to do to get it on the website, but this is a critical step for that to take place.

specs/registries/format.json
```diff
@@ -0,0 +1,375 @@
...
+        "examples": ["token"],
+        "deprecated": false,
+        "supportedBy": []
+    },
+    "time": {
+        "description": "Time as defined by full-time - RFC3339",
+        "definingBody": "JSON Schema",
+        "definition": "https://json-schema.org/draft/2020-12/json-schema-validation.html#name-dates-times-and-duration",
+        "types": ["string"],
+        "examples": ["13:45:30"],
```
GREsau, line 326: https://www.rfc-editor.org/rfc/rfc3339.html#section-5.6 defines `full-time` as:
```
time-numoffset  = ("+" / "-") time-hour ":" time-minute
time-offset     = "Z" / time-numoffset

full-time       = partial-time time-offset
```

`time-offset` is non-optional and non-empty, so I don't think `"13:45:30"` is valid in this scenario - it would need to be something like `"13:45:30Z"` or `"13:45:30+01:00"`

specs/jsonschema-validation.md
```diff
@@ -328,10 +328,10 @@ an annotation.
 
 Implementations:
 
-- SHOULD provide validation for each format attribute defined in this
-  document;
-- MAY support format values not defined in this document, but such support MUST
-  be configurable and disabled by default;
+- SHOULD provide validation for each format attribute defined in this document
+  and listed in the {{format-registry}};
```
jdesrosiers, line 332: I can see the registry growing to have hundreds of entries. I don't think suggesting that implementations should implement all of them is realistic.

specs/jsonschema-validation.md
```diff
@@ -328,10 +328,10 @@ an annotation.
...
+- MAY support format values not defined in this document or listed in the
+  registry, but such support MUST be configurable and disabled by default;
 - SHOULD use a common parsing library or a well-known regular expression for
   each format;
 - SHOULD clearly document how and to what degree each format attribute is
```
jdesrosiers, line 337: I don't think this bullet point makes sense. We don't allow partial validation anymore, so full validation is the only valid way to validate a format.

specs/jsonschema-validation.md
gregsdennis: We need a middle ground between SHOULD and MAY for the registry.  Kind of a "hey it's a good idea if you want to, but you don't, no big deal."

specs/jsonschema-validation.md
gregsdennis: I agree, but realistically implementations are going to support formats only as far as they care to, even if we include a bunch of edge-case tests.  I think having a note that requires implementations to declare their support (or maybe their limitations) covers that base.

specs/jsonschema-validation.md
GREsau: "Implementations are encouraged to X" is a pattern I've noticed in a few RFCs, is that the type of thing you're looking for?

e.g.
```
- SHOULD provide validation for each format attribute defined in this document;
- are encouraged to provide validation for format attributes listed in the {{format-registry}};
```

jdesrosiers: I like "encouraged". How about something like, "implementations are encouraged implement formats from the registry and SHOULD (MUST?) provide an extension mechanism for formats they don't support".

The idea is that you either need to provide support for every format in the registry or provide a mechanism for the user to plug in support for anything that's missing.

specs/jsonschema-validation.md
jdesrosiers: I see what you're saying, but I feel like it gives permission for incomplete validation, which we don't want.

> (or maybe their limitations) 

Yeah, that. I think that's better. You shouldn't have to say you support what you're required to support. You should only have declare when your support is lacking in some way.

specs/jsonschema-validation.md
gregsdennis: Updated

specs/jsonschema-validation.md
gregsdennis: Updated

jdesrosiers (approved):

PR #1576, opened 2025-01-17
Fix `links` not adhering to its meta-schema given invalid `dependencies`

jviotti:
In Draft 4, `dependencies` is either a schema or an array of unique strings.

See: https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.5.4.5
See: https://www.learnjsonschema.com/draft4/validation/dependencies/

### DIFF
--- a/links.json
+++ b/links.json
@@ -36,6 +36,6 @@
 	"required" : ["rel", "href"],
 	
 	"dependencies" : {
-		"enctype" : "method"
+		"enctype" : ["method"]
 	}
 }


jeremyfiel:
i don't see this file in `main`.  seems like they restructured at some point when they removed the hyper schema stuff

jeremyfiel:
https://github.com/json-schema-org/json-schema-spec/pull/1326

jviotti:
@jeremyfiel I'm targeting Draft 4 though. `main` always has `next` as far as I understand

jeremyfiel:
I suppose that's true  

But these changes should probably go to the hyper schema repo linked

gregsdennis:
I don't understand what you're asking for here.  The definition of `dependencies` in the draft 4 meta-schema is

```
        "dependencies": {
            "type": "object",
            "additionalProperties": {
                "anyOf": [
                    { "$ref": "#" },
                    { "$ref": "#/definitions/stringArray" }
                ]
            }
        }
```

This is consistent with the language of the specification: an object where each property (individually) must be a schema or an array of strings.

jviotti:
@gregsdennis 

> an object where each property (individually) must be a schema or an array of strings.

Correct, but on the schema I'm linking, it is set to just a string, and not an array of strings?

jdesrosiers:
@jviotti Good catch! A string was a valid value for `dependencies` in draft-03 and apparently this constraint didn't get updated when it was moved to the `links.json` schema in draft-04.

jdesrosiers (approved):

PR #1577, opened 2025-01-19
keywords only exhibit the behaviors they're defined with

### LINKED ISSUE #1447: Applicator, Assertion, and Annotation as behaviors not categories
[Core 7](https://json-schema.org/draft/2020-12/json-schema-core.html#section-7) describes keywords as belonging to "behavior categories."  However, I don't think "categories" is quite right.  It seems to imply that a keyword only behaves in one of these ways, which isn't the case.  (If that's not what it's saying, I'm still reading it that way, and it could be clearer.)

These are just behaviors.  Case in point: `properties` actually assumes all of these behaviors:

- it's an applicator because it contains subschemas
- it produces an annotation of the keywords it constrains
- it produces a validation result by AND-ing all of the applicable subschemas' validation results

I think this section could make it clearer that a keyword can (and often does) embody more than one of these behaviors.

### LINKED ISSUE #1540: ✨ Proposal: Not all keywords need to produce a validation result.
### Describe the inspiration for your proposal

The current specifications declare a validation result for every non-core keyword.  On keywords for which a validation result doesn't really make sense, it says that validation must always pass or that they must always return a "true" validation result.

Examples are `if`, `minContains`, and `maxContains`.  For the latter two, it's especially apparent since we've moved them into the Core spec alongside `contains`.  Specifically, the WIP version states that these two keywords "modify the behavior of `contains`".  Then it talks about validation for them.

If their purpose is to help `contains` do its job, why should they produce a validation result at all?

### Describe the proposal

The spec should state whether a keyword should be ignored for validation purposes.

I think instead of each keyword needing to specify, we just have a single statement at the top that says something like

> Keywords which do not specify a validation result do not participate in validation.  Implementations MAY consider them to always return a validation result of "true".

This would need to go in Core.

### Describe alternatives you've considered

No alternatives; just the proposal.

### Additional context

This would also apply to all of the "pure annotation" keywords, like `title` and `deprecated`.

I like the idea of keywords having disjoint roles.  It can help move us toward supporting other things people use JSON Schema for, like code/schema/form/docs generation.

gregsdennis:

clarification

- Closes #1540
- Closes #1447

### Summary

The previous language could imply that all keywords needed to have an assertion result, e.g. annotations would always produce a "true" assertion result.

For `min/maxContains`, I added explicit text that they do not produce an assertion result, emphasizing that the assertion comes from `contains` and that these keywords are informative.

This change clarifies that keywords only exhibit the behaviors they're defined with.

No.

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -415,19 +415,36 @@ keywords MUST NOT begin with this prefix.
 Implementations MUST refuse to evaluate schemas which contain keywords which
 they do not know how to process or explicitly choose not to process.
 
-## Keyword Behaviors
+## Keyword Behaviors {#keyword-behaviors}
 
-JSON Schema keywords fall into several general behavior categories. Assertions
-validate that an instance satisfies constraints, producing a boolean result.
-Annotations attach information that applications may use in any way they see
-fit. Applicators apply subschemas to parts of the instance and combine their
-results.
+JSON Schema keywords may exhibit one or more behaviors. This specification
+defines three such behaviors:
 
-Extension keywords SHOULD stay within these categories, keeping in mind that
+- Assertions validate that an instance satisfies constraints, producing a
+  boolean result: `true` if the constraints are satisfied; `false` otherwise.
+- Annotations attach information that applications may use in any way they see
+  fit.
+- Applicators apply subschemas to parts of the instance and combine their
+  results.
+
+Extension keywords SHOULD be defined using these behaviors, keeping in mind that
 annotations in particular are extremely flexible. Complex behavior is usually
 better delegated to applications on the basis of annotation data than
 implemented directly as schema keywords. However, extension keywords MAY define
 other behaviors for specialized purposes.
+
+Keywords which are not defined to exhibit a particular behavior MUST NOT affect
+that aspect of evalution. For example, a keyword which does not act as an
+assertion MUST NOT affect the validation result.
+
+For the purposes of this document, an instance "validating against a keyword"
+means that the keyword produces an assertion result of `true` if the instance
+satisfies the given constraint; otherwise an assertion result of `false` is
+produced.
+
+<!-- The next two paragraphs and the following section seem to have to more to
+do with schema evaluation than keywords specifically. I'd like to move them out
+of the "keyword behaviors" h2, but will do so separately. [TODO: GD] -->
 
 Evaluating an instance against a schema involves processing all of the keywords
 in the schema against the appropriate locations within the instance. Typically,
@@ -569,11 +586,11 @@ the keyword's value. Alternatively, an applicator may refer to a schema
 elsewhere in the same schema document, or in a different one. The mechanism for
 identifying such referenced schemas is defined by the keyword.
 
-Applicator keywords also define how subschema or referenced schema boolean
-[assertion](#assertions) results are modified and/or combined to produce the
-boolean result of the applicator. Applicators may apply any boolean logic
-operation to the assertion results of subschemas, but MUST NOT introduce new
-assertion conditions of their own.
+Applicator keywords also behave as assertions by defining how subschema or
+referenced schema boolean [assertion](#assertions) results are modified and/or
+combined to produce the boolean result of the applicator. Applicators may apply
+any boolean logic operation to the assertion results of subschemas, but MUST NOT
+introduce new assertion conditions of their own.
 
 [Annotation](#annotations) results from subschemas are preserved in accordance
 with {{collect}} so that applications can decide how to interpret multiple
@@ -1002,10 +1019,15 @@ identified schema. Its results are the results of the referenced schema.[^5]
 other keywords can appear alongside of `$ref` in the same schema object.
 
 The value of the `$ref` keyword MUST be a string which is an IRI reference.
+The value of the `$ref` keyword MUST be a string which is an IRI reference.
 Resolved against the current IRI base, it produces the IRI of the schema to
 apply. This resolution is safe to perform on schema load, as the process of
 evaluating an instance cannot change how the reference resolves.
 
+The resolved IRI produced by `$ref` is not necessarily a network locator, only
+an identifier. A schema need not be downloadable from the address if it is a
+network-addressable URL. Implementations which can access the network SHOULD
+default to operating offline.
 The resolved IRI produced by `$ref` is not necessarily a network locator, only
 an identifier. A schema need not be downloadable from the address if it is a
 network-addressable URL. Implementations which can access the network SHOULD
@@ -1476,8 +1498,8 @@ operators can contact the owner of a potentially misbehaving script.
 
 ## Keywords for Applying Subschemas
 
-This section defines a set of keywords that enable schema combinations and
-composition.
+This section defines a set of applicator keywords that enable schema
+combinations and composition.
 
 ### Keywords for Applying Subschemas in Place {#in-place}
 
@@ -1737,8 +1759,8 @@ The value of this keyword MUST be a non-negative integer.
 This keyword modifies the behavior of `contains` within the same schema object,
 as described below in the section for that keyword.
 
-Validation MUST always succeed against this keyword. The value of this keyword
-is used as its annotation result.
+This keyword produces no assertion result. The value of this keyword is used as
+its annotation result.
 
 ##### `minContains`
 
@@ -1747,8 +1769,8 @@ The value of this keyword MUST be a non-negative integer.
 This keyword modifies the behavior of `contains` within the same schema object,
 as described below in the section for that keyword.
 
-Validation MUST always succeed against this keyword. The value of this keyword
-is used as its annotation result.
+This keyword produces no assertion result. The value of this keyword is used as
+its annotation result.
 
 Per {{default-behaviors}}, omitted keywords MUST NOT produce annotation results.
 However, as described in {{contains}}, the absence of this keyword's annotation
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -434,7 +434,7 @@ implemented directly as schema keywords. However, extension keywords MAY define
 other behaviors for specialized purposes.
 
 Keywords which are not defined to exhibit a particular behavior MUST NOT affect
-that aspect of evalution. For example, a keyword which does not act as an
+that aspect of evaluation. For example, a keyword which does not act as an
 assertion MUST NOT affect the validation result.
 
 For the purposes of this document, an instance "validating against a keyword"
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -442,7 +442,7 @@ means that the keyword produces an assertion result of `true` if the instance
 satisfies the given constraint; otherwise an assertion result of `false` is
 produced.
 
-<!-- The next two paragraphs and the following section seem to have to more to
+<!-- The next two paragraphs and the following section seem to have more to
 do with schema evaluation than keywords specifically. I'd like to move them out
 of the "keyword behaviors" h2, but will do so separately. [TODO: GD] -->
 
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -394,6 +394,18 @@ Additional schema keywords MAY be defined by any entity. Save for explicit
 agreement, schema authors SHALL NOT expect these additional keywords to be
 supported by implementations that do not explicitly document such support.
 
+Extension keywords MUST NOT interfere with the operation of keywords defined by
+this document or the companion JSON Schema Validation specificiation, and SHOULD
+NOT interfere with the operation of keywords defined by other extension
+documents.[^11]
+
+[^11]: JSON Schema currently does not have a namespacing mechanism, which would
+allow multiple extensions to define the same keyword differently while also
+giving the schema author the ability to declare which definition is intended.
+Such a feature is planned for future releases. See the
+[Vocabularies / Extensions project](https://github.com/orgs/json-schema-org/projects/28/views/2)
+in GitHub for more information.
+
 Implementations MAY provide the ability to register or load handlers for
 keywords that they do not support directly. The exact mechanism for registering
 and implementing such handlers is implementation-dependent.
@@ -422,16 +434,15 @@ defines three such behaviors[^7]:
 
 - Assertions validate that an instance satisfies constraints, producing a
   boolean result: `true` if the constraints are satisfied; `false` otherwise.
-- Annotations attach information to instance locations that applications may use in any way they see
-  fit.
+- Annotations attach information to instance locations that applications may use
+  in any way they see fit.
 - Applicators apply subschemas to parts of the instance and combine their
   results.
 
 [^7]: This specification also defines several operational directive keywords,
-such as `$id` and `$schema`. As such, these keywords do not exhibit these
-behaviors. However, it is recommended that extensions avoid defining additional
-directive keywords as they could interfere with schema processing and produce
-unexpected or undesirable results.
+such as `$id` and `$schema`, which do not exhibit these behaviors. Instead,
+these keywords provide metadata that instruct implementations on how to
+interpret and process the schema.
 
 Extension keywords SHOULD be defined using these behaviors, keeping in mind that
 annotations in particular are extremely flexible. Complex behavior is usually
@@ -439,9 +450,7 @@ better delegated to applications on the basis of annotation data than
 implemented directly as schema keywords. However, extension keywords MAY define
 other behaviors for specialized purposes.
 
-Keywords which are not defined to exhibit a particular behavior MUST NOT affect
-that aspect of evaluation. For example, a keyword which does not act as an
-assertion MUST NOT affect the validation result.
+Implementations SHOULD NOT add unspecified behaviors to keywords.
 
 For the purposes of this document, an instance "validating against a keyword"
 means that the keyword produces an assertion result of `true` if the instance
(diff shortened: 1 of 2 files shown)


karenetheridge:
So, what is the result of:

what is the result of:

```
allOf: [
  { maxContains: 1 },
  { maxContains: 1 },
]
```

and

```
anyOf: [
  { maxContains: 1 },
  { maxContains: 1 },
]
```

and

```
oneOf: [
  { maxContains: 1 },
  { maxContains: 1 },
]
```

?

gregsdennis:
Good point.  The _keywords_ produce no assertions, but the _subschemas_ still need to.

Granted, this is true with 2020-12, too.  The validation spec doesn't actually define assertion results for any of the annotations, yet it's still considered a pass because there are no constraints.

I think this could be stated explicitly.

specs/jsonschema-validation.md
```diff
@@ -62,6 +62,21 @@ information. The {{format}} keyword is intended primarily as an annotation, but
 can optionally be used as an assertion. The {{content}} keywords are annotations
 for working with documents embedded as JSON strings.
 
+## Keyword Behaviors
+
+The keywords defined by this document exhibit one or more behaviors as defined by
+the [JSON Schema Core Specification](./jsonschema-core.md#keyword-behaviors).
+
+Keywords which are not defined to exhibit a particular behavior MUST NOT affect
+that aspect of the evalution outcome. In particular, the keywords defined in
```
jdesrosiers, line 71:
```suggestion
that aspect of the evaluation outcome. In particular, the keywords defined in
```

specs/jsonschema-core.md
```diff
@@ -564,11 +581,11 @@ the keyword's value. Alternatively, an applicator may refer to a schema
...
-Applicator keywords also define how subschema or referenced schema boolean
-[assertion](#assertions) results are modified and/or combined to produce the
-boolean result of the applicator. Applicators may apply any boolean logic
-operation to the assertion results of subschemas, but MUST NOT introduce new
-assertion conditions of their own.
+Applicator keywords also behave as assertions by defining how subschema or
+referenced schema boolean [assertion](#assertions) results are modified and/or
+combined to produce the boolean result of the applicator. Applicators may apply
+any boolean logic operation to the assertion results of subschemas, but MUST NOT
+introduce new assertion conditions of their own.
```
jdesrosiers, line 588: > but MUST NOT introduce new assertion conditions of their own.

I'm not following this. What is this trying to say? What are we protecting against be forbidding it?

jdesrosiers (changes requested):
I've always thought of all keywords having an assertion. Annotation-only keywords just always assert `true`. `$defs` is another example of a keyword always returns true although it's not an annotation.

I think the way this is worded is great because it allows for implementations to ignore non-assertions or just make them `true`. They can implement it however makes most sense for their implementation.

specs/jsonschema-core.md
gregsdennis: That text was already there.

It's just saying that applicator assertions can only be combinatorial on their subschemas.  For instance `anyOf` only ORs the results of its subschemas without asserting anything else.

I think this is a simplicity guard, kind of a JSON Schema version of SRP.  I suppose this means something like `requiredProperties` would be in violation of the spec since it's combining subschema results _and_ asserting that all of the properties are there.

specs/jsonschema-core.md
```diff
@@ -410,20 +410,37 @@ keywords MUST NOT begin with this prefix.
...
-## Keyword Behaviors
+## Keyword Behaviors {#keyword-behaviors}
 
-JSON Schema keywords fall into several general behavior categories. Assertions
-validate that an instance satisfies constraints, producing a boolean result.
-Annotations attach information that applications may use in any way they see
-fit. Applicators apply subschemas to parts of the instance and combine their
-results.
+JSON Schema keywords may exhibit one or more behaviors. This specification
+defines three such behaviors:
```
notEthan, line 416: This seems to say that all keyword behaviors in the specification are in these three categories, but the spec defines other behaviors for e.g. `$id` and `$anchor`/`$dynamicAnchor`. And `$schema` ... exists.

gregsdennis: That is an interesting point.  I was thinking that these keywords don't exhibit a _behavior_ so much as they are merely informative, like `maxContains` would be to `contains`.

specs/jsonschema-core.md
jdesrosiers: I don't think it's unnecessary for that to be a constraint. Your `requiredProperties` example is great. Although I'm not a fan of that keyword, I don't think there should be any spec reason why it shouldn't be allowed.

> or maybe even an informative note suggesting simplicity over complexity?

Yes. I think that's a great idea.

specs/jsonschema-core.md
jdesrosiers: This is a great point! My implementation doesn't even treat those as keywords. They're meta-data used to process the schema, but they're not directly part of the validation/annotation process. So, I wouldn't consider them similar to `maxContains` that is involved in validation.

I'm not sure what I'd suggest, but I think something needs to be called that these aren't normal keywords.

specs/jsonschema-core.md
gregsdennis: I've added a footnote and changes this to SHOULD NOT.

specs/jsonschema-core.md
```diff
@@ -415,20 +415,43 @@ keywords MUST NOT begin with this prefix.
...
-Annotations attach information that applications may use in any way they see
-fit. Applicators apply subschemas to parts of the instance and combine their
-results.
+JSON Schema keywords may exhibit one or more behaviors. This specification
+defines three such behaviors[^7]:
+
+- Assertions validate that an instance satisfies constraints, producing a
+  boolean result: `true` if the constraints are satisfied; `false` otherwise.
+- Annotations attach information that applications may use in any way they see
+  fit.
```
jdesrosiers, line 426:
```suggestion
- Annotations attach information to instance locations that applications may use in any way they see
  fit.
```

specs/jsonschema-core.md
```diff
@@ -415,20 +415,43 @@ keywords MUST NOT begin with this prefix.
...
+- Applicators apply subschemas to parts of the instance and combine their
+  results.
+
+[^7]: This specification also defines several operational directive keywords,
+such as `$id` and `$schema`. As such, these keywords do not exhibit these
+behaviors. However, it is recommended that extensions avoid defining additional
+directive keywords as they could interfere with schema processing and produce
+unexpected or undesirable results.
```
jdesrosiers, line 434: I don't understand what you mean by "interfering with schema processing".

The way I see it, people should avoid creating these kinds of keywords because it's unlikely that an implementation will be able to support them using typical extension mechanisms. At least my implementation wouldn't.

specs/jsonschema-core.md
```diff
@@ -415,20 +415,43 @@ keywords MUST NOT begin with this prefix.
...
-Extension keywords SHOULD stay within these categories, keeping in mind that
+Extension keywords SHOULD be defined using these behaviors, keeping in mind that
 annotations in particular are extremely flexible. Complex behavior is usually
 better delegated to applications on the basis of annotation data than
 implemented directly as schema keywords. However, extension keywords MAY define
 other behaviors for specialized purposes.
 
+Keywords which are not defined to exhibit a particular behavior MUST NOT affect
+that aspect of evaluation. For example, a keyword which does not act as an
+assertion MUST NOT affect the validation result.
```
jdesrosiers, line 444: I'm not sure what this means. It sounds like it's saying that if you say it doesn't have an assertion behavior, then it shouldn't have an assertion behavior. Is there some nuance that I'm missing?

gregsdennis: Nope, that's pretty much it.  I thought it would be good to be clear that a keyword that doesn't define a behavior shouldn't be made to exhibit that behavior.  If we don't state that, someone's gonna do it.  Maybe it can be reworded.

jdesrosiers: Maybe it needs to be reworded, but it's still not making sense to me. A keyword is defined in a specification. Whatever the specification says it does is what it does. I don't see how the spec can define a keyword to do something, but it does another. Unless this is a note implementers of that specification telling them they shouldn't add behaviors that aren't in the specification. But, again, that doesn't seem like it's something the spec needs to say. It seems self evident that if you implement a spec, it should do what the spec says. That's just the concept of implementing a spec.

Is it that you don't want people to write in their spec that their keyword is not an assertion but then define assertion behavior? That just means the spec isn't self consistent. That also doesn't seem like it needs to be said. We shouldn't have to say that a keyword spec shouldn't contradict itself.

jdesrosiers: Ok, so it was this one,

> Unless this is a note implementers of that specification telling them they shouldn't add behaviors that aren't in the specification.

I don't really think that needs to be said, but I'm ok with it as long as it can be reworded to make it more clear that that's what it means.

specs/jsonschema-core.md
gregsdennis: I've removed this sentence and expanded on the idea of a 'directive' keyword to explain why they're different.

specs/jsonschema-core.md
```diff
@@ -569,11 +601,17 @@ the keyword's value. Alternatively, an applicator may refer to a schema
...
 identifying such referenced schemas is defined by the keyword.
 
-Applicator keywords also define how subschema or referenced schema boolean
-[assertion](#assertions) results are modified and/or combined to produce the
-boolean result of the applicator. Applicators may apply any boolean logic
-operation to the assertion results of subschemas, but MUST NOT introduce new
-assertion conditions of their own.
+Applicator keywords also behave as assertions by defining how subschema or
+referenced schema boolean [assertion](#assertions) results are modified and/or
+combined to produce the boolean result of the applicator. Applicators may apply
```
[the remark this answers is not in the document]
notEthan: 'referenced' seems okay to me; $ref has a referenced schema that isn't a subschema - although maybe that is better mentioned above in the initial description of applicator keyword behavior.

The 'boolean' refers to an assertion result, not type of subschema, though the words/phrases run together a bit and it's difficult to parse how they are grouped (even knowing all the terms, much less while learning them).

```suggestion
Applicator keywords also behave as assertions, using the assertion results of
each subschema or referenced schema of the keyword. These boolean results are
modified (e.g. the `not` keyword negates its subschema's assertion) and/or
combined (e.g. `allOf` takes the conjunction of its subschemas' assertions)
to produce the boolean result of the applicator.
```

Not quite polished - not suggesting that be put in as is, just thinking through what I might find to read better. (The spec is always very light on examples, I assume intentionally so maybe those don't go in, but I do always find they make the more generalized spec language much easier to connect to what it is describing).

specs/jsonschema-core.md
```diff
@@ -415,20 +427,40 @@ keywords MUST NOT begin with this prefix.
...
+interpret and process the schema.
 
-Extension keywords SHOULD stay within these categories, keeping in mind that
+Extension keywords SHOULD be defined using these behaviors, keeping in mind that
 annotations in particular are extremely flexible. Complex behavior is usually
 better delegated to applications on the basis of annotation data than
 implemented directly as schema keywords. However, extension keywords MAY define
 other behaviors for specialized purposes.
 
+Implementations SHOULD NOT add unspecified behaviors to keywords.
```
notEthan, line 453: I'm not sure what this is aiming to prevent.

Behaviors outside the three above? I'm not sure what that would be. That also seems more the concern of the definition of an extension keyword rather than an implementation.

Or behaviors within the three, but that aren't in a specification? If so, that seems just the nature of a specification, that an implementation of it sticks to its specified behaviors.

specs/jsonschema-validation.md
```diff
@@ -62,6 +62,21 @@ information. The {{format}} keyword is intended primarily as an annotation, but
...
+
+Keywords which are not defined to exhibit a particular behavior MUST NOT affect
+that aspect of the evaluation outcome. In particular, the keywords defined in
+{{annotations}} produce no assertion result and therefore are not considered
+during validation.
+
+For the purposes of this document, an instance "validating against a keyword"
+means that the keyword produces an assertion result of `true` if the instance
+satisfies the given constraint; otherwise an assertion result of `false` is
+produced.
```
notEthan, line 78: This references Core Spec Keyword Behaviors but then also repeats, or is redundant with, a significant amount of that section.

specs/jsonschema-core.md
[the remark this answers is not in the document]
jdesrosiers: > 'referenced' seems okay to me; $ref has a referenced schema that isn't a subschema

`$ref` is just a keyword in a subschema. It's sufficient to call `{ "$ref": "..." }` a subschema. It's not something different because it happens to contain a reference.

(I wouldn't say the same thing before 2019-09. Back then I would call it something different than a schema.)

specs/jsonschema-core.md
jdesrosiers: See, [this](https://github.com/json-schema-org/json-schema-spec/pull/1577#discussion_r2040716835) thread.

gregsdennis: See [this discussion](https://github.com/json-schema-org/json-schema-spec/pull/1577#discussion_r2040716835).

specs/jsonschema-validation.md
gregsdennis: The validation spec uses the phrase "an instance validates against this keyword if...", or some variant of it, quite a lot.  I just wanted to define what that means.  The phrasing is different than "this keyword exhibits assertion behavior by...", which feels more clunky to me (and I'd have to change a lot of places).

specs/jsonschema-core.md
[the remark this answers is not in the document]
notEthan: sure, a schema containing a $ref isn't different than any other, but taking $ref as an applicator keyword, the schema it is doing application of is its referenced schema, not a subschema (I think I recall that your implementation does make it a subschema, but in the original document it isn't)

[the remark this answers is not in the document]
jdesrosiers: Oh, I see what you mean. `$ref` is an applicator but its value isn't a subschema, it's a reference. Good point!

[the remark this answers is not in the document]
gregsdennis: I do agree that "boolean" is redundant and "modified" is misleading at best.  You _could_ say that `not` "modifies" its subschema's result, but does it really?  I'd say it has its own result which is the inverse of its subschema.

I'll massage this a bit.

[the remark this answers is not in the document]
gregsdennis: Actually, @notEthan, I like what you have.  The examples do make both cases clear, even the "boolean" inclusion makes more sense.

specs/jsonschema-core.md
```diff
@@ -569,11 +601,19 @@ the keyword's value. Alternatively, an applicator may refer to a schema
...
-boolean result of the applicator. Applicators may apply any boolean logic
-operation to the assertion results of subschemas, but MUST NOT introduce new
-assertion conditions of their own.
+Applicator keywords also behave as assertions, using the assertion results of
+each subschema or referenced schema of the keyword. These boolean results are
+modified (e.g. the `not` keyword negates its subschema's assertion) and/or
+combined (e.g. `allOf` takes the conjunction of its subschemas' assertions)
+to produce the boolean result of the applicator.
+any boolean logic operation to the assertion results of subschemas, but SHOULD
+NOT introduce new assertion conditions of their own.[^2]
```
jdesrosiers, line 610: Looks like something broke in the last edit. There's a floating sentence fragment at the end.

(Otherwise I like the new wording.)

gregsdennis: Ah, yeah, that was pulling in a suggestion from @notEthan.  The start of the next sentence was omitted.

jdesrosiers (approved):

specs/jsonschema-core.md
```diff
@@ -569,11 +601,19 @@ the keyword's value. Alternatively, an applicator may refer to a schema
...
+combined (e.g. `allOf` takes the conjunction of its subschemas' assertions) to
+produce the boolean result of the applicator. Applicators may apply any boolean
+logic operation to the assertion results of subschemas, but SHOULD NOT introduce
+new assertion conditions of their own.[^2]
+
+[^2]: It is recommended that keywords have a single resposibility. An example of
+this in action is the separation between `properties`, which verifies object
+property values have the right data _if_ they exist, and `required`, which
+verifies that object properties exist. Separating these concerns allows for more
+reusable components.
```
Relequestual, line 616: More a broader comment that comes to mind with this change...

I know we have at times talked about the posibility of adding "shortcut" keywords, such as combining `properties` and `required`. I forget if this is something we explicitly decided we would or would not do. (I would be happy stating we will not do such, but I wanted to see what those with a better memory remember also!)

Relequestual (approved):
Looks great. Thanks for doing this work!

I've left one suggestion and one general comment.
The suggestion should by no means be a blocker, and I'm not even convinced I have a better phrasing, but it at least starts the discussion.

specs/jsonschema-core.md
```diff
@@ -394,6 +394,18 @@ Additional schema keywords MAY be defined by any entity. Save for explicit
 agreement, schema authors SHALL NOT expect these additional keywords to be
 supported by implementations that do not explicitly document such support.
 
+Extension keywords MUST NOT interfere with the operation of keywords defined by
+this document or the companion JSON Schema Validation specificiation, and SHOULD
+NOT interfere with the operation of keywords defined by other extension
+documents.[^11]
```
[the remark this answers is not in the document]
gregsdennis: I would say that a new applicator introducing a new subschema in which a `properties` keyword could appear is neither interfering with nor directly modifying `unevaluatedProperties`.  `unevaluatedProperties` is defined to consider subschemas of all applicators, not just specific ones.

Do you have another example?

"Directly modify" does also convey the purpose of this text, and it's language we've used before.

PR #1579, opened 2025-01-19
Split Overview into the two specific use cases (resubmission)

gregsdennis:

Replaces #1370

### Summary

> [!NOTE]
> This is a resubmission of @awwright's #1370.  Please refer there for extensive conversation.  The below text is also from that PR.

This is an alternative to https://github.com/json-schema-org/json-schema-spec/pull/1244, and like that issue, this one should be incorporated prior to https://github.com/json-schema-org/json-schema-spec/issues/1365. This PR makes a clearer distinction between validation uses and annotation uses that 1365 will draw on (it relies on classifying which use each keyword is being used for, since validation use is different from annotation use).

This PR moves vocabulary-related discussion deeper into the specification, and replaces Overview with a specific description of what a JSON Schema can do, especially in terms of its output.

Once its capabilities are described, the rest of the specification can describe how you write a schema that can do these things.

This replaces the Overview section, which is a little bit redundant with the above Abstract, then Introduction, sections.

No.

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -45,46 +45,45 @@ The terms "JSON", "JSON text", "JSON value", "member", "element", "object",
 "array", "number", "string", "boolean", "true", "false", and "null" in this
 document are to be interpreted as defined in [RFC 8259](#rfc8259).
 
-## Overview
+## Validation
 
-This document proposes a new media type `application/schema+json` to identify a
-JSON Schema for describing JSON data. It also proposes a further optional media
-type, `application/schema-instance+json`, to provide additional integration
-features. JSON Schemas are themselves JSON documents. This, and related
-specifications, define keywords allowing authors to describe JSON data in
-several ways.
+A JSON Schema document describes a validator (also known as a "recognizer" or
+"acceptor") which classifies a provided JSON document as "accepted" or
+"rejected."
 
-JSON Schema uses keywords to assert constraints on JSON instances or annotate
-those instances with additional information. Additional keywords are used to
-apply assertions and annotations to more complex JSON data structures, or based
-on some sort of condition.
+It supports "structural validation" (context-free grammars), and certain more
+complicated conditions. Validation follows JSON semantics, so two documents that
+are value-equal, but vary only by character escapes, property ordering, or
+whitespace, will validate with the same result.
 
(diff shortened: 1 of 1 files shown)


specs/jsonschema-core.md
```diff
@@ -45,46 +45,45 @@ The terms "JSON", "JSON text", "JSON value", "member", "element", "object",
...
-
-Further keywords for purposes such as structural validation or hypermedia
-annotation are defined in other documents. These other documents each define a
-dialect collecting the standard sets of keywords needed to write schemas for
-that document's purpose.
+## Validation
+
+A JSON Schema document describes a validator (also known as a "recognizer" or
+"acceptor") which classifies a provided JSON document as "accepted" or
+"rejected."
```
karenetheridge, line 52: This is introducing a lot of new terms, which could cause confusion -- generally we have been saying "evaluator" and "valid" or "invalid". I would prefer we stick to one set of terms rather than introducing new overlapping ones.

karenetheridge: Further, the JSON Schema document itself is not the evaluator/validator, but rather _describes rules for evaluation_, when applied against a provided JSON document ("data instance").

awwright: The goal of this phrasing is to relate the concepts behind JSON Schema to the general concepts that pre-exist JSON Schema. If you're from the world of computer languages, grammars, and parsers, then this will introduce you to the terms used in JSON Schema in terms that you're already familiar with. If you're from the world of JSON then this will introduce you to the terminology that you will inevitably end up seeing elsewhere.

karenetheridge: I've never seen "recognizer" or "acceptor" used, either in the world of JSON Schema or outside it.  "Validator" and "parser" have much higher recognition IMO.

awwright: Consider the page https://en.wikipedia.org/wiki/Finite-state_machine (of which the vast majority of keywords in JSON Schema fall within a https://en.wikipedia.org/wiki/Pushdown_automaton specifically) — the word "valid" doesn't even show up. This passage helps build a bridge to this audience.

gregsdennis:
I think @jdesrosiers' https://github.com/json-schema-org/json-schema-spec/pull/1627, which edits the intro considerably, overrides this.  Closing, but happy to continue discussion if desired.

PR #1588, opened 2025-02-25
Gregsdennis/improve scopes

### LINKED ISSUE #1566: 🧹 Clarification: Scopes section could be improved
### Specification section

Core section on lexical and dynamic scopes

### What is unclear?

This section isn't _wrong_, but it is quite confusing.  We need to rewrite this.

### Proposal

Not currently.

This is follow-on work from https://github.com/json-schema-org/json-schema-spec/pull/1555

### Do you think this work might require an [Architectural Decision Record (ADR)]? (significant or noteworthy)

No

gregsdennis:

clarification

- Closes #1566

### Summary

Improve readability, definitions, and explanations of lexical and dynamic scopes.

No.

There is a `MUST` requirement that was previously listed as undefined behavior before, but this `MUST` was stated in the scopes section anyway.  It's just now `MUST` in both places.

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -458,8 +458,9 @@ This specification defines two such scopes: lexical and dynamic.
 #### Lexical Scope
 
 The lexical scope of a keyword is determined by the nested JSON data structure
-of objects and arrays. The largest such scope is an entire schema document.  The
-smallest scope is a single schema object with no subschemas.
+of objects and arrays. The smallest such scope is a single schema object with no
+subschemas. The largest scope is an entire schema document, recursively
+including all of its subschemas.
 
 Keywords MAY be defined with a partial value which must be resolved against
 another value found within the lexical structure of the JSON document. The
@@ -476,12 +477,13 @@ root object.
 
 #### Dynamic Scope
 
-Other keywords may take into account the dynamic scope that exists during the
-evaluation of a schema, typically together with an instance document. The
-outermost dynamic scope is the schema object at which processing begins, even if
-it is not a schema resource root. The path from this root schema to any
-particular keyword (that includes any `$ref` and `$dynamicRef` keywords that may
-have been resolved) is considered the keyword's "evaluation path."
+The dynamic scope is the ordered collection of schema objects navigated during
+evaluation, starting at the root and ending at the subschema under evaluation.
+The outermost dynamic scope is the schema object at which processing begins,
+even if it is not a schema resource root. The path that evaluation takes,
+starting from this root schema to any particular subschema (including any `$ref`
+and `$dynamicRef` keywords that may have been resolved), is considered the
+"evaluation path".
 
 Lexical and dynamic scopes align until a reference keyword is encountered. While
 following the reference keyword moves processing from one lexical scope into a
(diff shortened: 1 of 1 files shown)


gregsdennis:
I still have some work to do here for `$dynamic*`, but I'm going to wait until #1589 is done so I can rebase on that.

gregsdennis:
Just did a close/open to rekick the build.  It had been a while and couldn't do it manually.

gregsdennis:
Reviewed the changes in #1589, and they seem to align with the rest of what I've updated here, so I'm going to leave `$dynamic*` as it is.

jdesrosiers (approved):
This is an improvement, but I still really dislike the way dynamic scope is defined. I think it should be the stack of schema _resources_ and not include subschemas. The extra complexity makes learning how dynamic references work more complicated.

Also, I desperately hope we wouldn't ever add another keyword that uses dynamic scope. My preference would be to remove this whole concept from the spec and address it only in the context of dynamic references and recommend that extension keywords not use dynamic scope.

All of that can be debated elsewhere and done as a follow up if there's consensus.

karenetheridge:
> This is an improvement, but I still really dislike the way dynamic scope is defined. I think it should be the stack of schema resources and not include subschemas.

I agree, and this is already how we define dynamic scope (unless you're intending to change that?) -- we say elsewhere in the spec, when processing `$dynamicRef` \[emphasis mine\], "...the initial URI MUST be replaced by the URI (including the fragment) for the outermost **schema resource** in the dynamic scope that defines an identically named fragment with "$dynamicAnchor".

Therefore, in my implementation, the current URI is pushed to the `dynamic_scope` stack (the array that is searched when processing `$dynamicRef`) only when processing the `id` keyword, not for every subschema.

I can't see any usecases for wanting other, non-resource, subschemas to be stored in the dynamic scope.

gregsdennis:
I've updated to use resources.  (This is actually what we agreed on when @jviotti was writing his posts on scopes.)

It does raise one question, though:  what if evaluation starts at a subschema?  I suppose the dynamic scope is still the resource, but the evaluation path would still start from the subschema?  This is how I have it now, and I could change it if we decide something else.

> I desperately hope we wouldn't ever add another keyword that uses dynamic scope. - @jdesrosiers

I don't know.  I think if explained properly (which is what this change is for), dynamic scopes could be utilized by others that need it.  I think it's a powerful concept and it gives extensions another trick.

jdesrosiers (approved):

karenetheridge:
> It does raise one question, though: what if evaluation starts at a subschema? I suppose the dynamic scope is still the resource, but the evaluation path would still start from the subschema? This is how I have it now, and I could change it if we decide something else.

This would be good to have as a test case in the test suite. Also, writing up an example of how this would work depending on which way it would be implemented would help us see which way is more "right".

gregsdennis:
> writing up an example of how this would work

What I'm really thinking of here is more related to what you might have in an OpenAPI description.  A resource root might be the entire document, so a schema wouldn't be at the resource root.

```yaml
openapi: 3.1
# ...
components:
  schemas:
    foo:
      $dynamicAnchor: root
      properties:
        bar: { $ref: "#/components/schemas/bar" }
    bar:
      $id: bar
      $dynamicAnchor: root
      properties:
        recurse: { $dymamicRef: "#root" }
```

Here, while evaluation starts at `/components/schemas/foo`, the evaluation path after a loop would look like `/properties/bar/$ref/properties/recurse/$dynamicRef`, and you'd be back at `/components/schemas/foo`.

The dynamic scopes, then, are:

- the document
- the `bar` schema
- the document again, ...

So even though the top-level scope is the full document, the evaluation path starts in the middle of it.  I think this is fine and probably expected.

PR #1589, opened 2025-02-25
Additional cleanup for dynamic references

### LINKED ISSUE #1140: Proposal: Remove initial resolution step for dynamic references

jdesrosiers:

Clarification

- Closes #1140

### Summary

This is mostly cleanup building on #1555, although I rewrote a couple of parts I though needed it.

The main objective is to achieve consistent and correct usage of terms around fragment identifiers. Here are the definitions I'm using.

- **fragment identifier** -- This is an identifier for a secondary resource (subschema) within the primary resource (schema resource).
- **IRI fragment** -- The fragment identifier that appears after the `#` character in an IRI.
- **plain name** -- A specific type of fragment identifier. `$anchor` creates a plain name fragment identifier.
- **fragment-only IRI** -- An IRI that starts with `#` and is followed by a fragment identifier.
- ~**fragment**~ -- This term by itself is ambiguous and imprecise. I replaced all usages with a more precise term from the above terms.

Here's a list of the other notable changes.

- Defining location-independent identifiers
  - I removed the language around constructing an IRI from the fragment identifier. All this section needs to do is define how plain name fragment identifiers are created. How to use a fragment identifier in any IRI is well established in RFC-3986/7 and doesn't need to be repeated here.
  - I tried to clarify that `$dynamicAnchor` does not create a fragment identifier. It's a special kind of identifier unique to the dynamic reference mechanism. Specifically, you can't use the identifier created by `$dynamicAnchor` in an IRI.
- Dynamic References with `$dynamicRef`
  - I provided a new introduction for dynamic references that I think make sense. I've always thought that description was awful and I think this an improvement.
- Example of recursive schema extension
  - I fixed a sentence that didn't make sense.

Everything else is minor grammar fixes or whitespace cleanup.

No. (Actually, the issue this closes is a breaking change, but this PR doesn't break anything. This is just cleanup related to and around that change.)

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -240,7 +240,7 @@ A schema that itself describes a schema is called a meta-schema. Meta-schemas
 are used to validate JSON Schemas and specify the set of keywords those schemas
 are using.
 
-#### Root Schema and Subschemas and Resources {#root}
+#### Root Schema, Subschemas, and Resources {#root}
 
 A JSON Schema resource is a schema which is
 [canonically](https://www.rfc-editor.org/info/rfc6596) identified by an
@@ -334,9 +334,8 @@ NCNameChar      = NCNameStartChar / "-" / "." / DIGIT
 All fragment identifiers that do not match the JSON Pointer syntax MUST be
 interpreted as plain name fragment identifiers.
 
-Defining and referencing a plain name fragment identifier within an
-`application/schema+json` document are specified in the [`$anchor`
-keyword](#anchors) section.
+Defining a plain name fragment identifier within an `application/schema+json`
+document is specified in the [`$anchor` keyword](#anchors) section.
 
 ## General Considerations
 
@@ -950,40 +949,35 @@ an [absolute IRI](https://www.rfc-editor.org/rfc/rfc3987.html#section-2.2)
 
 #### Defining location-independent identifiers {#anchors}
 
-Using JSON Pointer fragments requires knowledge of the structure of the schema.
-When writing schema documents with the intention to provide re-usable schemas,
-it may be preferable to use a plain name fragment that is not tied to any
-particular structural location. This allows a subschema to be relocated without
-requiring JSON Pointer references to be updated.
+Using JSON Pointers in IRI fragments to reference subschemas couples the IRI to
+the structure of the schema. Using plain name fragment identifiers in IRI
+fragments to identify subschemas is sometimes preferable because it is not tied
+to a particular structural location. This allows a subschema to be relocated
+without requiring references to be updated.
 
-The `$anchor` and `$dynamicAnchor` keywords are used to specify such fragments.
-They are identifier keywords that can only be used to create plain name
-fragments, rather than absolute IRIs as seen with `$id`.
+The `$anchor` and `$dynamicAnchor` keywords are used to define
+location-independent identifiers for subschemas within a schema resource.
 
-`$anchor` defines a reference target for `$ref`. The fragment defined by this
-keyword is appended to the IRI of the schema resource containing it. As
-discussed in {{id-keyword}}, this is either the nearest `$id` in the same or an
-ancestor schema object, or the base IRI for the document as determined according
-to [RFC 3987][rfc3987] and
-[RFC 3986][rfc3986].
-
-In contrast, `$dynamicAnchor` operates independently of resource IRIs and is
-instead dependent on the dynamic scope of the evaluation. `$dynamicAnchor`
-defines a reference target for the `$dynamicRef` keyword. This advanced feature
-makes it easier to extend recursive schemas such as the meta-schemas, without
-imposing any particular semantics on that extension. See {{dynamic-ref}} for
-details.
-
-In most cases, the normal fragment behavior both suffices and is more intuitive.
-Therefore it is RECOMMENDED that `$anchor` be used to create plain name
-fragments unless there is a clear need for `$dynamicAnchor`.
-
-If present, the value of these keywords MUST be a string and MUST conform to the
-plain name fragment identifier syntax defined in {{fragments}}.[^4]
+`$anchor` defines a plain name fragment identifier that can be used in IRI
+fragments as an alternative to JSON Pointers.[^4] See {{fragments}}.
 
 [^4]: Note that the anchor string does not include the "#" character, as it is
-not a IRI reference. An `$anchor`: "foo" becomes the fragment `#foo` when used
-in a IRI. See below for full examples.
+just a fragment identifier not an IRI reference. To reference the "foo"
+`$anchor` from the same schema resource, you would use the fragment-only IRI
+`#foo`. See below for full examples.
+
+`$dynamicAnchor` defines a different kind of fragment identifier that only has
+meaning when used with `$dynamicRef`. It's not a normal fragment identifier and
+therefore can't be used anywhere other than `$dynamicRef`. Normal [fragment
+identifiers](https://www.rfc-editor.org/rfc/rfc3986#section-3.5) identify the
+secondary resource (the subschema) while the rest of the IRI identifies the
+primary resource (the schema resource). The fragment identifiers defined by
+`$dynamicAnchor` are not normal fragment identifies because they identify both
+the primary resource and the secondary resource. See {{dynamic-ref}} for
+details.
+
+If present, the value of these keywords MUST be a string and MUST conform to the
+plain name fragment identifier syntax defined in {{fragments}}.
 
 #### Duplicate schema identifiers {#duplicate-iris}
 
@@ -1007,7 +1001,7 @@ identified schema. Its results are the results of the referenced schema.[^5]
 [^5]: Note that this definition of how the results are determined means that
 other keywords can appear alongside of `$ref` in the same schema object.
 
-The value of the `$ref` keyword MUST be a string which is a IRI reference.
+The value of the `$ref` keyword MUST be a string which is an IRI reference.
 Resolved against the current IRI base, it produces the IRI of the schema to
 apply. This resolution is safe to perform on schema load, as the process of
 evaluating an instance cannot change how the reference resolves.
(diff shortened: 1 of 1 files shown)


specs/jsonschema-core.md
```diff
@@ -950,40 +949,32 @@ an [absolute IRI](https://www.rfc-editor.org/rfc/rfc3987.html#section-2.2)
...
+location-independent identifiers for subschemas within a schema resource.
+
+`$anchor` defines a plain name fragment identifier that can be used in IRI
+fragments as an alternative to JSON Pointers.[^4] See {{fragments}}.
 
 [^4]: Note that the anchor string does not include the "#" character, as it is
-not a IRI reference. An `$anchor`: "foo" becomes the fragment `#foo` when used
-in a IRI. See below for full examples.
+just a fragment identifier not an IRI reference. To reference the "foo"
+`$anchor` from the same Schema Resource, you would use the fragment-only IRI
```
gregsdennis, line 966:
```suggestion
`$anchor` from the same schema resource, you would use the fragment-only IRI
```

We're not capitalizing this anywhere else.

specs/jsonschema-core.md
```diff
@@ -950,40 +949,32 @@ an [absolute IRI](https://www.rfc-editor.org/rfc/rfc3987.html#section-2.2)
...
+`#foo`. See below for full examples.
+
+`$dynamicAnchor` defines a different kind of identifier that only has meaning
+when used as a value for `$dynamicRef`. It's not a fragment identifier that can
+be used in an IRI. `$dynamicRef` uses a syntax similar to a fragment-only IRI,
+but the semantics are different. Unlike IRIs, the fragment identifies the
+primary resource (the schema resource) in addition to the secondary resource
+(the subschema within that schema resource). See {{dynamic-ref}} for details.
```
gregsdennis, line 974: If you're looking to distinguish a `$dynamicAnchor` fragment identifier from an `$anchor` fragment identifier, I don't think this is clear enough.  An IRI that has a fragment identifier also "identifies the primary resource (the schema resource) in addition to the secondary resource (the subschema within that schema resource)".

If that's not your intent, then I'm not sure what purpose this sentence serves.

specs/jsonschema-core.md
```diff
@@ -1002,12 +993,13 @@ applicators, applying the referenced schema to the instance.
 ##### Direct References with `$ref` {#ref}
 
 The `$ref` keyword is an applicator that is used to reference a statically
-identified schema. Its results are the results of the referenced schema.[^5]
+identified schema. Its results are the results of evaluating the referenced
+schema.[^5]
```
gregsdennis, line 997: This is less clear to me.  If you're going to require specificity that we're evaluating the referenced schema (which I think is sufficiently implied), then you need to require that same specificity of the `$ref` keyword results.

Personally I prefer this unchanged.

specs/jsonschema-core.md
```diff
@@ -1002,12 +993,13 @@ applicators, applying the referenced schema to the instance.
...
 
 [^5]: Note that this definition of how the results are determined means that
 other keywords can appear alongside of `$ref` in the same schema object.
 
-The value of the `$ref` keyword MUST be a string which is a IRI reference.
+The value of the `$ref` keyword MUST be a string which is an IRI reference.
```
gregsdennis, line 1002: I think I caught this too.  One of us is going to have to rebase. :)

specs/jsonschema-core.md
```diff
@@ -1019,28 +1011,25 @@ default to operating offline.
...
-resolution until runtime, at which point it is resolved each time it is
-encountered while evaluating an instance.
-
-Together with `$dynamicAnchor`, `$dynamicRef` implements a cooperative extension
-mechanism that is primarily useful to to create open schemas, where
-`$dynamicRef` defines the extension point and `$dynamicAnchor` defines the
-target.
+The `$dynamicRef` keyword is an applicator that is used when the referencing
+schema might need to override where a reference in the referenced schema will
+resolve. This is useful for authoring a recursive schema that can be extended or
```
gregsdennis, line 1016: I like having the word "primarily" in there.  It communicates that we recognize that these are not the only use cases.

specs/jsonschema-core.md
```diff
@@ -1205,19 +1194,19 @@ If an implementation has been configured to resolve that identifier to a schema
...
 Conceptually, a set of linked schema resources should behave identically whether
 each resource is a separate document connected with [schema
 references](#referenced), or is structured as a single document with one or more
 schema resources embedded as subschemas.
 
-Since IRIs involving JSON Pointer fragments relative to the parent schema
+Since IRIs with JSON Pointer fragments relative to the parent schema
 resource's IRI cease to be valid when the embedded schema is moved to a separate
 document and referenced, applications and schemas SHOULD NOT use such IRIs to
 identify embedded schema resources or locations within them.
```
gregsdennis, line 1212:
```suggestion
Since IRIs with JSON Pointer fragments are relative to the parent schema
resource's IRI, they cease to be valid when the embedded schema is moved to a separate
document and referenced. Because of this, applications and schemas SHOULD NOT use such IRIs to
identify embedded schema resources or locations within them.
```

It took me some time to parse this as a single sentence.  It makes sense, but I think the above is a bit cleaner.

(I haven't bothered to adjust for line length.)

gregsdennis (changes requested):

specs/jsonschema-core.md
jdesrosiers: See the [Compound documents](https://github.com/json-schema-org/json-schema-spec/blob/main/specs/jsonschema-core.md?plain=1#L1299) section. But, yes, it looks like that's the only place it's being capitalized. We should probably update that section for consistency as well.

specs/jsonschema-core.md
jdesrosiers: The sentence isn't referring to an IRI that has a fragment, it's referring to just the fragment or the IRI or IRI-like value.

In `"$ref": "example#foo"`, "example" determines the primary resource and "foo" determines the secondary resource. In `"$dynamicRef": "#foo"`, "foo" determines both the primary resource and the secondary resource (which is why the part before the `#` is always empty).

I think sentence is clear that it's referring to the fragment of the IRI, not the whole IRI, but let me know if you think there's a better way to make that more clear.

specs/jsonschema-core.md
jdesrosiers: That was how I originally wrote it, then changed it in the end. I do feel like just saying "the results" is less precise than it should be, but the revised sentence feels awkward. I'll revert the change.

specs/jsonschema-core.md
jdesrosiers: I don't like "primarily" because what it's primarily used for can and has changed over time. We originally thought it was primarily useful for extending recursive schemas, but then when we figured out it can be used to express generic schemas and that became a far more common and well known use case than extending recursive schemas. That forced a need to update the spec because extending recursive schemas was no longer the primary use case. Removing "primary", makes the spec more resilient to evolving usage. If another pattern becomes popular and eclipses the existing patterns, we don't necessarily have to update the spec to acknowledge that new reality.

I think "such as" communicates that we recognize that these aren't the only use cases while also not implying that these use cases are more common or more important than others.

specs/jsonschema-core.md
jdesrosiers: Agreed! I found this sentence difficult as well.

specs/jsonschema-core.md
gregsdennis: I think it's the "Unlike IRIs" part that throws it off.  Why make the comparison, especially when IRIs (ones with fragments, anyway) also do the thing you're saying these fragments do?

specs/jsonschema-core.md
gregsdennis: But now you're saying, "This is useful for use-case-X (recursive) and use-case-Y (generics)".  This is the part that is opened by the "primarily".  If we can put a "such as" in here, I'd be for that.

_This is useful for cases such as..._

The "such as" as you have it comes later and seems to expand solely on the generics case.  It doesn't suggest that there are use cases beyond the X and Y mentioned in the first part.

jdesrosiers: Ah! you're right. I'll fix it.

specs/jsonschema-core.md
jdesrosiers: I tried making a change. Let me know if you think it help. Honestly, I'm having trouble seeing what you're seeing in the sentence, so I'm not sure how to make it better.

> Why make the comparison, especially when IRIs (ones with fragments, anyway) also do the thing you're saying these fragments do?

This sentence is talking about the fragments in IRIs, not the whole of IRI capabilities. Is that the confusion?

specs/jsonschema-core.md
```diff
@@ -950,40 +949,33 @@ an [absolute IRI](https://www.rfc-editor.org/rfc/rfc3987.html#section-2.2)
...
+`$anchor` from the same schema resource, you would use the fragment-only IRI
+`#foo`. See below for full examples.
+
+`$dynamicAnchor` defines a different kind of identifier that only has meaning
+when used as a value for `$dynamicRef`. It's not a fragment identifier that can
+be used in an IRI. `$dynamicRef` uses a syntax similar to a fragment-only IRI,
+but the semantics are different. Unlike fragments in IRIs, the fragment in
+`$dynamicRef` identifiers identifies the primary resource (the schema resource)
+in addition to the secondary resource (the subschema within that schema
+resource). See {{dynamic-ref}} for details.
```
gregsdennis, line 975: This is better.

Maybe highlight that the difference is that the primary resource is being identified with `$dynamicRef`.

Maybe

> Fragments in IRIs can only identify a subschema within the resource identified by the IRI. That is, the fragment itself does not identify the primary schema resource. However, the fragment in `$dynamicRef` identifies both the primary schema resource and a subschema within it. See {{dynamic-ref}} for details.

(Also the "`$dynamicRef` identifiers identifies" is a bit redundant.)

jdesrosiers: I gave this another try. I ended up rewriting most of the paragraph taking a slightly different approach.

> primary schema resource

I specifically wanted to use the terms "primary resource" and "secondary resource" because those are the terms used to define fragment identifiers in RFC-3986. I added a link for people who might not be familiar with those terms.

gregsdennis: Yeah, your new paragraph is better.  Taking the time to fully explain things and not trying to be terse seems to be the trick here.

gregsdennis (approved):

PR #1592, opened 2025-03-18
Update dependencies

jdesrosiers:
This brings our dependencies up to date so we don't get security warnings any more.

One of the causes of those security warnings was the torchlight remark plugin we use for code highlighting/numbering. That project seems to be dead, so I had to replace it with something else.


jdesrosiers:
Merging early because this is just maintenance and isn't changing the spec in any way.

PR #1596, opened 2025-04-12
Gregsdennis/spec versioning

### LINKED ISSUE #1421: Spec Update: Specify how the stable URI works
The stable dialect URI identifies a schema as using the latest version of the standard dialect. What schema this URI identifies depends on what the implementation supports. Generally, the implementation will associate the stable URI with the meta-schema for the latest release they support. However, implementations may use a custom meta-schema that better represents what it supports. For example, if the implementation supports all of 2024 and some of 2025, they might want to use a custom meta-schema that includes only the 2025 keywords they support.

gregsdennis:

clarification

- Closes #1421 

### Summary

Update meta-schema IRIs and provide some explanation of the path meaning.  Probably could use a bit more explanation, but would like to get others' opinions.

No? Maybe?  I'm not sure how this applies since it's a necessary change.

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -873,16 +873,17 @@ default to using the current location if a default is desireable.
 
 ## The JSON Schema Core Keywords {#core}
 
-Keywords declared in this section, which all begin with "$", are essential to
-processing JSON Schema. These keywords inform implementations how to process any
-schema or meta-schema, including those split across multiple documents, or exist
-to reserve keywords for purposes that require guaranteed interoperability.
+Keywords declared in this section, which all begin with a dollar sign (`$`), are
+essential to processing JSON Schema. These keywords inform implementations how
+to process any schema or meta-schema, including those split across multiple
+documents, or exist to reserve keywords for purposes that require guaranteed
+interoperability.
 
 Support for these keywords MUST be considered mandatory at all times as they are
 necessary to navigate and process any schema.
 
-The "$" prefix is reserved for use by this specification. Extensions MUST NOT
-define new keywords that begin with "$".
+The `$` prefix is reserved for use by this specification. Extensions MUST NOT
+define new keywords that begin with `$`.
 
 ### Meta-Schemas
 
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -2250,7 +2250,7 @@ and only allows the "data" and "children" properties. An example instance with
 
 ```jsonschema "Tree schema, extensible"
 {
-  "$schema": "https://json-schema.org/draft/next/schema",
+  "$schema": "https://json-schema.org/1/2025",
   "$id": "https://example.com/tree",
   "$dynamicAnchor": "node",
 
@@ -2269,7 +2269,7 @@ and only allows the "data" and "children" properties. An example instance with
 
 ```jsonschema "Strict-tree schema, guards against misspelled properties"
 {
-  "$schema": "https://json-schema.org/draft/next/schema",
+  "$schema": "https://json-schema.org/1/2025",
   "$id": "https://example.com/strict-tree",
   "$dynamicAnchor": "node",
 
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -427,6 +427,12 @@ keywords MUST NOT begin with this prefix.
 Implementations MUST refuse to evaluate schemas which contain keywords which
 they do not know how to process or explicitly choose not to process.
 
+## Specification Versioning and Compatibility
+
+This specification is versioned by two values: iteration and release year.
+
+A schema written to conform with the requirements of a given version (iteration and release year) is compatible with specifications published with the same iteration value and either the same or greater release year value.  Thus, JSON Schema provides a guarantee of compatibility for future releases within an iteration.
+
 ## Keyword Behaviors {#keyword-behaviors}
 
 JSON Schema keywords may exhibit one or more behaviors. This specification
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -429,9 +429,13 @@ they do not know how to process or explicitly choose not to process.
 
 ## Specification Versioning and Compatibility
 
-This specification is versioned by two values: iteration and release year.
+This specification is identified collectively by two values: version and release
+year.
 
-A schema written to conform with the requirements of a given version (iteration and release year) is compatible with specifications published with the same iteration value and either the same or greater release year value.  Thus, JSON Schema provides a guarantee of compatibility for future releases within an iteration.
+A schema written to conform with the requirements of a given version is
+compatible with successive specifications, which are published with the same
+version and either the same or greater release year value. Thus, JSON Schema
+provides a guarantee of compatibility for future releases within a version.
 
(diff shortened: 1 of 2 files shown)


specs/jsonschema-core.md
```diff
@@ -415,6 +415,12 @@ keywords MUST NOT begin with this prefix.
 Implementations MUST refuse to evaluate schemas which contain keywords which
 they do not know how to process or explicitly choose not to process.
 
+## Specification Versioning and Compatibility
+
+This specification is versioned by two values: iteration and release year.
+
+A schema written to conform with the requirements of a given version (iteration and release year) is compatible with specifications published with the same iteration value and either the same or greater release year value.  Thus, JSON Schema provides a guarantee of compatibility for future releases within an iteration.
```
jdesrosiers, line 422: Is the linter not working? It should have flagged this line as too long.

specs/jsonschema-validation.md
```diff
@@ -88,16 +88,27 @@ specification.
 ## Meta-Schema {#meta-schema}
 
 The current IRI for the default JSON Schema dialect meta-schema is
-`https://json-schema.org/draft/next/schema`. For schema author convenience, this
-meta-schema describes a dialect consisting of all keywords defined in this
+`https://json-schema.org/1/2025`. This IRI encodes the specifications version
```
jdesrosiers, line 91:
```suggestion
`https://json-schema.org/1/2025`. This IRI encodes the specification's version
```

specs/jsonschema-validation.md
```diff
@@ -88,16 +88,27 @@ specification.
...
+and release year. Because all schemas written to conform to a given version are
```
jdesrosiers, line 92: I feel like there's some ambiguity here. You've defined the "version" as the iteration + release year. But, here you're calling the iteration "specification version".

```suggestion
`https://json-schema.org/1/2025`. This IRI encodes the specifications iteration
and release year. Because all schemas written to conform to a given version are
```

specs/jsonschema-core.md
```diff
@@ -415,6 +415,12 @@ keywords MUST NOT begin with this prefix.
 Implementations MUST refuse to evaluate schemas which contain keywords which
 they do not know how to process or explicitly choose not to process.
 
+## Specification Versioning and Compatibility
+
+This specification is versioned by two values: iteration and release year.
```
jdesrosiers, line 420: The term "iteration" doesn't feel right to me. It's not very intuitive and it's never defined what it means.

specs/jsonschema-validation.md
```diff
@@ -88,16 +88,27 @@ specification.
...
+guaranteed to be compatible with later releases within the same iteration, the
+meta-schema IRI `https://json-schema.org/1` is also recognized to represent the
```
karenetheridge, line 94: - what is "indicated iteration"?
- using a URI that can change meaning depending on the current state of the spec feels really dodgy to me -- just like "https://json-schema.org/latest" would be, because the mapping of this value can change at any time, unexpectedly, which can be disruptive to both schema authors and implementation authors.

specs/jsonschema-core.md
```diff
@@ -2188,7 +2195,7 @@ and only allows the "data" and "children" properties. An example instance with
 
 ```jsonschema "Tree schema, extensible"
 {
-  "$schema": "https://json-schema.org/draft/next/schema",
+  "$schema": "https://json-schema.org/1/2025",
```
karenetheridge, line 2198: Putting the iteration before the year is not going to sort very well, and I imagine it could easily be confused with a date.

karenetheridge (commented):
I'm not sold on this idea at all. What other ideas were considered? Is there a discussion that I missed?

gregsdennis:
> I'm not sold on this idea at all. What other ideas were considered? Is there a discussion that I missed? - @karenetheridge 

There is the issue linked at the top of this PR as well as [this discussion thread](https://github.com/orgs/json-schema-org/discussions/671#discussioncomment-8753266) (the rest of the discussion is worth reading as well).  We also have our [PROCESS.md file](https://github.com/json-schema-org/json-schema-spec/blob/main/PROCESS.md).

specs/jsonschema-core.md
gregsdennis: I had "iteration" at first because I wanted to avoid "version" since it seems it be such an overused term.  But rereading it a few days ago, it didn't feel right.  I thought I got all of them.

specs/jsonschema-core.md
gregsdennis: Hm.. maybe.  I'll update it.

specs/jsonschema-validation.md
gregsdennis: > using a URI that can change meaning depending on the current state of the spec feels really dodgy to me -- just like "https://json-schema.org/latest" would be, because the mapping of this value can change at any time, unexpectedly, which can be disruptive to both schema authors and implementation authors.

That's why we stated the compatibility guarantee.  Any `https://json-schema.org/1/x` schema is compatible with all `https://json-schema.org/1/x+i` specifications.  As long as the version (`1` in this case) is the same, that compatibility guarantee holds.  Because of this, it's safe to have `https://json-schema.org/1` as a `$schema` value.  Future releases within the same version will not break schemas.

specs/jsonschema-core.md
gregsdennis: The compatibility guarantee stated [here](https://github.com/json-schema-org/json-schema-spec/pull/1596#discussion_r2046201670) necessitates that the version be first.

specs/jsonschema-core.md
```diff
@@ -415,6 +415,16 @@ keywords MUST NOT begin with this prefix.
 Implementations MUST refuse to evaluate schemas which contain keywords which
 they do not know how to process or explicitly choose not to process.
 
+## Specification Versioning and Compatibility
+
+This specification is identified collectively by two values: version and release
+year.
+
+A schema written to conform with the requirements of a given version is
```
jdesrosiers, line 423:
```suggestion
A schema written to conform with the requirements of a given release is
```

jdesrosiers (approved):

specs/jsonschema-core.md
gregsdennis: I think this could go either way.  Is there a compelling reason you think it should be "release"?

jdesrosiers: "Release" makes perfect sense to me, but I'm having trouble convincing myself one way or another that "version" makes sense.

I don't think it makes sense to write a schema to conform with the requirements of a given version. If there are five releases for that version, which one are you writing your schema for? Saying I'm writing a v1 schema doesn't tell me if I can use a keyword introduced in 2026 (assuming there's also a 2025). The schema author is always has to be writing toward a release.

From the perspective of a validator, things are a little different. The validator would validate against a version and the release doesn't matter so much. But, this is a statement about the schema, not the validator.

karenetheridge (dismissed):
I find this representation of the version to be really confusing. It's not at all clear to me what the next version after "1/2025" would be. Is the version intended to be "1" and "2025" a subversion? or is the full version "1/2025"? What combinations of numbers are intended to be grouped into one family and which are to be compatible with each other?

Whatever we go with should be intuitive to observers without having to read the spec, and I don't think we have that with the current proposal.

PR #1599, opened 2025-04-26
docs and resources and external docs

### LINKED ISSUE #1513: Clarify how schemas embedded inside other document types are identified
[Core 4.3.5](https://json-schema.org/draft/2020-12/json-schema-core#section-4.3.5) has the following:

> The root schema is the schema that comprises the entire JSON document in question. The root schema is always a schema resource, where the URI is determined as described in section [9.1.1](https://json-schema.org/draft/2020-12/json-schema-core#initial-base).

along with the following as an editor's note:

> Note that documents that embed schemas in another format will not have a root schema resource in this sense. Exactly how such usages fit with the JSON Schema document and resource concepts will be clarified in a future draft.

I assume this is to cover cases like OpenAPI, which embed JSON Schema directly into their own document, which means that the entire document cannot be the root schema.

We need to clarify this.

### LINKED ISSUE #1538: Unsupported deprecated keywords are listed in the metaschema
As [discussed on Slack](https://json-schema.slack.com/archives/CT8QRGTK5/p1727902395528899), there are some deprecated elements listed in the metaschema that are not actually supported. For example the [draft 2020-12 metaschema](https://json-schema.org/draft/2020-12/schema) includes `$recursiveAnchor` and `$recursiveRef`.

gregsdennis:
…ernal docs should specify their usage of json schema

- Closes #1513 

### Summary

Adds a clarification around docs and resources, and recommends that docs specified by third parties which use json schema should specify how they use it.

no

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -256,7 +256,7 @@ A root schema, which may comprise the entire JSON document, is the top-level
 schema that is identified by the absolute IRI. The root schema is always a
 schema resource, where the IRI is determined as described in {{initial-base}}.
 
-Documents formats which embed JSON Schemas within them will not necessarily have
+Document formats which embed JSON Schemas within them will not necessarily have
 a single root schema in this sense. How root schemas are identified within such
 documents SHOULD be defined by the specifications which govern them.
 
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -252,13 +252,13 @@ is identical to the primary resource. This can occur with the empty fragment, or
 when one schema resource is embedded in another. Any such IRIs with fragments
 are considered to be non-canonical.
 
-The root schema is the schema that comprises the entire JSON document in
-question. The root schema is always a schema resource, where the IRI is
-determined as described in {{initial-base}}.[^1]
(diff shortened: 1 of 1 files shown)


specs/jsonschema-core.md
```diff
@@ -252,13 +252,13 @@ is identical to the primary resource. This can occur with the empty fragment, or
...
-question. The root schema is always a schema resource, where the IRI is
-determined as described in {{initial-base}}.[^1]
+A root schema, which may comprise the entire JSON document, is the top-level
+schema that is identified by the absolute IRI. The root schema is always a
+schema resource, where the IRI is determined as described in {{initial-base}}.
 
-[^1]: Note that documents that embed schemas in another format will not have a
-root schema resource in this sense. Exactly how such usages fit with the JSON
-Schema document and resource concepts will be clarified in a future draft.
+Documents formats which embed JSON Schemas within them will not necessarily have
```
karenetheridge, line 259:
```suggestion
Document formats which embed JSON Schemas within them will not necessarily have
```

jdesrosiers (approved):

Relequestual (approved):

PR #1610, opened 2025-06-22
re-unite meta-schema

### LINKED ISSUE #1538: Unsupported deprecated keywords are listed in the metaschema

### LINKED ISSUE #1575: Reunite the meta-schemas
For 2019-09, the meta-schema was split along the vocab boundaries, and each vocab got its own, and the top-level meta-schema simply referenced these individual ones.

With vocabs being put into the feature life cycle process as a proposal, it doesn't make sense to release the meta-schemas split this way.

Should we reunite them?  I don't think that we need to, but the split doesn't really make sense outside of the vocabularies concept.

Our process says that a proposed keyword should be added to the meta-schema with a `true` value to define it for implementations.  (Implementation need to "know about" it.)  This would apply to `$vocabulary` as well.  This means that we can leave the `$vocabulary` keyword in, and implementations that don't support the feature should just ignore it (because they have to "know about" proposal keywords, even if they don't support them).

To be more explicit, the core vocab schema would be updated to

```jsonc
{
  // ...
  "properties": {
    // ...
    "$vocabulary": true,
    // ...
  },
  // ...
}
```

and the (top-level) meta-schema would remain unchanged, still containing a `$vocabulary` keyword.

gregsdennis:

Refactor

- Closes #1575 
- Closes #1538 

### Summary

Re-unites the meta-schema into a single file since vocabularies are a proposal now.  We will likely split it back up again when we figure out vocabs, but for now it doesn't make sense to have them separated.

Also:
- removes any deprecated keywords
- reserves the names of existing proposals
- reserves `x-*` keywords
- reserves `$*` keywords


specs/meta/meta.json
```diff
@@ -0,0 +1,191 @@
...
+        "$vocabulary": {
+            "$comment": "Proposed keyword: https://github.com/json-schema-org/json-schema-spec/blob/main/specs/proposals/vocabularies.md"
+        },
+        "propertyDependencies": {
+            "$comment": "Proposed keyword: https://github.com/json-schema-org/json-schema-spec/blob/main/specs/proposals/propertyDependencies.md"
+        }
+    },
+    "patternProperties": {
+        "^x-": true,
+        "^(?!(?:\\$id|\\$schema|\\$ref|\\$anchor|\\$dynamicRef|\\$dynamicAnchor|\\$comment|\\$defs)$)\\$.*": false
```
Relequestual, line 151: Isn't this superfluous?

Relequestual (approved):
Looks good. Thanks for doing this work.
Left one comment, but not essential.

specs/meta/meta.json
```diff
@@ -0,0 +1,191 @@
+{
+    "$schema": "https://json-schema.org/version/year/meta",
+    "$id": "https://json-schema.org/version/year/meta",
```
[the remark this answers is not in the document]
gregsdennis: > What's "meta"?

it was late.  It should just be `/version/year`

> Also, what's the reason for using a placeholder for "version" and "year"?

It's not published yet.  Just wanted to communicate the idea.

specs/meta/meta.json
notEthan: I think the meta-schema should adhere to jsonschema-core's "given the high disparity in regular expression constructs support, schema authors SHOULD limit themselves to the following regular expression tokens: ...".
Negative lookahead is definitely a feature with much disparity in support.
But I don't know another way to express this constraint for `patternProperties`.
Maybe `propertyNames` would be better suited to validate this?

gregsdennis: good catch.  Maybe this is one of those requirements where we just don't enforce it with the meta-scheam.  I'm okay with that.

jdesrosiers: > Maybe `propertyNames` would be better suited to validate this?

I think this would work,

```json
    "propertyNames": {
      "pattern": "^[^\\$]|^\\$(id|schema|ref|anchor|dynamicRef|dynamicAnchor|comment|defs)$"
    },
```

It doesn't interfere with `unevaluatedProperties` and the error messaging is reasonable.

gregsdennis:
@jdesrosiers I've updated the $-keyword restriction using your suggestion.  It's a bit direct, but simple is probably better anyway.

PR #1616, opened 2025-07-30
Update python dependencies with security vulnerabilities

jdesrosiers:
This brings the python dependencies up to date patching a couple security vulnerabilities.


gregsdennis (approved):
Just out of curiosity, what kinds of security vulnerability could be leveraged against this?  We're just running lint and build.

jdesrosiers:
> Just out of curiosity, what kinds of security vulnerability could be leveraged against this?

Honestly, I didn't even think about it. It's good hygiene to keep dependencies up to date, so I took the dependabot alerts I've been seeing as my queue that it was time to update.

PR #1618, opened 2025-08-07
New Security Considerations

### LINKED ISSUE #1231: Security considerations should mention treating URIs as URLs (from $ref and $schema)

jdesrosiers:
Resolves #1231 
Replaces #1600 

This is a full rewrite of the Security Considerations sections of the Core spec. It retains most of the original content plus a lot more. The only thing I left out from the original is the part about `$comment`, which I don't think makes sense. If there's an argument for keeping it, I can add it back in.

This new Security Consideration section is inspired by the guidelines and examples in [RFC 3552 - Guidelines for Writing RFC Text on Security Considerations](https://datatracker.ietf.org/doc/html/rfc3552). Some principles I'm trying to follow are,

- This section should not make SHOULD/MUST requirements. Instead, it should state the implications of the requirements made elsewhere in the document.
- This section may discuss threats that are mitigated by spec requirements, but also makes recommendations for mitigating threats that aren't specifically addressed by the spec.
- The audience of the spec is implementers. Schema authors should only be referred to in terms of how implementations should handle their schemas.

Something worth mentioning is that this PR advises not to use `file:` URIs in `$id`. However, we use file URIs in a couple tests in the official test suite. We should consider changing, removing, or moving those tests to optional.

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -2076,8 +2076,8 @@ the paper or impose limits on dynamic reference resolution.
 
 Infinite loops can occur when evaluating schemas that produce cycles during
 reference resolution. These cycles may involve multiple schemas. Not all
-recursive schemas create loops, but implementations are advised to detect and
-break these cycles when they are encountered.
+recursive schemas create loops, but implementations are advised to detect these
+cycles and terminate evaluation when they are encountered.
 
 ### Schema Identity and Collisions
 
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -2029,32 +2029,129 @@ SHOULD use the terms defined by this document to do so.
 
 ## Security Considerations {#security}
 
-Both schemas and instances are JSON values. As such, all security considerations
-defined in [RFC 8259][rfc8259] apply.
+While schemas and instances are not always represented as JSON text, they are
+defined in terms of the JSON data model. As such, the security considerations
+defined in [RFC 8259][rfc8259] may still apply in environments where text-based
+representations are used, particularly those considerations related to parsing,
+number precision, and structural limitations.
 
-Instances and schemas are both frequently written by untrusted third parties, to
-be deployed on public Internet servers. Implementations should take care that
-the parsing and evaluating against schemas does not consume excessive system
-resources. Implementations MUST NOT fall into an infinite loop.
+Schemas and instances are frequently authored by untrusted parties.
+Implementations that accept or evaluate such inputs may be exposed to several
+classes of attack, particularly denial-of-service (DoS) by means of resource
+exhaustion.
 
-A malicious party could cause an implementation to repeatedly collect a copy of
-a very large value as an annotation. Implementations SHOULD guard against
-excessive consumption of system resources in such a scenario.
+### Nested `anyOf`/`oneOf`
 
-Servers MUST ensure that malicious parties cannot change the functionality of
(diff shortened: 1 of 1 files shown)


specs/jsonschema-core.md
```diff
@@ -2029,32 +2029,129 @@ SHOULD use the terms defined by this document to do so.
...
+doing so can cause the size of the schema to grow exponentially. Implementations
+should be aware of this risk and may wish to implement the method described in
+the paper or impose limits on dynamic reference resolution.
+
+### Infinite Loops and Cycles
+
+Infinite loops can occur when evaluating schemas that produce cycles during
+reference resolution. These cycles may involve multiple schemas. Not all
+recursive schemas create loops, but implementations are advised to detect and
+break these cycles when they are encountered.
```
gregsdennis, line 2080: Earlier in the spec, implementations are _required_ to halt when encountering infinite cycles.

How would one break such a cycle?

jdesrosiers: I didn't intend to imply that one could somehow continue evaluating the schema. I'll reword to make it more clear that it should stop evaluation when it detects a cycle.

specs/jsonschema-core.md
```diff
@@ -2029,32 +2029,129 @@ SHOULD use the terms defined by this document to do so.
...
+Schemas may declare an `$id` to identify themselves or have embedded schemas
+that declare an `$id`. An attacker may attempt to register a schema with an
+`$id` that collides with a previously registered schema, or that differs only by
+case, encoding, or other URI normalization quirks. Such collisions could result
+in overwriting or shadowing of trusted schemas.
+
+Implementations should consider rejecting schemas that have identifiers
+(including embedded schema identifiers) that conflict with registered schemas
+and should apply consistent URI normalization and comparison logic to detect and
+prevent conflicts.
```
karenetheridge, line 2093: When would we not want to allow that? I think we should upgrade this to a MUST.

karenetheridge (commented):
one comment, otherwise 👍🏼

Relequestual (approved):
Looks pretty good. Thanks for your work on this. Sorry for the delay in reviewing.

specs/jsonschema-core.md
gregsdennis: I think that comes down to an implementation option to allow it.  The default behavior should be to reject an overwrite attempt.

jdesrosiers:
It seems like it might be incorrect for us to suggest that implementers to normalize identifiers (see https://github.com/json-schema-org/json-schema-spec/pull/1618#discussion_r2301950193)

For now, I made a slight change to the wording about URI normalization and I suggest we merge now and consider the question of normalization separately. I'll create a follow up issue.

specs/jsonschema-core.md
```diff
@@ -2029,32 +2029,129 @@ SHOULD use the terms defined by this document to do so.
...
+  reject `$id` values that use the `file:` scheme. If `file:` URIs are permitted
+  internally, implementations are advised to sanitize them (for example, by
+  converting them to relative URIs) to avoid exposing host filesystem structure
+  to users.
+
+### Vocabulary-Specific Risks
+
+Third-party JSON Schema vocabularies may introduce additional risks.
+Implementers are advised to consult the specifications of any extensions they
+support and take into account their security considerations as well.
```
gregsdennis, line 2154: We should update this to use "extension" language since we've removed "vocabulary".

PR #1625, opened 2025-08-31
Clarify IRI comparison for identity tokens (#1623)

### LINKED ISSUE #1623: URI normalization
In the https://github.com/json-schema-org/json-schema-spec/pull/1618 review discussion, I noticed that we may not be doing the right thing regarding URI normalization. RFC 3987 says,

> Applications using IRIs as identity tokens with no relationship to a protocol MUST use the Simple String Comparison (see [section 5.3.1](https://www.rfc-editor.org/rfc/rfc3987.html#section-5.3.1)). All other applications MUST select one of the comparison practices from the Comparison Ladder (see [section 5.3](https://www.rfc-editor.org/rfc/rfc3987.html#section-5.3)

JSON Schema describes our use of IRIs as identity tokens, not locators, which means implementations should limit themselves to Simple String Comparison. That's not what our spec currently says. It says to use the normalization procedure defined in 5.3, which isn't quite correct either. We'd need to be more specific about which comparisons to use. For example, we wouldn't require scheme-based or protocol-based normalization because it would be too onerous a requirement for implementers to implement scheme-based requirements for any scheme they might encounter.

riyap03:
…ison (#1623)

Documentation / specification clarification

- Closes #1623

### Summary

This PR updates the JSON Schema specification to clarify that when IRIs are used as identity tokens (not locators), implementations **must use Simple String Comparison** according to RFC 3987 section 5.3.1.  

Full normalization or scheme-based comparison is **not required** in this context.  

Example for clarity:

- `https://example.com/schema/α`  
- `https://example.com/schema/α`  

These two IRIs are considered identical only if their strings match exactly, without performing additional normalization.  

This fixes the inconsistency between the spec and RFC 3987 recommendations.

No. This is a documentation/specification update and does not affect existing implementations.

### DIFF
--- a/specs/jsonschema-core.md
+++ b/specs/jsonschema-core.md
@@ -978,9 +978,18 @@ To differentiate between schemas in a vast ecosystem, schema resources are
 identified by
 [absolute IRIs](https://www.rfc-editor.org/rfc/rfc3987.html#section-2.2)
 (without fragments). These identifiers are used to create references between
-schema resources. When comparing IRIs for the purposes of resource
-identification, implementations SHOULD first follow the IRI normalization
-procedures defined in [RFC 3987][rfc3987], section 5.3.
+schema resources. ### IRI comparison for identity tokens
+
+When IRIs are used as **identity tokens** (not locators), implementations **MUST use Simple String Comparison** as defined in RFC 3987 section 5.3.1.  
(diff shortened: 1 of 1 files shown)


jdesrosiers:
This issue is not ready for PR. It has some pretty significant implications that we need to discuss and decide on what action (if any) to take.

Issue #1185, opened 2022-02-17
jsonschema: declarative language or media type?


ioggstream:
## I wish

to clarify whether json schema is a declarative language or 

> a JSON-based format for describing the structure of JSON data

https://github.com/json-schema-org/json-schema-spec/blob/707431358d529dc8892610e4d8593cfdd1f090a3/jsonschema-core.xml#L63

## note 

This may be not trivial ;)

Relequestual:
Why not both?
The details of what JSON Schema IS, depends on who I'm talking to.

awwright:
Any "format for describing the structure of (other) data" sounds like a type of declarative language. Is your understanding different?

ioggstream:
When I first read the spec, I found that defining json schema as a media type was confusing, since iiuc a media type is a "label" for some content. 

Stating that it's either a  "declarative language" or a "format for describing the structure" is probably more clear to a first-time reader. I think the definition provided by @jdesrosiers here, is a perfect fit for an Abstract/Introduction.

https://github.com/ietf-wg-httpapi/mediatypes/blame/main/draft-ietf-httpapi-rest-api-mediatypes.md#L319

jdesrosiers:
This is definitely confusing, because there is JSON Schema the language and JSON Schema the media type. They are two distinct things, but have the same name.

JSON Schema the media type is the JSON media type with a few extensions including fragment identification, a media type parameter, and `$schema`. That's all it is. In theory, someone could use this media type for something that has nothing to do with JSON Schema the language.

JSON Schema the language is everything else including the keywords, references, and any other design decisions.

At least that's how I see it.

When the media type is registered, we can remove all the media type stuff from the JSON Schema spec for the next release. When we're doing that, we can make things more clear since we only have to define the language and not the media type.

awwright:
> They are two distinct things, but have the same name.

When I say JSON Schema is a media type, among the other components of its media type registration (the `application/schema+json` name, fragment semantics, etc), is the complete specification of its semantics. You can't really talk about a media type without talking about how a computer program processes it, in the normative.

And when I talk about media types in general, there's a large variety of them— images, data, scripting languages, executable code, or a combination of all of these. Of these, I would call JSON Schema a declarative, JSON-based vocabulary.

gregsdennis:
I think there's a distinction here that should be made.  There are two parts at work here:

1. a declarative language which allows multiple functions, including validation of data that is structured according to the JSON data model
[...]

PR #1627, opened 2025-09-08
Decouple Core from media types

jdesrosiers:
Resolves https://github.com/json-schema-org/json-schema-spec/issues/1185

This PR moves us away from talking about JSON Schema as a media type and toward talking about it as a language that can be implemented in a number of media types. For example, there could be an `application/schema+yaml` as well as `application/schema+json`.

The media type definitions have been moved to a separate document written as an I-D with the intention of it being adopted by the HTTP-APIs IETF working group when we're ready. HTTP-APIs already has a version of this [here](https://github.com/ietf-wg-httpapi/mediatypes/blob/main/draft-ietf-httpapi-jsonschema-mediatypes.md), but this has been completely rewritten and I thought it would be better for it to be maintained here until we're ready. Specifically, we need to publish the spec first before we can register media types that reference them.

In addition to removing media type language where necessary, I also rewrote the Abstract, Introduction, and Overview. Once I removed the language about media types (as well as some language about vocabularies that we missed previously) there was little difference between these sections. I rewrote them to be more distinct and serve separate purposes. I'm pretty happy with the Abstract and Introduction, but wasn't quite sure what to put in the Overview. I chose to make it a general overview of how JSON Schema works. I think it works, but let me know your thoughts.

~NOTE: This depends on https://github.com/json-schema-org/json-schema-spec/pull/1624 and includes those commits, but that PR should be merged first and then this PR will should only contain the spec updates. For now, you can just look at the `.md` files and ignore the rest as noise.~ (The infrastructure changes have been merged now)


specs/jsonschema-core.md
```diff
@@ -47,51 +84,29 @@ document are to be interpreted as defined in [RFC 8259][rfc8259].
...
-implementation, and cannot be disabled. These keywords are each prefixed with a
-"$" character to emphasize their required nature. These keywords are essential
-to the functioning of the `application/schema+json` media type.
+Schema evaluation is a recursive process. Some keywords contain one or more
+subschemas. These keywords can be used to create complex constraints or to
+describe compound values like arrays and objects. For example, to describe a
+JSON object, a schema can use the `type` keyword to declare that the value MUST
+be an object, and the `properties` keyword to apply separate schemas to each of
+the object's properties. This allows for the evaluation of a complex JSON
+document using a uniform recursive algorithm.
```
gregsdennis, line 98: I wonder if it would be good to add a note that the presence of `properties` doesn't imply that the instance must be an object.

specs/jsonschema-core.md
```diff
@@ -929,21 +900,25 @@ steps.
...
+3. External context - In some contexts, there is a default dialect or a way to
+   declare the dialect external from the schema resource in question. Examples
+   include the following.
+    - If the schema is embedded in a non-schema document (such as an OpenAPI
+      Document), the semantics for determining the dialect MUST be determined by
+      the specification that applies to that document.
+    - If the media type of the schema is known and the media type defines a
+      default dialect or a way to declare a dialect, the dialect MUST be
+      determined by the rules of that media types. For example, the
+      [application/schema+json] media type includes a `schema` parameter that
```
gregsdennis, line 915: Is this pre-emptive reference to a media type that doesn't yet exist okay?

specs/jsonschema-core.md
```diff
@@ -1461,14 +1451,17 @@ custom keywords.
...
-[^10]: These scenarios are analogous to fetching a schema over HTTP but
-receiving a response with a Content-Type other than `application/schema+json`.
-An implementation can certainly try to interpret it as a schema, but the origin
-server offered no guarantee that it actually is any such thing. Therefore,
-interpreting it as such has security implication and may produce unpredictable
-results.
+When a schema is resolved over HTTP or another protocol that declares the media
+type of the response, implementations SHOULD refuse to evaluate schemas whose
+declared media type is not a known and supported JSON Schema media type such as
+[application/schema+json].[^10]
```
gregsdennis, line 1459: pre-emptive reference

gregsdennis (approved):
Looks really good.  I don't think the overview is redundant.

Just a couple comments.

specs/jsonschema-core.md
jdesrosiers: I think that would be a reasonable thing to put in the definition of `properties`, but I think it's getting too deep in the details for the overview section.

However, it could make sense to cover the general principle that some keywords only apply when the instance is of a particular type. But, don't think we want to get too into architectural principles in the overview. It could balloon as we decide that other architectural principles should be mentioned as well.

specs/jsonschema-core.md
gregsdennis: That fine.  I just wanted to raise it.

PR #1640, opened 2025-10-21
Add proposal for removing leading # from  values

### LINKED ISSUE #1632: Proposal: Spec authors for the stable release
I propose that we completely drop the idea of the spec having "authors" and replace it with "contributors". The spec is a collaborative work and no one should be confused that they have any ownership of the spec because they are listed as an "author". We currently have an "Acknowledgements" appendix listing contributors. I suggest we move anyone listed as an author to the contributors list and just present that.

If we want, we can acknowledge the "editors" for a release as well. The role of "editor" would be a transient role that acknowledges who's doing the work to move the spec forward at any particular time. You get the title because you're doing the work, not because had the title at some point in the past. For example, Austin hasn't played a significant role in spec development since draft-06, yet we continue to list him as an author because he was a significant contributor in the past. Continuing to list him misleads about his level of involvement in the project. Listing the current editors for a release makes more sense.

I feel very strongly that dropping "authors" in favor of "contributors" is definitely something we should do. I think that acknowledging "editors" is probably ok as long as it's clear that it's only an indicator of who worked on that release and not some kind of indicator of something larger such as being the leader or owner of the project as whole. But, I wouldn't be opposed to leaving out "editors" and avoiding that slippery slope.

AnvayKharb:
It fixes #1632  
Proposal: Deprecate Leading # in $dynamicRef Values

This change addresses a long-standing inconsistency where $dynamicRef uses fragment-only IRI syntax ("#foo") while $dynamicAnchor uses plain text ("foo"). Following the removal of bookending requirements in previous versions, there's no technical necessity for $dynamicRef values to be formatted as URIs.

We propose deprecating the leading # while maintaining backward compatibility, allowing both "$dynamicRef": "node" (preferred) and "$dynamicRef": "#node" (deprecated). This provides a migration path toward consistent, simplified syntax while preserving existing schema validity during the transition period.

### DIFF
--- a/specs/proposals/dynamicref-leading-hash-removal.md
+++ b/specs/proposals/dynamicref-leading-hash-removal.md
@@ -1,0 +1,157 @@
+# JSON Schema Proposal: Remove Leading `#` from `$dynamicRef` Values
+
+## Status
+**Proposal** - Under consideration
+
+## Summary
+This proposal addresses the inconsistency between `$dynamicRef` and `$dynamicAnchor` syntax by removing the requirement for leading `#` in `$dynamicRef` values, transitioning from fragment-only IRI syntax to plain text identifiers.
+
+## Motivation
+
+### Current Inconsistency
+Currently, JSON Schema has an acknowledged inconsistency between related keywords:
+- `$dynamicAnchor`: uses plain text (e.g., `"node"`)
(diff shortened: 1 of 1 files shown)


jdesrosiers:
Please don't create PRs in the spec repo without being asked to. If you want to contribute here, we need a lot more people discussing the issues and sharing their opinions in the issues.

PR #1661, opened 2026-01-13
Clarify repository purpose for new contributors.

Rushda971:
This PR adds a short clarification note based on first-time contributor experience.

The goal is to help new contributors distinguish between stable specifications
and work-in-progress drafts, and to clarify that local builds are not required
for documentation-only contributions.


jdesrosiers:
There's nothing here that we can use. The first part is irrelevant. The second part says exactly the same thing as the introduction. The third part goes without saying. I'm going to close this. If there are changes you want to see, create an issue and let's discuss it first.

PR #1671, opened 2026-02-13
Attempt to make the spec look stylish

jdesrosiers:
Our current build just uses [water.css](https://watercss.kognise.dev/) because it does a reasonable job with no effort and I'm not a designer. I decided to try and see if I could make something that looks more professional with the help of LLMs. I think it looks ok and certainly better than what we have.

I also added a few new features. I added special highlighting for requirement level keywords like "SHOULD" and "MUST". I also added a scroll spy TOC, but it only works in Chromium-based browsers because it uses a brand new CSS feature.

I feel like a lot of this still needs a more professional eye to clean up the rough edges. I'm still hoping someone will come along and can do better.

image


gregsdennis:
I think this is a nice job.  I wonder if we should let one of our more experienced web contributers have a go at the CSS.

---

Things really stand out when it's formatted.  I've already noticed a few formatting issues in the spec text.  (The renderer looks like it did the right thing.)

- minor formatting issue in a list
- we link to keywords sometimes and sometimes not

Ahmedmhmud:
I've pulled down the changes @jdesrosiers made and worked on it to keep the design features and enhance the style and format, here is what I've done:
  1. added Pico CSS url as a stylesheet 
  2. changed in specs.css to prevent overriding the Pico variables 
  3. enhanced the padding of the content
  4. enhanced the code block and the dark mode colors

Here are some After results:
Screenshot 2026-02-14 173424

Screenshot 2026-02-14 173217

If it seems good, I'd like to create a PR with the newest version

jdesrosiers:
>  I wonder if we should let one of our more experienced web contributers have a go at the CSS.

I hope so! I talked to @Ahmedmhmud about trying to use Pico.css to replace water.css as an alternative to this new look. I'm interested in seeing what that looks like, but I would still like to get someone with experience have a go at cleaning up some of the rough edges. I know the TOC and the code blocks don't look quite right, but I don't really know what would look better.

jdesrosiers:
I did some clean up for code blocks. It includes several fixes as well as some better styling. Most importantly, it changes to using semantic HTML tags like figure and figcaption. It also fixes some of the markdown for a few code blocks. @Ahmedmhmud, sorry this will probably conflict a bit with what you're working on. Hopefully you can merge it ok with your own ideas.

image

jdesrosiers:
Closing in favor of #1672

PR #1675, opened 2026-02-20
Fix remark headings map to avoid Object.prototype collisions (fixes #1673)

### LINKED ISSUE #1673: Bug: Build crash due to unsafe object property access in remark plugins
## Description
While working with the local build pipeline, I discovered a bug in the custom `remark` plugins that can cause the entire build process to crash with an unhandled `TypeError`.

The issue stems from how heading IDs are tracked in `remark/remark-headings.js`. Currently, they are stored in a standard JavaScript object:

`file.data.headings = {};`

Later, in `remark/remark-reference-links.js`, the code checks for the existence of an ID using the `in` operator:
```javascript
if (!(id in file.data.headings)) {
  throw Error(`ReferenceLinkError: No header found with id "${id}"`);
}
```

Because `file.data.headings` is a plain object, it implicitly inherits properties from `Object.prototype` (like `constructor`, `toString`, or `__proto__`). If a document contains a heading or a reference that normalizes to one of these built-in property names (for example, a heading titled "Constructor"), the `in` check erroneously returns `true`. 

Instead of throwing the intended `ReferenceLinkError`, the plugin attempts to treat the native function (e.g., `Object.prototype.constructor`) as an AST node, leading to a fatal crash downstream:
`TypeError: Cannot read properties of undefined (reading 'section')`

## Why this matters
This is a critical Developer Experience (DX) issue. In technical specifications, terms like "Constructor" or "Prototype" are extremely common as headings. If an author refers to one of these via a reference link (e.g., `{{constructor}}`), the CI build will fail with an obscure error message that provides no hint to the author about what went wrong in their markdown.

## Steps to Reproduce
1. Create a markdown file with a heading named `constructor`:
   ```markdown
   # constructor
   Reference: {{constructor}}
   ```
2. Run the build command: `npm run build`
3. The build will crash with `TypeError: Cannot read properties of undefined (reading 'section')`.

NandiniDhanrale:
## Summary

Fixes a build crash in the docs remark pipeline when a heading ID collides with an `Object.prototype` property (for example, `constructor`), causing `remark-reference-links` to treat a prototype property as a heading and later crash when reading its metadata.

This PR ensures the headings collection is stored and accessed safely, so only real headings are considered.

Fixes #1673.

## Problem

Previously, `remark-headings.js` stored headings on `file.data.headings` as a plain object.

`remark-reference-links.js` then used:

- `id in file.data.headings` to check if a heading exists, and
- `file.data.headings[id]` to read heading metadata.

If a document contained a heading like:

```markdown
# constructor

Reference: {{constructor}}
```

then `id === "constructor"` would match via `Object.prototype.constructor`, so the check would pass but the stored value would not be a proper heading node. Later access to heading properties (such as `.data.section`) could fail with a `TypeError` during spec build.

## Implementation

This PR:

- Changes `remark-headings.js` to initialize `file.data.headings` as `Object.create(null)`.
- Updates heading existence checks to `Object.prototype.hasOwnProperty.call(...)` instead of `in`.
- Keeps direct keyed access (`file.data.headings[id]`) but only after the safe own-property check.
- Leaves external markdown/reference-link behavior unchanged, except prototype property names are no longer treated as valid headings.

## Tests

Added a regression test in `remark/remark-reference-links.test.js` that covers:

1. Missing-heading reference: `{{constructor}}` with no matching heading throws `ReferenceLinkError`.
2. Real-heading reference: a real `constructor` heading plus `{{constructor}}` resolves correctly without crashing.

Local validation:

- `node --test remark/remark-reference-links.test.js`
- `npm run build -- specs`

Both commands pass (exit code 0).

### DIFF
--- a/remark/remark-headings.js
+++ b/remark/remark-headings.js
@@ -75,11 +75,11 @@ const remarkNumberHeadings = (options) => (tree, file) => {
     file.data = {};
   }
 
-  file.data.headings = {};
+  file.data.headings = Object.create(null);
 
   visit(tree, "heading", (headingNode) => {
     if (headingNode.data?.id) {
-      if (headingNode.data.id in file.data.headings) {
+      if (Object.prototype.hasOwnProperty.call(file.data.headings, headingNode.data.id)) {
         file.message(`Found duplicate heading id "${headingNode.data.id}"`);
       }
       file.data.headings[headingNode.data.id] = headingNode;
--- a/remark/remark-reference-links.js
+++ b/remark/remark-reference-links.js
@@ -8,12 +8,13 @@ const referenceLink = /\{\{(?<id>.*?)\}\}/ug;
 const remarkReferenceLinks = () => (tree, file) => {
   findAndReplace(tree, [referenceLink, (value, id) => {
     // file.data.headings comes from ./remark-headings.js
-    if (!(id in file.data.headings)) {
+    if (!Object.prototype.hasOwnProperty.call(file.data.headings, id)) {
       throw Error(`ReferenceLinkError: No header found with id "${id}"`);
     }
 
-    const headerText = nodeToString(file.data.headings[id]);
-    const linkText = text(file.data.headings[id].data.section);
+    const headingNode = file.data.headings[id];
+    const headerText = nodeToString(headingNode);
+    const linkText = text(headingNode.data.section);
     return link(`#${id}`, headerText, [linkText]);
   }]);
 };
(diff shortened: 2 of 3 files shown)


h30s:
Since the root cause was prototype collisions, maybe we could also add a regression test for __proto__ to ensure full coverage of inherited keys?

NandiniDhanrale:
I’ve added regression coverage for inherited-key behavior in `remark-reference-links.test.js`, including `constructor`, `proto`, and `__proto__` cases. This includes both failure-path tests (no matching heading throws `ReferenceLinkError`) and success-path tests where real headings resolve correctly to their `#id` anchors. The new tests follow the existing `constructor` patterns, and `node --test remark/remark-reference-links.test.js` is passing. [github](https://github.com/json-schema-org/json-schema-spec/pull/1675)

PR #1676, opened 2026-02-21
fix: avoid Object.prototype collisions in remark headings map to fix …

### LINKED ISSUE #1673: Bug: Build crash due to unsafe object property access in remark plugins

h30s:

Bugfix

- Closes #1673 

### Summary

This PR addresses a build crash caused by heading IDs colliding with `Object.prototype` built-in properties like `constructor`. 

When an author uses `{{constructor}}`, the `(id in file.data.headings)` check natively evaluates to `true`, causing the plugin to crash because it tries to map a native JS function as a remark AST Node.

The problem is resolved by explicitly using a null-prototype map via `Object.create(null)` in `remark-headings.js` and changing presence checks in `remark-reference-links.js` to strictly use `Object.prototype.hasOwnProperty.call`. I have also added testing to confirm standard headers and colliding headers (like `constructor`) map cleanly without triggering ReferenceLinkErrors.

No breaking change.


h30s:
@jdesrosiers PTAL

jdesrosiers (changes requested):
As much as it pains me to say it, you can remove the tests. Adding tests in a repo that isn't setup to run tests are just a waste. They won't get run just become stale.

h30s:
@jdesrosiers I have removed the tests as requested. Please let me know if any further changes are needed.

jdesrosiers (changes requested):

h30s:
@jdesrosiers PTAL

jdesrosiers (changes requested):

jdesrosiers (approved):

PR #1679, opened 2026-02-25
fix: escape dot in rehype-link-transformer regex

### LINKED ISSUE #1677: Bug: `rehype-link-transformer.js` fails on unescaped regex and deprecated API usage
### Description

While reviewing the build pipeline, I noticed a few issues in `remark/rehype-link-transformer.js` that can lead to silent link corruption or build crashes.

### The Problems

1.  **Unescaped Regex Dot**: On line 12, the regex `/.md$/` uses an unescaped dot. This means it matches any character followed by `md` (like `readme-cmd`), rather than specifically `.md` files.
2.  **Deprecated API**: It uses `url.parse()`, which has been deprecated since Node 11 in favor of the WHATWG `URL` API.
3.  **Null Safety**: There's no check for missing `href` attributes. While `url.parse(undefined)` is currently "safe" (returns nulls), switching to the modern `URL` constructor would cause a crash without a guard.

### Why it matters
These issues make the link transformation fragile. The regex bug specifically can cause links to be rewritten incorrectly without any build errors, which is hard to track down once the site is deployed.

h30s:
Fixes #1677

Escape the dot in `/.md$/` → `/\.md$/` so it matches only literal `.md`
extensions, not any character followed by `md`

### DIFF
--- a/remark/rehype-link-transformer.js
+++ b/remark/rehype-link-transformer.js
@@ -9,7 +9,7 @@ const rehypeLinkTransformer = () => (tree, vfile) => {
     if (node.tagName === "a") {
       const href = url.parse(node.properties.href);
       if (href.hostname === null && href.pathname?.endsWith(".md") && existsSync(vfile.history[0])) {
-        href.pathname = basename(href.pathname).replace(/.md$/, ".html");
+        href.pathname = basename(href.pathname).replace(/\.md$/, ".html");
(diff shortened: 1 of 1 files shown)


h30s:
@jdesrosiers PTAL.

jdesrosiers:
See, that "Request" link? Please use that instead of tagging people to request a review.

image

PR #1680, opened 2026-02-26
fix: add missing 'annotation' property to output unit schema

### LINKED ISSUE #1504: Output unit schema inconsistent with spec
The 2020-12 spec [says this](https://json-schema.org/draft/2020-12/json-schema-core#section-12.3.4-4) about successful validations:

> The JSON key for failed validations is "error"; for successful validations it is "annotation".

A [later sentence](https://json-schema.org/draft/2020-12/json-schema-core#section-12.3.5-2) draws an explicit contrast for the case of nested results :

> Note the plural forms, as a keyword with nested results can also have a local error or annotation.

But the [output schema](https://json-schema.org/draft/2020-12/output/schema#$defs/outputUnit/properties) only defines a property for the plural `annotations` and not the singular `annotation`.

whocoder11:
The 2020-12 spec text states that the JSON key for successful validations
is 'annotation', but the output unit schema only defined 'error' (singular)
without the corresponding 'annotation' property.

Added 'annotation': true since annotation values can be of any type.

Fixes #1504

### DIFF
--- a/output/schema.json
+++ b/output/schema.json
@@ -28,6 +28,7 @@
         "error": {
           "type": "string"
         },
+        "annotation": true,
         "errors": {
           "$ref": "#/$defs/outputUnitArray"
         },


jdesrosiers:
@whocoder11 There's another step to completing this task. You'll also need to make a PR to the website repo and update the git submodule that points to this 2020-12 branch. Are you prepared to make that change as well? If so, I'll go ahead and merge this so you can make that change as well.

whocoder11:
@jdesrosiers Yes you can merge it

Issue #1299, opened 2022-09-19
State that implementations SHOULD accept a schema retrieval IRI / initial base IRI


handrews:
While we discuss initial base IRIs, at no point do we make it clear that it's advantageous to allow an application to supply a retrieval IRI or other IRI for use as a base IRI in the absence of an absolute `$id`.  Implementations vary with respect to how they handle this (or don't).

This strikes me as a SHOULD requirement rather than a MUST.  It's advantageous to allow, but JSON Schema is usable without it, and the exact mechanism is not something we should specify.  It's also conceivable that a specialized implementation would have a reason to skip this on the grounds of minimizing code or knowledge that external IRIs of this sort will not be useful in its intended execution context.

CLARIFICATION:  "Initial base URI" is not the same thing as RFC 3986 §5.1.4's "Default base URI".  An initial base URI can come from any of the sources described in RFC 3986 §5.1.2 - 5.1.4.  This issue doesn't have anything to do with setting a broader §5.1.4 default base URI.  Several comments related to that concept have been marked off-topic.

gregsdennis:
For example of this, see the playground at https://json-everything.net/json-schema.  There is an option that allows the user to specify a default base URI.

mwadams:
Agreed. I have been toying with this thought.

I have a "rebase as root" option which allows you to take a JSON island and rebase it as if it were an isolated document root; and I also have a "don't rebase this but treat this island as a base schema embedded in a non-schema document" (which is essentially the "OpenAPI" case).

But I don't have a "use this iri as the base for a schema, instead of the (potentially local filesystem-based) iri you actually came from.

gregsdennis:
I definitely think that we need to include language that states this is only a fallback for when `$id` is absent.  `$id` MUST NOT be overridden.

handrews:
@gregsdennis good point.  Should be obvious, but I am well aware that hardly anyone actually reads RFC 3986 no matter how much I xref it 😛

handrews:
@gregsdennis thinking on this more, there are four cases.   "Associate the schema with X" means that a schema author will be able to `"$ref": X"` or any other way of looking up schemas by IRI.  Assume "retrieval IRI" always means "retrieval or other initial base IRI" since really there's no way to tell where the IRI came from.

1. `$id` and no retrieval IRI:  MUST associate the schema with `$id` and set the base IRI
[...]

PR #1689, opened 2026-03-19
Remove "initial base iri" concept in favor of retrieval + default base IRI

Abha220804:
This change removes the “initial base IRI” model from the spec and aligns base IRI handling with RFC-3986’s concepts of retrieval IRI (identification) and default base IRI (relative reference resolution base).

The previous “initial base IRI” input/compressed model is confusing and ambiguous: an implementation cannot know whether a provided IRI is meant to be referenceable as a retrieval/identification IRI or only used as the default base for resolving relative references. This PR updates the json-schema-core wording to explicitly separate these roles and to encourage implementations to accept them independently (per #1299).

### DIFF
--- a/remark/remark-headings.js
+++ b/remark/remark-headings.js
@@ -62,12 +62,10 @@ const remarkNumberHeadings = (options) => (tree, file) => {
       headingNode.children.splice(0, 0, text(`${sectionNumber}. `));
     }
 
-    if (!("id" in headingNode.data)) {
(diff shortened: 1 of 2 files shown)


jdesrosiers:
If you actually read the issue, it should have been obvious that there was no consensus reached or agreement to move forward.

Also, we don't allow first time contributors to work on the specification. It's something that requires experience and a deep knowledge of the specification. Please always wait to be assigned an issue before submitting a PR.

PR #1694, opened 2026-04-05
fix: resolve #1647 — ✨ Proposal: `recommended` keyword similar to `required`

tudragon154203:
## Summary

fix: resolve #1647 — ✨ Proposal: `recommended` keyword similar to `required`

## Problem

**Severity**: `Medium` | **File**: `specs/proposals/recommended.md`

Create a new proposal document following the repository's proposal template structure. This document will formally capture the `recommended` keyword proposal from issue #1647, providing structured analysis of the problem, proposed solutions (both the object-key list variant and the schema variant), alternatives, and examples. This follows the same pattern as existing proposals like `propertyDependencies.md` and `vocabularies.md` in the same directory.

## Solution

Create the file with sections matching the proposal-template.md structure: Summary, Motivation (hand-written JSON workflows, RFC 2119 "SHOULD" semantics), Detailed Explanation (both variants from the issue), Examples (journal entry schema, library metadata schema showing recommended vs required fields), Drawbacks (complexity, implementation burden, ambiguous warning semantics across ecosystems), Alternatives Considered (annotations, custom vocabularies, `default` combined with validation), and Unresolved Questions (whether this belongs in core vs a vocabulary, exact warning output format, interaction with `required`). Reference RFC 2119 for "SHOULD" semantics and cross-link to the existing `required` keyword in jsonschema-validation.md.

## Changes

- `specs/proposals/recommended.md` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
*Generated by [ContribAI](https://github.com/tang-vu/ContribAI) v5.8.1*

PR #1696, opened 2026-04-05
fix: resolve #1585 — Issue: subset schemas don't work with `unevaluatedProperties` across `oneOf` branches

### LINKED ISSUE #1585: Issue: subset schemas don't work with `unevaluatedProperties` across `oneOf` branches
I can't figure out how to set up mutual exclusion of subset schemas elegantly.

I thought `unevaluatedProperties` was the solution so that I can easily reject properties of the superset schema that don't appear in the subset schema, but I can't find the right pattern to set it up right.

### Subset and Superset schemas

Let's define subset schema as such:

> A schema is a subset schema of another schema if any instance of the first schema also is valid also in the second schema.

Similarly,

> A schema is a superset schema of another schema if the other schema is a subset schema of it.

### Rationale

I have a real world schema with 2 kinds of values. But the first option happens to be a subset of the second
option. I'd like to make it so that if the subset schema validates, the superset schema will not validate.

```json
{
  "foo": (Either subset schema or superset schema)
}
```

## Attempt 1 - Use `not` - PASS

```yaml
oneOf:
- required: ['a']
  properties:
    a: {}
  not:
    required: ['b']
    properties:
      b: {}
- required: ['a', 'b']
  properties:
    a: {}
    b: {}
```

This one works as expected, where `{"a": true, "b": true}` only validates against the second of the oneOf options, but I have to nest the meat of the second option inside a not of the first. This is ugly, but effective. When I compose my schemas, I don't want to reference all possible superset schemas in the subset schema.

## Attempt 2 - Use `anyOf` - PASS

```yaml
anyOf:
- required: ['a']
  properties:
    a: {}
- required: ['a', 'b']
  properties:
    a: {}
    b: {}
```

This approach gives up on the "typing" quality of the oneOf where consumers of the instance can learn about the "type" of the schema by inspecting which branch validates. Consumers need more sophistication since both validate for `{"a": true, "b": true}`.

## Attempt 3 - Use `unevaluatedProperties` at the top level - FAIL

```yaml
unevaluatedProperties: false
oneOf:
- required: ['a']
  properties:
    a: {}
- required: ['a', 'b']
  properties:
    a: {}
    b: {}
```

For this instance, `{"a": true, "b": true}`, validation will fail on the `oneOf` since both branches validate and the collected annotations from the valid oneOf branches make both `"a"` and `"b"` evaluated properties.

## Attempt 4 - push unevaluatedProperties to child schemas - FAIL

If I push down `unevaluatedProperties: false`, I get the same issue as we have with `additionalProperties: false`, schemas can not be extended if a parent "tacks on" some more properties.

```yaml
oneOf:
- unevaluatedProperties: false
  required: ['a']
  properties:
    a: {}
- unevaluatedProperties: false
  required: ['a', 'b']
  properties:
    a: {}
    b: {}
properties:
  c: {}
```

If I try this JSON, the second oneOf will fail on `unevaluatedProperty`: `"c"`

```json
{"a": true, "b": true, "c": true}
```

## Attempt 5 - Introduce new in-place applicator `firstOf`: PASS

Let's define a new in place applicator, `firstOf`

> This keyword's value MUST be a non-empty array. Each item of the array MUST be a valid JSON Schema.
>
> An instance validates successfully against this keyword if it validates successfully against at least one schema defined by this keyword's value.
>
> Annotations are only collected from the first valid sub-schema.
>
> Consumers are to ignore subsequent valid subschemas beyond the first valid subschema

```yaml
unevaluatedProperties: false
firstOf:
- required: ['a']
  properties:
    a: {}
- required: ['a', 'b']
  properties:
    a: {}
    b: {}
```

Both branches validate this instance, `{"a": true, "b": true}`, but only annotations from the first branch
are collected. This causes `unevaluatedProperties` to disregard annotations from the second branch and thereby reject property `"b"` as "unevaluated".

tudragon154203:
## Summary

fix: resolve #1585 — Issue: subset schemas don't work with `unevaluatedProperties` across `oneOf` branches

## Problem

**Severity**: `High` | **File**: `specs/proposals/firstOf.md`

The issue describes a fundamental limitation with `unevaluatedProperties` when used with `oneOf` across subset/superset schema branches. When multiple `oneOf` branches validate, annotations from ALL valid branches are collected, which means `unevaluatedProperties: false` cannot distinguish between them. The user proposes a `firstOf` keyword that would only collect annotations from the first valid subschema, enabling proper mutual exclusion of subset schemas without the verbose `not` pattern.

## Solution

Create a new proposal document following the template in specs/proposals/proposal-template.md. The proposal should define `firstOf` as an in-place applicator keyword whose value is a non-empty array of schemas, where instances validate if they match at least one schema, but annotations are collected only from the first valid subschema. Include: (1) the problem statement with concrete examples showing the current `oneOf` + `unevaluatedProperties` failure, (2) the proposed keyword semantics, (3) interaction with `unevaluatedProperties` and `unevaluatedItems`, (4) comparison with alternatives like `anyOf` + discriminator patterns, and (5) examples showing the expected behavior for the subset/superset use case from the issue.

## Changes

- `specs/proposals/firstOf.md` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
*Generated by [ContribAI](https://github.com/tang-vu/ContribAI) v5.8.1*

### DIFF
--- a/specs/proposals/firstOf.md
+++ b/specs/proposals/firstOf.md
@@ -1,0 +1,371 @@
+# `firstOf` Keyword Proposal
+
+## Summary
+
+Introduce a `firstOf` applicator keyword that validates an instance against an array of subschemas, accepting the instance if at least one subschema validates, but collecting annotations only from the first valid subschema.
+
+## Motivation
+
+### The Problem
+
+The `unevaluatedProperties` keyword depends on the set of properties that have been "evaluated" by other keywords in the schema. When using `oneOf`, annotations from **all** valid subschemas are collected, even though only one branch is logically selected. This creates a fundamental problem when `oneOf` branches represent subset/superset relationships.
+
+Consider a schema that must distinguish between two related object shapes:
+
+```json
+{
+  "oneOf": [
+    {
+      "type": "object",
+      "properties": {
+        "kind": { "const": "base" },
+        "foo": { "type": "string" }
+      },
+      "required": ["kind", "foo"]
+    },
+    {
+      "type": "object",
+      "properties": {
+        "kind": { "const": "extended" },
+        "foo": { "type": "string" },
+        "bar": { "type": "integer" }
+      },
+      "required": ["kind", "foo", "bar"]
+    }
+  ],
+  "unevaluatedProperties": false
+}
+```
+
+An instance `{ "kind": "extended", "foo": "hello", "bar": 42 }` validates against the second subschema. However, it also validates against the first subschema if `kind` and `foo` are present. Because `oneOf` collects annotations from all valid branches, the property `bar` is marked as evaluated by the second branch. The `unevaluatedProperties: false` check passes, which is correct in this case.
+
+The problem arises when the branches are true subsets — where one schema's valid instances are a strict subset of another's:
+
+```json
+{
+  "oneOf": [
+    {
+      "type": "object",
+      "properties": {
+        "foo": { "type": "string" }
+      },
+      "required": ["foo"]
+    },
+    {
+      "type": "object",
+      "properties": {
+        "foo": { "type": "string" },
+        "bar": { "type": "integer" }
+      },
+      "required": ["foo", "bar"]
+    }
+  ],
+  "unevaluatedProperties": false
+}
+```
+
+For the instance `{ "foo": "hello", "bar": 42, "baz": true }`:
+
+1. The first subschema validates (it only requires `foo`)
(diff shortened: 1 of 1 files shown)


gregsdennis:
For the same reasons as https://github.com/json-schema-org/json-schema-spec/pull/1694, I'm closing this.

PR #1697, opened 2026-04-05
fix: resolve #1607 — ✨ Proposal: Provide a catalog of reusable types to replace most formats

### LINKED ISSUE #1607: ✨ Proposal: Provide a catalog of reusable types to replace most formats
### Describe the inspiration for your proposal

In the discussion about the format registry, someone proposed that formats in the registry include a schema that validates them. Implementations could then use that schema as a fallback if they didn't support the format. It occurred to me that that's just a different kind of referencing system specific to `format` and if that's the case, why not just use our existing referencing system and save `format` for things that can't be fully expressed by a schema such as dates. This approach may make the need for additional formats small enough that a registry is not longer necessary.

See https://github.com/json-schema-org/json-schema-spec/issues/1552#issuecomment-2954321107

### Describe the proposal

JSON Schema would maintain and host a catalog of schemas representing standard types that can be used in any schema. This catalog would be only for types that can point to a standard or be otherwise unambiguous. These could replace most of the `format`s we define in the spec and or the [registry](https://spec.openapis.org/registry/format/). Only those formats that can't be fully expressed with a schema such as dates would remain formats.

This would not be a replacement for SchemaStore.org which is generally used for things like schemas for configuration files. That would be out-of-scope for this catalog. Possibly we could include schemas for Schema.org types because they are standardized, but otherwise I don't expect complex types to be included.

Implementations would have this catalog built-in and/or have a way to pull down the catalog periodically to get recent updates. Retrieving individual schemas from the catalog would be possible but discouraged just like with meta-schemas. In fact, meta-schemas could be included in the catalog as well.

The main drawback of this approach is usability. It's a lot easier for a user to remember and type `format: "ipv4"` than something like `$ref: "https://json-schema.org/catalog/ipv4"`. Therefore, I'm proposing a custom URI scheme to access the catalog. Something like `$ref: "catalog:ipv4"`.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

tudragon154203:
## Summary

fix: resolve #1607 — ✨ Proposal: Provide a catalog of reusable types to replace most formats

## Problem

**Severity**: `Low` | **File**: `specs/proposals/type-catalog-adr.md`

Following the pattern in the `adr/` directory, this proposal warrants an Architecture Decision Record once the proposal reaches a decision point. The ADR should capture the context (format registry discussions), the decision, and consequences of choosing a catalog approach over alternatives.

## Solution

Create `specs/proposals/type-catalog-adr.md` following the pattern of existing ADRs like `vocabularies-adr.md`. This would document: 1) Context of the format registry problem and limitations, 2) Decision to pursue catalog approach (if accepted), 3) Consequences including implementation requirements, backward compatibility considerations, and impact on the format keyword's future role.

## Changes

- `specs/proposals/type-catalog-adr.md` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
*Generated by [ContribAI](https://github.com/tang-vu/ContribAI) v5.8.1*

### DIFF
--- a/specs/proposals/type-catalog-adr.md
+++ b/specs/proposals/type-catalog-adr.md
@@ -1,0 +1,22 @@
+# Type Catalog Architecture Decision Record
+Status: Accepted
+## Context
+The `format` keyword has been discussed as a possible registry-like extension point for reusable string formats and related validation behavior. Those discussions exposed limitations in using `format` as the primary mechanism for broader type discovery:
+
+- `format` is already a validation keyword, so expanding its role would blur the line between assertion and cataloging.
+- Different implementations need a place to publish and consume reusable entries without changing existing schema semantics.
+- A registry-style solution needs stable identifiers, discoverability, and room for implementation-defined entries.
+- The design should preserve existing behavior for schemas that already use `format`.
+
+## Decision
+We will pursue a catalog-based approach rather than expanding `format` into a general-purpose registry.
+
+The catalog will serve as the mechanism for naming, publishing, and discovering reusable type entries. The `format` keyword will retain its existing role unless a separate specification changes it.
+
+## Consequences
+- Implementations that support the catalog will need registration and lookup behavior for catalog entries.
+- Catalog entries must have stable identifiers and clear semantics so they can be shared across implementations.
+- Existing schemas that use `format` will continue to work without modification.
+- This decision preserves backward compatibility by avoiding changes to the current meaning of `format`.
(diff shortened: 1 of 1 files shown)


gregsdennis:
For the same reasons as https://github.com/json-schema-org/json-schema-spec/pull/1694, I'm closing this.

PR #1724, opened 2026-07-01
Fix specs-ietf cert bundle path

### LINKED ISSUE #1723: CI: specs-ietf fails on main due to kramdown-rfc cert-store check
`specs-ietf` is failing on `main`, even for editorial only PRs.

Examples:
- https://github.com/json-schema-org/json-schema-spec/actions/runs/28474453171
- https://github.com/json-schema-org/json-schema-spec/actions/runs/26911430558

In both runs, `specs-markdown` passes and `specs-ietf` fails in `npm run build-ietf` with:

```text
*** Configuration problem with OS certificate store.
```

This does not appear to be caused by the PR content. The failure also reproduces from a fresh checkout of `main`.

## Steps to reproduce:

1. Clone the repository.
2. Check out the current `main` branch.
3. Run:

```sh
npm run build-ietf
```

Observed result:
- `docker-compose run --rm build` starts.
- The IETF build aborts with `*** Configuration problem with OS certificate store.`
- The failure happens before spec content changes matter.

## Notes:

The current IETF build uses floating inputs in the Dockerfile:
- `FROM fedora`.
- `gem install kramdown-rfc`.

The error string comes from `kramdown-rfc` itself. It's current release checks `OpenSSL::X509::DEFAULT_CERT_FILE` and exits if that file is missing, which makes the build sensitive to environment drift rather than repository content.

RedZapdos123:
## Description:

This PR fixes the IETF build on current Fedora images by ensuring Ruby OpenSSL's default certificate file path exists in the build container before `kramdown-rfc` runs.

Closes #1723

### Checks run, locally:
- [x] `git diff --check`.
- [x] `npm run build-ietf`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 FROM fedora
 
 RUN dnf install -y ruby python3-pip
+# Fedora's CA bundle exists, but Ruby OpenSSL defaults to /etc/pki/tls/cert.pem.
+# kramdown-rfc aborts early if that default path is missing.
+RUN ln -sf /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/pki/tls/cert.pem
 RUN gem install kramdown-rfc
 RUN pip install xml2rfc
 


jdesrosiers (approved):
Thanks for tracking this one down!

PR #1736, opened 2026-07-16
fix: resolve meta-schema path as file URL for Windows compatibility

### LINKED ISSUE #1730: bug: Vitest suite crashes on Windows due to invalid path backslashes
I encountered an issue running the meta-schema tests on a Windows environment. Running `npm test` fails immediately on all 7 test cases with `Error: Invalid IRI-reference`.

### Root Cause
The path resolution in `specs/meta/test-suite/meta-schema.test.js` resolves the meta-schema path using a string template:
```javascript
`${import.meta.dirname}/../meta.schema.json`
```

On Windows, this resolves with backslashes (e.g., `D:\path\to\specs\meta\test-suite/../meta.schema.json`). Because the internally used `@hyperjump/uri` library strictly validates references against RFC 3986 (where backslashes are illegal URI characters), the parser throws an exception and crashes the test run.
### Steps to Reproduce
1. Clone the repository on Windows.
2. Run `npm install` and `npm test`.
3. The test execution fails with the following traceback:
```
Error: Invalid IRI-reference: D:\VS Code Projects\OSS\json-schema-spec\specs\meta\test-suite/../meta.schema.json
 ❯ Object.parseReference node_modules/@hyperjump/uri/lib/index.js:225:11
 ❯ node_modules/@hyperjump/uri/lib/index.js:83:78
 ❯ getSchema node_modules/@hyperjump/json-schema/lib/schema.js:33:24
```
### Screenshot:

Image

### Proposed Solution
To ensure cross-platform compatibility, we can resolve the file path as a standard file URL:
```javascript
const metaSchemaUrl = new URL("../meta.schema.json", import.meta.url).href;
```

I have verified this fix locally on Windows and all tests pass.

ApurveKaranwal:
## What kind of change does this PR introduce?

- [x] Bugfix (non-breaking change that fixes an issue)

---

## Issue & Discussion References

Closes #1730

---

## Summary

This fixes a Windows-only issue where the meta-schema test suite crashes before running.

The test currently resolves the meta-schema using:

```js
`${import.meta.dirname}/../meta.schema.json`
```

On Windows, `import.meta.dirname` contains backslashes, which results in an invalid URI being passed to `@hyperjump/uri`, causing the following error:

```text
Error: Invalid IRI-reference
```

The fix is to resolve the file relative to `import.meta.url` instead:

```js
const metaSchemaUrl = new URL("../meta.schema.json", import.meta.url).href;
```

This generates a valid `file://` URL on Windows, macOS, and Linux, avoiding the mixed path separator issue.

### Verification

- Reproduced the failure on Windows.
- Verified that the test suite no longer crashes after the change.
- Confirmed all 7 meta-schema tests pass successfully on Windows.

---

## Does this PR introduce a breaking change?

**No.**

This change only updates how the test file resolves the local meta-schema. It does not affect validator behavior or any runtime functionality.

## Screenshot

Screenshot (902)

### DIFF
--- a/specs/meta/test-suite/meta-schema.test.js
+++ b/specs/meta/test-suite/meta-schema.test.js
@@ -2,7 +2,6 @@ import { describe, expect, test } from "vitest";
 import fs from "node:fs/promises";
 import path from "node:path";
 
-const metaSchemaUrl = new URL("../meta.schema.json", import.meta.url).href;
 
 for (const entry of await fs.readdir(`${import.meta.dirname}/tests`, { withFileTypes: true })) {
   if (entry.isDirectory() || path.extname(entry.name) !== ".json") {
@@ -16,9 +15,9 @@ for (const entry of await fs.readdir(`${import.meta.dirname}/tests`, { withFileT
     for (const testCase of suite.tests) {
       test(testCase.description, async () => {
         if (testCase.valid) {
-          await expect(testCase.schema).to.matchJsonSchema(metaSchemaUrl);
+          await expect(testCase.schema).to.matchJsonSchema("specs/meta/meta.schema.json");
         } else {
-          await expect(testCase.schema).to.not.matchJsonSchema(metaSchemaUrl);
+          await expect(testCase.schema).to.not.matchJsonSchema("specs/meta/meta.schema.json");
         }
       });
     }
--- a/specs/meta/test-suite/meta-schema.test.js
+++ b/specs/meta/test-suite/meta-schema.test.js
@@ -2,6 +2,7 @@ import { describe, expect, test } from "vitest";
 import fs from "node:fs/promises";
 import path from "node:path";
 
+const metaSchemaUrl = new URL("../meta.schema.json", import.meta.url).href;
 
 for (const entry of await fs.readdir(`${import.meta.dirname}/tests`, { withFileTypes: true })) {
   if (entry.isDirectory() || path.extname(entry.name) !== ".json") {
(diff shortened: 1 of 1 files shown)


ApurveKaranwal:
hey @jdesrosiers sir, please review my PR.

specs/meta/test-suite/meta-schema.test.js
```diff
@@ -15,9 +16,9 @@ for (const entry of await fs.readdir(`${import.meta.dirname}/tests`, { withFileT
     for (const testCase of suite.tests) {
       test(testCase.description, async () => {
         if (testCase.valid) {
-          await expect(testCase.schema).to.matchJsonSchema(`${import.meta.dirname}/../meta.schema.json`);
+          await expect(testCase.schema).to.matchJsonSchema(metaSchemaUrl);
```
jdesrosiers: Let's just skip the URI resolution altogether.

```suggestion
          await expect(testCase.schema).to.matchJsonSchema("specs/meta/meta.schema.json");
```

jdesrosiers (changes requested):

specs/meta/test-suite/meta-schema.test.js
ApurveKaranwal: Thanks @jdesrosiers sir, Updated the PR to pass "specs/meta/meta.schema.json" directly as suggested. Tested locally and all tests pass.

jdesrosiers (approved):
Thank you.

And thank you for your patience.

PR #1767, opened 2026-08-23
Fix typos in decouple-from-IETF ADR

RedZapdos123:
This PR fixes two editorial typos in `adr/2022-09-decouple-from-ietf.md`.


jdesrosiers (approved):
While I appreciate fixing typos, please focus your attention away from ADRs. They're intended to be historical records and shouldn't really be changing. I'll merge this, but in the future, let ADR typos go.

PR #1769, opened 2026-08-23
fix(meta-test-suite): require test case fields

RedZapdos123:
## Description:

This PR requires `description`, `schema`, and `valid` in each test case entry in `specs/meta/test-suite/test-suite.schema.json`.

### Checks run, locally:

- [x] `git diff --check`
- [x] `npm test`
- [x] `npm run lint`


jdesrosiers (changes requested):
Adding the `required` keyword makes sense.

But, I don't think it's necessary to have tests for the test-suite schema. If you want to add something to the github actions that checks the tests against against the test-suite schema, that would be more helpful.

RedZapdos123:
Removed the synthetic test from this PR. Actual suite file validation is already covered by merged #1770 so this PR now only keeps the required field change.

jdesrosiers (approved):

Issue #28, opened 2016-05-10
Add logical "if" schema


TaylanUB:
It would be great if there were a logical `if` schema in the vein of `and`/`or`/`not`. It could look like

```
{ "if": [ conditional, consequent ] }
```

whose semantics would be exactly equivalent to:

```
{ "or": [ { "not": conditional },
          { "and": [ conditional, consequent ] } ] }
```

Thank you!

davek2:
How would this be used?   Schemas are declarative languages, not procedural - there are no variables and no flow of control.  "and" and "or" do not exist in JSON Schema; "not" exists, but not as a logical operator, it is more like unary negation:

       { "not": { "type": "string" } }

validates all values except strings.  An example illustrating how this would be used to accomplish a schema-related task would be helpful.

TaylanUB:
David Kay <notifications@github.com> writes:

> How would this be used? Schemas are declarative languages, not
> procedural - there are no variables and no flow of control. "and" and
> "or" do not exist in JSON Schema; "not" exists, but not as a logical
> operator, it is more like unary negation:
>
>    { "not": { "type": "string" } }
>
> validates all values except strings. An example illustrating how this
> would be used to accomplish a schema-related task would be helpful.

Oops, with 'and' and 'or' I meant 'allOf' and 'anyOf'.

Propositional logic is declarative, and has 'if'.  The proposition

    If cats are blue, then the moon is made of cheese.

evaluates to true, because the proposition "cats are blue" is false.
(See also "vacuous truth".)  The proposition

    If cats have fur, then dogs have four legs.

evaluates to true, because both sub-propositions are true.

Anyhow, here's a useful schema example:

    { "type": "object",
      "required": [ "productType" ],
      "properties":
        { "productType": { "enum": [ "book", "vhs" ] },
          "numberOfPages": { "type": "integer" },
          "duration": { "type": "number" } },
      "allOf":
        [ { "if": [ { "properties": { "productType": { "enum": [ "book" ] } } },
                    { "required": [ "numberOfPages" ] } ] },
          { "if": [ { "properties": { "productType": { "enum": [ "vhs" ] } } },
                    { "required": [ "duration" ] } ] } ] }

This validates

    { "productType": "book",
      "numberOfPages": 242 }

and

    { "productType": "vhs",
      "duration": 9330 }

but not

    { "productType": "book",
      "duration": 9430 }

or

    { "productType": "vhs",
      "numberOfPages": 230 }

[...]

PR #1777, opened 2026-09-14
fix: clarify regular expression guidance

### LINKED ISSUE #1760: bug: meta.schema.json enforces valid regex for pattern and patternProperties even though the spec only says SHOULD
## Description:

While validating schemas against `specs/meta/meta.schema.json` on the latest `main` (`8980615`, August 10, 2026), I found that the meta-schema rejects invalid regular expressions for both:

- `pattern`
- `patternProperties` property names

That behavior appears stricter than the current spec text.

In the meta-schema:

- `specs/meta/meta.schema.json:120` sets `pattern` to `"format": "regex"`
- `specs/meta/meta.schema.json:66` sets `patternProperties.propertyNames` to `"format": "regex"`

But the spec text says these regular expressions `SHOULD` be valid, not `MUST` be valid:

- `specs/jsonschema-validation.md:238`
- `specs/jsonschema-core.md:1757-1758`

So right now the meta-schema seems to turn a `SHOULD` requirement into an enforced validation failure.

## How to reproduce:

From a fresh checkout on the latest `main`:

```bash
npm ci
node --input-type=module -
```

Then run:

```js
import { matchJsonSchema } from "./node_modules/@hyperjump/json-schema-coverage/src/vitest/json-schema-matchers.js";

const ctx = {
  isNot: false,
  equals: Object.is,
  utils: {
    printReceived: (v) => JSON.stringify(v),
    printExpected: (v) => JSON.stringify(v),
  },
};

const cases = [
  ["pattern", { "$schema": "https://json-schema.org/v1", "pattern": "(" }],
  ["patternProperties", { "$schema": "https://json-schema.org/v1", "patternProperties": { "(": true } }],
];

for (const [name, schema] of cases) {
  const result = await matchJsonSchema.call(ctx, schema, "./specs/meta/meta.schema.json");
  console.log(`> case: ${name}`);
  console.log(`pass: ${result.pass}`);
  console.log(result.message());
}
```

### Example 1: `pattern`

```json
{
  "$schema": "https://json-schema.org/v1",
  "pattern": "("
}
```

### Example 2: `patternProperties`

```json
{
  "$schema": "https://json-schema.org/v1",
  "patternProperties": {
    "(": true
  }
}
```

## Reproduction:

Image

## Actual behavior:

Both examples fail meta-schema validation.

For `pattern`, the failure points to:

- `https://json-schema.org/v1/2026#/properties/pattern/format`

For `patternProperties`, the failure points to:

- `https://json-schema.org/v1/2026#/properties/patternProperties/propertyNames/format`

So invalid regex syntax is currently being enforced by the meta-schema itself.

## Expected behavior:

Based on the current spec wording, these cases should not fail meta-schema validation solely because the regex is invalid, since the spec currently says the regex `SHOULD` be valid rather than `MUST` be valid.

If the intended behavior is to enforce valid regex in the meta-schema, then the spec wording may need to be tightened to match that behavior.

RedZapdos123:
## Description:

This PR clarifies that regular expression validity is guidance for schema authors and links the `regex` format to the shared Core guidance.

Closes #1760.

### Checks run, locally:

- [x] `git diff --check`
- [x] `npm test`
- [x] `npm run lint`
- [x] `npm run build -- specs/jsonschema-core.md specs/jsonschema-validation.md`


specs/jsonschema-core.md
```diff
@@ -323,9 +323,9 @@ describable by JSON.
 ### Regular Expressions {#regex}
 
 Keywords MAY use regular expressions to express constraints, or constrain the
-instance value to be a regular expression. These regular expressions SHOULD be
-valid according to the regular expression dialect described in [ECMA-262,
-section 21.2.1](https://www.ecma-international.org/ecma-262/11.0/index.html).
+instance value to be a regular expression. Schema authors SHOULD use regular
+expressions that are valid according to the regular expression dialect described
+in [ECMA-262, section 21.2.1](https://www.ecma-international.org/ecma-262/11.0/index.html).
```
gregsdennis, line 328: This should direct readers to the Core regex section also.  All regex usages should point there.

gregsdennis (changes requested):

specs/jsonschema-core.md
RedZapdos123: Thanks for the review. The guidance points directly to the Core regex section now.

gregsdennis: This wasn't addressed. This reference still points directly to ecma 262.

RedZapdos123: Since this is the main Core regex section itself, should it use `{{regex}}` or keep the ECMA-262 link?

jdesrosiers: Yep, this is the core regex section 😄. This is the only place the ECMA-262 reference should be needed.

jdesrosiers, line 328: With this version of the specification, we've be moving away from placing requirements on users. The specification should instead describe what implementations should do. So, the "schema authors SHOULD" language is problematic. It should instead indicate that implementations SHOULD support ECMA-262 regexes and MUST raise an error if it encounters regex syntax it doesn't support.

@gregsdennis, should it also say that it MUST raise an error if given regex syntax that isn't compatible with ECMA-262?

jdesrosiers (changes requested):

specs/jsonschema-core.md
gregsdennis: > It should instead indicate that implementations SHOULD support ECMA-262 regexes and MUST raise an error if it encounters regex syntax it doesn't support.

The point of my comments are that we need to steer away from mentioning ECMA 262 everywhere and point people to the Core regex section instead.  Core doesn't say that ECMA 262 MUST be fully supported.  It gives a subset of features.

Also, yes, the language should be aimed at implementation support instead of authors.  Thanks for the reminder.

>  should it also say that it MUST raise an error if given regex syntax that isn't compatible with ECMA-262?

I think it should say that it MUST raise an error if the given regex isn't supported by the implementation.  Core doesn't say that ECMA 262 MUST be fully supported.  It gives a subset of features.

gregsdennis: @RedZapdos123 my mistake.  I didn't see where this was.

jdesrosiers: Actually, I was referring to something a little different. If a Python implementation gets a schema with a regex that Python understands but ECMA-262 doesn't, should it be allowed for the Python implementation to evaluate that regex, or should it detect that the regex isn't ECMA-262 compatible and throw an error?

I think it should. The only question I have is, how difficult is it to detect if a regex is ECMA-262 compatible in a language other than JavaScript? If that's a relatively easy thing to do in any language, I think the requirement should be a MUST, otherwise we can make it a SHOULD.

gregsdennis: > I think it should.

Which?  You presented two options then said yes. 😆

I don't think that's a trivial task.

gibson042: Note also #28, and in particular that [RFC 9485 I-Regexp](https://datatracker.ietf.org/doc/html/rfc9485) was intended to address precisely this kind of issue.

jdesrosiers: > Which? You presented two options then said yes. 😆

Ha! You're right. Sorry. I meant I think it should throw an error.

> I don't think that's a trivial task.

I'm leaning that way as well. Let's make it a "SHOULD".

jdesrosiers: @gibson042 Yes, we're aware of I-Regexp, but that's a whole other discussion for another time.

